PR #274, opened 2014-05-12
Add unzip kmz files to collada files module:

huanjet:
modify the lib file location
delete extra files


pjcozzi:
@fabrobinet do we still want this pull request?  It doesn't seem like something we need in the core converter.  It seems like it would be better to build a separate kmz/kml tool on top of COLLADA2GLTF.

pjcozzi:
Closing, this is an old pull request and something that should be done separate from the core converter.

PR #345, opened 2015-01-07
Fixed compilation error.

tfili:
Compilation was failing on Windows because it didn't know about the type String. Added COLLADAFW:: and all is good again.


fabrobinet:
Thanks Tom, merging.

oh yeah, those COLLADAFW::String are annoying, I feel like we should replace all that by just std::string in OpenCOLLADA, that was a way to eventually replace by another type of string (from what I understood) but this is not happening and we should get rid of those now...

PR #346, opened 2015-01-13
Images weren't be checked for transparency

tfili:
Fixed issue where if the cwd isn't a models directory the parent path gets prepended to the texture data uri. This causes the png data to not be read and we set the blending flag incorrectly.


fabrobinet:
looks good thanks.

PR #352, opened 2015-02-09
Image files with spaces don't embed correctly.

tfili:
When we try to embed images with spaces we don't decode the URI so the %20's are in the path and we fail to find it. This change decodes the path before we try to open it.


pjcozzi:
@fabrobinet can we merge this?

Issue #359, opened 2015-02-22
Add minimal gltf file


adius:
It'd be cool if you could add something like a minimal gltf file!
Maybe just a unit-cube or a tetrahedron with minimal shading?
Or can I find something like this somewhere else?

pjcozzi:
The closest example model is the [duck](https://github.com/KhronosGroup/glTF/tree/master/model/duck), which is a single mesh/material/texture.

adius:
But the whole techniques part isn't really necessary to render something, or is it?

pjcozzi:
The `technique` and `material` are both needed.  Think of a material as an instance of a technique, and think of a technique as the shader and its inputs that describes the physical appearance.

adius:
…and there is no defined default value?
I hoped that just specifying the mesh would be enough to get a rendering.

pjcozzi:
No.  Just like WebGL, geometry along is not enough.  We need to describe how the geometry is shaded.

pjcozzi:
@adius see the simple model in #365.

adius:
Cool, thanks!

fabrobinet:
Thanks @pjcozzi closing now.

adius:
Maybe next time you leave it open until the pull request is merged. ;-)

pjcozzi:
I think merging the pull request automatically closed this.

emackey:
#365 is not merged yet.

fabrobinet:
Oops yet, done.

PR #365, opened 2015-03-06
Simple box model

pjcozzi:
This is a simple test model we use in Cesium from AGI and [@planetpuncher](https://twitter.com/planetpuncher).

Would be good for #359.


fabrobinet:
Thanks @pjcozzi

Issue #369, opened 2015-03-17
Converter generates incorrect shaders or attributes?


pjcozzi:
@fabrobinet can you look at this?  The model is attached here: https://groups.google.com/forum/#!topic/cesium-dev/pi20u3DWPTc

What looks to be a valid COLLADA model (well, at least it loads in Mac Preview and Xcode) generates either incorrect shaders and/or incomplete (and extra) vertex attributes.

This model generates three meshes, each of which has one primitive.  These primitives have the following attributes:
```
Geometry0: POSITION, NORMAL, TEXCOORD_0
Geometry1: POSITION
Geometry2: POSITION
```
(Not the original COLLADA model does not have any textures, perhaps this is the problem?)

However, the generated vertex shader expects `POSITION` and `NORMAL` (so `Geometry0` has extra attributes and the other geometries are missing the `NORMAL` attribute):
```
precision highp float;
varying vec3 v_normal;
uniform vec4 u_ambient;
uniform vec4 u_diffuse;
uniform vec4 u_specular;
uniform float u_shininess;
void main(void) {
vec3 normal = normalize(v_normal);
vec4 color = vec4(0., 0., 0., 0.);
vec4 diffuse = vec4(0., 0., 0., 1.);
vec4 ambient;
vec4 specular;
ambient = u_ambient;
diffuse = u_diffuse;
specular = u_specular;
diffuse.xyz *= max(dot(normal,vec3(0.,0.,1.)), 0.);
color.xyz += diffuse.xyz;
color = vec4(color.rgb * diffuse.a, diffuse.a);
gl_FragColor = color;
}
```
This causes Cesium to fail to render it.

Loading and saving in Xcode does not help.

fabrobinet:
@pjcozzi will look at it this WE.

pjcozzi:
Thanks so much @fabrobinet!

PR #371, opened 2015-03-22
https://github.com/KhronosGroup/glTF/issues/369

fabrobinet:
@pjcozzi I checked the generated file it looks fine - but I can't test the viewer right now. Can you check with Cesium before merging ? 

The generated technique hashcode wasn't taking into account the vertex attributes - which was quite wrong, and possibly a regression.

Note: for the shader that only contains one position, the generated code is quite dumb and could be much better but we can track that separately. For now the scope of this fix was make sure we generate a shader consistent with the vertex attributes.


pjcozzi:
CC #369 

> the generated code is quite dumb and could be much better but we can track that separately

Agreed, created #372.

pjcozzi:
Works!  Thanks so much @fabrobinet.

Issue #142, opened 2013-10-18
Animation: scope for channel target id/path


pjcozzi:
This is the first in a series of questions (over the next week or so) to help me nail down my glTF animation implementation and properly word the spec.

In our current design, given a channel's target, e.g.,

``` javascript
"target": {
  "id": "_-1236436",
  "path": "translation"
}
```

Am I correct that:
- `id` can only reference a `node`?
- `path` must be `"translation"`, `"rotation"`, or `"scale"`?

I am OK with only having TRS animations for glTF 1.0, but are we confident that we won't need breaking changes to make this support `instanceTechnique` parameters for example?

pjcozzi:
Given that this is not a breaking change, isn't widely used AFAIK, and I don't think we would have time to implement it, I suggest we push this from 1.0.

carlsonsolutiondesign:
Events, data and color binding and animation.  Strike out animation from the spec and show me how your events, color and data binding are better than D3 + X3DOM???

pjcozzi:
@carlsonsolutiondesign I am not following.  We would not remove animation from the spec.  My suggestion is to keep animation targets as they are currently scoped (node transforms) and then widen the scope (e.g., to target material parameters) in an upcoming glTF version when we have enough time to implement it and make sure the design is solid.

@tparisi I still suggest we go this route.

tparisi:
@pjcozzi I'm not sure what the issue is here any more. Maybe close this and open others e.g. 1) add the ability to animate arbitrary properties and 2) add texture matrices as properties that could be animated.

pjcozzi:
I'm going to leave this open, but drop the `1.0` label since it has a lot of useful discussion.  When we revisit, we can divide it into two issues if needed.

carlsonsolutiondesign:
My apologies.  I googled glTF animations and I thought this was the word on animations for 1.0.  node animations are likely enough.


How do you do animations? Is it in the JSON or otherwise?

John
On Sep 16, 2015, at 12:56 PM, Patrick Cozzi <notifications@github.com<mailto:notifications@github.com>> wrote:


@carlsonsolutiondesign<https://github.com/carlsonsolutiondesign> I am not following. We would not remove animation from the spec. My suggestion is to keep animation targets as they are currently scoped (node transforms) and then widen the scope (e.g., to target material parameters) in an upcoming glTF version when we have enough time to implement it and make sure the design is solid.

[...]

PR #375, opened 2015-03-31
Fixed a few random crashes

tfili:
Fixes crashes caused by

1. Animations that don't target nodes (such as animation targeting the diffuse color of an effect)
2. Empty source arrays. Not sure why this keeps coming up by I've seen multiple models that have an `<source>` element used for normals. Regardless we shouldn't crash.


pjcozzi:
Thanks @tfili.

> Animations that don't target nodes (such as animation targeting the diffuse color of an effect)

FYI, glTF will support this in the future, #142

PR #377, opened 2015-04-02
Fix help text

### LINKED ISSUE #310: -e command-line argument is not documented
This is used to embed resources in the glTF file.

emackey:
This fixes some issues with the built-in help text.

   * The number of help options is now automatically computed, so none are left off the end.
   * The `-a` option help text is commented out, to match code supporting the option which is also currently commented out.
   * The `-e` option has an improved description, revealed now by the above fix.

Fixes #310.


pjcozzi:
Thanks @emackey

PR #378, opened 2015-04-22
Added a plugin point for the writer object

tfili:
Made GLTFWriter abstract and made the previous implementation the default in GLTFDefaultWriter. This will allow people to implement their own writers (minified, better double precision, output to string, etc).


pjcozzi:
Thanks @tfili

PR #382, opened 2015-05-05
Helped with performance for apps using this as a library

tfili:
Added ability to load a collada file from memory and read hints about image transparency from collada extra tags. This helps especially with data uris.


pjcozzi:
Thanks @tfili

PR #385, opened 2015-05-11
URI fix

tfili:
Fixed issue where we don't use OpenCOLLADA URI.getFileDir funcion correctly. On Windows this would lead to issues with not writing bin/glsl files.

PR #386, opened 2015-05-19
Fixed linux build.

tfili:
For some reason the include of `GLTFOpenCOLLADAUtils.h` was inside `#if __cplusplus <= 199711L`. I don't have a linux machine but I'm assuming this is the fix for the linux compile issue mentioned by @JHanley85 in #385.


pjcozzi:
Thanks @tfili

PR #388, opened 2015-05-22
Specify MIME types for transfer to client.

emackey:
The new MIME type is meant to fit in with the official list of [MIME types for models](http://www.iana.org/assignments/media-types/media-types.xhtml#model).  It also makes use of the [`+json` suffix](https://tools.ietf.org/html/rfc6839#page-4) indicating the format of the data.

I've removed the old "assetLayout" graphic as it has the obsolete `.json` extension burned into it, and the information there was redundant with what's written directly above it in the same README.

Are there any other places where the official MIME type & extension can or should be specified outside of this README?

/cc @shunter


pjcozzi:
Thanks @emackey.  I would say that we should merge this into the `spec-9` branch, but I think this is fine since we did not change the format, are still at `0.x`, and never previously specified the MIME type.

PR #394, opened 2015-07-03
Threejs anim fix

amirebrahimi:
I have two branches on my fork: one w/ r71 and one w/o. I went back and created this separate threejs-anim-fix branch, so that the resources loading issue could be pulled back without any of the r71 changes. If you hop on the threejs-r71-fix branch you'll see there are a lot more of WebGL errors that require fixing.

Additionally, I've added a new model, tom_plnt01, that can be tested against. I checked this into master on my branch, so if you checkout that and run it against the pre-existing loader you'll see the issue happen (e.g. will be a 'm undefined' issue, which is a matrix that doesn't exist.


tparisi:
Thanks Amir, I will play around with both branches and let you know. It's the holiday here and then I am traveling next week; will be able to dig in early next week.

pjcozzi:
@tparisi what's the plan with this pull request?

pjcozzi:
@tparisi is this OBE?

tparisi:
Don't know, I'll be getting back to this and have a look along with the skinning issues. Stay tuned.

pjcozzi:
@tparisi any update here?

pjcozzi:
@tparisi any update here? Should we close this?

tparisi:
@pjcozzi not yet. I will look into this again after all the 1.0 changes and my own r71 upgrade or even might upgrade to r73. Those have to come first too disruptive atm otherwise

pjcozzi:
@tparisi any update here? Should we close this?

tparisi:
I think we should close this, there were so many breaking changes along the way. If this rears its head again we can open another issue.

pjcozzi:
Agreed.  Thanks to @amirebrahimi nonetheless.

PR #396, opened 2015-07-07
Add option to output minified glTF JSON

kainino0x:
As-is, this changes the default behavior to ugly-print and allows enabling pretty-print. Not sure whether that should change. @tfili

### DIFF
--- a/converter/COLLADA2GLTF/main.cpp
+++ b/converter/COLLADA2GLTF/main.cpp
@@ -82,6 +82,7 @@ static const OptionDescriptor options[] = {
 	{ "h",              no_argument,        "-h -> help" },
 	{ "r",              no_argument,        "-r -> verbose logging" },
 	{ "e",				no_argument,		"-e -> embed resources (bin, shaders, available textures) in glTF file" },
+    { "M",              no_argument,        "-M -> output minified glTF JSON" },
     { "n",              no_argument,        "-n -> don't combine animations with the same target" }
 };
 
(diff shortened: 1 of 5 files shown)


converter/COLLADA2GLTF/main.cpp
```diff
@@ -82,6 +82,7 @@ static const OptionDescriptor options[] = {
 	{ "h",              no_argument,        "-h -> help" },
 	{ "r",              no_argument,        "-r -> verbose logging" },
 	{ "e",				no_argument,		"-e -> embed resources (bin, shaders, available textures) in glTF file" },
+    { "P",              no_argument,        "-P -> pretty print glTF JSON" },
```
tfili: I'm not sure if we should change the default print style since this is the open source converter. @pjcozzi what do you think?

I'm thinking we should have it default to pretty print. Give an option to ugly-print it (I wouldn't call it ugly-print, maybe minimize size or something like that).

pjcozzi: Yes, default to pretty print and change the ugly-print name to whatever is standard.

pjcozzi:
@tfili any other comments on the code?

kainino0x:
Ready for review.

kainino0x:
I'm not sure how, but it turns out this still does change the default behavior. I am working on a fix for that plus a small refactor.

kainino0x:
rapidjson does not allow for polymorphism between `PrettyWriter` and `Writer` (`Writer` doesn't have virtual methods, so this actually doesn't work (which slipped by in my apparently-crappy testing before) and this is more complex than I thought. I would either have to write wrapper classes with dynamic dispatch or add a lot of if/else statements to `GLTFWriter.cpp`. As such, perhaps we should just back out the merge and not think about this now.

kainino0x:
(Or, I guess, use awful macros....)

tfili:
@kainino0x This isn't really a must have feature, so it may just be easier to back it out.

kainino0x:
I agree, let's do that. Should I open another PR or will you just do it on master?

PR #395, opened 2015-07-06
bglTF benchmarking results and analysis

kainino0x:
https://github.com/kainino0x/glTF/blob/new-extensions/extensions/CESIUM_binary_glTF/README.md

https://github.com/kainino0x/glTF/blob/new-extensions/extensions/CESIUM_binary_glTF/BenchData/README.md

PR #397, opened 2015-07-08
bglTF benchmarking results and analysis

kainino0x:
Re-created version of #395.

https://github.com/kainino0x/glTF/blob/new-extensions/extensions/CESIUM_binary_glTF/README.md

https://github.com/kainino0x/glTF/blob/new-extensions/extensions/CESIUM_binary_glTF/BenchData/README.md

@pjcozzi


pjcozzi:
Any thoughts on what I asked in #395:

> Do you want to write a short tech blog about this, e.g., "Do not embed textures in your gzipped models?"

kainino0x:
I wouldn't mind doing that. I would have to do a bit more testing to really understand what is going on with textures in gzipped bgltf, though. The test model that should probably show it best (LargeTex) seems to actually not be very well-compressed to begin with, as gzip seems to reduce the size of the png.

pjcozzi:
OK, let's hold on the blog post until after we do geometric compression, which I think will yield deeper insights here as well.

pjcozzi:
@kainino0x this is ready, right?  No need to hold this up for the blog post.

kainino0x:
Yes, it is ready.

Issue #167, opened 2013-10-28
Guarantee typed array compatible byte alignment


pjcozzi:
This is a serious (but easy to fix) issue that we need to consider ASAP.

We want a glTF implementation to be able to create a single arraybuffer from a glTF `buffer` and then use the glTF `bufferView` and `accessor` (CC #161) to create a typed array view (e.g., `Float32Array`) into that arraybuffer without copying it.  However, to do so, the byte offset of the typed array view must be a multiple of the size of the type (e.g., `4` for `Float32Array`).  For homogenous arraybuffers where every element is the same type, this, of course, just works, but for heterogeneous arraybuffers, it does not work in the general case.

The glTF spec needs to guarantee this byte alignment and the converter needs to add the required padding, which will only have a trivial impact on the total size.

**JSON example**

Say we have this buffer view:
```json
"bufferView_393": {
  "buffer": "anim.bin",
  "byteLength": 36608,
  "byteOffset": 46728
}
```
which is then referenced from this animation parameter
```json
"translation": {
  "bufferView": "bufferView_393",
  "byteOffset": 312,
  "count": 78,
  "type": "FLOAT_VEC3"
}
```
In order to make a typed array for `translation`, we want to write code like
```javascript
new Float32Array(buffer, bufferView_393.byteOffset + translation.byteOffset, translation.count);
```
For this code to work, `bufferView_393.byteOffset + translation.byteOffset` must be division by `4` (which it is in this case).

pjcozzi:
@tparisi here is text for the draft 1.0 spec:

---

## Accessor Attribute Size

The following tables can be used to compute the size of an accessor's attribute type.

| `componentType` | Size in bytes |
|:-:|:-:|
| `5120` (BYTE) | 1 |
| `5121`(UNSIGNED_BYTE) | 1 |
| `5122` (SHORT) | 2 |
| `5123` (UNSIGNED_SHORT) | 2 |
| `5126` (FLOAT) | 4 |

| `type` | Number of components |
|:-:|:-:|
| `"SCALAR"` | 1 |
| `"VEC2"` | 2 |
| `"VEC3"` | 3 |
| `"VEC4"` | 4 |
| `"MAT2"` | 4 |
| `"MAT3"` | 9 |
| `"MAT4"` | 16 |

The size of an accessor's attribute type, in bytes, is `(size in bytes of the 'componentType') * (number of components defined by 'type')`.

For example:
```javascript
"accessor_1": {
[...]

PR #400, opened 2015-08-21
[binary_glTF] First Version

mlimper:
Hi all binary_glTF contributors and maintainers :-)

This first pull request contains our modifications on the binary file layout of binary glTF, as done for our SRC format. The changes can be previewed here:
https://github.com/mlimper/glTF/blob/header-preamble-naming/extensions/binary_glTF/README.md

Changes:
* renamed from CESIUM_binary_glTF to binary_glTF
* Added contributors and references to SRC resources
* Changed wording for file layout: preamble, header, body
* Added entry for header format in preamble, added note that currently only JSON is allowed (value '0')
* Made header directly follow the preamble (removed jsonOffset field), added some sentence on that

Of course, all of these points are open for discussion right now (maybe directly inside this pull request?).


extensions/binary_glTF/README.md
```diff
@@ -171,12 +184,17 @@ Using the 1200 12th Ave model (thanks to [Cube Cities](http://cubecities.com/)),
...
 
 ## Resources
 
 * Discussion - [#357](https://github.com/KhronosGroup/glTF/issues/357)
 * base64-encoded data in glTF - [#68](https://github.com/KhronosGroup/glTF/issues/68)
 * [Faster 3D Models with Binary glTF](http://cesiumjs.org/2015/06/01/Binary-glTF/) article on the Cesium blog
+* SRC project page (paper, background, basic writer) - [http://x3dom.org/src/](http://x3dom.org/src/)
 
 <a name="BenchData">
 * [1] Raw data for benchmarks using compression available in [BenchData](BenchData/README.md) supplemental.
```
pjcozzi: Can you tweak `BenchData/README.md` to say that the tests were done with `CESIUM_binary_glTF`, but the difference is insignificant?

extensions/binary_glTF/README.md
```diff
@@ -1,10 +1,12 @@
-# CESIUM_binary_glTF
+# binary_glTF
```
pjcozzi: We'll still want to leave `CESIUM_binary_glTF` around at least for a bit since we have _lots_ of Cesium users depending on it.  Our goal is to move completely to `binary_glTF` and then not try to ratify or promote `CESIUM_binary_glTF`.

pjcozzi:
> Made header directly follow the preamble (**removed jsonOffset field**), added some sentence on that

@tfili can you confirm that we are OK with removing the `jsonOffset` field?  We originally added it to ease the tool implementation, but I believe we were able to workaround needing it.  Right?

pjcozzi:
> Added entry for header format in preamble, added note that currently only JSON is allowed (value '0')

Do you think we still need this in the context of glTF?  What else would the header be?

pjcozzi:
Just those comments.  It is awesome how remarkably similar our approaches are!

tfili:
@pjcozzi We didn't still needed the offset and length in our extension. Making the json first should let us remove that.

pjcozzi:
> Made header directly follow the preamble (removed jsonOffset field), added some sentence on that

@mlimper do you have an elegant approach for assigning the glTF `bufferview.byteOffset` property with this approach since the byte offset depends on the size of the glTF, which varies depending on the number of characters needed for the `byteOffset` properties.  Do you add padding?

pjcozzi:
@mlimper we're also not sure that we should call the glTF section the header. AFAIK a file header usually comes first.  Is preamble/header widely used?  Can you point me to some examples?

mlimper:
> Added entry for header format in preamble, added note that currently only JSON is allowed (value '0')
> Do you think we still need this in the context of glTF? What else would the header be?

In the paper we mentioned binary JSON (http://bsonspec.org/) and protocol buffers.
We felt that having something different than JSON was not necessary in the first working implementation. But there have been requests and discussions, from the very beginning, to keep the possibility of having different, more efficient header encodings. Therefore, I think it would make sense to have this possibility, and I would vote for allowing only JSON in the first version.

mlimper:
> @mlimper we're also not sure that we should call the glTF section the header. AFAIK a file header usually comes first. Is preamble/header widely used? Can you point me to some examples?

There is something called header in BMP, for example:
https://en.wikipedia.org/wiki/BMP_file_format#Bitmap_file_header

Wikipedia's definition:
https://en.wikipedia.org/wiki/Header_(computing)

The word "preamble" for the "pre-header" was proposed by Leonard in his SRC spec draft, and I liked it, so I just used it in this initial pull request.

We are totally open for other proposals. Initially, we just called the glTF-like information "header" as we thought about it as a bit of additional information about the actual data, which was called the "body". We were very focused on transporting raw data (buffers for mesh geometry, plus textures) and the JSON really just served as a description of the data. We have different JSON formats (or X3D) to describe the scene data.
This is a conceptual difference between glTF and SRC: With glTF, it is possible to describe a whole scene, including lighting, node hierarchy, etc. inside the JSON - with this in mind, calling the JSON part just "header" really sounds a bit weird, I agree.

In the CESIUM_binary_glTF proposal, what we called "preamble" is called "header":
https://github.com/KhronosGroup/glTF/blob/new-extensions/extensions/CESIUM_binary_glTF/layout.png

Should I just change my fork to go back to this original idea? Then the question would be how the structured glTF data ("JSON part") is called. I don't like calling it JSON because I would like to keep the possibility to use different (text or binary) encodings later on. Any ideas?

extensions/binary_glTF/README.md
mlimper: I changed it back for now.
Is there a schedule for moving to binary_glTF? Is this the deadline end of september?

extensions/binary_glTF/README.md
mlimper: Thanks for the hint, I added a remark at the end of the "Experimental Results" section:
https://github.com/mlimper/glTF/tree/header-preamble-naming/extensions/CESIUM_binary_glTF#SRCPaper

pjcozzi:
>>> Added entry for header format in preamble, added note that currently only JSON is allowed (value '0')

>>Do you think we still need this in the context of glTF? What else would the header be?

> In the paper we mentioned binary JSON (http://bsonspec.org/) and protocol buffers.

> We felt that having something different than JSON was not necessary in the first working implementation. But there have been requests and discussions, from the very beginning, to keep the possibility of having different, more efficient header encodings. Therefore, I think it would make sense to have this possibility, and I would vote for allowing only JSON in the first version.

Ah, I suppose [msgpack](http://msgpack.org/index.html) is another option; however, I would hesitate to use anything that is decoded in JavaScript instead of by the browser.

How would this work when adding a new format?  Wouldn't we have to release a new binary glTF extension with a new version?  Instead of overbuilding this extension, why not wait until we have an implementation that shows the benefit of multiple formats, and then introduce it with a new version of the binary glTF extension?

pjcozzi:
> Should I just change my fork to go back to this original idea? Then the question would be how the structured glTF data ("JSON part") is called. I don't like calling it JSON because I would like to keep the possibility to use different (text or binary) encodings later on. Any ideas?

Open to suggestions, but perhaps `[header][scene][body]`?

extensions/binary_glTF/README.md
pjcozzi: Instead of changing it back, I think we should just leave the `CESIUM_binary_glTF` extension as is, and then copy and paste it to create the `binary_glTF` extension in a new directory.

I plan to move Cesium to `binary_glTF` as fast as possible - end of September or end of October.  Leaving the Cesium-specific extension around for now should have no impact on our official extension.

mlimper:
> How would this work when adding a new format? Wouldn't we have to release a new binary glTF extension with a new version? Instead of overbuilding this extension, why not wait until we have an implementation that shows the benefit of multiple formats, and then introduce it with a new version of the binary glTF extension?

I think the main argument for introducing the format word with the very first version of binary_glTF would be that old .bgltf data, and respective parsers, can remain unchanged. Backwards compatibility with older parsers is probably important.

The alternative would be that we define "Binary glTF version 1.0 uses JSON and has no word for specifying a different header encoding", but if we define such a word later on, parsers have to adapt to a different header size for this.

Clearly both is possible, I would personally vote for an explicit "header format" word right from the beginning.

mlimper:
(ok, there is now a separate directory, and cesium_binary_gltf should be in the original state again:
https://github.com/mlimper/glTF/blob/header-preamble-naming/extensions/binary_glTF/README.md
https://github.com/mlimper/glTF/blob/header-preamble-naming/extensions/CESIUM_binary_glTF/README.md )

pjcozzi:
> I think the main argument for introducing the format word with the very first version of binary_glTF would be that old .bgltf data, and respective parsers, can remain unchanged. Backwards compatibility with older parsers is probably important.

Ah, I see, so it would allow current Binary glTF content to be forward compatible with a future Binary glTF version assuming we don't change the header for some other reason.

OK with me.

Is there anything else to discuss for this?  Have you implemented it?

mlimper:
No, we haven't implemented the current proposal right now. It should not be too much work to adjust our existing implementation. It seems for now there are no open points. I'll use "header-scene-body" for the proposal now, and we'll have a "scene format" entry inside the header.

We will now adjust the implementation. We can provide a binary_glTF example file along with this pull request - the ultimate goal would then probably be that we can export this and you can load it, and vice versa ;-)

pjcozzi:
All sounds good.  @tfili and I will start work on the Cesium loader ASAP.

mlimper:
There is one more issue I'd like to discuss, now that we agreed that the glTF scene description always comes first:
Could we treat all buffer offsets within the glTF scene description relative to the beginning of the binary body?

With the header always coming before the binary body, it seems easier to simply split the initial arraybuffer response into three subsequent parts, represented as three arraybuffer views:
* header (fixed size)
* glTF scene (size known from header) 
* body (remaining size, also known from header)

Offsets of buffers within the body would always be specified relative to the beginning of the binary body. This would allow to write the scene without having to know how large the scene itself will be.

If OK like this, we'll update the js example code within README.md accordingly.

mlimper:
(... this would also mean that the special buffer entitled 'binary_glTF' does not point to the whole file, but just to the binary body)

pjcozzi:
> Offsets of buffers within the body would always be specified relative to the beginning of the binary body. This would allow to write the scene without having to know how large the scene itself will be.

I agree, I think this would really simplify tools that write binary glTF.  @tfili can you confirm?

> (... this would also mean that the special buffer entitled 'binary_glTF' does not point to the whole file, but just to the binary body)

Just to confirm, this mean that the `"binary_glTF"` buffer points to the binary body and everything else remains unchanged.  `bufferViews` that point to this buffer implicitly have their offsets now relative to binary body, and, as usual, there could still be other `buffer`s (and `bufferView`s) pointing to external data?

pjcozzi:
I renamed the extension to `EXT_binary_glTF` (601a329559be9953e453fc2dc4b6130034fb04d8) so it now follows the official glTF extension process (#404).

I didn't do it yet, but we'll also have to rename the `"extensions"` sub-property:
```javascript
"extensionsUsed" : [
    "binary_glTF"
]
...
"a_shader" : {
    "extensions" : {
        "binary_glTF" : {
            "bufferview" : ...
        }
    }
}
```
to
```javascript
"extensionsUsed" : [
    "EXT_binary_glTF"
]
...
"a_shader" : {
    "extensions" : {
        "EXT_binary_glTF" : {
            "bufferview" : ...
        }
    }
}
```
This will require another rename after ratification:
```javascript
"extensionsUsed" : [
    "KHR_binary_glTF"
]
...
"a_shader" : {
    "extensions" : {
        "KHR_binary_glTF" : {
            "bufferview" : ...
        }
    }
}
```
Although this will require changes to converters and runtimes, I think this is the most solid approach as opposed to, for example, allowing:
```javascript
"a_shader" : {
    "extensions" : {
        "binary_glTF" : {
            "bufferview" : ...
        }
    }
}
```
Which could then lead to this awkward case (in general, for extensions):
```javascript
"extensionsUsed" : [
    "KHR_binary_glTF",
    "EXT_binary_glTF"
]
...
"a_shader" : {
    "extensions" : {
        "binary_glTF" : {
            // EXT or KHR?  This would be especially problematic if KHR changed the EXT extension. 
        }
    }
}
```
Any thoughts before I do the rename to `EXT_`?

tfili:
@pjcozzi Yeah, that would definitely simplify creation. As long as you make the .bin file follow the glTF Scene you wouldn't have to adjust the offsets of existing buffers and bufferviews. The old way we had to adjust these regardless.

pjcozzi:
In order to finalize this extension so it can enter ratification, I believe we have these remaining tasks:

* [ ] Create the JSON schema for the extension
* [ ] Figure out "[preamble][header][binary body]" naming (this only affects the spec wording).  How is "[header][scene][binary body]?"
* [ ] Update spec so the `binary_glTF` buffer points to the binary body, not the entire Binary glTF
* [ ] Add `EXT_` prefix to `extensions` sub-property
* [ ] Can we just claim the `model/vnd.gltf.binary` MIME type?  Or is there a MIME type registry?

@mlimper is there anything else?

tparisi:
My $.02

* I like [header][scene][binary] terminology

* Don't forget, we should not be using "vnd" in our MIME type. I changed the main one to model/gltf+json, I think we should go with model/gltf+binary ? I'm pretty dumb wrt MIME naming; let's make an issue to look into this. It's OK if we can't solve it before submitting to KHR, but we should get on top of this now so that we can finalize it before ratification.

* I have an issue with the .bgltf extension and would like to discuss an alternative.

mlimper:
> Just to confirm, this mean that the "binary_glTF" buffer points to the binary body and everything else remains unchanged. bufferViews that point to this buffer implicitly have their offsets now relative to binary body, and, as usual, there could still be other buffers (and bufferViews) pointing to external data?

Yes, that's what I also had in mind.

I agree renaming to EXT_ makes sense.

> @mlimper is there anything else?
Except for an example I currently don't see anything... I'm currently re-writing README.md, after this most of the TODOs should be done. I'll commit this in a few hours for review.

> I have an issue with the .bgltf extension and would like to discuss an alternative.

I agree, having five characters for a filename extension seems a bit uncommon.
Maybe something like ".bgt", or ".bgl"?

tparisi:
Maybe .glb for the extension?

tparisi:
@pjcozzi I am leaning to .glb - should start with GL the way glTF does.

Or .gltb but may as well just be three letters.

pjcozzi:
@shunter any chance you can advise on the MIME type naming?  @tparisi said:

> Don't forget, we should not be using "vnd" in our MIME type. I changed the main one to model/gltf+json, I think we should go with model/gltf+binary ? I'm pretty dumb wrt MIME naming; let's make an issue to look into this. It's OK if we can't solve it before submitting to KHR, but we should get on top of this now so that we can finalize it before ratification.

pjcozzi:
> I'm currently re-writing README.md, after this most of the TODOs should be done. I'll commit this in a few hours for review.

Great, I am finishing the schema for the extension now.  Please let me know when the spec is ready for final review and I will look at it right away.

pjcozzi:
For the extension, `.glb` is OK with me:
```
.gltf // stock glTF
.glb  // binary glTF
```

pjcozzi:
@mlimper I added the schema in dd6fcf99c730e04b2b49875ffc4d8c832424a720.  Please link to it from the README.md.

mlimper:
Here's a first updated version of the README.md for review.

One TODO / question:
Should there be a more sophisticated way of specifying the version, such as major-minor-maintenance, or major-minor? Maybe major-minor as two 16 bit values?
The current proposal just uses a uint32 to specify a single unsigned integer.

mlimper:
.. ah, and now I also updated the mimetype to `model/gltf+binary`, but of course this can be changed after we got more feedback

mlimper:
... one more thing, do you have an up-to-date example file (glTF or binary_glTF) that uses the upcoming glTF 1.0 schema? Just looking for a very simple example (collada duck, stanford bunny or so) to ease my writer implementation.

pjcozzi:
@mlimper looks great.  I made minor edits in ddf045260b50db23ee879abd90249327cb0d4f82 and 31552579db8a4c04041291884e5a3f62b0edc0bc.  I'm going to do one final read shortly.

> Should there be a more sophisticated way of specifying the version, such as major-minor-maintenance, or major-minor? Maybe major-minor as two 16 bit values?
The current proposal just uses a uint32 to specify a single unsigned integer.

I think it is total overkill for this.

> one more thing, do you have an up-to-date example file (glTF or binary_glTF) that uses the upcoming glTF 1.0 schema? Just looking for a very simple example (collada duck, stanford bunny or so) to ease my writer implementation.

Not yet, but I plan to write a 0.8 to 1.0 spec guide (#406).  The best thing to look at now is the [diff of the schemas](https://github.com/KhronosGroup/glTF/compare/spec-1.0#diff-1987f7d964a53ee1f4da85d71580f5f0R3).

However, there is more [very minor changes](https://github.com/KhronosGroup/glTF/labels/1.0) to come.

pjcozzi:
I made some more minor edits in 69b48c49cd37316fba34ae9e4354134ac1934b06.

I am going to merge this so we can reduce the number of branches we have as I finalize the writeup for extensions (#404) in the `spec-extensions` branch.

In the near term, future pull requests for this extension can be opened into that branch, and then into `spec-1.0` once we merge `spec-extensions` - hopefully tonight.

Thanks for your contribution, @mlimper.  This is better than the initial CESIUM_binary_glTF extension.  I'm looking forward to working with you on more extensions with you.

pjcozzi:
@mlimper we usually delete branches after we merge them, but I'll leave it up to you to delete your `header-preamble-naming` branch.

shunter:
> @shunter any chance you can advise on the MIME type naming?

The proposed new types seem logical to me.  They would need to be registered with IANA, however.  `vnd` is not necessarily a problem - e.g. Collada uses `model/vnd.collada+xml` as the official type.

pjcozzi:
Thanks @shunter.  @mlimper do you or any of your collaborators have experience registering MIME types?

pjcozzi:
@mlimper @tfili I added the following in 112a232b6ec665560bc45c79093e5493ff28db6e:

> The start of `body` is 4-byte aligned to ease its use with JavaScript Typed Arrays.

This was in the original CESIUM_binary_glTF extension.  I think we still want it.  See #167.

mlimper:
> @mlimper @tfili I added the following in 112a232:
> 
> The start of body is 4-byte aligned to ease its use with JavaScript Typed Arrays.
> This was in the original CESIUM_binary_glTF extension. I think we still want it. See #167.

Oh yeah, sorry, after checking #167 and wondering why we didn't have this problem, I figured out in our code we only create and use UInt8 TypedArray views, which we then directly pass to the GPU. We never used Float32Arrays or something like that in JS when dealing with GPU buffers.
For the general case, it makes sense to have the body aligned to 4 byte boundaries.

mlimper:
> Thanks @shunter. @mlimper do you or any of your collaborators have experience registering MIME types?

Me not, but I'll ask around

pjcozzi:
All good, thanks.

pjcozzi:
@mlimper I added a clarification about the 4-byte alignment in 550d14f9da80a285eff4acc1034bad61ba6b31df based on feedback from @tfili, who is currently implementing it.

PR #416, opened 2015-09-17
Make JSON examples consistent

pjcozzi:
@tparisi this is a PR into `spec-1.0`.

Using "```javascript" allows us to put comments in the examples, which is not valid JSON, but is handy.

This also fixes white spaces.


pjcozzi:
@tparisi can you merge this?

Issue #84, opened 2013-04-28
fallbacks for techniques


RemiArnaud:
According to the space, the following technique names can be provided to provide more guidance:

Lambert
Phong
Blinn-Phong

1) missing from that list: constant

2) the converter should follow the spec, specifically for duck.json, 'Blinn-Phong' should be used instead of 'technique1'

3) regarding the spec, this is not a name, this is an id.

pjcozzi:
As far as I can tell, this is duplicate with #361 (let me know if someone thinks otherwise), and we all agree with @tparisi's comment above:

> Here is my view: a valid glTF material could have any of the following.

> 1. No shader code supplied - just technique info and param values: requires
application to have an implementation of the technique;
2. Shader code is  supplied with the content, along with technique info;
the application can use the supplied shader code or its own implementation
if it has one;
3. Shader only is supplied, no common technique information; the
application must use the shader code supplied.

> Converters and exporters can provide options for what to output.

RemiArnaud:
What's the size difference between 2 and 3 ?
is it worth the added complexity? (I'd remove option 3 to simplify design/loader)

pjcozzi:
Removing 3 will depend on how #361 comes out.  If the materials are described in an extension, we wouldn't be able to remove 3 since then it would mean that glTF requires the extension.  I also don't think we should remove 3 since a model could contain shaders that can't be represented by the fixed-function materials, and the converter would have the burden of creating some type of material even if the engine doesn't use it, which is the case in Cesium.

RemiArnaud:
3) prevent simple applications to take a peak at the model. 

I was not proposing that the shader provided and the information provided in the technique need to provide same result. Its in fact not possible in the general case, so why bother?

pjcozzi:
> 3) prevent simple applications to take a peak at the model.

Not sure what you mean.

> I was not proposing that the shader provided and the information provided in the technique need to provide same result.

It would still be burdensome to have to generate anything at all when all the runtime wants is the shader.

RemiArnaud:
ok, so what about just 1) then ?

pjcozzi:
Just 3 prevents engines from generating their own shaders at runtime like Three.js does, see #361.

RemiArnaud:
I'm confused. 

If you want engines to generate their own sharers, then we're back to my original post, do not offer 3) always provide 1) or 2)

pjcozzi:
[...]

PR #419, opened 2015-09-30
Update material and technique schema

pjcozzi:
As discussed in https://github.com/KhronosGroup/glTF/issues/84#issuecomment-142185519


pjcozzi:
@tparisi for your review.

Issue #180, opened 2013-11-09
More models


pjcozzi:
Any chance we can add a few more freely-available models to the [models](https://github.com/KhronosGroup/glTF/tree/master/model) directory in the repo?  We could really use the examples for learning.

Specifically:
* Animation
* Skinning
* Morphs
* Compression
* Multi-pass

With these I think we would cover all the glTF features.

pjcozzi:
There are a few more models in #395.

pjcozzi:
Added a few more test models in #422.

pjcozzi:
We have a reasonable set of models in the `sampleModels` directory in the `spec-1.0` branch.  In #421, we will provide multiple glTF versions of each model: separate resources, embedded resources, KHR_binary_glTF, and KHR_materials_common.

PR #422, opened 2015-10-01
Sample models

pjcozzi:
@tparisi this is a pull request into `spec-1.0` for your review.

* Rename `model` to `sampleModels`
* Adds three new test models from Cesium:
   * `boxTextured` - textures
   * `CesiumMilkTruck` - keyframe animation
   * `Cesium_Man` - skinning


pjcozzi:
For #180.

PR #425, opened 2015-10-07
Changed rotations to quaternions.

tfili:


pjcozzi:
Thanks @tfili.  Can we cherry pick any other changes we need from [dev-0.9](https://github.com/KhronosGroup/glTF/compare/dev-0.9) to `dev-1.0` so we can delete that branch?

Issue #37, opened 2013-03-22
JSON refer other JSON?


pjcozzi:
In our quest to balance simplicity and flexibility, I'd like to open the discussion for if glTF JSON should be able to reference other JSON.  This would be useful for storing large node hierarchies across several files, where only a few of them may need to be loaded based on visibility, LOD, etc.  It would also be useful for sharing parts of the JSON, e.g., materials, among multiple assets, which can reduce overall bandwidth.

I'm happy to draft up some example JSON (probably after GDC).

My feeling on this is yes because of the above use cases and that it matches the flexibility with everything else in glTF - buffers, images, and shaders.

I'm interested in other thoughts...

pjcozzi:
@mlimper let's use this issue to discuss your approach to this with SRC and how we could use it in glTF or an extension (it should probably be core spec though).

mlimper:
Within SRC, there is a two-level addressing schema for meshes / mesh data and textures:

* A hash symbol is used to address a mesh or texture: http://foo.bar.myDataFile.src#myMesh
* A dot symbol is used to address a specific attribute of a mesh: http://foo.bar.myDataFile.src#myMesh.colors

(also see Figure 7 in the paper, for example: http://x3dom.org/src/files/Web3d2014_SRC.pdf)

This assumes that meshes and textures are first-class citizens of the SRC file, which share the same scope (and hence cannot have similar names). For the original idea of using SRC to externalize X3D Shape Data, this was sufficient - but this approach is probably not general enough for glTF, I suppose.

pjcozzi:
Thanks @mlimper.  I will need to implement this in Cesium before I am comfortable suggesting it for the core glTF spec.  I'm not sure I can do that in time for the draft 1.0 spec, but we could propose it as an extension, and then put it in the 1.1 spec.

I agree that the SRC approach is not general enough, but I think it generalizes easy enough.  Here's an example (assuming an `external_glTF` extension):

**commonMaterials.gltf**

```javascript
{
    "materials": {
        "glass" : {
            // ...
        }
    }    
}
```

**model.gltf**

```javascript
{
    "extensions": {
        "external_glTF" : {
            "commonMaterials" : "commonMaterials.gltf"
        }
    },
    "meshes" : {
        "a-geometry" : {
            "primitives" : [{
                "material" : "commonMaterials#glass"
            }]
        }
    }
}
```

Basically, any referenced property name with an `#` identifies an object in an external glTF file.

[...]

PR #426, opened 2015-10-08
Web3D SRC Extension

mlimper:
This is a pull request containing the proposed extension WEB3D_SRC.

It depends on EXT_binary_glTF and EXT_quantized_attributes.

This extension mainly provides the possibility to use (binary) glTF assets within other structured 3D scene descriptions, such as X3D or VRML content. As such, it does not introduce _any_ changes to the glTf schema (which is uncommon for an extension). However, it proposes a referencing scheme that allows to refer to binary glTF shape content (`meshes`, `materials`, `textures`) from the outside, and it defines some rules how binary glTF content for shapes  should be treated when just a URL is specified.

That said, it is an open question whether this should really be seen as an extension, since it does not introduce any changes to the core glTF spec.
Instead, this proposal could also become part of the core glTF specification, for example, providing a referencing scheme and building a bridge to the VRML / X3D world.


mlimper:
Preview:
https://github.com/mlimper/glTF/tree/WEB3D_SRC_Extension/extensions/Vendor/WEB3D_SRC

pjcozzi:
Thanks @mlimper.  I am at a conference today (actually two conferences) and will look at this tomorrow.

> That said, it is an open question whether this should really be seen as an extension, since it does not introduce any changes to the core glTF spec.
Instead, this proposal could also become part of the core glTF specification, for example, providing a referencing scheme and building a bridge to the VRML / X3D world.

Interesting...will think about this.

pjcozzi:
This references glTF lights in a few places.  I suggest removing the references since lights are in an extension now (#84).

pjcozzi:
EXT_binary_glTF should be KHR_binary_glTF.

pjcozzi:
In the **Addressing Scheme for SRC content** section, I think it is worthwhile to explicitly mention that all these types (meshes, materials, and textures - and all top-level glTF objects) share one namespace so, for example, there will never be a mesh and material with the same name.

pjcozzi:
```json
"extensions" : {},
"extras" : {}   
```
Remove these from the example to keep them clean.

pjcozzi:
I suggest that we
* Add a short non-normative section to the glTF 1.0 spec that recommends using `#` when referencing a top-level glTF object (mesh, material, etc.) by its id and the glTF's file, e.g., `model.gltf#material_id`.
* Do not include WEB3D_SRC as an extension.  Instead, the extension - with some wording tweaks - could be a standalone spec for SRC.
* Keep this all in mind when we finish the extension in #37.

CC @tparisi

mlimper:
Thanks for the feedback. I have updated the proposal.

@DrX3D This could be a kind of spec describing the integration of (binary) glTF content into X3D, entitled SRC. Since you have already been working on this topic, let me add you to this conversation.

It will be interesting to figure out together where we finally put the information from this proposal - glTF spec, standalone document hosted / provided together with glTF spec, standalone document somewhere else, ...?

brutzman:
SRC providing common capabilities with binary glTF keeps looking better and better.  Thanks for all of the great efforts on this work.

SRC will not only be critical for the forthcoming X3D Efficient Binary Encoding (EBE) but we expect that it will help all X3D file encodings and language APIs.  Such shape interoperability with binary glTF can become a critical new capability that hopefully can be maximized for use by all Web authors.

http://www.web3d.org/specifications
http://www.web3d.org/specifications/X3dSpecificationRelationships.png
http://www.web3d.org/specifications/X3dSpecificationRelationships.pdf

mlimper:
Maybe the content proposed within this pull request would better fit into some document on the "X3D side" of this connection, it could then also contain the relationship diagram, for example.

The file extension will then probably be ".glb", the regular binary glTF extension... a mimetype should also be on the way, at least it is planned, I think.

@pjcozzi What do you think?

pjcozzi:
@mlimper I agree.

I added the following to the glTF spec:

> It is recommended that container formats reference a glTF object by ID by concatenating the asset's path, `#`, and the ID, e.g., `asset.gltf#material_id`.

a09353f43ab99f062830a3c0553a38b88619941b

mlimper:
Thanks - looks fine!

PR #431, opened 2015-10-11
Move lights schema to extension

pjcozzi:
Pull request into `spec-1.0`.

KHR_common_materials still needs more schema work to finish lights and add materials.  This PR's goal is to remove the lights schema from the main spec.

This also moves the `node` property for a technique's parameter to the extension.  I don't have a use case for putting this in the main spec (it is used to transform lights).  If we have a good use case, it can go in the main spec.

I also suggest that the extension explicitly state that ambient lights are never referenced from a `node` since they do not require an explicit transform.


tparisi:
Roger that.

tparisi:
@pjcozzi @tfili hey guys rethinking this thing about `node` - I think we still want the semantic in there. Shaders might want to use the modelview info for a node, not necessarily for a light. Like in the current example in the spec (search for light0transform), that is done using a shader. Exported shaders might want that info. IMO it's also cleaner to just have this info in a base spec rather than having an extension define a semantic.

pjcozzi:
Just to clear up terminology, this is not a semantic.  We are talking about the `node` property for a parameter, which indicates that the parameter should use the node's transform.  Right?

I think you are making the case that a pre-generated shader might want a node's transform since the shader could be using the node for lighting even though the asset doesn't explicitly have any lights, is this correct?

tparisi:
Yes - my bad - this is for a parameter. And that is the case I am making. The current example in the spec is used by a user-defined shader, does not depend on the materials extension. So I think this capability should be in the base spec, the language should stand, and you should revert that part of the schema. OK?

pjcozzi:
Yes, I agree.  This also implies that empty nodes (e.g., no camera, skin, or meshes) can be useful since they may provide inputs to parameters.  We already have the node schema written that way.

I will update the spec/schema in the `spec-1.0` branch.  There may be minor merge conflicts when the `spec-1.0-copyedit branch comes in`, but I will fix them.

CC @slchow
CC @tfili

tparisi:
huzzah - I await copy edit.

PR #433, opened 2015-10-12
WEB3D_streaming Extension

mlimper:
This is a first version of the extension WEB3D_streaming, which introduces `bufferChunks` as an additional concept to facilitate streaming of mesh data with binary glTF (requiring the `KHR_binary_glTF` extension).

Small remark about textures not being mentioned here:
In our original paper, we also proposed a concept to progressively deliver MIP levels of textures within several chunks. We believed that this could be useful when using compressed textures, since you'll usually deliver pre-computed, compress MIP levels. It introduced an additional layer between texture and chunks, called "TextureImage" (i.e., one MIP level). However, for the sake of simplicity, and to keep things well-structured, this was left out for now, it could be discussed and added as part of a different extension.


mlimper:
sorry, wrong target branch

PR #438, opened 2015-10-15
Add node back to technique parameter

pjcozzi:
See https://github.com/KhronosGroup/glTF/tree/parameter-node/specification#reference-technique.parameters

As discussed in https://github.com/KhronosGroup/glTF/pull/431#issuecomment-148096066


pjcozzi:
Merging into `spec-1.0`.

PR #452, opened 2015-10-18
Converter changes for v1.0-draft spec

tfili:


pjcozzi:
Thanks @tfili!

@tparisi this includes your work-in-progress for Three.js loader (which I think is fine since it is still better than the pre-0.8 loader in master now).  Can you please verify nothing broke with the merges?

Issue #448, opened 2015-10-18
Sample glTF models with KHR_materials_common


pjcozzi:

pjcozzi:
Now there is a `glTF-MaterialsCommon` subdirectory for each sample model.

PR #465, opened 2015-10-28
Move sample models into subdirectories

pjcozzi:
Each model now has three subdirectories: `collada`, `gltf`, and `gltf-embedded`.  Now, we can cleanly add new subdirectories for binary glTF (#464) and the materials extension (#448).

@tfili is [convertAll.sh](https://github.com/KhronosGroup/glTF/blob/master/sampleModels/convertAll.sh) useful?  Should we update or remove it?


tfili:
@pjcozzi Not currently. It won't run on Windows right now. Shouldn't be too hard to modify. Looks like we can use `uname -s` to detect gitbash on Windows.

PR #468, opened 2015-10-29
Added models that use the KHR_materials_common extension.

tfili:
@tparisi This has all the models converted to use the materials_common extension.


pjcozzi:
Fixes #448.

PR #471, opened 2015-11-02
Changed asset.version to be a string as per the 1.0 spec.

tfili:
The converter was never fixed to make this a string.


pjcozzi:
Really good catch, thanks @tfili.

PR #480, opened 2015-11-20
Added simple rigged model

lilleyse:
I created a very basic rigged model that has a simple animation. This is a simpler alternative to the CesiumMan.

I created the model in Blender and converted it with the Collada to glTF converter.


pjcozzi:
Awesome, thanks @lilleyse!

PR #482, opened 2015-11-22
Added a gltf_uriType tag to each URI indicating which uri type is expected

MattMcMullan:


pjcozzi:
Can you please put this new property below (or above) the `gltf_detailedDescription` property?  I want to keep all the official JSON schema properties together, then all the `gltf_` custom properties together at the bottom of each object.

MattMcMullan:
Sure, that makes sense.

pjcozzi:
Thanks!

PR #483, opened 2015-11-25
Add a C# reference loader

MattMcMullan:
The state of the loader is somewhat basic, but quite functional. It can read gltf models other than glb. There is a generator program that reads the schema files and generates corresponding C# classes for each type defined.These classes are then used to load the model. Custom converters are added for all types that enforce the constraints listed in the schema. Currently only array min/max are unenforced at load time.  Contents of URIs and data URIs are decoded into their proper types (byte[] for raw data, Bitmap for images, or string for text/shaders).

At this time, the loader is useful for many applications, but the following work has yet to be completed
 - Reference renderer
 - glb loading
 - Array min/max enforcing at load time
 - Proper extension support
 - URIs representing web sources (http/ftp/etc)


pjcozzi:
Wow, this is awesome!  Thanks so much @MattMcMullan and @yukicode!

A few questions:
* Do you think this should be in a separate repo?  We, of course, would still link to it from the README.md in this repo as a the reference C# loader.  Long term, I have even been thinking about moving COLLADA2GLTF into a separate repo and keeping this repo spec only.  What do you think?
* Can you add a LICENSE.md to this subdirectory?  The license is up to you.
* It would help users if the README.md had a bit more explicit instructions and an example if possible.
* Just curious, do you plan to write a C# reference renderer in the future? With Direct3D? With [OpenGL](https://www.opengl.org/wiki/Language_bindings#C.23)?

MattMcMullan:
You're welcome, though I can assure you that our motivations are primarily greedy. We intend to use this in our primary side-project.

A few answers:

 - The reason that we put it in this repo is because it relies directly on the .spec files.  Putting it in a separate repo means adding an additional layer of complexity to making it work.  Other than that, I'm generally in favor of having related things together.  If you decide that you don't agree, we'll leave it in the current fork and keep pulling in changes from here.
 - I'll add a new copy of the BSD license here.
 - Sure, I'll expand upon the readme. For now, I'll probably just point to gltfLoaderUnitTests/SchemaTest.cs for an example. Maybe I'll add a test there that just loads one schema file too. It's really as simple as Gltf model = Interface.LoadModel("PathToFile.gltf").
 - I plan on doing everything listed at the end of the comment, roughly in that order.  We're planning on OpenGL for the reference renderer, since all of my experience is with OpenGL and Direct3D would seem rather silly here anyway.  Though most of my OpenGL work has been in C++, I have become familiar with the OpenTK C# bindings for OpenGL recently for the aforementioned primary side-project.

When the reference renderer is done, it's my intent to use it as a bit of a playground to test out some ideas for adding tessellation to the standard before submitting a proposal.  I'd also like to get your thoughts on collision primitives, as they seem to be within the defined scope of the project, applicable to web-applications, and not yet in the standard.

pjcozzi:
> The reason that we put it in this repo is because it relies directly on the .spec files.

It would be a reasonable setup to include this repo as a submodule in a dedicated repo or even as a third-party directory since the schema should be pretty static.  This would also make it easier to manage issues, e.g., it keeps the loader issues separate from spec discussion.

If you really want it in this repo, we can go for it with the caveat that we might end up moving it later if we also move COLLADA2GLTF.

Let me know your preference.

> I plan on doing everything listed at the end of the comment, roughly in that order...

Awesome!  It is several years old at this point, but here is a bunch of C# and OpenTK code I worked on: https://github.com/virtualglobebook/OpenGlobe

> When the reference renderer is done, it's my intent to use it as a bit of a playground to test out some ideas for adding tessellation to the standard before submitting a proposal.

Great, here's the info on extensions if you haven't seen it yet: https://github.com/KhronosGroup/glTF/blob/master/extensions/README.md

> I'd also like to get your thoughts on collision primitives, as they seem to be within the defined scope of the project, applicable to web-applications, and not yet in the standard.

Yes, this would be a great extension!

We want to use this in Cesium too, e.g., add a bounding sphere or box to a node and then use it for collision detection and response.  We hard-coded a bounding box for NORAD Track Santa: https://cesiumjs.org/2013/12/23/Building-A-WebGL-Santa-with-Cesium-and-glTF/

MattMcMullan:
> Let me know your preference.

Those do sound like valid ways to do things. The presence of COLLADA2GLTF and the ThreeJS loader in the generic "loaders" directory seem to tell me that this is not yet the canonical way of doing things.  The loaders directory in particular screams out for contributions. Personally, I find that sharing a repository helps to keep the project as a whole more cohesive and cooperative. I would rather keep with the currently set convention of keeping things in-repo until we make a unified decision to move all of the components out. If that happens and things like COLLADA2GLTF and the ThreeJS loader are moving, I'll certainly dig up your suggestions on how to use the spec files without sharing a repository and make something happen.

> Awesome! It is several years old at this point, but here is a bunch of C# and OpenTK code I worked on

Thanks for the sample, but I'm afraid that it's already coming too late.  OpenTK isn't too complicated and I've already submitted a pull request to improve it and add a more modern AZDO-type OpenGL 4 example. Sadly, they're much slower at responding than you are.

> Great, here's the info on extensions if you haven't seen it yet:

Thanks! I hadn't. I'm still unclear on how I would go about adding enumeration values to existing properties though.  For example, I would probably need to add 36487 and 36488 to the shader type enum to represent tessellation evaluation and control shaders.

> Yes, this would be a great extension!

I'm glad that you support the idea :)

pjcozzi:
>  I would rather keep with the currently set convention of keeping things in-repo until we make a unified decision to move all of the components out.

OK.

> I'm still unclear on how I would go about adding enumeration values to existing properties though. For example, I would probably need to add 36487 and 36488 to the shader type enum to represent tessellation evaluation and control shaders.

We could start with the collision extension, which should plug in pretty easily. :)

As for enums, extensions can add properties to objects, e.g.,
```javascript
"a_shader" : {
    "extensions" : {
        "EXT_tessellation_shader" : {
            "type" : // ...
        }
    }
}
```
What's ugly about this approach is the shader object still needs to define a `type` even though it won't be used because extensions are not allowed to make required properties optional.

In a future version of glTF - or future OpenGL profile - the new enum values would become part of the allowed values for `type`.  We could also think through if an extension is allowed to add to the list of allowed values, which may be OK.

Open a new issue when you want to discuss this further.

Issue #484, opened 2015-11-25
Incorrect Skeletons in Cesium Man Sample Model


tparisi:
The `skeletons` property in Cesium Man looks wrong. It has the whole array of joints in it instead of just the root of the skeleton. I had to add logic to the Three.js loader to only use the first element of the array instead of adding every skeleton, otherwise it did not animate correctly (unless of course I've got a bug somewhere which is totally possible). Here's the content:

```
            "skeletons": [
                "locator047Node",
                "locator051Node",
                "locator055Node",
                "locator059Node",
                "locator063Node",
                "locator071Node",
                "locator074Node",
                "locator079Node",
                "locator082Node",
                "locator087Node",
                "locator090Node",
                "locator103Node",
                "locator106Node",
                "locator111Node",
                "locator114Node",
                "locator119Node",
                "locator122Node",
                "locator135Node",
                "locator138Node"
            ],
```

tparisi:
@pjcozzi @tfili

tparisi:
@lilleyse can you also reconvert RiggedFigure? Seems to have the same `skeletons` issue where the property has all the joints instead of the just the root node...

pjcozzi:
Can we close this?

pjcozzi:
I believe we can close this.  Please reopen if not.

PR #485, opened 2015-11-25
Fixed CesiumMan, monster, and brainsteam

lilleyse:
I made some tweaks to these files in Blender and then re-exported them. I verified that they work in Cesium now.


lilleyse:
Related to: https://github.com/AnalyticalGraphicsInc/cesium/issues/3244

tparisi:
@lilleyse good work! monster looks right but Brainsteam doesn't. The skinning seems to work but the model itself appears to be missing several parts. Does your mode look like this?
brainsteamscreenshot

pjcozzi:
CC #484.

pjcozzi:
Brainsteam works OK in Cesium.

(Sorry about the resolution).

image

tparisi:
Gotcha - so it's on my end. Will open a new issue and investigate.

PR #493, opened 2015-11-29
Add the ability to save glTF files to the C# loader

MattMcMullan:
Also added primitive but functional support for KHR_binary_glTF


pjcozzi:
Rock on!

PR #498, opened 2015-12-02
Various bugfixes for the C# loader

MattMcMullan:
- Clean up the package
- Package now builds only in release
- Make the generator see the difference between a texture sampler and an animation sampler
- Make the generator run automatically before building the loader


pjcozzi:
Wow, you guys are fast!

PR #499, opened 2015-12-02
Spec tweaks

pjcozzi:
@tparisi for your review.  Just tightens up some language and adds an example.


pjcozzi:
@tparisi are you good with these tweaks?

PR #505, opened 2015-12-07
New cad examples, converted from JT

mlimper:
... using Okino Polytrans.
Thanks a lot to Robert Lansdale.


pjcozzi:
Wow, thanks @mlimper - and Robert!

Great test datasets!

Just FYI, I may rename the directories to camel case after merging to keep everything consistent.

mlimper:
> Just FYI, I may rename the directories to camel case after merging to keep everything consistent.

Cool, thanks for that.

PR #506, opened 2015-12-08
KHR_materials_common: Added naming of mesh attribs according to semantics

mlimper:
This topic was briefly discussed [here](https://github.com/AnalyticalGraphicsInc/cesium/pull/3294)

PR #509, opened 2015-12-14
KHR_materials_common: Two sentences and example for naming of mesh attributes

mlimper:
(originally here: https://github.com/KhronosGroup/glTF/pull/506)


pjcozzi:
:+1: 

@tparisi this just clarifies semantics/names and adds an example.  Are you good with this?

PR #510, opened 2015-12-14
Common material spec: handling of specular exponent

mlimper:
As discussed [here](https://github.com/AnalyticalGraphicsInc/cesium/pull/3294#issuecomment-163057175), this PR adapts the spec to clarify handling of different values for the specular exponent.

Just a strawman proposal - it would probably also be a good solution to leave this special case of adapting values from [0,1] completely out. I guess that could facilitate writing correct reader / renderer implementations even more.

### DIFF
--- a/extensions/Khronos/KHR_materials_common/README.md
+++ b/extensions/Khronos/KHR_materials_common/README.md
@@ -115,8 +115,7 @@ where
 * `H` – Half-angle vector,calculated as halfway between the unit Eye and Light vectors, using the
 equation H= normalize(I+L)
 
-The `shininess` exponent is typically expected to be in a range of [0, 128], as in the OpenGL fixed function pipeline (larger exponents are still possible, though).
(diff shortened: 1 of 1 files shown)


pjcozzi:
@mlimper this sounds like the client has to map `[0, 1]` to `[0, 128]`?  Why can't this be done at conversion time?  That is the glTF way: keep the client simple.

mlimper:
> Why can't this be done at conversion time? That is the glTF way: keep the client simple.

Yes, I would also prefer that!
Actually, that's what I meant:
> I guess that could facilitate writing correct reader / renderer implementations even more.

But then we need to adapt the current spec and remove the remark about handling different ranges differently (the passage about the Blinn-Torrance-Sparrow model).

pjcozzi:
That is all good with me.

mlimper:
Ok, thanks - I updated that.

Now, the PR removes the mentioned part of the spec about different interpretations of `shininess`.
Instead, an implementation note is added, clarifying the interpretation and illustrating typical usage by example (could maybe be more concise, I don't know).

pjcozzi:
OK with me.  @tparisi are you good with this?

PR #527, opened 2016-01-23
Replace old aspect_ratio example with aspectRatio

pjcozzi:
@tparisi for your review.  This is a small fix to an example in the spec.


pjcozzi:
Found by @leerichard42.

PR #532, opened 2016-02-04
Fixes blank textures

### LINKED ISSUE #530: collada2gltf blank textures
Hello, 

I'm trying to convert collada model to gltf with last released version of collada2gltf converter. Generally it works  OK, but some parts of model are blank. But if you use converter with option -i (invert transparancy) you can see parts of model which are was blank. This behavior you can see with this model, for example https://3dwarehouse.sketchup.com/model.html?id=d133fa016f4e452b8287fae9e4c703ae

After normal convertation
2016-02-01_17h12_50

After convertation with option -i
2016-02-01_17h11_43

Online converter works fine though. Is this a problem of collada2gltf converter or there is needs some prepocessing? I would be apprciate to get any advice.

tfili:
We didn't check whether transparent was a color or texture and we just use the alpha value which was in its initial state of -1. This caused meshes to be transparent when they weren't suppose to be. The real fix would be to handle a transparency map, but that is a decent undertaking. This fix ignores the transparent texture like before, it just doesn't introduce a -1 into the equation.

Its a band aid for #530

### DIFF
--- a/COLLADA2GLTF/COLLADA2GLTFWriter.cpp
+++ b/COLLADA2GLTF/COLLADA2GLTFWriter.cpp
@@ -219,32 +219,29 @@ namespace GLTF
         
         opaqueMode = effectCommon->getOpaqueMode();
 
+        ColorOrTexture transparent = effectCommon->getTransparent();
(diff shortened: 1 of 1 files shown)


pjcozzi:
Thanks @tfili!

PR #533, opened 2016-02-10
Fix a JSON-like code sample for Render States.

emackey:


pjcozzi:
Good catch, thanks @emackey!

PR #537, opened 2016-02-19
Uniform scaling of zero fix

gbeatty:
prevent a uniform scaling of zero from causing invalid values to be returned from a matrix decompose


pjcozzi:
Nice, thanks @gbeatty!  @tfili can you merge when ready?

tfili:
Looks great. Thanks @gbeatty.

PR #552, opened 2016-03-17
Fixed default rotation in spec.

dgough:


pjcozzi:
Good catch, thanks @dgough!

Issue #417, opened 2015-09-26
COLLADA2GLTF doesn't build with rapidjson 1.0.2


reneengelhard:
Hi,

we in Debian just upgraded our rapidjson package to the 1.0.2 release they released in spring. Because of the following only to experimental, though :)

I should note that in our collada2gltf packaging we fixed the build to use everything from the system, including rapidjson. (see http://anonscm.debian.org/cgit/pkg-multimedia/collada2gltf.git/tree/debian/patches/system-rapidjson.diff)

Unfortunately, the (already deprecated in 0.11/0.12 git) FileStream class/filestream.h got removed completely, so the build of COLLADA2GLTF with rapidjson 1.0.2 fails:

`
[  2%] Building CXX object CMakeFiles/collada2gltfConvert.dir/COLLADA2GLTFWriter.cpp.o
/usr/bin/c++   -DLIBXML_STATIC_FOR_DLL -DSTATIC_COLLADA2GLTF -DUSE_LIBPNG -DUSE_OPEN3DGC -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2  -std=c++11 -I/home/rene/tmp/collada2gltf-20140924 -I/home/rene/tmp/collada2gltf-20140924/GLTF -I/home/rene/tmp/collada2gltf-20140924/JSON -I/home/rene/tmp/collada2gltf-20140924/shaders -I/home/rene/tmp/collada2gltf-20140924/helpers -I/home/rene/tmp/collada2gltf-20140924/convert -I/home/rene/tmp/collada2gltf-20140924/assetModifiers -I/usr/include/rapidjson -I/usr/include/opencollada/COLLADAFramework -I/usr/include/opencollada/COLLADABaseUtils -I/usr/include/opencollada/COLLADASaxFrameworkLoader -I/usr/include/opencollada/GeneratedSaxParser -I/home/rene/tmp/collada2gltf-20140924/extensions/o3dgc-compression -I/usr/include/o3dgc    -fPIC -o CMakeFiles/collada2gltfConvert.dir/COLLADA2GLTFWriter.cpp.o -c /home/rene/tmp/collada2gltf-20140924/COLLADA2GLTFWriter.cpp
In file included from /home/rene/tmp/collada2gltf-20140924/COLLADA2GLTFWriter.h:30:0,
                 from /home/rene/tmp/collada2gltf-20140924/COLLADA2GLTFWriter.cpp:28:
[...]

PR #559, opened 2016-04-06
Removed requirement of rapidjson

tfili:
Removed rapidjson from all public headers. Its now only used for reading the json and as the default writer. This allows us to use a different version or other json library if we specify a custom writer from our own app.


pjcozzi:
Will this also fix #417?

pjcozzi:
Doesn't look that, was just curious.

Issue #556, opened 2016-03-21
Invalid fields in sample models


lexaknyazev:
I've found some invalid data in [sampleModels/duck/glTF/duck.gltf](https://github.com/KhronosGroup/glTF/blob/master/sampleModels/duck/glTF/duck.gltf) while testing my glTF loader.
1. `asset.version` is `int` instead of `string` [L72](https://github.com/KhronosGroup/glTF/blob/master/sampleModels/duck/glTF/duck.gltf#L72)
2. `aspectRatio` is spelled as `aspect_ratio` [L99](https://github.com/KhronosGroup/glTF/blob/master/sampleModels/duck/glTF/duck.gltf#L99)

pjcozzi:
Good catch, thanks @lexaknyazev!

@tfili I noticed this in several sample models, including some from Cesium (but not all).  Do you think these were converted with an old version of COLLADA2GLTF or is this still a bug in COLLADA2GLTF?  In either case, can you update the models, including the glb versions?

@lexaknyazev - btw, what glTF loader are you working on?

lexaknyazev:
Looks like bug is in the COLLADA2GLTF:
https://github.com/KhronosGroup/glTF/blob/master/COLLADA2GLTF/COLLADA2GLTFWriter.cpp#L1154
https://github.com/KhronosGroup/glTF/blob/master/COLLADA2GLTF/COLLADA2GLTFWriter.cpp#L1162
https://github.com/KhronosGroup/glTF/blob/master/COLLADA2GLTF/COLLADA2GLTFWriter.cpp#L1167

lexaknyazev:
Just tested other samples (except `MaterialsCommon` and `Binary`) for validity. Found two more issues:

1. `node.camera` in [`boxAnimated/glTF/glTF.gltf`](https://github.com/KhronosGroup/glTF/blob/master/sampleModels/boxAnimated/glTF/glTF.gltf#L307) is an empty string (according to schema it must be undefined or at least 1 char length).
2. `techniques.technique2.states.functions.depthMask` in [`vc/glTF/vc.gltf`](https://github.com/KhronosGroup/glTF/blob/master/sampleModels/vc/glTF/vc.gltf#L20112) is `boolean` instead of `boolean[1]`

Here is validator's output:
```
Loading 2_cylinder_engine/glTF/2_cylinder_engine.gltf...
[WARNING] glTF: cameras.default-camera.perspective: Unexpected property aspect_ratio

[...]

Issue #343, opened 2015-01-04
Bug: aspectRatio vs. aspect_ratio


bernstein:
The spec for perspective cameras uses "aspectRatio" but the converter writes out "aspect_ratio".

glTF/specification/cameraPerspective.schema.json  uses "aspectRatio"

glTF/converter/COLLADA2GLTF/COLLADA2GLTFWriter.cpp uses "aspect_ratio"

pjcozzi:
Thanks for the report, @bernstein.

The spec is correct and consistent with the glTF naming convention.  Looks like a converter bug, @fabrobinet.

fabrobinet:
indeed, good catch.

pjcozzi:
The spec and converter are in-sync in the 1.0 branch.

pjcozzi:
This was updated.  Thanks for the report, @bernstein.

MarcoHutter:
Are there any considerations to update the sample models accordingly? I stumbled over https://github.com/KhronosGroup/glTF/blob/master/sampleModels/duck/glTF/duck.gltf#L99 recently. I'm not sure whether this will be fixed in the current repo (in view of https://github.com/KhronosGroup/glTF/issues/502 ). But it's a bit unfortunate that the sample models can not be parsed with a spec-compliant parser (and I wonder how many non-compliant parsers are currently being written based on the fact that it's called `aspect_ratio` in the sample...)

pjcozzi:
@MarcoHutter yes, the sample models will be updated as part of #556.  Let us know if you run into any other issues with them.

PR #565, opened 2016-04-15
"aspect_ratio" should be "aspectRatio"

lasalvavida:
@fabrobinet did this in #343, but it looks like it never made it into master.
Also related to #556.


tfili:

pjcozzi:
Thanks @lasalvavida!

PR #567, opened 2016-04-19
Depth mask should be written as an array

lasalvavida:
Part of the fixes for #556.
Tested with the vc model and the output looks correct 
[vc.zip](https://github.com/KhronosGroup/glTF/files/226135/vc.zip)


tfili:
`depthMask` is now an array like the spec says and the model loads in Cesium.

Thanks @lasalvavida

PR #568, opened 2016-04-19
Remove TEXCOORD vertexAttribute when it is unnecessary

lasalvavida:
In reference to #556.

The TEXCOORD field is unnecessary unless the material for the node references a texture.

I've tested this with the box model and it successfully removed unused TEXCOORD fields. I've also tested it with the milk truck model and it correctly identifies that the TEXCOORD fields are used and doesn't remove them.

Also adds the ```removeVertexAttributes``` method to GLTFPrimitive.

Edit: After reviewing a bit, I realized that this still leaves the unused accessor and data for TEXCOORD. While that doesn't break anything, they should be removed to reduce the size of the model. Feel free to comment on what's here already, and I should have a fix in for the aforementioned soon.

### DIFF
--- a/COLLADA2GLTF/GLTF/GLTFAsset.cpp
+++ b/COLLADA2GLTF/GLTF/GLTFAsset.cpp
@@ -67,12 +67,6 @@ namespace GLTF
         asset->convertionResults()->setUnsignedInt32("verticesCount", totalVerticesCount);
 
         GLTF::JSONValueVector primitives = mesh->getPrimitives()->values();
-
-        bool isCompressed = false;
-        if (mesh->contains(kExtensions)) {
-            isCompressed = mesh->getExtensions()->contains("Open3DGC-compression");
-        }
-
         unsigned int primitivesCount = (unsigned int)primitives.size();
         for (unsigned int j = 0; j < primitivesCount; j++) {
             shared_ptr<GLTF::GLTFPrimitive> primitive = static_pointer_cast<GLTFPrimitive>(primitives[j]);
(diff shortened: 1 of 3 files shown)


COLLADA2GLTF/GLTF/GLTFAsset.cpp
```diff
@@ -941,29 +941,54 @@ namespace GLTF
         }
         
         meshesUIDs = meshes->getAllKeys();
+		shared_ptr <GLTF::JSONObject> materials = this->_root->createObjectIfNeeded(kMaterials);
```
tfili: It looks like your editor is inserting tabs instead of spaces. In Visual Studio go to Tools->Options->Text Editor->All Languages->Tabs. Select `Insert Spaces` and set `Tab Size

tfili:
Looks OK. I'll test it when you get the other fix in.

COLLADA2GLTF/GLTF/GLTFAsset.cpp
lasalvavida: Thank you

lasalvavida:
Unused accessor elements are removed from the final model with this update. I'm working on a final step to remove unused chunks of the buffer.

lasalvavida:
@tfili This is ready for testing. I tested with vc since it has fields that need to be left alone and fields that need to be removed. The produced model behaves correctly in model inspector and is clean in @lexaknyazev's [validator](http://lexaknyazev.github.io/gltf-demo/).

pjcozzi:
@tfili any update here?

tfili:
@pjcozzi I'm reviewing it now.

tfili:
Models load and the size differences of the `box.bin` file seem to match up with the size of the removed accessor.

PR #574, opened 2016-04-21
Added empty camera node workaround

lasalvavida:
Part of #556. 

boxAnimated.dae has a camera node that references a camera that doesn't exist. It ends up with an empty string for the camera which isn't allowed in the glTF schema.

This empty camera field can be safely removed.


pjcozzi:
@tfili can you please review?

tfili:

Issue #199, opened 2013-12-02
Converter compiles with warnings


pjcozzi:
Even with a newly cloned repo and following the instructions on the [wiki](https://github.com/KhronosGroup/glTF/wiki/converter).

![image](https://f.cloud.github.com/assets/782098/1655229/b3b09562-5b5e-11e3-80ab-83fe9a1660c0.png)

There are also 40 build warnings, which makes us look sloppy.  A lot of them look trivial.

pjcozzi:
Replaced by new problems, lol, https://github.com/KhronosGroup/COLLADA2GLTF/issues/91, that will be fixed!

PR #577, opened 2016-04-25
Fixed warnings when building

TomPed:
For #199 

Fixed trivial warnings. Still need to fix four more `C4503` warnings.

### DIFF
--- a/COLLADA2GLTF/COLLADA2GLTFWriter.cpp
+++ b/COLLADA2GLTF/COLLADA2GLTFWriter.cpp
@@ -95,7 +95,7 @@ namespace GLTF
         else
         {
             // We have the data in memory so use it
-            if (!root.loadDocument(asset->getInputFilePath(), fileData.data(), fileData.size())) {
+            if (!root.loadDocument(asset->getInputFilePath(), fileData.data(), (int)fileData.size())) {
                 delete _extraDataHandler;
                 return false;
             }
@@ -150,7 +150,7 @@ namespace GLTF
                             size_t minor = version.find(".", major + 1);
                             if (minor != string::npos) {
                                 version = version.substr(0, minor);
-                                float sketchUpVersion = atof(version.c_str());
+                                float sketchUpVersion = (float)atof(version.c_str());
                                 if (sketchUpVersion < 7.1) {
                                     asset->converterConfig()->config()->setBool("invertTransparency", true);
                                     this->_asset->log("WARNING: Fixing transparency - by inverting it - as we convert an asset generated from SketchUp version:%s \n", version.c_str());
@@ -453,10 +453,10 @@ namespace GLTF
                 Math::Real angle;
                 Math::Vector3 axis;
                 rotationQuat.toAngleAxis(angle, axis);
-                rotation[0] = axis.x;
-                rotation[1] = axis.y;
-                rotation[2] = axis.z;
-                rotation[3] = angle;
+                rotation[0] = (float)axis.x;
+                rotation[1] = (float)axis.y;
+                rotation[2] = (float)axis.z;
+                rotation[3] = (float)angle;
             }
 
             // Scale distance units if we need to
@@ -1286,7 +1286,7 @@ namespace GLTF
             case COLLADAFW::Light::SPOT_LIGHT: {
                 glTFLight->setString(kType, "spot");
 
-                float fallOffAngle =  (float)light->getFallOffAngle().getValue() * radianPerDegree;
+                float fallOffAngle =  (float)(light->getFallOffAngle().getValue() * radianPerDegree);
                 float fallOffExponent = (float)light->getFallOffExponent().getValue();
                 
                 description->setValue("constantAttenuation", shared_ptr <JSONNumber> (new JSONNumber(constantAttenuation)));
@@ -1463,7 +1463,7 @@ namespace GLTF
 
         //
         shared_ptr <GLTFBufferView> weightsView = createBufferViewWithAllocatedBuffer(weightsPtr, 0, skinAttributeSize, true);
-        shared_ptr <GLTFAccessor> weightsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize(bucketSize)));
+        shared_ptr <GLTFAccessor> weightsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize((int)bucketSize)));
         
         weightsAttribute->setBufferView(weightsView);
         weightsAttribute->setByteStride(componentSize * bucketSize);
@@ -1472,7 +1472,7 @@ namespace GLTF
         glTFSkin->setWeights(weightsAttribute);
         
         shared_ptr <GLTFBufferView> jointsView = createBufferViewWithAllocatedBuffer(bonesIndices, 0, skinAttributeSize, true);
-        shared_ptr <GLTFAccessor> jointsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize(bucketSize)));
+        shared_ptr <GLTFAccessor> jointsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize((int)bucketSize)));
         
         jointsAttribute->setBufferView(jointsView);
         jointsAttribute->setByteStride(componentSize * bucketSize);
--- a/COLLADA2GLTF/COLLADA2GLTFWriter.cpp
+++ b/COLLADA2GLTF/COLLADA2GLTFWriter.cpp
@@ -150,7 +150,7 @@ namespace GLTF
                             size_t minor = version.find(".", major + 1);
                             if (minor != string::npos) {
                                 version = version.substr(0, minor);
-                                float sketchUpVersion = (float)atof(version.c_str());
+                                double sketchUpVersion = atof(version.c_str());
                                 if (sketchUpVersion < 7.1) {
                                     asset->converterConfig()->config()->setBool("invertTransparency", true);
                                     this->_asset->log("WARNING: Fixing transparency - by inverting it - as we convert an asset generated from SketchUp version:%s \n", version.c_str());
@@ -1402,7 +1402,7 @@ namespace GLTF
         }
         
         size_t index = 0;
-        size_t bucketSize = 4;
+        int bucketSize = 4;
         size_t componentSize = sizeof(float);
         size_t skinAttributeSize = componentSize * vertexCount * bucketSize;
         float *weightsPtr = (float*)malloc(skinAttributeSize);
@@ -1463,7 +1463,7 @@ namespace GLTF
 
         //
         shared_ptr <GLTFBufferView> weightsView = createBufferViewWithAllocatedBuffer(weightsPtr, 0, skinAttributeSize, true);
-        shared_ptr <GLTFAccessor> weightsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize((int)bucketSize)));
+        shared_ptr <GLTFAccessor> weightsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize(bucketSize)));
         
         weightsAttribute->setBufferView(weightsView);
         weightsAttribute->setByteStride(componentSize * bucketSize);
@@ -1472,7 +1472,7 @@ namespace GLTF
         glTFSkin->setWeights(weightsAttribute);
         
         shared_ptr <GLTFBufferView> jointsView = createBufferViewWithAllocatedBuffer(bonesIndices, 0, skinAttributeSize, true);
-        shared_ptr <GLTFAccessor> jointsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize((int)bucketSize)));
+        shared_ptr <GLTFAccessor> jointsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize(bucketSize)));
         
         jointsAttribute->setBufferView(jointsView);
         jointsAttribute->setByteStride(componentSize * bucketSize);
--- a/COLLADA2GLTF/GLTF/GLTFAsset.cpp
+++ b/COLLADA2GLTF/GLTF/GLTFAsset.cpp
@@ -819,7 +819,7 @@ namespace GLTF
             if (jointsCount > 0) {
                 if (meshExtras == nullptr)
                     meshExtras = std::make_shared<JSONObject>();
-                meshExtras->setUnsignedInt32("jointsCount", jointsCount);
+                meshExtras->setUnsignedInt32("jointsCount", (int)jointsCount);
             }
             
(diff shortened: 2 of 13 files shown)


COLLADA2GLTF/COLLADA2GLTFWriter.cpp
```diff
@@ -150,7 +150,7 @@ namespace GLTF
                             size_t minor = version.find(".", major + 1);
                             if (minor != string::npos) {
                                 version = version.substr(0, minor);
-                                float sketchUpVersion = atof(version.c_str());
+                                float sketchUpVersion = (float)atof(version.c_str());
```
lasalvavida: I think maybe just make this variable a double instead of casting to a float.

COLLADA2GLTF/COLLADA2GLTFWriter.cpp
```diff
@@ -1463,7 +1463,7 @@ namespace GLTF
 
         //
         shared_ptr <GLTFBufferView> weightsView = createBufferViewWithAllocatedBuffer(weightsPtr, 0, skinAttributeSize, true);
-        shared_ptr <GLTFAccessor> weightsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize(bucketSize)));
+        shared_ptr <GLTFAccessor> weightsAttribute(new GLTFAccessor(profile, "FLOAT", GLTFUtils::getTypeForVectorSize((int)bucketSize)));
```
lasalvavida: It's probably better to just declare the variable as an int instead of a size_t, as opposed to casting it everywhere.

lasalvavida: Same comment throughout

TomPed:
Updated. I've only declared the variables if I needed to cast _that_ variable more than once. Is this okay, or should I make this change for variables that only get cast once currently as well.

COLLADA2GLTF/GLTF/GLTFMesh.cpp
```diff
@@ -83,7 +83,7 @@ namespace GLTF
             size_t attributesCount = this->getMeshAttributesCountForSemantic(semantic);
             for (size_t j = 0 ; j < attributesCount ; j++) {
                 shared_ptr <GLTF::GLTFAccessor> selectedMeshAttribute = this->getMeshAttribute(semantic, j);
-                unsigned int indexSet = j;
+                unsigned int indexSet = (int)j;
```
lasalvavida: I see why you left this one and a few of the others this way since it would involve changing a few method stubs as well. @tfili, what do you think? I'm fine with a few casts, but it seems like it may be wise to switch over the size_t variables to ints throughout to avoid future warnings of this nature.

lasalvavida:
@TomPed, let's see what @tfili says. It's not a huge deal to leave this largely alone and cast the variables where necessary, but there's really not a good reason that I'm aware of to use size_t instead of int.

COLLADA2GLTF/GLTF/GLTFMesh.cpp
tfili: Yeah, switching it over to int (or unsigned int) should be fine. Passing an int to a function that takes size_t shouldn't throw a warning since size_t is bigger.

TomPed: Made this change.

TomPed:
Updated.

COLLADA2GLTF/GLTF/GLTFAsset.cpp
```diff
@@ -819,7 +819,7 @@ namespace GLTF
             if (jointsCount > 0) {
                 if (meshExtras == nullptr)
                     meshExtras = std::make_shared<JSONObject>();
-                meshExtras->setUnsignedInt32("jointsCount", jointsCount);
+                meshExtras->setUnsignedInt32("jointsCount", (int)jointsCount);
```
lasalvavida: You should be able to just make jointsCount an int

COLLADA2GLTF/GLTF/GLTFMesh.cpp
```diff
@@ -81,7 +81,7 @@ namespace GLTF
         for (unsigned int i = 0 ; i < allSemantics.size() ; i++) {
             GLTF::Semantic semantic = allSemantics[i];
             size_t attributesCount = this->getMeshAttributesCountForSemantic(semantic);
-            for (size_t j = 0 ; j < attributesCount ; j++) {
+            for (unsigned int j = 0 ; j < attributesCount ; j++) {
                 shared_ptr <GLTF::GLTFAccessor> selectedMeshAttribute = this->getMeshAttribute(semantic, j);
                 unsigned int indexSet = j;
```
lasalvavida: You don't need this variable anymore, just use j.

COLLADA2GLTF/extensions/o3dgc-compression/GLTF-Open3DGC.cpp
```diff
@@ -326,8 +326,8 @@ namespace GLTF
             
             for (size_t j = 0 ; j < attributesCount ; j++) {
                 shared_ptr <GLTFAccessor> meshAttribute = mesh->getMeshAttribute(semantic, j);
-                vertexCount = meshAttribute->getCount();
-                size_t componentsPerElement = meshAttribute->componentsPerElement();
+                vertexCount = (long)meshAttribute->getCount();
```
lasalvavida: You shouldn't need to cast here because there is no loss of precision from size_t -> long.

COLLADA2GLTF/extensions/o3dgc-compression/GLTF-Open3DGC.cpp
```diff
@@ -326,8 +326,8 @@ namespace GLTF
...
+                long componentsPerElement = (long)meshAttribute->componentsPerElement();
```
lasalvavida: Same comment

COLLADA2GLTF/extensions/o3dgc-compression/GLTF-Open3DGC.cpp
```diff
@@ -265,7 +265,7 @@ namespace GLTF
         O3DGCSC3DMCPredictionMode jointPrediction = _predictionModeForString(config->stringForKeyPath("extensions.Open3DGC.quantization.JOINT"));
         
         GLTFOutputStream *outputStream = asset->createOutputStreamIfNeeded(kCompressionOutputStream).get();
-        size_t bufferOffset = outputStream->length();
+        int bufferOffset = (int)outputStream->length();
```
lasalvavida: This cast is unnecessary as well

COLLADA2GLTF/extensions/o3dgc-compression/GLTF-Open3DGC.cpp
```diff
@@ -405,7 +405,7 @@ namespace GLTF
         BinaryStream bstream(vertexCount * 8);
         SC3DMCEncoder <unsigned short> encoder;
         shared_ptr<JSONObject> compressedData(new JSONObject());
-        compressedData->setInt32("verticesCount", vertexCount);
+        compressedData->setInt32("verticesCount", (int)vertexCount);
```
lasalvavida: Same comment

COLLADA2GLTF/helpers/geometryHelpers.cpp
```diff
@@ -371,7 +371,7 @@ namespace GLTF
         for (unsigned int i = 0 ; i < allSemantics.size() ; i++) {
             GLTF::Semantic semantic = allSemantics[i];
             size_t attributesCount = sourceMesh->getMeshAttributesCountForSemantic(semantic);
-            for (size_t j = 0 ; j < attributesCount ; j ++) {
+            for (unsigned int j = 0 ; j < attributesCount ; j ++) {
                 shared_ptr <GLTF::GLTFAccessor> selectedMeshAttribute = sourceMesh->getMeshAttribute(semantic, j);
                 unsigned int indexSet = j;
```
lasalvavida: You can get rid of this variable too

COLLADA2GLTF/extensions/o3dgc-compression/GLTF-Open3DGC.cpp
```diff
@@ -540,33 +540,33 @@ namespace GLTF
...
+        int byteOffset = 0;
         bool shouldEncodeOpen3DGC = CONFIG_STRING(asset, "compressionType")  == "Open3DGC";
         GLTFOutputStream *outputStream = shouldEncodeOpen3DGC ? asset->createOutputStreamIfNeeded(kCompressionOutputStream).get() : asset->createOutputStreamIfNeeded(asset->getSharedBufferId()).get();;
-        byteOffset = outputStream->length();
+        byteOffset = (int)outputStream->length();
         parameter->setUnsignedInt32("byteOffset", byteOffset);
         
         if (shouldEncodeOpen3DGC) {
-            size_t componentsCount = profile->getComponentsCountForType(parameter->getString(kType));
+            long componentsCount = (long)profile->getComponentsCountForType(parameter->getString(kType));
```
lasalvavida: This is size_t -> long, you shouldn't need a cast

COLLADA2GLTF/extensions/o3dgc-compression/GLTF-Open3DGC.cpp
```diff
@@ -540,33 +540,33 @@ namespace GLTF
...
             if (componentsCount) {
-                encodeDynamicVector((float*)buffer, parameterSID, componentsCount, cvtAnimation->getCount(), asset);
+                encodeDynamicVector((float*)buffer, parameterSID, componentsCount, (long)cvtAnimation->getCount(), asset);
                 
-                byteLength = outputStream->length() - byteOffset;
+                byteLength = (int)outputStream->length() - byteOffset;
```
lasalvavida: Make byteLength an int

COLLADA2GLTF/COLLADA2GLTFWriter.cpp
```diff
@@ -95,7 +95,7 @@ namespace GLTF
         else
         {
             // We have the data in memory so use it
-            if (!root.loadDocument(asset->getInputFilePath(), fileData.data(), fileData.size())) {
+            if (!root.loadDocument(asset->getInputFilePath(), fileData.data(), (int)fileData.size())) {
```
lasalvavida: Were you seeing a warning here? I don't see one locally and I can't think of a reason why this would need to be cast.

COLLADA2GLTF/extensions/o3dgc-compression/GLTF-Open3DGC.cpp
tfili: It actually will give a warning. In Visual Studio `size_t` is actually an `unsigned __int64` (8 bytes) and `long` is the same size as an `int` (4 bytes). `long`s only requirement that they must at least be the same size as an `int`.

COLLADA2GLTF/COLLADA2GLTFWriter.cpp
TomPed: I see this warning when I build all or just compile that one file.

COLLADA2GLTF/extensions/o3dgc-compression/GLTF-Open3DGC.cpp
lasalvavida: @tfili Thank you for the correction, I'll edit the comments where I pointed this out

COLLADA2GLTF/COLLADA2GLTFWriter.cpp
tfili: It's a `size_type` to `int`, which is 8 bytes to 4 bytes. I get a warning if I remove it. @lasalvavida did you build the 32-bit or 64-bit version of the converter?

TomPed:
As for the `size_t` to `long` I get the warning if I remove the casts. Also after some googling stack overflow says that in some cases `size_t` will be larger than `long` and `unsigned long`: 

> The main downside of replacing size_t with unsigned long is that unsigned long is not guaranteed to be large enough to contain every possible value of size_t, and on Windows 64 it is not large enough. So you might find that you still have warnings.

lasalvavida:
@TomPed My apologies, I was building x86. I removed the comments marking that, and left the ones that should probably still be done.

TomPed:
No problem. Made those edits.

lasalvavida:
Looks good to me, thanks @TomPed!

TomPed:
I am now only getting these warning `4503`. I've done some Googling and there are a few stack overflow answers that says that only way to stop this error is to silence this type of warning (`4503`), unless you want to remake the whole data structure. @tfili any ideas for this?

lasalvavida:
@TomPed It's actually a pretty easy fix. The issue is caused in [GLTFAsset.h](https://github.com/KhronosGroup/glTF/blob/master/COLLADA2GLTF/GLTF/GLTFAsset.h#L58).

```c++
typedef std::map<std::string, std::shared_ptr <MaterialBindingsPrimitiveMap> > MaterialBindingsForMeshUID;
```

ends up expanding out to

```c++
typedef std::map<std::string, std::shared_ptr <std::map<unsigned int, std::shared_ptr <COLLADAFW::MaterialBinding> > > > MaterialBindingsForMeshUID;
```

typedef essentially just creates a macro and this ends up generating some huge type definitions. You can fix this by creating making MaterialBindingsPrimitiveMap a subclass of map.

```c++
class MaterialBindingsPrimitiveMap : public std::map<unsigned int, std::shared_ptr <COLLADAFW::MaterialBinding>> {};
```

Try it out, it seems to work for me locally.

TomPed:
I've added that line to the header file and the warnings have gone away. But now I seem to be getting some errors now: `redefinition; different basic types`.

lasalvavida:
@TomPed I'm not seeing any errors. Did you remember to delete the old typedef?

TomPed:
Updated, thanks for the help @lasalvavida!

TomPed:
@tfili this is ready for a look.

tfili:
@TomPed these changes all look good.

I'm getting 2 warnings in `main.cpp`

```
main.cpp(268): warning C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C++ conformant name: _strdup.

main.cpp(275): warning C4067: unexpected tokens following preprocessor directive - expected a newline
```

Pretty easy fixes.

For the first you just need to define `_CRT_NONSTDC_NO_DEPRECATE` on windows in the `CMakeLists.txt` file. Look for the `if (WIN32)/endif()` around line 80.

For the second just replace
```
#ifndef WIN32 || WIN64
```
with
```
#ifndef WIN32
```
since WIN32 should always be defined for any windows app.

tfili:
Other than that this project is warning free on my machine.

TomPed:
@tfili, I've made the second change to problem, but when I define `_CRT_NONSTDC_NO_DEPRECATE` on windows in the `CMakeLists.txt` file i get his error: `Error 7 error C1083: Cannot open source file: '_CRT_NONSTDC_NO_DEPRECATE'`. Currently trying to find a solution.

tfili:
@TomPed how did you try to add it. Look for the line

`add_definitions(-D_CRT_SECURE_NO_WARNINGS)`

And add the following below it

`add_definitions(-D_CRT_NONSTDC_NO_DEPRECATE)`

TomPed:
Ahh, I put it before that line. Updated, no more warnings.

tfili:
Looks good. Thanks @TomPed.

Issue #571, opened 2016-04-20
Sample models don't have uniform names


lasalvavida:
There are a lot of different naming conventions in the sample models, and I think for consistency, one style should be adopted throughout. 

I'm proposing capital camel-case as in "CesiumMan" or "GearboxAssy".

There are also some models whose internal names are different from the name of the directory they live in, and those should have the same name.

pjcozzi:
Fixed in #579.

Issue #524, opened 2016-01-12
Add vc (virtual city) to list of sample models


pjcozzi:
https://github.com/KhronosGroup/glTF/tree/master/sampleModels

It's there, it's just not listed in the README.md.

pjcozzi:
Fixed in #579.

Issue #517, opened 2016-01-06
We should have all flavors of converted models in sampleModels folder (Binary, Materials_Common etc)


tparisi:
@pjcozzi @tfili can we get glTF-Embedded, glTF-MaterialsCommon and glTF-Binary versions of each model in the sampleModels folder? This is for completeness and also will help me test the data: URI feature with Three.js.

FWIW I'm not implementing the Binary extension right now, so that won't help me for Three.js; however we should have them for completeness.

pjcozzi:
There already are all of these variations for several of the models, for example: https://github.com/KhronosGroup/glTF/tree/master/sampleModels/boxAnimated

tparisi:
Understood but it would be good to have them all if possible. Not highest priority. But for completeness.

pjcozzi:
Fixed in #579.

PR #564, opened 2016-04-14
Clean up sample models

lasalvavida:
In reference to #556 and #571.
Edit: also resolves #524 and #517

PR #579, opened 2016-04-25
Fix sample models

### LINKED ISSUE #556: Invalid fields in sample models

lasalvavida:
This should resolve #556, #571, #524, and #517.


pjcozzi:
Replaces #564.

pjcozzi:
@tomped would you like to test this by dragging and dropping each model (including each format variation in the sub-directories) onto http://cesiumjs.org/convertmodel.html.  I _think_ the ones with separate resource files (.bin, images, shaders) will need to be tested explicitly in Sandcastle.

Also validate each one with http://lexaknyazev.github.io/gltf-demo/

sampleModels/README.md
```diff
@@ -1,4 +1,4 @@
-Sample models are provided in as many of the following formats as possible:
+bSample models are provided in as many of the following formats as possible:
```
pjcozzi: "bSample" should be "Sample"

pjcozzi: Also, up to you, but it would not be a bad idea to replace each README.txt with README.md for consistency:

>  See the `README.txt` in each model's directory for usage restrictions.

Maybe even include the thumbnail and short description here also in the README.md for each model.

pjcozzi:
I **love** the new animated gifs for the animated/skinned models!

sampleModels/README.md
pjcozzi: Also update this .md:

> A README.txt with any usage restrictions.

lasalvavida: Good catch. And I think that's a good idea for the READMEs, I'll add that when I get a chance.

pjcozzi:
The issues in #556 look like they have been addressed.

Just waiting on @tomped's testing and any additional feedback from @lexaknyazev before merging.

TomPed:
> each model (including each format variation in the sub-directories) onto http://cesiumjs.org/convertmodel.html.

@pjcozzi I only see one type of COLLADA file to send to the converter

TomPed:
I am getting some warnings about Unexpected propertys when I send some `.gltf` files into the glTF Validator. Is this a problem?

lasalvavida:
@TomPed it depends which versions you're talking about. glTF and glTF-Embedded should be totally clean. glTF-MaterialsCommon has some warnings like:

[WARNING] glTF: materials/Effect-inner/extensions/KHR_materials_common: Unexpected property doubleSided

from the validator. These are issues with the validator (and the spec for the extension which needs some updating and clarification).

glTF-Binary has the following:

[SEVERE] glTF: buffers/binary_glTF/uri: Unsupported MIME type: text/plain

As best I can tell, this is also a validator issue.

TomPed:
Same. I see the same results.

lexaknyazev:
`KHR_materials_common` files have 2 kinds of issues:

1. Unexpected properties (`doubleSided`, `jointCount` and `transparent`) in the extension object. Schema says (maybe wrongly) they must be inside `values`.

2. Textures for `ambient` property in `VC.gltf`. Schema permits only `FLOAT_VEC4`.

Those MIME errors come from `"uri": "data:,"` fields because `data:` URI without explicit MIME type implies `text/plain` which is incorrect type for buffers or images.
I'll remove this validation on empty data URIs.

`/ReciprocatingSaw/glTF-Embedded/ReciprocatingSaw.gltf` is an empty file.

lasalvavida:
The KHR_materials_common spec hasn't been ratified, and does need some attention. I'm inclined to follow the implementation and update the spec to match, but that's probably up to @pjcozzi.

Thanks @lexaknyazev! Good catch on the empty file

pjcozzi:
@lasalvavida for KHR_materials_common, I agree, update the spec.
* See https://github.com/KhronosGroup/glTF/issues/424#issuecomment-148527783
   * It would be worth going through that entire issue and seeing if anything else in the extension spec needs to be updated.
* Are both COLLADA2GLTF and Cesium in-sync?

TomPed:
> I think the ones with separate resource files (.bin, images, shaders) will need to be tested explicitly in Sandcastle.

You are right. I've tested all other models and they work. I will now be testing these types of models.

lexaknyazev:
Validator's MIME type checks fixed.

TomPed:
Alright so I've tested everything. All looks good but a few issues.

1. `sampleModels/BoxSemantics/glTF-MaterialsCommon/CesiumLogoFlat.jpg` is missing
   * The file names in `sampleModels/BoxSemantics/glTF-MaterialsCommon/` should also be changed to start with a capital letter for consistency
2. Origin of GearboxAssy is off (for all types of models)
3. Origin of Monster is off (for all types of models)

lasalvavida:
@TomPed 
1. Good catch! Names are fixed and the missing image is added
2/3. I think that's just how the models are. Choosing the origin for a model is a somewhat subjective topic, and as far as I'm aware, there's not a really a correct choice.

pjcozzi:
@lasalvavida @TomPed is anything else needed before merging this?

TomPed:
From me no, all formats worked.

pjcozzi:
Thanks so much @lasalvavida @TomPed @lexaknyazev!  This is a great contribution.

Issue #570, opened 2016-04-20
uint32 indices


pixelnerve:
Hello,

I've noticed that collada2gltf is splitting meshes and that accessor.componentType doesn't mention int32/uint32 types. Is there any plans to change that anytime soon ?

lexaknyazev:
Neither WebGL 1.0, nor OpenGL ES 2.0 supports int32 indices without `OES_element_index_uint` extension. So core glTF spec doesn't mention int32 type.

pixelnerve:
Yes. Will it be supported ? And how soon ?

lexaknyazev:
Formally speaking, spec should already support int32 indices through extensions. It just lacks explicit examples of how to do it properly. Hope to have them with 1.0.1 update.
I wouldn't expect collada2gltf to implement that before spec update.

Maybe @pjcozzi or @lasalvavida could shed some light?

pixelnerve:
Any pointers on how to make that work are welcome. Thanks @lexaknyazev

pjcozzi:
As @lexaknyazev mentioned, given that uint32 indices are not in the core WebGL 1.0 spec, they would not be in the core glTF 1.0 spec.  However, the WebGL extension is widely supported and we could do a very simple glTF extension for it so a glTF renderer knows to (1) enable the WebGL extension and (2) allow uint32 as a validate datatype for indices.

@lexaknyazev or @pixelnerve would you like to try to write up the extension?  For details and examples, see the [glTF Extension Registry](https://github.com/KhronosGroup/glTF/blob/master/extensions/README.md).

lexaknyazev:
@pjcozzi Since we don't have any formal example of WebGL extensions in glTF I've written some considerations on this topic [here (comment)](https://github.com/KhronosGroup/glTF/issues/563#issuecomment-213421633).

pixelnerve:
32bit indices would need new data types: int32 and uint32 and disabling the splitting meshes method. Would also write the extension to the extension list.

pjcozzi:
Related: #535

pjcozzi:
Latest discussion is in https://github.com/KhronosGroup/glTF/issues/563#issuecomment-224095245

pjcozzi:
See pull request #619.

lexaknyazev:
@pjcozzi Can this issue be labeled as `resolved` or we need to wait for `collada2gltf` support?

pjcozzi:
[...]

PR #583, opened 2016-04-29
Added UNSIGNED_INT accessor.componentType

abwood:
UNSIGNED_INT was missing from the list of valid accessor componentTypes.  Without this, 32bit indices are not valid in gltf models.

I also noticed the default values for blendFuncSeparate did not match the defaults specified by the es2 spec.  See: http://docs.gl/es2/glBlendFuncSeparate

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -415,6 +415,7 @@ The following tables can be used to compute the size of an accessor's attribute 
 | `5121`(UNSIGNED_BYTE) | 1 |
 | `5122` (SHORT) | 2 |
 | `5123` (UNSIGNED_SHORT) | 2 |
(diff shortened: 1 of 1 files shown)


abwood:
Closing after seeing the discussion in issue #570.  uint32 indices is not in the es2 or webgl spec.

emackey:
Were there smaller fixes included here that should be broken out into a separate PR?

abwood:
Yes, already done in https://github.com/KhronosGroup/glTF/pull/585

pjcozzi:
Thanks for investigating, @abwood!  You are welcome to propose the (simple) extension for #570.

PR #585, opened 2016-04-29
Fixed default values for blendFuncSeparate.

abwood:
The ordering of parameters for blendFuncSeparate are incorrect.  See http://docs.gl/es2/glBlendFuncSeparate

Also removed the unused lighting section from the Table of Contents.

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -36,7 +36,6 @@ Copyright (C) 2013-2016 The Khronos Group Inc. All Rights Reserved. glTF is a tr
   * [Geometry and Meshes](#geometry-and-meshes)
   * [Materials and Shading](#materials-and-shading)
   * [Cameras](#cameras)
-  * [Lights](#lights)
   * [Animations](#animations)
   * [Metadata](#metadata)
   * [Specifying Extensions](#specifying-extensions)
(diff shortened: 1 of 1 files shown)


lasalvavida:
@abwood The collada2gltf implementation in [commonProfileShaders.cpp](https://github.com/KhronosGroup/glTF/blob/master/COLLADA2GLTF/shaders/commonProfileShaders.cpp#L386) needs to be modified to match this.

Also, this would be a breaking change for some existing models, so it would have to be part of 1.0.1.

abwood:
Great catch, you are correct that any models out in the wild that have been exported with blendFuncSeparate will have the mismatched ordering.  I discovered this when I was fiddling with a Gltf loader and assumed it was just a typo in the specification document.

lasalvavida:
@pjcozzi If this is a change we want to make, can you make a branch for 1.0.1?

pjcozzi:
Good catch @abwood!

This is just a spec/COLLADA2GLTF bug for the default value so I am fine merging this into 1.0 as a fix.

@lasalvavida can you open a new PR with the COLLADA2GLTF fix?

PR #586, opened 2016-04-29
Fix for #584

### LINKED ISSUE #584: Edge case for removing unused TEXCOORD fields
I ran into an issue with an AGI model that had a primitive with multiple TEXCOORD fields declared where some of those weren't used by the material/technique. In this case, it isn't enough to just check if the material uses textures, we need to make sure that the specific TEXCOORD is used in the technique.

lasalvavida:
Fixes #584 by checking if TEXCOORD is bound to a technique


lasalvavida:
@tfili This results in clean validation from that AGI model, and it renders correctly in cesium

tfili:
Looks good.

Issue #589, opened 2016-05-03
COLLADA2GLTF Assumes that all primitives in a mesh have the same attribute accessors


lasalvavida:
This seems to have been true for a while, and in general wasn't a problem until recently because it is usually true; typically COLLADA2GLTF will make a separate sub-mesh if the geometry is different. 

However, with my recent change that removes unused TEXCOORD attributes, it broke one of the assumptions made during this process, exposed by an AGI model that failed conversion. I can't include the model, but here is the breaking case:

``` json
"primitives": [
    {
        "attributes": {
            "NORMAL": "accesor_1",
            "POSITION": "accessor_2"
        },
        "attributes": {
            "NORMAL": "accesor_1",
            "POSITION": "accessor_2"
        },
        "attributes": {
            "NORMAL": "accesor_1",
            "POSITION": "accessor_2",
            "TEXCOORD_0": "accessor_3"
        }
    }
]
```

The internal _semanticToMeshAttributes in GLTFMesh gets populated like this:

| NORMAL | POSITION | TEXCOORD |
| --- | --- | --- |
| accessor_1 | accessor_2 | accessor_3 |
| accessor_1 | accessor_2 | null |
| accessor_1 | accessor_2 | null |

So, when we try to get TEXCOORD for the primitive at index 2, we get null instead, so the accessor is never created. The quick fix for this issue is to just pull from index 0 all of the time, since the attributes are never different currently.

This does need to be more thoroughly fixed by cleaning up GLTFMesh and GLTFAsset.

pjcozzi:
Workaround in #590.  @lasalvavida do you want to leave this open?

lasalvavida:
@pjcozzi, yes, leave this open for now.

pjcozzi:
Closing since a workaround was added and the COLLADA2GLTF 2.0 branch is a rewrite.

PR #590, opened 2016-05-03
Quick fix for #589

lasalvavida:


lasalvavida:
This is a temporary fix for #589, the long term fix will involve more refactoring

tfili:
Looks good. Thanks @lasalvavida

PR #595, opened 2016-05-18
Create unique mesh instances for collada models with instance geometry

lasalvavida:
Thanks to @gbeatty for pointing this out.

If a COLLADA model uses instance_geometry it should create new meshes along with the nodes so that material binding can be different.

See the attached zip [convertedModels.zip](https://github.com/KhronosGroup/glTF/files/271165/convertedModels.zip)

The COLLADA file currently produces the incorrect output in 1m_cube.gltf. With these changes it produces 1m_cube_fixed.gltf which displays the correct behavior of having different materials on each side.


lasalvavida:
@tfili take a look at this when you get a chance

tfili:
Looks good

pjcozzi:
Very nice, thanks @lasalvavida @gbeatty!

PR #599, opened 2016-05-24
Only make unique mesh instances if the material is an instance

lasalvavida:
There were some test failures in an AGI project dependent on COLLADA2GLTF after my last change: #595. 

While largely cosmetic, it is still better to only create new mesh instances if they are necessary, so this change checks if the bound material is an instance first.

I have run this change against that AGI project and the tests all pass, and @gbeatty's box model with instance materials on each side also converts correctly, so this should be good to go.


lasalvavida:
@tfili, Can you take a look at this when you get a chance?

tfili:

PR #612, opened 2016-06-02
Clarify that sceneLength in KHR_binary_glTF must be greater than zero

pjcozzi:
I consider this a clarification - not a breaking change that would require a new version - since if `sceneLength === 0`, then the binary glTF payload would not contain a glTF asset.

@tparisi @mlimper @lexaknyazev do you agree?


mlimper:
Yes, I could even imagine valid cases where the body could be empty.

pjcozzi:
> I could even imagine valid cases where the body could be empty.

Agreed.

pjcozzi:
Thanks for the prompt feedback.  We're good here.

Issue #563, opened 2016-04-14
Thoughts on tightening up the spec


lexaknyazev:
### Base spec
- [x] 1. There is no integers in JS, but other languages can have them. Spec is silent on implicit type casts of integers provided for values declared as `FLOAT` or `FLOAT_VEC*`. I believe that a little permissive comment about that would be enough.
  - No spec update, https://github.com/KhronosGroup/glTF/issues/563#issuecomment-223148438
- [x] 2. `MODELVIEWINVERSETRANSPOSE` and `MODELINVERSETRANSPOSE` technique parameter semantics are declared as `FLOAT_MAT3`. At the same time presence of `node` field requires parameter to be `FLOAT_MAT4`. Spec is silent on this ambiguity. (see `boxSemantics/glTF/boxSemantics.gltf`)
  - Updated spec, f548e60e335a0ea3b7c039c5950dd823a83fd31b
- [x] 3. Is it OK to have technique parameters that are mentioned neither in `technique.uniforms` nor in `technique.attributes`?
  - Updated spec, ff4ac9797082bce00011e337f2790793f7a52b51
- [x] 4. Can technique parameter be referenced in `technique.attributes` and have a `value` property at the same time?
  - Updated spec, 537c356cc4023b86084b10a90b4d998908f6896a
- [x] 5. Can technique parameter with `FLOAT` (or other scalar) type and `"count": 3` be linked with `FLOAT_VEC3` accessor? (More details in the comment below.)
  - Updated spec, #614
- [x] 6. Are `{POSITION|NORMAL|...}_%d` semantics valid? Provided samples use only `TEXTURE_%d`.
  - Updated spec, b93fa35991ba58f8b1715a68f409ff749da0e2bc
- [x] 7. There is no explicit requirement on the `mesh.primitive.indices` accessor to not to be `FLOAT` and to reference `bufferView` with `ELEMENT_ARRAY_BUFFER` target.
  - Updated spec, #615
- [x] 8. Default material example (Appendix A) has an extra `a_texcoord_0` attribute.
  - Updated example, #608
- [x] 9. Also I think that spec could be more verbose on required semantics for mesh to be rendered with default material. Without `NORMAL` runtime can try to calculate them on-the-fly, but without  `POSITION` there's nothing to render at all.
  - Updated example, #608
- [x] 10. Do we need 2D versions of default material shaders for pure 2D content? Or in that case users must provide shader code themselves?
[...]

Issue #258, opened 2014-03-26
need size for uniforms


RemiArnaud:
split from #100 

need to add count  to uniforms when it is a multiple of the size defined in the type.

for example

```
"jointMat": {
                    "semantic": "JOINT_MATRIX",
                    "type": 35676
                },
```

will become

```
"jointMat": {
                    "semantic": "JOINT_MATRIX",
                    "type": 35676,
                    "count": 7
                },
```

PR #614, opened 2016-06-02
Clarify that array parameters are not the same as vectors

pjcozzi:
For https://github.com/KhronosGroup/glTF/issues/563#issuecomment-210993081

@lexaknyazev what do you think of this clarification?

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -3008,6 +3008,8 @@ Additional properties are not allowed.
 
 When defined, the parameter is an array of count elements of the specified type.  Otherwise, the parameter is not an array.  When defined, `value` is an array with length equal to count, times the number of components in the type, e.g., `3` for `FLOAT_VEC3`.
 
+An array parameter of scalar values is not the same as a vector parameter of the same size; for example, when `count` is `2` and `type` is `5126` (FLOAT), the parameter is an array of two floating-point values, not a `FLOAT_VEC2`.
+
 * **Type**: `integer`
 * **Required**: No
 * **Minimum**:` >= 1`
--- a/specification/schema/technique.parameters.schema.json
+++ b/specification/schema/technique.parameters.schema.json
@@ -9,7 +9,7 @@
             "type" : "integer",
             "description" : "When defined, the parameter is an array of count elements of the specified type.  Otherwise, the parameter is not an array.",
             "minimum" : 1,
-            "gltf_detailedDescription" : "When defined, the parameter is an array of count elements of the specified type.  Otherwise, the parameter is not an array.  When defined, `value` is an array with length equal to count, times the number of components in the type, e.g., `3` for `FLOAT_VEC3`."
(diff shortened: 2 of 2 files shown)


lexaknyazev:
That clears possible ambiguity. 
Two more points:

1. Should `parameter.count` property be allowed for attribute parameters?
2. The only use of `count` property in sample models is `jointMat` parameter with `JOINTMATRIX` semantic. Maybe it should be stated in spec, that parameters with other spec-defined semantics can't have `count` property.

pjcozzi:
See #258.

In Cesium, we currently only support attribute arrays for skinning: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Scene/Model.js#L1976-L1981

Actually, the same is true for uniforms: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Scene/Model.js#L2508-L2511

We want it to be as easy as possible to write a correct (and _fast!_) loader for glTF.  At the same time, we don't want to limit widely used GL features.

Given this, do you think the best trade off is to only allow `count` for `JOINTMATRIX` to simplify what it takes to write a complete loader?

Also, a change here will need to be a separate PR since it would need to be glTF 1.0.1, not just a clarification in the 1.0 spec.

lexaknyazev:
> We want it to be as easy as possible to write a correct (and fast!) loader for glTF. At the same time, we don't want to limit widely used GL features.

Spec doesn't define how to treat arrays for known semantics such as `PROJECTION` or `MODELVIEW`. So I think it would be safe to say that all uniform semantics except `JOINTMATRIX` can't be arrays (or define that behavior otherwise). There are unrestricted custom semantics for application-specific use.

As for attribute arrays I honestly don't quite understand what it means since GLSL explicitly forbids them.

pjcozzi:
All sounds good, updated in d2f69457dc46a9c00307c27faf9e4cde4a5896d0.

For reference, from Section `4.3.3 Attribute` of the [GLSL spec](https://www.khronos.org/registry/gles/specs/2.0/GLSL_ES_Specification_1.0.17.pdf):

 > Attribute variables cannot be declared as arrays or structures.

pjcozzi:
@lexaknyazev OK to merge this?

lexaknyazev:
Lgtm
Is it intended that this PR goes to `master`, while https://github.com/KhronosGroup/glTF/commit/d2f69457dc46a9c00307c27faf9e4cde4a5896d0 goes to `1.0.1`?

pjcozzi:
Yes, I think it is pretty clear-cut that this is just a clarification to the spec, where d2f69457dc46a9c00307c27faf9e4cde4a5896d0 is really a spec change.  It is a subtle call here, but I'd rather stay conservative when adding clarifications to the 1.0 spec.

pjcozzi:
@lexaknyazev is this ready to merge?

lexaknyazev:
:+1:

PR #615, opened 2016-06-02
Make indices accessor constraints explicit

pjcozzi:
Part of #563

>  There is no explicit requirement on the mesh.primitive.indices accessor to not to be FLOAT and to reference bufferView with ELEMENT_ARRAY_BUFFER target.

@lexaknyazev what do you think of this word?

Also, I think this is fine for the 1.0 spec since these are the constraints that are needed in practice; this is just a clarification and does not need to wait for 1.0.1

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -2385,6 +2385,8 @@ A dictionary object of strings, where each string is the ID of the accessor cont
 
 The ID of the accessor that contains the indices.  When this is not defined, the primitives should be rendered without indices using `drawArrays()`.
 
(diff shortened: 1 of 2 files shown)


lexaknyazev:
Lgtm
Btw should we also consider it in the context of uint indices (#570 and WebGL 2) to minimize future word changes?

pjcozzi:
> Btw should we also consider it in the context of uint indices (#570 and WebGL 2) to minimize future word changes?

Yes, this is next on my list - will hopefully work on it this afternoon.

Issue #610, opened 2016-06-02
WEB3D_quantized_attributes decodedMax/decodedMin/decodeMatrix constraints


pjcozzi:
@mlimper @lasalvavida what do you think about this feedback from @lexaknyazev (originally in #563)?

* [ ] 1. Should `decodedMax` and `decodedMin` array lengths be aligned with accessor's `type`?
* [ ] 2. Can we enforce `decodeMatrix` length to be just one of [4, 9 16] instead of [1-16]?

lasalvavida:
I agree with both of these

pjcozzi:
Sounds great, thanks for the fast response!

@lasalvavida can you open a pull request to update the extension spec?

lasalvavida:
@pjcozzi Can we close this?

pjcozzi:
Yes, fixed in #616.

Issue #588, opened 2016-05-02
WEB3D_quantized_attributes Specification


lasalvavida:
There are a few things that came up while writing support for the proposed [WEB3D_quantized_attributes](https://github.com/KhronosGroup/glTF/blob/master/extensions/1.0/Vendor/WEB3D_quantized_attributes/README.md) extension into Cesium, that I'd like to discuss here.

First off, the implementation can be found on this [branch](https://github.com/lasalvavida/cesium/tree/web3d-quantized-attributes) of my fork of Cesium, and any comments and suggestions are welcome in my [pull request](https://github.com/AnalyticalGraphicsInc/cesium/pull/3891).

So, firstly, the specification calls for the addition of a decodeMatrix, and optional decodeMin and decodeMax properties.

I'm not certain this is the best approach, largely because in practice, the formula with the decode matrix ends up looking like this:

![latex_4d3d54230e5ddd06cd376254f7054947](https://cloud.githubusercontent.com/assets/2489570/14960873/d67ac85c-1064-11e6-95d4-bc240a3b338c.png)

I can't think of a case where any individual component would combine with any of the others, so I think that this could be more concisely stated. Perhaps by just making the accessor max and min attributes mandatory if the extension is to be used, and then including precision information in the extension attributes.

@mlimper Your input on this would be greatly appreciated

MarcoHutter:
Just some remarks, summarized / derived from the README:

- The `min`/`max` accessor attributes (basically the bounding box of the model, when referring to the positions) are not necessarily equal to the decoding parameters.
- The [SRC format](http://x3dom.org/src/), where this quantization was originally proposed, indeed used only vector properties (`decodeOffset` and `decodeScale`, basically just a slightly different representation of the min/max)
- The justification for choosing a matrix was that it may easily be integrated into (i.e. multiplied with) the MODELVIEW matrix, so that the shaders remain unaffected by the quantization

However, I also think that the matrix could simply be *constructed* at runtime, just like you described, but I may have overlooked something here.

mlimper:
@lasalvavida Thanks a lot for the nice explanation, and thanks @MarcoHutter for your answers.

I guess marco summarized well what we also had in mind when writing the extension:

[...]

Issue #593, opened 2016-05-18
Require accessor min/max for POSITION?


pjcozzi:
See https://github.com/AnalyticalGraphicsInc/cesium/issues/3925

pjcozzi:
Labeling as resolved.  See fdf433752303593af461e565ee6efd035c318637 in the `1.0.1` branch.

lexaknyazev:
Just to be clear, why are `min|max` always required in https://github.com/KhronosGroup/glTF/commit/fdf433752303593af461e565ee6efd035c318637, not just for `POSITION`?
Are there any common use cases for `NORMAL` or `TEXCOORD`?

lexaknyazev:
One more possibly worth mentioning constraint: `min|max` length must be one of [1, 2, 3, 4, 9, 16] and match accessor's `type`.

pjcozzi:
> One more possibly worth mentioning constraint: min|max length must be one of [1, 2, 3, 4, 9, 16] and match accessor's type.

See #617.

> Just to be clear, why are min|max always required in fdf4337, not just for POSITION?
Are there any common use cases for NORMAL or TEXCOORD?

This is for simplicity and consistency.  Early on, we received some glTF feedback that *all* properties should be required.  We didn't agree as it can make things verbose; however, experience has shown that having some optional and some required properties has caused some implementation confusion for clients - which goes against one of our main goals: making glTF easy to implement consistency.  So I think requiring `min`/`max` only for `POSITION` starts to make the required/optional rules too complex and could lead to a less robust ecosystem.

lexaknyazev:
> So I think requiring min/max only for POSITION starts to make the required/optional rules too complex and could lead to a less robust ecosystem.

I was curious because of opposite approach in [#616 (diff)](https://github.com/KhronosGroup/glTF/pull/616/files#diff-b20cdac71cf65c6672d2cc43c3f88861R94).

pjcozzi:
Ah, yes, I think we should make [#616 (diff)](https://github.com/KhronosGroup/glTF/pull/616/files#diff-b20cdac71cf65c6672d2cc43c3f88861R94) apply to all attributes, not just `POSITION`.

@mlimper @lasalvavida what do you think?

>> Just to be clear, why are min|max always required in fdf4337, not just for POSITION?
Are there any common use cases for NORMAL or TEXCOORD?

[...]

PR #616, opened 2016-06-03
Updated quantized attributes extensions

lasalvavida:
For #610 and #588 
@pjcozzi @mlimper

### DIFF
--- a/extensions/Vendor/WEB3D_quantized_attributes/README.md
+++ b/extensions/Vendor/WEB3D_quantized_attributes/README.md
@@ -90,8 +90,8 @@ As already discussed in section [Choosing a Range for Normalization](#choosing-a
 To obtain information such as bounding boxes, glTF accessors already contain the optional `min` and `max` values.
 However, this values might not have been provided, and even if they are provided, they apply to the encoded, compressed data.
 One could obtain original values by multiplying the decode matrix with given `min` and `max` values, but it might desirable to save the overhead of matrix multiplication.
-Therefore, this extension introduces two new, optional properties,`decodedMin` and `decodedMax`, which contain the extreme values for the decoded accessor. These properties are required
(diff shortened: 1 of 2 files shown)


pjcozzi:
+1 from me.

@lasalvavida also add an `Acknowledgements` section at the bottom (like the glTF spec) and include @lexaknyazev.

@mlimper can you confirm and I will merge?

lasalvavida:
Updated

mlimper:
Confirmed, thanks for updating and merging

pjcozzi:
@mlimper did you also see the discussion in https://github.com/KhronosGroup/glTF/issues/593#issuecomment-223651243 where I want to change

> These properties are required if the quantized attribute uses the POSITION semantic. This simplifies the client implementation where `min` and `max` are used to compute bounding volumes.

to always requiring `decodedMax` and `decodedMin` since `max`/`min` will always be required in glTF 1.0.1.  What do you think?

mlimper:
It doesn't seem to cause a lot of work for writer implementations, but it looks like it would make reader implementations more straightforward.
+1!

pjcozzi:
> It doesn't seem to cause a lot of work for writer implementations, but it looks like it would make reader implementations more straightforward.
+1!

Great!

@lasalvavida can you make the update?

Is anything else needed before merging?

lasalvavida:
If nobody else has any comments, this should be ready to go.

pjcozzi:
Thanks for all your work on this @lasalvavida and @mlimper!

lexaknyazev:
Should [L96](https://github.com/KhronosGroup/glTF/blame/d17eb2b1f7d6a7c14709ac0925cb4bdecb105ef0/extensions/Vendor/WEB3D_quantized_attributes/README.md#L96) say "two" instead of "three", or one example is missing?

lasalvavida:
Nope I missed that, good catch @lexaknyazev. There were three, but there are two now. I will put in a PR for that.

PR #617, opened 2016-06-03
More specific constraints for accessor min and max

pjcozzi:
For https://github.com/KhronosGroup/glTF/issues/593#issuecomment-223636136

@lexaknyazev what do you think?

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -1246,14 +1246,14 @@ Specifies if the attribute is a scalar, vector, or matrix, and the number of ele
 
 ### accessor.max
 
-Maximum value of each component in this attribute.  When both min and max arrays are defined, they have the same length.  The length is determined by the value of the type property.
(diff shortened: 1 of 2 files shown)


lexaknyazev:
Maybe it's too obvious, but should we also demand all values of `min|max` array to align with `accessor.componentType`?

pjcozzi:
It is implicit that the min/max values are bound to the size of `componentType` since they are the min/max of the components which are all of `componentType`.  I don't think we should define them as having the same datatype since they are really a JSON number not a GL datatype.

lexaknyazev:
Sounds fair since `min|max` values are computed from actual typed data.
Will keep validator's check on this anyway.

pjcozzi:
Thanks!

lexaknyazev:
Since both properties are required now, maybe we can replace
>  When both min and max arrays are defined, they have the same length.
with something like
>  Both min and max arrays have the same length.

pjcozzi:
Agreed.  Update the `1.0.1` branch in 16ace82b9bdc1604cee7f7c7f1f782a253bbe268.

PR #619, opened 2016-06-07
Draft of uint32 support

pjcozzi:
PR into `1.0.1` branch for #570.

Adds `webglExtensionsUsed` property and `5125` (UNSIGNED_INT) `accessor.componentType` value.

For example:
```json
{
    "webglExtensionsUsed " : ["OES_element_index_uint"],
    "accessor_id": {
        "bufferView": "bufferView_id",
        "byteOffset": 0,
        "byteStride": 0,
        "componentType": 5125,
        "count": 36,
        "type": "SCALAR"
    }
}
```

### DIFF
--- a/specification/schema/examples/glExtensionsUsed.json
+++ b/specification/schema/examples/glExtensionsUsed.json
@@ -1,5 +1,5 @@
 {
-    "webglExtensionsUsed " : ["OES_element_index_uint"],
+    "glExtensionsUsed " : ["OES_element_index_uint"],
     "accessor_id": {
         "bufferView": "bufferView_id",
         "byteOffset": 0,
--- a/specification/schema/glTF.schema.json
+++ b/specification/schema/glTF.schema.json
@@ -199,7 +199,7 @@
             "default" : {},
             "gltf_detailedDescription" : "A dictionary object of textures.  The name of each texture is an ID in the global glTF namespace that is used to reference the texture."
         },
-        "webglExtensionsUsed" : {
+        "glExtensionsUsed" : {
             "type" : "array",
             "description" : "Names of WebGL extensions required to render this asset.",
             "items" : {
--- a/specification/schema/glTF.schema.json
+++ b/specification/schema/glTF.schema.json
@@ -7,13 +7,12 @@
     "properties" : {
         "extensionsUsed" : {
             "type" : "array",
-            "description" : "Names of extensions used somewhere in this asset.",
+            "description" : "Names of glTF extensions used somewhere in this asset.",
             "items" : {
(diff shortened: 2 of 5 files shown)


lexaknyazev:
What do you think of dropping `web` from `webglExtensionsUsed`? glTF was webglTF once.

pjcozzi:
Good call.  Updated.

pjcozzi:
@lexaknyazev any other feedback here?

lexaknyazev:
This PR defines `glExtensionsUsed` with only one possible value (`["OES_element_index_uint"]`).
Maybe it's OK for now (to merge PR) but should we consider other possibly valid values (like `OES_standard_derivatives`)?

pjcozzi:
> should we consider other possibly valid values (like OES_standard_derivatives)?

We should at some point, but I think it is outside the scope of 1.0.1.

We just want to be confident that this new schema will extend well, which I think it does, but, of course, some WebGL extension will also require glTF extensions to extend the schema.

pjcozzi:
@lexaknyazev is this ready to merge into the `1.0.1` branch?

lexaknyazev:
:+1:

specification/schema/glTF.schema.json
```diff
@@ -199,6 +198,18 @@
...
             "default" : {},
             "gltf_detailedDescription" : "A dictionary object of textures.  The name of each texture is an ID in the global glTF namespace that is used to reference the texture."
+        },
+        "glExtensionsUsed" : {
+            "type" : "array",
+            "description" : "Names of WebGL extensions required to render this asset.",
+            "items" : {
+                "type" : "string"
+            },
+            "enum" : ["OES_element_index_uint"],
```
nehmebilal: This enum should be inside "items"

specification/schema/examples/glExtensionsUsed.json
```diff
@@ -0,0 +1,11 @@
+{
+    "glExtensionsUsed " : ["OES_element_index_uint"],
```
nehmebilal: there is a white space at the end of "glExtensionsUsed "

pjcozzi:
Good eye, @nehmebilal.

@lexaknyazev could update directly in the `1.0.1` branch?  No need for a PR for such minor changes.

lexaknyazev:
Done.
https://github.com/KhronosGroup/glTF/commit/f405891c5df97076a2fd2684c7469343e11292ac https://github.com/KhronosGroup/glTF/commit/ecfc8f69bb5c6e5d92aaf995831c7287d8083b50

PR #622, opened 2016-06-10
'Three' should be 'Two'

lasalvavida:
From @lexaknyazev on #616


pjcozzi:
Thanks @lasalvavida and @lexaknyazev!

PR #636, opened 2016-07-01
Check for unbound textures

### LINKED ISSUE #635: CesiumMan doesn't have texture coordinates after conversion
[CesiumMan.zip](https://github.com/KhronosGroup/glTF/files/343904/CesiumMan.zip)

lasalvavida:
Fixes #635. CesiumMan converts correctly after this change and CesiumMilkTruck still has its unused TEXCOORD fields removed.

If you look at the collada between these two models the issue is that in CesiumMilkTruck, the texture is bound to a texture coordinate set:

```xml
<diffuse>
    <texture texture="Image0001-sampler" texcoord="Texture" />
</diffuse>
```
while in CesiumMan, it is not:

```xml
<diffuse>
    <texture texture="Cesium_Man_jpg-sampler"/>
 </diffuse>
```

The TEXCOORD removal optimization now jumps ship if a texture is unbound since there's no way to tell what is used and what isn't.


pjcozzi:
Thanks for tracking this down, @lasalvavida!

@tfili can you please review and merge?

tfili:
Looks good to me. Thanks @lasalvavida.

pjcozzi:
Thanks for the quick fix, @lasalvavida!

Issue #639, opened 2016-07-06
Required top-level object properties


pjcozzi:
For example, just asset metadata and geometry objects so a valid glTF asset doesn't need to include materials and friends.

@lexaknyazev would you be able to open a PR into the `1.0.1` branch with a proposal?

lexaknyazev:
> would you be able to open a PR into the `1.0.1` branch with a proposal?

As soon as we agree on specifics:

- `asset` dictionary is required (this implies existence of `asset.version`).
- `meshes` dictionary is required and must not be empty.
- `mesh.primitives` array is required and must not be empty. (Btw, what's the point in empty array here?)
- `[mesh.]primitive.attributes` dictionary is required. (Again, why isn't it already?) This implies requirements for non-empty `accessors`, `bufferViews` and `buffers` dictionaries.
- `[mesh.]primitive.material` is not required.

Not sure about `nodes`.

@pjcozzi What do you think?

---

Is there any valid use case for explicit `null` anywhere in glTF?

pjcozzi:
All looks good.  As for the previously optional properties, I think we were just being too aggressive with defaults.

I am tempted to require `nodes`, but there have been requests to use glTF for purely geometry, e.g., in one global coordinate system.

> Is there any valid use case for explicit null anywhere in glTF?

What do you mean?  Do you have an example?

lexaknyazev:
>> Is there any valid use case for explicit `null` anywhere in glTF?

> What do you mean? Do you have an example?

JSON values can be `null` (instead of `undefined`). It could be used as a placeholder for app-specific usage or as a stub for required fields. For example, `buffer.uri` for `KHR_binary_glTF` could be like this:

```js
{
  "buffers": {
    "binary_glTF": {
      "byteLength": 12345,
      "uri": null // instead of "data:,"
    }
}
```

lexaknyazev:
At the current state of spec, default material is defined as a material without `technique` (or `extensions`).
Should `primitive` without `material` property be rendered with default material or should it be excluded from rendering? 

Since default material doesn't have any `values`, I propose to refine that part like this:

- `primitive` with no `material` renders with default material (here's one more possible use case for `null`: use `"material": null` to disable object rendering).

- `material` can't have `values` if `material.technique` is undefined.

pjcozzi:
> Is there any valid use case for explicit null anywhere in glTF?

[...]

PR #642, opened 2016-07-08
Mandatory dictionaries

lexaknyazev:
See #639.
Also added a note about uint indices.


pjcozzi:
Very nice, thanks @lexaknyazev!

Issue #605, opened 2016-06-01
Spec changes from 1.0 to 2.0


pjcozzi:
This issue will be updated as we work on the 1.1 spec (also see the spec pull request, #784)

**1.0.1/1.1 was renumbered to 2.0**

**JSON format changes:**
- Removed indirection from `animation`: `sampler` now refers directly to accessors, #712 
- `material.parameter.value` and `technique.parameter.value` must be an array, #690
- Node can have only one mesh #821
- Top-level JSON objects are arrays now
- Added reqs on JSON encoding
- Added reqs on binary data alignment https://github.com/KhronosGroup/glTF/issues/802#issuecomment-270406859

**Additions:**
- Added `accessor.normalized`, #691, #706
- Added `"STEP"` as valid `animation.sampler.interpolation` value, #712 
- Added `glExtensionsUsed` property and `5125` (UNSIGNED_INT) `accessor.componentType` value, #619
- Added `extensionsRequired` property, #720, #721

**Removals:**
- Removed revision number from `profile.version`, #709
- Removed `buffer.type`, #786, #629
- Removed `technique.functions.scissor` and removed `3089` (`SCISSOR_TEST`) as a valid value for `technique.states.enable`, #681 
- Techniques, programs, and shaders were moved out to KHR_technique_webgl extension.

**Other edits:**
- `asset` is now required, #642
- `accessor.min` and `accessor.max` are now required, #593, and clarified that the JSON value and binary data must be the same, #628.
- Clarified `animation.sampler` and `animation.channel` restrictions, #712 
- `buffer.byteLength` and `bufferView.byteLength` are now required, #560.
- `skin.inverseBindMatrices` is now optional, #461.
- Attribute parameters can't have a `value` defined in the `technique` or `parameter`, https://github.com/KhronosGroup/glTF/issues/563#issuecomment-223155322.
- Only `TEXCOORD` and `COLOR` attribute semantics can be written in the form `[semantic]_[set_index]`, https://github.com/KhronosGroup/glTF/issues/563#issuecomment-223270789.
- `TEXCOORD` and `COLOR` attribute semantics **must** be written in the form `[semantic]_[set_index]`, e.g., just `TEXCOORD` should be `TEXCOORD_0`, and just `COLOR` should be `COLOR_0`, #649
[...]

PR #643, opened 2016-07-08
[WIP] PBR Materials extension

mlimper:
This is our (currently WIP) PBR material extension - this PR is currently mainly intended to serve as a place for discussion.

Direct Preview:
https://github.com/tsturm/glTF/tree/master/extensions/Vendor/FRAUNHOFER_materials_pbr


selim-bekkar-sb:
Cool ! Merci du partage, j’implémente ça de ce pas.

pjcozzi:
@erich666 would you or any of your colleagues be interested in reviewing this proposed PBR materials extension to glTF?

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
```diff
@@ -0,0 +1,146 @@
...
+* Maik Th&ouml;ner, Fraunhofer IGD, [@mthoener](https://twitter.com/mthoener)
+* Max Limper, Fraunhofer IGD, [@mlimper_cg](https://twitter.com/mlimper_cg)
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 1.0 spec.
```
pjcozzi: You'll want this to be using the 1.0.1 spec, which may require no changes to the extension.  We plan to finish 1.0.1 for SIGGRAPH.  See #605.

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Overview
+
+The glTF 1.0 allows the definition of materials by instancing `techniques`. A `technique`, as defined in glTF 1.0, is a verbose description of shader parameters combined with actual shader code. Typically, shader code is engine specific and, as such, may not be used across systems which do not share the same rendering pipeline.
```
pjcozzi: To help better position `FRAUNHOFER_materials_pbr` in the glTF ecosystem, consider adding a sentence here about the [KHR_materials_common](https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_materials_common) extension, which defines non-PBR materials.

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+
+## Material models
+
+<p style="text-align:justify;">A material model defines a set of parameters used to describe a material. This extension supports two material models commonly used in PBR, namely the specular-glossiness and the metal-roughness models. An implementation of this extension should support both models.</p>
+
+### Specular - Glossiness
+
+<img src="figures/specular_glossiness_2.png" align="middle" height="220" style="display: block; margin: 0 auto; padding: 20px 0 10px 0;">
```
pjcozzi: Nice figure.  Did you create it?  If not, do we need permission to use it here?

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
```diff
@@ -0,0 +1,146 @@
...
+* Specular:   Specular color of the material
+* Glossiness: Glossiniess of the material
+
+<p style="text-align:justify;">The diffuse value represents the reflected diffuse color of the material. Raw metals have a diffuse value of (0, 0, 0). The specular value defines specular reflectance at normal incidence (F0). The Glossiness value is a factor between 0.0 (rough surface) and 1.0 (perfectly smooth surface) and represents the surface irregularities that cause light diffusion. Figure 1 shows the three components of the specular-glossiness model and the rendered result.</p>
+
+The following table lists the allowed types and ranges for the specular-glossiness model:
+
+| Property | Type | Range | Default Value | Description |
+|:------------:|:----:|:-----:|:-:|:-----:|:-----------:|
+| `diffuseFactor`   | `FLOAT_VEC4` | [0, 1] for all components | [1,1,1,1] | The RGB components of the reflected diffuse color of the material. For raw metals the diffuse color is black (0.0). The fourth component (A) is the `opacity` of the material. |
```
pjcozzi: I don't have experience with PBR materials, but I just want to confirm that it is common to use the alpha channel of the diffuse factor for opacity.

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
```diff
@@ -0,0 +1,146 @@
...
+| `specularFactor`  | `FLOAT_VEC3` | [0, 1] for all components | [1,1,1]   | The specular RGB color of the material. |
+| `glossinessFactor`| `FLOAT`      | [0, 1]                    | 1         | The glossiness of the material surface (0 is glossiness, 1 is full glossiness). |
+| `diffuseTexture`  | string       | valid texture identifier  |           | Texture with RGB components of the reflected diffuse color of the material. For raw metals the diffuse color is black (0.0). If the fourth component (A) is present, it represents the `opacity` of the material. Otherwise, an `opacity` of 1 is assumed. |
```
pjcozzi: Here and throughout, "valid texture identifier" means "id of the glTF texture", right?  The glTF spec says `id`, not `identifier`, and we can minimize confusion by being consistent here.

pjcozzi: > "Texture with RGB components"

This would be more precisely written as "Texture with RGB or RGBA components"

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+<strong>Usage Example:</strong>
+
+```javascript
+"materials": {    
+    "rough_gold2": {
+        "extensions": {
+            "KHR_materials_pbr" : {
+                "technique" : "PBR_specular_glossiness",
```
pjcozzi: It would be insightful to include the `PBR_specular_glossiness` technique in this example.

pjcozzi: Ah, I think I see now; this identifies the shading model.  The extension spec should explain this before showing the example.  It would also be useful to show texture references here so readers see a complete example.

pjcozzi: `"technique"` might also not be the best name given it is easily confused with glTF's `technique`.  What is the best PBR term?  `"shadingModel"` or `"materialModel"`?  The allowed values could be `"SPECULAR_GLOSSINESS"` and `"METAL_ROUGHNESS"` to be consistent with other glTF string enums like `accessor.type`.

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
pjcozzi: Same comment below, of course.

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
```diff
@@ -0,0 +1,146 @@
...
+| `baseColorTexture`          | string       | valid texture identifier  |           |Texture with the RGBA components of the base color (RGB) of the material. If the fourth component (A) is present, it represents the `opacity` of the material. Otherwise, an `opacity` of 1 is assumed. |
+| `metallicRoughnessTexture`  | string       | valid texture identifier  |           |Texture with two (or more) components, containing the metallic-ness of the material (first component) and its roughness (second component). |
+
+The factors (baseColorFactor, metallicFactor, roughnessFactor) scale the components given in the respective textures (baseColorTexture, metallicRoughnessTexture).
+If a texture is not given, all respective texture components are assumed to have a value of 1. This is similar to the handling of factors and texture within the specular-glossiness model.
+
+
+<strong>Usage Example:</strong>
+
+```javascript
```
pjcozzi: Same comments as above for this example.

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
```diff
@@ -0,0 +1,146 @@
...
+* [specular-glossiness](schema/materials_pbr_specular_glossiness.schema.json)
+* [metal-roughness](schema/materials_pbr_metal_roughness.schema.json)
+
+
+## Appendix
+
+An introduction to PBR concepts is provided in the [appendix](Appendix.md).
+
+
+## Known Implementations
```
pjcozzi: Would you consider contributing a stage to [gltf-pipeline](https://github.com/AnalyticalGraphicsInc/gltf-pipeline) to convert a glTF model with PBR materials using this extension to a glTF model with generated materials/techniques/shaders?

This would:
* Create a low barrier-to-entry for users to get started with this extension
* Serve as a reference implementation for those implementing the extension (JavaScript developers could probably use parts of it as is)
* Help ensure consistent implementations of this extension across engines

extensions/Vendor/FRAUNHOFER_materials_pbr/schema/materials_pbr_metal_roughness.schema.json
```diff
@@ -0,0 +1,33 @@
+{
+    "$schema" : "http://json-schema.org/draft-03/schema",
+    "title" : "material",
+    "type" : "object",
+    "description" : "An attribute or uniform input to a technique, and an optional semantic and value.",
+    "extends" : { "$ref" : "glTFProperty.schema.json" },
+    "properties" : {
+    	"baseColorFactor" : {
+    		"type" : { "$ref" : "arrayValues.schema.json" },
```
pjcozzi: Mixed tabs and spaces here?

extensions/Vendor/FRAUNHOFER_materials_pbr/schema/materials_pbr_metal_roughness.schema.json
```diff
@@ -0,0 +1,33 @@
+{
+    "$schema" : "http://json-schema.org/draft-03/schema",
+    "title" : "material",
+    "type" : "object",
+    "description" : "An attribute or uniform input to a technique, and an optional semantic and value.",
```
pjcozzi: Bad copy and paste?

extensions/Vendor/FRAUNHOFER_materials_pbr/schema/materials_pbr_metal_roughness.schema.json
```diff
@@ -0,0 +1,33 @@
+{
+    "$schema" : "http://json-schema.org/draft-03/schema",
+    "title" : "material",
```
pjcozzi: Same comment.

extensions/Vendor/FRAUNHOFER_materials_pbr/schema/materials_pbr_metal_roughness.schema.json
```diff
@@ -0,0 +1,33 @@
...
+    "type" : "object",
+    "description" : "An attribute or uniform input to a technique, and an optional semantic and value.",
+    "extends" : { "$ref" : "glTFProperty.schema.json" },
+    "properties" : {
+    	"baseColorFactor" : {
+    		"type" : { "$ref" : "arrayValues.schema.json" },
+    		"description" : "The material's base color factor.",
+            "default" : [1,1,1,1]
+    	},
+        "metallicFactor" : {
```
pjcozzi: Throughout, JSON schema can define min/max (exclusive and inclusive) values.  For example, see https://github.com/KhronosGroup/glTF/blob/master/specification/schema/camera.perspective.schema.json#L14-L27

extensions/Vendor/FRAUNHOFER_materials_pbr/schema/materials_pbr_specular_glossiness.schema.json
```diff
@@ -0,0 +1,33 @@
+{
```
pjcozzi: Same comments in this schema as well.

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
pjcozzi: Same comments for the equation figures in Appendix.md.

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
```diff
@@ -0,0 +1,146 @@
...
+## glTF Schema
+
+
+* [specular-glossiness](schema/materials_pbr_specular_glossiness.schema.json)
+* [metal-roughness](schema/materials_pbr_metal_roughness.schema.json)
+
+
+## Appendix
+
+An introduction to PBR concepts is provided in the [appendix](Appendix.md).
```
pjcozzi: This is a nice overview; perhaps link to it again in the intro for readers who need an overview of PBR.

pjcozzi:
That's all my initial comments.  Once there's more community feedback, I can help do editing if you want.

vorg:
Hi, great work! What's the motivation for supporting both material models (specular-glossiness and the metal-roughness)? I did a research when deciding on the implementation myself few months back and metallic was definitely winning as it takes less memory.

screen shot 2016-07-09 at 20 01 47

selim-bekkar-sb:
Hello, nice work !

I have a comment:

- Why force to merge the Specular and Glossiness and or the Metallic and Roughness textures ? Is it to decrease the server request number ? Memory gain in graphic card ? Or a particular reason ?
It's common for artists to work on each channel separately (as you show in your schema). 
So I don't think that's a good idea to merge this channels because we could loose clarity. And in some particular case loose gain. (if 5 different materials shared the same roughness texture but not the same metallic texture, thus the roughness texture will be present 5 time instead of storing/loading it only 1 time) So maybe allow the both way (merged or not) could be useful.

pjcozzi:
> Hi, great work! What's the motivation for supporting both material models (specular-glossiness and the metal-roughness)?

@vorg thanks for the review.  I was actually wondering the same, e.g., is the better served with two extensions or just one shading model.  I don't know enough about PBR to provide advice here.

pjcozzi:
> Why do you force to merge the Specular and Glossiness and or the Metallic and Roughness textures ? Is it to decrease the server request number ?

@selim-bekkar-sb thanks for the review!

The spirit of glTF is that it is easy and fast for an engine to load and render; it does not need to match how a modeling tool wants to store data since it is a "final stage" format that is intended for engines, not modeling tools.

With that said, there may be other optimal and still clear ways to organize the data, and that is up to the extension authors.

jeffdr:
> Hi, great work! What's the motivation for supporting both material models (specular-glossiness and the metal-roughness)?

In theory a tool could automatically convert assets from metal-roughness to specular-glossiness, but the reverse is not possible in all cases. Spec/gloss can do certain combinations that metalness can't, and it can be hard to infer a metalness value from spec/gloss. There are also some potential artifacts with the interpolation of metalness values along boundaries that some people do not like. So if you really wanted to have just one, you should stick with spec/gloss.

That said I think both should be supported here in some fashion, given the memory benefits metalness provides as well as the general popularity of that workflow.

jeffdr:
Also, some thought should be given to how gloss and roughness values are to be interpreted in some specific BRDFs. These [0,1] values can be used in quite different ways, depending on what encoding is convenient. You might want as part of the specification to explain how such values might be used in, say, a GGX or Phong functions.

pjcozzi:
Really insightful feedback, thanks @jeffdr!  It sounds like keeping both material models in this extension is the way to go.

jeffdr:
One more thought: I'm not sure if this is implicit from the glTF spec this is written against already, but this document should probably explicitly mention that the diffuse, specular and baseColor RGB values should always be in the sRGB color space (and that gloss and opacity values should be linear). This confers the usual benefits of fidelity, reduction in banding, and so on.

mlimper:
> That said I think both should be supported here in some fashion, given the memory benefits metalness provides as well as the general popularity of that workflow.

Thanks a lot for sharing your thoughts on this topic!
We did receive some other feedback saying people would actually prefer to have only one parameter set instead of two alternatives. Both approaches have advantages and disadvantages.

I just opened a ticket to discuss the parameter set separately.
https://github.com/KhronosGroup/glTF/issues/696

mlimper:
More points for separate discussion:
https://github.com/KhronosGroup/glTF/issues/697 (example / reference code)
   https://github.com/KhronosGroup/glTF/issues/698 (specify more precisely how textures / variables interact)
   https://github.com/KhronosGroup/glTF/issues/699 (bump maps and normal maps)
   https://github.com/KhronosGroup/glTF/issues/700 (gamma / sRGB)

extensions/Vendor/FRAUNHOFER_materials_pbr/README.md
mlimper: Yes, we created those figures (also for the equations). We used them in the respective ACM Web3D paper, so we should double-check that it is OK to use them for non-commercial purposes. I would guess so, but I am not 100% sure, so: good point, we need to check (or simply create replacements specifically for this extension).

mlimper:
@pjcozzi Thanks for the first review, most of the points are now incorporated into the draft

cedricpinson:
What is missing to merge it to the spec ? We would like to try it but needs PBR materials. Any timeline to have it merged ?

mlimper:
We need to clarify the issues mentioned above (#696, #697, #698, #699, #700) - then we're probably ready to merge. Please feel free to join the discussions!

erich666:
> We used them in the respective ACM Web3D paper, so we should double-check that it is OK to use them for non-commercial purposes. I would guess so, but I am not 100% sure, so: good point, we need to check (or simply create replacements specifically for this extension).

By default the ACM owns your images if you publish with them. My advice to people is to make your own copies, when it comes to 3D content: http://www.realtimerendering.com/blog/you-may-want-to-own-your-own-images/

The ACM has loosened its policy on images, but you have to claim copyright at the time of publication: http://www.realtimerendering.com/blog/loosening-of-acms-copyright-policy/

mlimper:
Wow, thanks for the nice pointers, Eric!

I would probably also go for the "make another" solution, that shouldn't be too much effort - can be done nicely in conjunction with the public demo application, for example.

mlimper:
Thanks for the feedback, and thanks for providing the links.

I just adapted the example glTF model a bit to use tangents and binormals (generated with a slightly modified version of gltf-pipeline... will do a separate PR at some point). With that, I do currently still have a non-PBR (blinn-phong) shading, but using the provided tangent-space normal map:

hammer

Next, I'll update the shader to use a true PBR shading model.

@emackey Pleased to see the Fire Hydrant model... I have been searching for an example object that is not a weapon, which was maybe only half successful ;-)

> Also, a simpler shader a la this one lets people mess with the shader directly, to understand it (though you can add textures for some channels, but there's a powerful shader recompiler powering all that). Parameters baked in to the textures is pretty to look at, but less educational.

I agree that there should also be a more simple example, completely without textures, so I'd go for both: one example that is really as simple as possible, one more advanced example with textures, and maybe even one with IBL and prefiltered environment maps (later).

pjcozzi:
> generated with a slightly modified version of gltf-pipeline... will do a separate PR at some point

You rock, that would be fantastic and much appreciated.

mlimper:
(continuing discussion regarding the example in #697)

extensions/Vendor/FRAUNHOFER_materials_pbr/example/untextured-pbr/index.html
```diff
@@ -0,0 +1,462 @@
...
+                // material definition
+
+                vec4 diffuseColor = vec4(diffuse, 1.0);
+
+                float roughnessFactor = roughness;
+                float metalnessFactor = metalness;
+
+                PhysicalMaterial material;                
+                material.diffuseColor      = diffuseColor.rgb * ( 1.0 - metalnessFactor );
+                material.specularRoughness = clamp( roughnessFactor, 0.04, 1.0 );
```
sbtron: Really great work here with the implementation. It makes it really easy to understand how to go from the equations to the implementation. 
I am trying to understand why roughness min is set to 0.04? Is this just a temporary workaround to avoid hiding the directional lights when roughness is 0?  
Thanks

mlimper: Thanks!

The simple answer would be "It was copied like that from the Three.JS PBR code, which served as a basis". The more satisfying answer, however, is that this value is a commonly used approximation for real-world dielectric materials. Quoting the allegorithmic PBR guide:

> The value for common dielectrics is around 2-5% based on the F0 as computed by the Index of Refraction (IOR)

In this version, that quote and the respective explanation are on page 10:
https://www.allegorithmic.com/system/files/software/download/build/PBR_Guide_Vol.1.pdf

sbtron: I think you were referring to the next line for the 4% specular coefficient ? 
material.specularColor     = mix( vec3( DEFAULT_SPECULAR_COEFFICIENT ), diffuseColor.rgb, metalnessFactor );

I was curious about the roughness value and the min of 0.04: 
material.specularRoughness = clamp( roughnessFactor, 0.04, 1.0 );
The roughness shouldn't be related to the Fresnel calculations?

sbtron: Was able to track this down in three.js. Looks like it used to be some different values -https://github.com/mrdoob/three.js/blob/b1df0234204f9a43eaa3c80da91803e87ded52d8/src/renderers/shaders/ShaderChunk/lights_physical_fragment.glsl#L3 
Based on disney's remapping - http://blog.selfshadow.com/publications/s2012-shading-course/burley/s2012_pbs_disney_brdf_notes_v3.pdf 
Page 16, Specular G details: "This remapping was based on comparisons with measured data as well as artist feedback that the specular was just “too hot” for small roughness values." 

Three.js was then updated with the new mapping here - https://github.com/mrdoob/three.js/blob/5a6df635c32071bec92df393f85c0d72ba34b560/src/renderers/shaders/ShaderChunk/lights_physical_fragment.glsl#L3

I am assuming the update was for similar reasons based on artist feedback. The 0.04 value confused me too since it happens to be the same as Specular Coefficient but they seem to exist for very different reasons?

mlimper: Thanks for the very nice summary, and for the good catch!

It would be interesting to know where the new mapping used by Three.JS (and by the example application) actually comes from

@WestLangley Could you explain why this was changed from the disney remapping to the current version for specularRoughness, and where the 0.04 value comes from in this case? Thanks a lot in advance!

WestLangley: @mlimper See https://github.com/mrdoob/three.js/pull/7420

mlimper:
@erich666 @cedricpinson @bghgary @sbtron 
Just added you as contributors (see [here](https://github.com/tsturm/glTF/tree/master/extensions/Vendor/FRAUNHOFER_materials_pbr)) - hope that's fine for your? Let me know if you like to have any changes / additions (such as twitter account, additional affiliations, etc.).

@pjcozzi 
This is - fortunately - becoming a multi-vendor proposal. What do you think would be the best way of naming it? Would it be an option to change it to `EXT_materials_PBR` right away?

pjcozzi:
> This is - fortunately - becoming a multi-vendor proposal. What do you think would be the best way of naming it? Would it be an option to change it to EXT_materials_PBR right away?

Yes.  Also, given that some contributors are Khronos members, it is likely that it could be `KHR_materials_PBR`.  Let me confirm.

cedricpinson:
@mlimper would be great if you could add my twitter contact https://twitter.com/trigrou

pjcozzi:
One of my students, @jian-ru, added support for the current version of this extension to his Vulkan engine: https://github.com/jian-ru/laugh_engine#gltf-support

pjcozzi:
Also WIP WebGL glTF PBR reference implementation (raw WebGL, no engine) by @moneimne is here: http://www.seas.upenn.edu/~moneimne/WebGL-PBR/

Repo: https://github.com/moneimne/WebGL-PBR

AurL:
From https://github.com/KhronosGroup/glTF/issues/817
> We have also decided that the metallic-roughness PBR material model will be in the core glTF spec (specular-glossiness will remain in the extension)

Is there any material schema example that shows the diff between a metal-roughness material vs a specular-glossiness one ?

It seems to be a small change regarding the schema but as I am working on an exporter using PBR, I would like to be sure that it's done well. From what I understand, only the specular-glossiness material description would be under as `EXT_materials_PBR` extension ?

pjcozzi:
@AurL the JSON schemas are here:
* [materials_pbr_metal_roughness.schema.json](https://github.com/tsturm/glTF/blob/de6f79617c61ade232bf0302b3da247fafeb0419/extensions/Vendor/FRAUNHOFER_materials_pbr/schema/materials_pbr_metal_roughness.schema.json)
* [materials_pbr_specular_glossiness.schema.json](https://github.com/tsturm/glTF/blob/de6f79617c61ade232bf0302b3da247fafeb0419/extensions/Vendor/FRAUNHOFER_materials_pbr/schema/materials_pbr_specular_glossiness.schema.json)

@mlimper other than moving metallic-roughness into core spec, are these fully up-to-date?

mlimper:
> other than moving metallic-roughness into core spec, are these fully up-to-date?

Three points coming to my mind right now:

* @sbtron, @bghgary and colleagues suggested to make some additional comments, such as specifying priorities for the different maps, which should help a client that does not support the required number of channels.

* There should be the possibility to use an F0 texture, to specify F0 values for non-metallic parts. That case is a bit special, but we should cover it.

* I'll provide a separate document for additional maps (normals, occlusion), and I would like to link to that one in both specs, `materials_common` and `materials_pbr`.

The last point will take some more days, but the first two points can be done relatively quickly.

@bghgary Would it be possible for you and your colleague (Sebastien, I believe) to specify the prioritization of the different maps as we discussed it and open a PR for that?

pjcozzi:
If there is anything @moneimne can do to help finish the spec/extension, please let us know.

bghgary:
Yes, though for the PR, I would expect this information to go with the separate document for additional maps that you mentioned. I can wait until you are done or you can just incorporate this into your document directly.

List of additional maps in priority order:

| Map | What happens if it’s not there |
|:--:|:--:|
| Tangent Space Normal | Geometry will appear less detailed than authored
| Ambient Occlusion | Model will appear brighter in areas that should be darker
| Emission	 | Model with lights will not be lit (e.g. the headlights of a car model will be off instead of on)
| F0 (for dielectrics) | Some non-metallic materials, such as ruby and diamond, will be less shiny than it’s supposed to be

Resource-bound implementations should drop maps from the bottom to the top.

mlimper:
> Yes, though for the PR, I would expect this information to go with the separate document for additional maps that you mentioned. I can wait until you are done or you can just incorporate this into your document directly.

Yes, I'll provide that one. Please feel free to open the PR and already incorporate the table into the spec right now.

javagl:
@selim-bekkar-sb mentioned in https://github.com/KhronosGroup/glTF/pull/643#issuecomment-231550995

> Why do you force to merge the Specular and Glossiness and or the Metallic and Roughness textures ?

The answer by @pjcozzi  in https://github.com/KhronosGroup/glTF/pull/643#issuecomment-231551617 was:

> The spirit of glTF is that it is easy and fast for an engine to load and render; it does not need to match how a modeling tool wants to store data since it is a "final stage" format that is intended for engines, not modeling tools.

Is it safe to assume that combining these textures will be *enforced*? 

I'm asking because, for example, the "Helmet" test model and the work in https://github.com/moneimne/WebGL-PBR in general seem to operate on separate textures. I'd like to create a "baseline" version (not an "official reference" - only for personal use - but to be published at https://github.com/javagl/gltfTestModels/tree/master/DamagedHelmetModified ), and now assume that I'd create a combined version of the respective textures.

mlimper:
> Is it safe to assume that combining these textures will be enforced?

Yes, and also agreed with the comment of @pjcozzi

Imagine we would allow both, we would have the risk that each combination "exporter/renderer" implements only one of the possible ways, which would lead to assets not being fully interchangeable across engines

@moneimne
you probably used the separate textures as an intermediate step towards the final version of the demo, I guess?

javagl:
So then I hope that https://github.com/javagl/gltfTestModels/blob/master/DamagedHelmetModified/glTF/Default_MetalSmooth_metallicRoughness.jpg can serve as a basis for further tests.

moneimne:
>you probably used the separate textures as an intermediate step towards the final version of the demo, I guess?

Yes, I have been using the Helmet test model as is for now, but I can definitely use the combined version of the model once the changes have been made.

Thank you for the combined texture, @javagl! Do you also plan on updating the [glTF file](https://github.com/javagl/gltfTestModels/blob/master/DamagedHelmetModified/glTF/DamagedHelmetModified.gltf#L111-L125) to reflect this change?

javagl:
Sure. I just started working on all that, and intended to create a shader implementation that implements the spec as closely as possible. I started reading the Frostbite PBR document, but it's difficult to flesh out the relevant parts here. E.g. the spec does not cover normals/bump maps or ambient occulsion, and I still have to figure out how they are integrated eventually (there are some open issues for the discussion about these points). 

In the meantime, I have updated the glTF and did some minor updates to the shaders, to use the combined metallic+roughness texture, and to take the factors (base color, metallic, roughness) into account: https://github.com/javagl/gltfTestModels/commit/66827933648b974d787ca2a12a6fa86d89b1801d

pjcozzi:
Metal Roughness is being added to the core glTF 2.0 spec in #830.

javagl:
You should define the variables used in the [Appendix](https://github.com/tsturm/glTF/blob/master/extensions/Vendor/FRAUNHOFER_materials_pbr/Appendix.md). 

For the Schlick approximation, there is equation 3:
equation3

In the [example implementation](https://raw.githubusercontent.com/tsturm/glTF/e0e419a11ceae9edea541c58a512f20bf1e8f2c7/extensions/Vendor/FRAUNHOFER_materials_pbr/example/untextured-pbr/index.html), the Schlick function is defined as

    vec3 F_Schlick(const in vec3 specularColor, const in float dotLH)

Although it's certainly *possible* to figure out
- What v is
- What h is
- What L is
- What H is
- Whether the formulations are "equaivalent" in any way

something like this can be confusing (for PBR-noobs like me). I know, the experts may now say: That's all clear and obvious. And indeed, there *seem* to be some common practices for naming certain vectors `L`, `V`, `H`, `N`, etc, but it does not seem to be consistent (particularly when referring to the '94 paper of Schlick, who used a different notation),

tparisi:
@pjcozzi can we have a succinct recap of the logic behind putting specular-glossiness into an extensions? @cedricpinson how does the decision relate to Sketchfab's PBR model out of curiosity?

pjcozzi:
@tparisi this was discussed at length with broad community input, see https://github.com/KhronosGroup/glTF/issues/696#issuecomment-270477627

cedricpinson:
@tparisi well nothing special we adapted our implementation to fit the extension. I would have been happy to have this in the core, but it does not really change a lot. I understand that simplicity with metalness,roughness is good for everybody.

lexaknyazev:
Should we close this PR, or it will be published as glTF 1.0 extension?

mlimper:
I think we could close it.

@pjcozzi What is your opinion?

pjcozzi:
OK with me.  If there are any TODOs in comments, please capture them in a separate issue.

mlimper:
I believe everything that was at some point identified as a separate TODO has already been captured in separate issues, which are all mentioned in this thread.
The ones that are still open are:
https://github.com/KhronosGroup/glTF/issues/697
https://github.com/KhronosGroup/glTF/issues/699

Issue #624, opened 2016-06-13
Tightening up skinning


lexaknyazev:
Relevant parts of `RiggedSimple.gltf` (all default values omitted), issues below:

``` js
{
  "nodes": {
    "Armature": {
      "children": ["Bone"]
    },
    "Bone": {
      "children": ["Bone_001"],
      "jointName": "Bone"
    },
    "Bone_001": {
      "jointName": "Bone_001"
    },
    "Cylinder": {
      "meshes": ["Cylinder-mesh"],
      "skeletons": ["Bone"],
      "skin": "Armature_Cylinder-skin"
    },
    "node_4": {
      "children": ["Armature", "Cylinder"],
      "matrix": [/*Y_UP_Transform matrix*/]
    }
  },
  "skins": {
    "Armature_Cylinder-skin": {
      "inverseBindMatrices": "IBM_Armature_Cylinder-skin",
      "jointNames": ["Bone", "Bone_001"]
    }
  },
  "techniques": {
    "technique0": {
      "parameters": {
        "joint": {
          "semantic": "JOINT",
          "type": 35666
        },
        "jointMat": {
          "count": 2,
          "semantic": "JOINTMATRIX",
          "type": 35676
        },
        "weight": {
          "semantic": "WEIGHT",
          "type": 35666
        }
      }
    }
  }
}
```
1. Here we have one geometry node (`"Cylinder"`) and a skeleton node tree (`"Armature" -> "Bone" -> "Bone_001"`). Can any node from skeleton tree have `meshes`, `skeletons`, `camera` or `skin` properties?
2. Spec [says](https://github.com/KhronosGroup/glTF/tree/1.0.1/specification#skin-instances): 
   
   > The `skeletons` property contains one or more skeletons, each of which is the root of a node hierarchy.
   
   A top-level node of that tree is `"Armature"` not `"Bone"`. Is it a converter bug? If not why do we need that empty unused `"Armature"` node with one child?
3. Spec [spells](https://github.com/KhronosGroup/glTF/tree/1.0.1/specification#nodeskeletons) `skin.joints` instead of `skin.jointNames`.
4. Namespace of `node.jointName`. If it's glTF global, why not use `id` instead? If it is skeleton-tree limited, it can cause conflicts when `node.skeletons` contains two trees with the same `jointName`'s inside.
[...]

PR #644, opened 2016-07-08
Small skinning fixes

lexaknyazev:
small fixes for #624, add skinning attributes definitions from glTF Overview


pjcozzi:
Thanks for the clarifications, @lexaknyazev!

Issue #656, opened 2016-07-14
'memcpy' Not Declared


KermMartian:
`#include <cstring>` is missing from two files in order to define memcpy() for g++-based builds:
- COLLADA2GLTF/JSON/JSONNumber.cpp
- COLLADA2GLTF/helpers/geometryHelpers.cpp

skrat:
Same problem here, Arch Linux, up to date

pjcozzi:
Thanks for the report and fix.

@lasalvavida can you make this small change?

lasalvavida:
@pjcozzi, @skrat has a pull request open #657 to resolve this issue already.

pjcozzi:
> @pjcozzi, @skrat has a pull request open #657 to resolve this issue already.

Ah yes, I am still catching up since SIGGRAPH.  Thanks for the report and fix, @KermMartian and @skrat!

PR #657, opened 2016-07-15
Fill in missing includes #656

skrat:
This solves the build issue on Linux (I had it on latest Arch, @timknip on Ubuntu)


skrat:
This is the output I've had before filling in missing cstring include

https://gist.github.com/skrat/aa289edf88639ac363d9c07ab97fc011

pjcozzi:
Thanks @skrat!

Issue #659, opened 2016-07-20
Regenerate sample models for 2.0


lexaknyazev:
As far as I can tell from the updated validation output for the 1.0.1 spec version, sample models need just a few tweaks to be valid 1.0.1:
1. Change `version` to `1.0.1`
2. Add `min` and `max` to all accessors.
3. #648

lexaknyazev:
@lasalvavida could you implement #690 and update sample models?

lexaknyazev:
Closed, since models have now a separate repo.

PR #666, opened 2016-08-01
min and max are generated for all accessors

lasalvavida:
Part of the changes for #659.

@tfili Could you look at this when you get a chance?


tfili:
Thanks. @lasalvavida. Looks good to me.

PR #678, opened 2016-08-10
Fixed CG_profile effect. Default material applied in this case

mycheckgithub:
openCOLLADA, which is the core library used by glTF converter, does not seem capable to load the profile_CG data from the COLLADA file, which in turns makes it very difficult for the glTF converter to not behave correctly. 

To have the gltf converter not to crash, this fix will create a defaultMaterial in this case.

### DIFF
--- a/COLLADA2GLTF/GLTF/GLTFAsset.cpp
+++ b/COLLADA2GLTF/GLTF/GLTFAsset.cpp
@@ -619,6 +619,31 @@ namespace GLTF
         return materialBindingKey;
     }
 
+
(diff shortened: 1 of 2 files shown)


pjcozzi:
@lasalvavida can you please test, review, and merge this?

lasalvavida:
@mycheckgithub Do you have a sample model you could share that doesn't convert correctly currently?

mycheckgithub:
Here is a sample model:
[zombie_male_a.dae.txt](https://github.com/KhronosGroup/glTF/files/422655/zombie_male_a.dae.txt)

selim-bekkar-sb:
Any news about merging ?

pjcozzi:
@lasalvavida can you look at this when you do the COLLADA2GLTF updates for 1.0.1?

lexaknyazev:
Closed due to moving COLLADA2GLTF into [separate repo](https://github.com/KhronosGroup/COLLADA2GLTF).

CC @lasalvavida

PR #687, opened 2016-08-20
Add xml3d.js to webgl engines supporting glTF

ksons:


pjcozzi:
Cool, thanks for sharing!

PR #688, opened 2016-08-22
Added links to Java glTF libraries

javagl:
All JglTF libraries are still work in progress, but some of them might already be interesting for others.


pjcozzi:
Wow, you have been busy!  Thanks @javagl!

PR #689, opened 2016-08-22
Rename scene to content in Binary glTF extension spec

### LINKED ISSUE #653: Terms collision in KHR_binary_glTF
` KHR_binary_glTF` refers to the glTF data as "scene", while glTF spec has its own definition of "scene" (and can have many of them).

Maybe, we could come up with something better?

pjcozzi:
Fixes #653.

@lexaknyazev for your review.


lexaknyazev:
Lgtm

Issue #704, opened 2016-09-07
Tightening up animations


lexaknyazev:
1. Specify `componentType` for animation accessors (suggest only `FLOAT`).
2. Remove `count` from [example](https://github.com/KhronosGroup/glTF/tree/1.0.1/specification#animations).
3. To be more explicit, 
   
   > All parameters must have the same number of elements.
   
   could be changed to something like
   
   > All accessors referenced by the same animation must have the same `count` property.

CC @pjcozzi @javagl @lasalvavida

lexaknyazev:
More animation stuff:

- `TIME` should be the only valid value for `sampler.input` [#573 (comment)](https://github.com/KhronosGroup/glTF/issues/573#issuecomment-226160783);
- `TIME` values must be sorted in ascending order.

Maybe it's useful to allow some output accessors to have less elements than `TIME` samples (if one channel stops before others).

javagl:
I think whether it's called `TIME` or not is not sooo important - it does not have a "semantic" in that sense, and could equally be called `timeKeyFrames` or whatever. 

The constraint of time values being monotonically increasing is already pointed out for `sampler.input` at https://github.com/KhronosGroup/glTF/tree/master/specification#samplerinput-white_check_mark .

(One could consider requiring it to be *strictly* monotonic. Handling the special case of `time[n+1] == time[n]` , could be fiddly and error-prone. But one would have to investigate further whether this special case is really never required)

lexaknyazev:
> I think whether it's called `TIME` or not is not sooo important - it does not have a "semantic" in that sense, and could equally be called `timeKeyFrames` or whatever.

In that case we should demand all samplers to use the same `input` parameter. And that parameter still has to have `TIME` "semantic" (even when called otherwise).

> The constraint of time values being monotonically increasing is already pointed out for `sampler.input`.

I've overlooked it, thanks.

> One could consider requiring it to be _strictly_ monotonic. Handling the special case of `time[n+1] == time[n]` , could be fiddly and error-prone. But one would have to investigate further whether this special case is really never required

I'd say it otherwise: we should forbid such case or define exact behavior (like always use last output value).

pjcozzi:
> 1. Specify componentType for animation accessors (suggest only FLOAT).

Do you mean for the examples?

> 2. Remove count from example.

Yes, please open a PR into master.

[...]

PR #708, opened 2016-09-08
Remove obsolete "count" property from animations example

lexaknyazev:
#704


pjcozzi:
Once again, good catch!

Issue #710, opened 2016-09-10
Spec should state TRS order for node transform


pjcozzi:
via @mre4ce 

> Afaik the order should be T * R * S, but I didn’t find anything in the spec.

@lexaknyazev could you please update the spec?  PR into `master` would be fine since this is just a spec fix.

Here is the code Cesium uses: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/Matrix4.js#L357

PR #711, opened 2016-09-10
Clarify TRS order

lexaknyazev:
See #710.


pjcozzi:
Great, thanks @lexaknyazev and @mre4ce!

PR #718, opened 2016-09-15
Fixed compile issue when linking into another exe

tfili:
I get a few compile errors about missing `memcpy` and `memcmp`.


pjcozzi:
Thanks @tfili.

PR #719, opened 2016-09-16
Overwrite node OID when it is an empty string

lasalvavida:
@tfili This should fix the model converter issues we were having


lasalvavida:
To elaborate, this resolves an issue where collada root nodes don't have explicit names. COLLADA2GLTF assigns a new node id when it is referred to by the root transform as a child, but still wrote the node with an empty string as the id.

Issue #695, opened 2016-08-25
Handling unsupported extensions


emackey:
How should user agents handle models that declare unknown or unsupported extensions?

Currently, Cesium [scans the extensions list](https://github.com/AnalyticalGraphicsInc/cesium/blob/1.24/Source/Scene/Model.js#L3430-L3439) and will `throw` if it finds anything it doesn't recognize.  This is problematic for example with glTF models exported from the [Blender exporter](https://github.com/Kupoman/blendergltf), which define some as-yet-unused Blender-specific extensions, all of which should be perfectly safe for Cesium to disregard.

My proposal is that glTF should specify that user agents follow the typical web behavior of making a best-effort to display the provided content, and not fail due to the mere presence of unknown extensions.  Likewise, glTF authoring software should strive to produce models that are at least minimally useful (or at least valid glTF) even if the extension block(s) were completely removed from the file.

In the case of the `BLENDER_Physics` extension, this is trivial:  Blender encodes settings needed by its physics engine into the model, and rendering engines are free to display the model outside of any physics engine, safely ignoring the unknown extension.

Things get more complicated with proposed material extensions.  Ideally, a glTF author could provide a simple fallback material, of lower visual quality, to be a stand-in for rendering engines where the high-quality material extension is unavailable.  If this is not practical, I believe there may be a [default (unlit gray) rendering strategy](https://github.com/KhronosGroup/glTF/issues/84#issuecomment-145932150) that could be used in lieu of the missing material extension.

lexaknyazev:
> My proposal is that glTF should specify that user agents follow the typical web behavior of making a best-effort to display the provided content, and not fail due to the mere presence of unknown extensions. 

"Best-effort" result may vary depending upon the extension. E.g., what should runtime do with accessors having (unsupported) `WEB3D_quantized_attributes`?

Maybe, an asset could declare some extensions as "mandatory" or "recommended" to ensure desired compatibility / interoperability level.

emackey:
> Maybe, an asset could declare some extensions as "mandatory" or "recommended"

I like this idea so long as the "mandatory" ones must declare themselves to be so (by whatever means, inside the glTF).  I think all of the existing known extensions are non-mandatory, and the default behavior should be to proceed with rendering.

[...]

PR #721, opened 2016-09-18
Add more extensions docs

lexaknyazev:


pjcozzi:
Part of #695.

@lexaknyazev please add any additional bullets to #605 if needed.

PR #732, opened 2016-09-28
Translation animations now include node translation if only a single axis is specified

lasalvavida:
@tfili This should solve our problem with the attached conditioned CesiumMan collada file
[Model.zip](https://github.com/KhronosGroup/glTF/files/498692/Model.zip)


lasalvavida:
It looks like VC has unresolved node references in master that this branch doesn't fix, looking into this now.

lasalvavida:
This PR now fixes the VC issue as well.

tfili:
Thanks @lasalvavida

PR #740, opened 2016-10-07
Point THREE.js GLTF loader directory toward active development repo

richtr:
This PR points people towards the THREE.js project's own [GLTFLoader.js](https://github.com/mrdoob/three.js/blob/dev/examples/js/loaders/GLTFLoader.js) and deprecates the old THREE.js glTF loaders from this repo.

This will avoid confusion for developers on which glTF loader to use in their THREE.js projects and where active development of the loader should happen.


emackey:
+1 for this.  But it would be great if the ThreeJS demo could load more (all?) of the demo models.

richtr:
I guess we just need a PR to the THREE.js example [here](https://github.com/mrdoob/three.js/blob/dev/examples/webgl_loader_gltf.html)? FWIW, all the sampleModels in this repo work great with the new loader.

richtr:
I also think this is a great initiative from @cx20 https://github.com/cx20/gltf-test. Maybe we want to get all the models up there?

emackey:
Wow, that's fantastic!  I'm filing #741 to get that into the main glTF readme.

emackey:
@cx20 would you be open to adding more models to your gltf-test?  Also I'm thinking about ways to make the load code more centralized per-engine, so you don't have so many copies of index.js, are you already planning anything like that?

cx20:
@emackey I can add more models to gltf-test. For now, I have been created a index.html and index.js for each model. However, I have no plan (idea) for the centralized management of model.
Please tell me if there is a better way.

pjcozzi:
+1 on removing this here to avoid confusion.

@richtr thanks so much for your major glTF contribution to three.js.

@cx20 it would probably be simple enough to pass a query parameter pointing to the model to an index.html/index.js per engine you are testing

pjcozzi:
@richtr are you up for reviewing the three.js labeled issues in this repo and recommending which ones can be closed? https://github.com/KhronosGroup/glTF/labels/three.js

PR #747, opened 2016-10-15
Fixed ordering of blendFuncSeparate

lilleyse:
The ordering was incorrect in this area. Uncovered through this post on the Cesium forum: https://groups.google.com/forum/#!msg/cesium-dev/BP_fs8K19ww/n2SBwzEsBAAJ

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -3166,7 +3166,7 @@ Integer values passed to `blendEquationSeparate()`. [rgb, alpha]. Valid values c
 
 ### functions.blendFuncSeparate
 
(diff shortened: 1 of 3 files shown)


pjcozzi:
Very nice catch,

Please update the schema as well: https://github.com/KhronosGroup/glTF/blob/master/specification/schema/technique.states.functions.schema.json#L44

Looks like the example is OK: https://github.com/KhronosGroup/glTF/blob/master/specification/schema/examples/techniques.json

Did you also spot check the models in the `sampleModels` directory?

lexaknyazev:
Related: https://github.com/KhronosGroup/glTF/pull/585

lilleyse:
@pjcozzi Thanks for noticing those other areas. My update fixes the remaining issues. The sample models are all fine.

pjcozzi:
Thanks @lilleyse!

This and #585 were my bad for sure.  Good catch!

Issue #745, opened 2016-10-12
Remove node.matrix


mre4ce:
A node can currently store a separate translation, rotation and scale or it can store a matrix. I vote for removing the matrix because it forces an implementation to implement code to decompose the matrix into a separate translation, rotation and scale in case an animation only changes some but not all three. Afaik this is the only reason to even have code to decompose a matrix. If for some reason source data stores matrices then it'd be better to do the decomposition in the converter as opposed to adding complexity to the glTF format.

lexaknyazev:
Can `node.matrix` contain something besides TRS (such as skew)?

pjcozzi:
Engines should not need to decompose the matrix for animation; the spec requires that when a node is targeted for animation that it uses TRS since animations can only target those properties.

@mre4ce would you be OK with just a spec update here that make this more explicit?

mre4ce:
Making this more explicit in the spec would be good thanks. I would still prefer to get rid of node.matrix just to reduce complexity but keeping it is not the end of the world :)

pjcozzi:
I added a spec clarification in #750.

donmccurdy:
Addressed in glTF 2.0.

Issue #743, opened 2016-10-10
Clarification of transform order


erich666:
In the [transforms section](https://github.com/KhronosGroup/glTF/blob/master/specification/README.md#transforms) it is probably worth adding something along the lines that "The TRS order means that first the scale is applied to the object's data, then the rotation, then the translation, to get the world position of the model." This is the norm, 90% of implementors will get it right, but we should go for 100%.

I would also, to emphasize this, put the code example in order of scale, rotation, and translation, not in the rotation, scale, translation order it's currently given.

```
    "rotation": [
        0,
        0,
        0,
        1
    ],
    "scale": [
        1,
        1,
        1
    ],
    "translation": [
        -17.7082,
        -11.4156,
        2.0922
    ]
```

Change to

```
    "scale": [
        1,
        1,
        1
    ],
    "rotation": [
        0,
        0,
        0,
        1
    ],
    "translation": [
        -17.7082,
        -11.4156,
        2.0922
    ]
```

It may be worth noting that the order TRS (reversed) is _always_ used, regardless of the order of the transforms in the file itself. This is often a point of confusion for my students using three.js, that the order they set these transforms does not influence the order in which these are applied.

If you really wanted to get fancy, I'd give an interesting set of scales, rotation, and translation (right now scale and rotation are default "do nothing"), and then also give the corresponding matrix that would do the equivalent concatenated transform. This then can help implementors as a reality check.

javagl:
I also stumbled over this (and think that I saw a "wrong" implementation a while ago (was it in three.js? not sure about that). This point has explicitly been addressed in a recent commit https://github.com/KhronosGroup/glTF/commit/eb4afeaae8c3d9e320b84906d10b368cc0f26b3e but indeed, the example could be updated as well.

erich666:
Thanks for the confirmation that it's not just me. Three.js is TRS (that is, read from right to left: scale first, then rotate, then translate).  [This lesson on](https://www.udacity.com/course/viewer#!/c-cs291/l-91073092/m-117955991) covers that. There's some Euler angle ordering that you can modify in three.js, but you can't change the overall TRS ordering. glTF doesn't have to worry about Euler angles since it uses quaternions.

lexaknyazev:
> It may be worth noting that the order TRS (reversed) is always used, regardless of the order of the transforms in the file itself.

Wouldn't t it be too explicit, since JSON objects don't have any particular properties order? 

[...]

PR #750, opened 2016-10-17
Improve node.matrix/node.[TRS] spec

pjcozzi:
For
* Clarification of transform order, #743
* Remove node.matrix, #745

@lexaknyazev please merge when ready.

CC @erich666, @mre4ce


lexaknyazev:
Please note that this PR could cause a conflict with `1.0.1` branch, because `animation.parameters` were removed. Maybe we should keep animation examples intact?

pjcozzi:
I can do the merge and conflict resolve into `1.0.1`.

If you really don't want the examples changed, I can do that too.

PR #754, opened 2016-10-22
Fix data URI detection to be case-insensitive.

DigiTec:
Several changes here, but the functional change is updating the data URI regex to be case-insensitive.

I've also cleaned up the resolvePathIfNeeded function to not rely on a separate, private function, not require a match (which is more expensive than a test), and added some commenting and considerations for future improvements.


pjcozzi:
Thanks @DigiTec.

Just curious, what are you building with glTF?

DigiTec:
@pjcozzi I'll be doing a bunch of syntax and perf work. I'm one of the members of the new Carmel team and we are evaluating ways to use glTF in the context of the browser. To that extent, we would like to understand the performance and ergonomics of parsing, loading and rendering the format.

pjcozzi:
Awesome, if you have the time, it would be great to write a tech article or blog post on your findings.

Issue #398, opened 2015-07-15
Proposal: Open3DGC mesh compression extension schema


kainino0x:
# `mesh_compression_open3dgc` proposal & mock-up

I have put together this proposal, with lots of [Patrick Cozzi](/pjcozzi)'s advice, in order to try to solve the problem of how to specify and used compressed mesh data in glTF.  This is based on the extension specification and implementation work of [Fabrice Robinet](/fabrobinet), which is in turn based on the Open3DGC algorithm and library work by [Khaled Mamou](/kmammou). It is designed to support meshes, but not animations.

This proposal should not require any changes to the core glTF specification, and _most or all_ changes are within `"extensions"` objects. This avoids churn in the core specification, and also has future-proofing benefits in format flexibility: While the three libraries I've looked at (Open3DGC, OpenCTM, webgl-loader) all work similarly, there could easily be other libraries which would be very difficult to shoehorn into this structure.

The semantics of this extension are also designed to provide a good design baseline for possible mesh compression extensions. Though I haven't thoroughly researched on the structure of other compression libraries, they should be able to use similar schemata. Some changes may be necessary to this extension schema, as it has not been implemented or formalized, but I will try to maintain structural compatibility with other compression libraries.

Please reply to this issue with any feedback you might have on this proposal.

(cc #208 #230)

[Rendered](https://gist.github.com/kainino0x/c082067f61ac9356ac04) (Gist for change tracking purposes)

pjcozzi:
Thanks for this detailed proposal, @kainino0x.

I'm going to take a second pass through this, but here are my initial comments (all are minor):

* We'll talked about having a generic mesh compression extension framework and then specific compression extensions.  Here, you have just used the generic glTF extension framework.  Perhaps this is a better approach since it gives each compression extension full control over how it fits into the glTF schema, but your extension still provides a concrete example that _may_ form the foundation for other extensions.  What do you think?
* To be consistent with glTF naming, `extensions.GLTF_mesh_compression_open3dgc.decompressedBuffer` should be plural, e.g., the top-level glTF `buffers` property is not `buffer`.
* `GLTF_mesh_compression_open3dgc` - we can drop the `GLTF` prefix.  Perhaps `mesh_compression_open3dgc` or `open3dgc_mesh_compression`.

> [--------------model1---------------][--------------model2---------------]

Should `model` be `primitive` or `mesh`?

> [----indices----][positions][normals]

[...]

PR #756, opened 2016-10-26
Fixed issue when FIFO PopFirst call resurns incorrect value; it was h…

proteamer:
…appening when number of PopFirst call exceeds allocated FIFO size.

I'm not sure that it's right repo for PR but I hope to get some help here.

I tried to convert my 3d models with `collada2gltf -f model.dae -o compressed -k -c Open3DGC -m binary` . And faced with problem when compression for some models caused errors like this:

`collada2gltf: /home/vagrant/dev/glTF/COLLADA2GLTF/dependencies/o3dgc/src/o3dgc_common_lib/inc/o3dgcAdjacencyInfo.h:123: long int o3dgc::AdjacencyInfo::Begin(long int) const: Assertion 'element < m_numElements' failed.`

I figured out the reasons of that error and fix it (see 3f58ffc) after that I checked my models but there was second error: `Segmentation fault`.

After some time of my research I found the main reason of error it was `const long O3DGC_MAX_TFAN_SIZE = 256;` limitation see code: [link#1](https://github.com/KhronosGroup/glTF/blob/master/COLLADA2GLTF/dependencies/o3dgc/src/o3dgc_common_lib/inc/o3dgcCommon.h#L44) and [link#2](https://github.com/KhronosGroup/glTF/blob/master/COLLADA2GLTF/dependencies/o3dgc/src/o3dgc_encode_lib/inc/o3dgcTriangleListEncoder.inl#L575-L576). But `TFAN` size for my models could reach 4k.

So my question is whtat is thre right way to fix it? I suppose to compute `O3DGC_MAX_TFAN_SIZE` value for particular case.

Thanks for attention.

### DIFF
--- a/COLLADA2GLTF/dependencies/o3dgc/src/o3dgc_common_lib/inc/o3dgcFIFO.h
+++ b/COLLADA2GLTF/dependencies/o3dgc/src/o3dgc_common_lib/inc/o3dgcFIFO.h
@@ -66,7 +66,7 @@ namespace o3dgc
                                     unsigned long current = m_start++;
                                     if (m_start == m_allocated) 
                                     {
-                                        m_end = 0;
+                                        m_start = 0;
                                     }
                                     return m_buffer[current];
                                 };


pjcozzi:
@proteamer are you picking up the work in #398?  That would be amazing!

You are right that this is not the right source repo for Open3DGC.  It might be here: https://github.com/amd/rest3d/tree/master/server/o3dgc

@kmammou would know.

proteamer:
@pjcozzi thanks for reply. I've made the duplicate [PR](https://github.com/amd/rest3d/pull/47)

And I'm interested in `mesh_compression_open3dgc` extention and I'll look into #398.

lexaknyazev:
Closed, due to moving COLLADA2GLTF into [separate repo](https://github.com/KhronosGroup/COLLADA2GLTF).

Issue #490, opened 2015-11-28
Remove dummy loader


pjcozzi:
https://github.com/KhronosGroup/glTF/tree/master/loaders/dummy

This is not up-to-date with the latest spec or sample models, and is not very illustrative.

I plan to remove it.  @tparisi let me know if you think otherwise.

pjcozzi:
Still +1 from me especially since this was not updated to 2.0.  Any objections?

emackey:
The whole `loaders` folder can go.  The other two already moved out.

pjcozzi:
Agreed.  Can you open a PR for this?

PR #757, opened 2016-10-27
Updates syntax for consistency

DigiTec:
This change tries to ensure a consistent syntax across brace styles, comparisons, line breaking and usage of console functions appropriate to the conditions they are flagging.

There are a couple of small functional changes exposed by:
1. Ordering of if statements to remove redundant checks.
2. An if check on uri to make sure it exists before we try to resolve it.
3. Object.keys always returns non-null so removed a null check.

Still unsure how to properly run tests, will investigate that for future pulls. Perhaps even try to set up testing on pull requests?

### DIFF
--- a/loaders/dummy/glTF-parser.js
+++ b/loaders/dummy/glTF-parser.js
@@ -54,12 +54,14 @@ var global = window;
         // only CommonJS-like enviroments that support module.exports,
         // like Node.
         factory(module.exports);
-    } else if (typeof define === 'function' && define.amd) {
+    }
+    else if (typeof define === 'function' && define.amd) {
         // AMD. Register as an anonymous module.
         define([], function () {
             return factory(root);
         });
-    } else {
+    }
+    else {
         // Browser globals
         factory(root);
     }
@@ -283,7 +285,8 @@ var global = window;
                         }
                     };
                     jsonfile.send(null);
-               } else {
+                }
+                else {
                     if (callback) {
                         callback(this._json);
                     }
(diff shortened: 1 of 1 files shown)


DigiTec:
Next set of changes will start functional fixes. I've spotted a few bugs while doing this once over and will be adding fixes for those once I create some local tests that hit the issues.

pjcozzi:
@DigiTec are you finding this dummy loader valuable (seems like yes)?  When I first started with glTF, I started with it, then quickly moved to my own loader.  I _thought_ - perhaps incorrectly - that this dummy loader was not that useful to the community so we were thinking of removing it as we start to separate this repo into a few repos to better organize the spec, COLLADA2GLTF converter, and sample models.

However given your recent contributions here, I'm starting to rethink this.  Should we keep it and continue to maintain it as you have?

pjcozzi:
CC #490

pjcozzi:
@DigiTec any thoughts here?

> @DigiTec are you finding this dummy loader valuable (seems like yes)? When I first started with glTF, I started with it, then quickly moved to my own loader. I thought - perhaps incorrectly - that this dummy loader was not that useful to the community so we were thinking of removing it as we start to separate this repo into a few repos to better organize the spec, COLLADA2GLTF converter, and sample models.
>
> However given your recent contributions here, I'm starting to rethink this. Should we keep it and continue to maintain it as you have?

pjcozzi:
This will most likely be removed.

Issue #749, opened 2016-10-17
Trailing zeros storage


lexaknyazev:
Should glTF require presence of accessor's trailing zeros? E.g.:

**two points, ubyte attribute, stride of 4:**

```
vertexAttribPointer(index, 1, UNSIGNED_BYTE, false, 4, *pointer);

Data layout:
|00|01|02|03|04|05|06|07|
|-----------------------|
|xx|00|00|00|xx|00|00|00|
```

Must bytes 5-7 be stored in glTF buffer or not?

CC @pjcozzi @mre4ce

pjcozzi:
I suppose it is not required (but could be there) as long as the data is true to the `bufferView` offset and length.  Does GL require this for the last attribute?

lexaknyazev:
> Does GL require this for the last attribute?

Looks like it doesn't. So valid `bufferView.byteLength` for a given accessor is a range, not a single value:
```c
int byteStride = accessor.byteStride == 0 ? COMPONENTS * COMPONENT_TYPE_LENGTH : byteStride;

int minLength = byteStride * (accessor.count - 1) + COMPONENTS * COMPONENT_TYPE_LENGTH;
int maxLength = byteStride * accessor.count;
```
Or should we limit only minimal length?

pjcozzi:
What is the right approach to keep the client as simple as possible, just say that:
```
byteLength  === byteStride * accessor.count
```

lexaknyazev:
If a client just passes values to the GL, it should be unaffected by either approach.
_Maybe_, tooling ecosystem will benefit a little from fixed byteLength.

pjcozzi:
> If a client just passes values to the GL, it should be unaffected by either approach.

Yes, but what if, for example, the client is doing mesh decompression.

Perhaps this fixed approach keeps things simple without loosing any useful use cases.

lexaknyazev:
OK, let's be more precise there.

pjcozzi:
Please open a PR into `1.0.1`

lexaknyazev:
__two verts, vec3 position + vec3 normal interleaved and aligned, float attribute, stride of 32 (2 x 4 x 4):__
```json
{
  "accessors": {
    "accessor_pos": {
      "byteOffset": 0,
      "byteStride": 32,
      "componentType": 5126,
      "count": 2,
      "type": "VEC3"
    },
    "accessor_nor": {
      "byteOffset": 16,
      "byteStride": 32,
      "componentType": 5126,
[...]

PR #758, opened 2016-10-28
Specify accessor's data layout

lexaknyazev:
See #749.


pjcozzi:
OK with me, will wait to hear from @mre4ce (given https://github.com/KhronosGroup/glTF/issues/749#issuecomment-256969594) before merging.

PR #762, opened 2016-10-30
Update repo layout

lexaknyazev:
- Removed COLLADA2GLTF and sampleModels
- glTF `1.0` spec version moved to `specification/1.0`
- Updates main README.md with links to new repos for converter and sampleModels


pjcozzi:
In the main README.md, I would strongly prefer that the "glTF Specification" link be renamed to "glTF 1.0 Specification" and link directly to https://github.com/lexaknyazev/glTF/tree/new-repo-layout/specification/1.0 so we can avoid the indirection of going through https://github.com/lexaknyazev/glTF/blob/new-repo-layout/specification/README.md, which I think can be deleted or just be a list of bullet points with links to each spec version.

pjcozzi:
Does the image link in the Overview need to be updated: https://github.com/lexaknyazev/glTF/tree/new-repo-layout#overview

image

pjcozzi:
Just those comments.  Is anything else needed before this can be merged?  For example, do the new repos need anything else?

pjcozzi:
For branding, can you add the glTF logo to the top of the COLLADA2GLTF and sample models repo like here: https://github.com/KhronosGroup/glTF/blob/master/README.md

lexaknyazev:
Two little comments:
- Maybe we should keep glTF logo (gltf.png) in its current location (with no version in URL)?
- What's the versioning scheme for gltfOverview? Why do we have 0.1.3 and 0.2.0 in this repo?

pjcozzi:
> What's the versioning scheme for gltfOverview? Why do we have 0.1.3 and 0.2.0 in this repo?

@javagl would know.  I think we just need the latest one.

pjcozzi:
> Maybe we should keep glTF logo (gltf.png) in its current location (with no version in URL)?

Could be fine so we are not copying it everywhere.

javagl:
Until now, The versioning scheme of the overview is not aligned with glTF itself. It uses its "own" semantic versioning.  

But to reduce confusion, I'd align it with glTF. So the upcoming updated version will be "1.1.0-beta" for glTF 1.1. 

(The "patch level" of the semantic version number then cannot be counted up as usual. So if there will be an overview "1.2.1" for glTF "1.2.1", and this requires a patch (e.g. for a typo), then it could be "1.2.1**b**", to disambiguate it from the overview for glTF "1.2.2")

---

BTW: Currently, the PNGs are huge (>2.8MB), and I wonder whether this has to be duplicated. I think there are at least three options:

- Copy the full image and the preview image into the glTF repo, as it is done now (independent of the gltfOverview)
- Create the preview image in the glTF repo, and only link to the full image in the gltfOverview releases section ( https://github.com/javagl/gltfOverview/releases/download/v0.2.0/gltfOverview-0.2.0.png )
- Add a preview image in the future gltfOverview releases, and from the glTF readme, only link to this preview *and* full image

Any preferences here?

pjcozzi:
Up to you guys, but perhaps this is easiest to maintain:

> Add a preview image in the future gltfOverview releases, and from the glTF readme, only link to this preview and full image

lexaknyazev:
I think it would be better to have only Khronos-owned repos dependencies in long-term.
Is it possible to have glTF-Overview repo within KhronosGroup GitHub organization? Otherwise i'd keep duplicating PNGs.

javagl:
Then the short-term solution could be to copy the PNGs.

Later, we could transfer the overview repo. According to https://help.github.com/articles/transferring-a-repository-owned-by-your-personal-account/ , it does not seem to be such a great effort, once the organizational part on Khronos side is done (but it might be better to do this after the other restructuring tasks have settled)

pjcozzi:
Whatever you guys think is OK with me.  It might also be reasonable to just move the overview repo to a folder in this directory since I think it is fine for it to be considered part of the glTF learning material and fewer repos will be easier for users to navigate.

Issue #766, opened 2016-11-02
Default material values should be contained in technique


javagl:
While trying to create a *really* minimal glTF asset, I stumbled over the default material. 

According to https://github.com/KhronosGroup/glTF/blob/master/specification/README.md#appendix-a-default-material , the default values for the `emission` property should be a 50% gray, and I assume that *this information should be contained in the technique.*

(The material that is given there as an example is probably supposed to override the default 50% `emission` `value` with an 80% one)

So the `technique0.parameters.emission.value` should probably given as

    "techniques": {
        "technique0": {
            ...
            "parameters": {
                ...
                "emission": {
                    "type": 35666,
                    "value": [
                        0.5,
                        0.5,
                        0.5,
                        1
                    ]                
                },
                ...

(Just drop me a note if this is correct and I should issue a PR)

pjcozzi:
@javagl I believe you are correct, good catch!  I might have written that JSON example by hand.  Please open a PR.  Just move the emission value to the technique (and change it to 50%) and delete it from the material.

PR #767, opened 2016-11-02
Default material values should be contained in technique

javagl:
Fixes https://github.com/KhronosGroup/glTF/issues/766


pjcozzi:
Very nice, thanks!

PR #768, opened 2016-11-08
Fix typo in spec.

Kangz:
PTAL!

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -240,7 +240,7 @@ The node hierarchy is defined using a node's `children` property, as in the foll
     },
 ```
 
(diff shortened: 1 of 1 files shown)


pjcozzi:
Either is probably OK, but I think this reads better, thanks!

emackey:
Actually I don't think this is correct.  "Their" is plural (for example, "Nodes can have their own children"), but "Each of these nodes" refers to individual nodes.  So each individual node can have "its" (singular possessive) own children.  It was correct the way it was before, not now.  Sorry!

pjcozzi:
OK, reverted in https://github.com/KhronosGroup/glTF/commit/8a6e755e4ca846e649187eb8d23d1dc9dbdaa498#diff-4638956650d3a085b3af675c3dbd3d53

Issue #252, opened 2014-03-12
Investigate JSON Schema 4


pjcozzi:
The glTF schema is written using JSON Schema 3.

I'll look into [JSON Schema 4](http://json-schema.org/documentation.html).  We'll use it if it adds value, which will mean switching the validator from [JSV](https://github.com/garycourt/JSV) to [JJV](https://github.com/acornejo/jjv) (or another validator that supports Schema 4).

pjcozzi:
I don't plan on moving from JSON Schema 3 to 4 right away, but here's:
* [A guide on moving from 3 to 4](https://github.com/json-schema/json-schema/wiki/ChangeLog)
* [A guide on JSON schema](http://spacetelescope.github.io/understanding-json-schema/) in general

sbtron:
Wanted to hit this up again as we are looking into implementing a native glTF parser using [rapidJSON](https://github.com/miloyip/rapidjson/).  rapidJSON needs JSON Schema 4. I can work on updating to the schema needed to go from 3 to 4. Are there any other updates needed around validators etc. if the schema was updated from 3 to 4?

pjcozzi:
This would be fantastic.

For some examples using JSON Schema 4, check out how we use it for 3D Tiles (including validation):

https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/schema#3d-tiles-json-schema

It would not be required right away, but we would also very much welcome a contribution to update wetzel:

https://github.com/AnalyticalGraphicsInc/wetzel

which is a small Node.js tool I wrote to generate the property references in the spec from the schema.

There is also a C# loader that was generated from the schema that we would have to determine the best path forward for; I'm not sure if it is being maintained:

https://github.com/KhronosGroup/glTF/tree/master/loaders/CSharp

pjcozzi:
Fixed in #769 and friends

PR #769, opened 2016-11-08
glTF schema v4 upgrade and inheritance issue fix

icechen1:
There's an issue in the gltf schema related to inheritance that is causing schema validators to fail (I verified this with [rapidjson](https://github.com/miloyip/rapidjson/) and [json-schema-validator](https://github.com/daveclayton/json-schema-validator). A description of the problem can be found [here](http://stackoverflow.com/questions/27410216/json-schema-and-inheritance) and [here](http://stackoverflow.com/questions/22689900/json-schema-allof-with-additionalproperties).

I implemented a workaround that fixes the issue (hopefully a future version of json schema will provide a more elegant solution). The solution consists of passing the name of the fields defined in the "parent" down to the children and define them as empty objects. The advantage of this approach is that there is no duplication involved and the actual schema of the fields in question is only defined in the parent. This way the schema will not fail at additionalProperties and validators seem to behave as expected.

I've also added v4 schema upgrades as well. See https://github.com/json-schema/json-schema/wiki/ChangeLog for changes between schema v3 and v4

### DIFF
--- a/specification/schema/accessor.schema.json
+++ b/specification/schema/accessor.schema.json
@@ -1,15 +1,15 @@
 {
-    "$schema" : "http://json-schema.org/draft-03/schema",
+    "$schema" : "http://json-schema.org/draft-04/schema",
     "title" : "accessor",
     "type" : "object",
     "description" : "A typed view into a bufferView.  A bufferView contains raw binary data.  An accessor provides a typed view into a bufferView or a subset of a bufferView similar to how WebGL's `vertexAttribPointer()` defines an attribute in a buffer.",
-    "extends" : { "$ref" : "glTFChildOfRootProperty.schema.json" },
+    "allOf" : [ { "$ref" : "glTFChildOfRootProperty.schema.json" } ],
     "properties" : {
         "name" : {},
         "extensions" : {},
         "extras" : {},
         "bufferView" : {
-            "extends" : { "$ref" : "glTFid.schema.json" },
+            "allOf" : [ { "$ref" : "glTFid.schema.json" } ],
             "description" : "The ID of the bufferView.",
             "required" : true
         },
--- a/specification/schema/accessor.schema.json
+++ b/specification/schema/accessor.schema.json
@@ -5,6 +5,9 @@
     "description" : "A typed view into a bufferView.  A bufferView contains raw binary data.  An accessor provides a typed view into a bufferView or a subset of a bufferView similar to how WebGL's `vertexAttribPointer()` defines an attribute in a buffer.",
     "extends" : { "$ref" : "glTFChildOfRootProperty.schema.json" },
     "properties" : {
+        "name" : {},
+        "extensions" : {},
+        "extras" : {},
         "bufferView" : {
             "extends" : { "$ref" : "glTFid.schema.json" },
             "description" : "The ID of the bufferView.",
--- a/specification/schema/accessor.schema.json
+++ b/specification/schema/accessor.schema.json
@@ -10,14 +10,12 @@
         "extras" : {},
         "bufferView" : {
             "allOf" : [ { "$ref" : "glTFid.schema.json" } ],
-            "description" : "The ID of the bufferView.",
-            "required" : true
+            "description" : "The ID of the bufferView."
         },
         "byteOffset" : {
             "type" : "integer",
             "description" : "The offset relative to the start of the bufferView in bytes.",
             "minimum" : 0,
-            "required" : true,
             "gltf_detailedDescription" : "The offset relative to the start of the bufferView in bytes.  This must be a multiple of the size of the data type.",
             "gltf_webgl" : "`vertexAttribPointer()` offset parameter"
         },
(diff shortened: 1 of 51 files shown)


pjcozzi:
Thanks for the workaround, @icechen1!

@lexaknyazev this is OK with me, does it impact the 1.1 schema?

@icechen1 this is likely to impact the [C# loader generated from the schema](https://github.com/KhronosGroup/glTF/tree/master/loaders/CSharp) and [wetzel](https://github.com/AnalyticalGraphicsInc/wetzel) used to generate parts of the spec.  Did you test them?

We could also retarget this to a new `schema-improvements` branch (since I imagine JSON Schema 4 updates are coming? #252) and then evaluate changes to the C# loader and wetzel after all the schema updates.

nehmebilal:
@pjcozzi We could certainly combine this PR and the V4 changes (which @icechen1 has already implemented) in one PR if you prefer? 

That being said, the changes proposed in this PR are fixes to actual issues in the schema and I feel that master could benefit from such fixes? What do you think?

pjcozzi:
If the v4 changes are imminent, can we just wait so we only have to propagate changes to the 1.1 spec and doc tool?

Otherwise, if this is holding you up, we can merge and just wait on those other changes.

nehmebilal:
Yes, v4 changes are ready. Should we add these changes to this PR? A different branch?

We can certainly wait but we're hoping to merge the changes before other changes to the GLTF schema are made to avoid diverging and having to deal with painful merges.

icechen1:
I've tested the changes with the C# loader and Wetzel and noticed that the changes are breaking. C# causes crashes (due to key duplication) and Wetzel generates [blanks fields](https://cloud.githubusercontent.com/assets/3333944/20155451/cdb140a4-a680-11e6-91de-64f5114410aa.png) due to the workaround. 

I believe the workaround conforms the json schema [specs](http://json-schema.org/latest/json-schema-validation.html#rfc.section.5.22), so changes are needed on the loader and Wetzel in order to support this change. The C# loader seems to merge the schema referenced in 'extends' creating the duplicate field issue. However, that seems not to be the expected behavior according to the spec, rather, the 2 documents needs to be parsed independently.

pjcozzi:
Adding the v4 changes to this branch would be OK.

icechen1:
Okay, I've pulled in the v4 schema updates as well. Will update the PR name and description.

sbtron:
@MattMcMullan could you take a look at the issues with C# loader that @icechen1 mentions above

pjcozzi:
Looks good, thanks so much!  Also, thanks for updating the extensions.

Fixes #252.

Should we merge after this:

> @MattMcMullan could you take a look at the issues with C# loader that @icechen1 mentions above

The wetzel updates can be separate, of course.

specification/schema/accessor.schema.json
```diff
@@ -1,12 +1,15 @@
...
+    "$schema" : "http://json-schema.org/draft-04/schema",
     "title" : "accessor",
     "type" : "object",
     "description" : "A typed view into a bufferView.  A bufferView contains raw binary data.  An accessor provides a typed view into a bufferView or a subset of a bufferView similar to how WebGL's `vertexAttribPointer()` defines an attribute in a buffer.",
-    "extends" : { "$ref" : "glTFChildOfRootProperty.schema.json" },
+    "allOf" : [ { "$ref" : "glTFChildOfRootProperty.schema.json" } ],
     "properties" : {
+        "name" : {},
+        "extensions" : {},
+        "extras" : {},
```
lexaknyazev: Why do we have both `"allOf"` and `"name", "extensions", "extras"` here?

lexaknyazev:
Some schemas have `"dependencies"`. Its format was [changed](https://github.com/json-schema/json-schema/wiki/ChangeLog#dependencies) with v4. 

@icechen1 Could you fix that also?

MattMcMullan:
Sure, I'll take a look

specification/schema/accessor.schema.json
nehmebilal: The author of the json schema explains why in this [stackoverflow post](http://stackoverflow.com/questions/22689900/json-schema-allof-with-additionalproperties).

icechen1:
@lexaknyazev Updated the schema with the updated `dependencies` and `required` syntax. Thanks for pointing it out!

lexaknyazev:
@icechen1 Great!
Besides C# loader, is it ready for merge?

icechen1:
`wetzel` also needs to be updated, but otherwise all the schema upgrades are pushed.

lexaknyazev:
@icechen1 
Should we change `required` in [WEB3D_quantized_attributes.accessor.schema.json](https://github.com/icechen1/glTF/blob/fa22eb33f61731dd3f1997981dbf791ed3e857e4/extensions/Vendor/WEB3D_quantized_attributes/schema/WEB3D_quantized_attributes.accessor.schema.json)?

---

> `wetzel` also needs to be updated

CC @pjcozzi

lexaknyazev:
Also [CESIUM_RTC.schema.json](https://github.com/icechen1/glTF/blob/fa22eb33f61731dd3f1997981dbf791ed3e857e4/extensions/Vendor/CESIUM_RTC/CESIUM_RTC.schema.json).

pjcozzi:
> wetzel also needs to be updated, but otherwise all the schema upgrades are pushed

This can be merged first.

icechen1:
@lexaknyazev Whoops, missed those two. I'll make the change now.

PR #770, opened 2016-11-09
Sample model syntax fix

icechen1:
I noticed that the .glb sample model for `BrainStem` has an extra field (`"disable":[]`) in `techniques.states`, causing it to fail schema validation. 

I don't think that field is defined anywhere in official specs (only `enable` is defined in v1.0), so I edited the glb file and removed it, as it was not needed anyways. The `disable` field is absent from other versions of the model (gltf, gltf-embedded, etc).


pjcozzi:
Good catch!  `disable` was removed from the spec before it was finalized.

@lexaknyazev should we re-open this PR in [glTF-Sample-Models](https://github.com/KhronosGroup/glTF-Sample-Models) since that is the future home of the sample models?

lexaknyazev:
This error was introduced with https://github.com/KhronosGroup/glTF/commit/9d97cfcf451f0931a9392417c36f74d8be53199c.
CC @lasalvavida 

Since we're going to remove all models from this repo, PR should go straight to [glTF-Sample-Models](https://github.com/KhronosGroup/glTF-Sample-Models).

nehmebilal:
I'm curious, why not remove the sample models from this repo right away and provide a link to the new repo to avoid confusion and potentially diverging models?

lexaknyazev:
> I'm curious, why not remove the sample models from this repo right away and provide a link to the new repo to avoid confusion and potentially diverging models?

@nehmebilal 
It'll happen very-very soon (1-2 days).

pjcozzi:
@lasalvavida could you look at this soon?  It would help us with glTF 1.1.

lexaknyazev:
See https://github.com/KhronosGroup/glTF-Sample-Models/issues/2.

PR #771, opened 2016-11-10
Updated example links

xeolabs:
xeoEngine was recently renamed to "xeogl", so this change fixes the links to glTF examples for that framework.


pjcozzi:
Thanks for the update, @xeolabs!

PR #778, opened 2016-11-23
Update schema definition asset for schema 0.4 consistency.

xelatihy:
This pull request updates asset.schema.json for schema 0.4 consistency.


lexaknyazev:
Good catch, thanks @xelatihy!

Issue #632, opened 2016-06-25
Finishing KHR_materials_common


lexaknyazev:
Main discussion was in #424.
#### Small fixes:
- [ ] Extension defines `transparent` and `doubleSided` flags inside `values`. COLLADA2GLTF puts them just into the extension object.
- [ ] Extension doesn't mention `jointCount` property.
- [ ] Extension shouldn't mention `JOINTMATRIX` semantic because it's a uniform parameter not attribute. #633
- [ ] `light.distance` was once removed, then returned back (by mistake?). JSON schemes don't have it. #633
- [ ] Extension shouldn't mention `direction` for Directional and Spot lights in descriptions after Table 2. #633
- [ ] Extension should mention `quadraticAttenuation` for Point and Spot lights in descriptions after Table 2. #633
- [ ] `light.color` type is defined as FLOAT_VEC4 in spec, but FLOAT_VEC3 in JSON schemes. #633
- [ ] Default `light.color` value is defined as `[0, 0, 0, 1]` in spec, but `[1, 1, 1]` in JSON schemes. #633
- [ ] Extension needs JSON schemes for materials.
- [ ] Links to JSON schemes for lights are broken. #633
- [ ] #611. Included in #633
- [ ] Default spec attenuation values aren't in sync with schemes. #633
- [ ] Wording of `CONSTANT` shading could be improved. It says: "a material that describes a constantly shaded surface that is independent of lighting". However, reflected color depends actually on `ambient` light contributions.

---

Since Extension hasn't been ratified yet, maybe it's possible to discuss a few more things:
1. Should Extension define light's affection distance (to turn it off for all objects that are outside of  affection range)?
2. Should Extension define per-object light affection matrix (which lights affects which objects)?
3. Should Extension support "Projector" lights with texture instead of a simple color?
4. Maybe, we can allow `light.color` values over `1.0`?

lexaknyazev:
> Should Extension define light's affection distance (to turn it off for all objects that are outside of affection range)?

This can be easily done via `node.boundingVolume` (#507) later (1.1). 

> Should Extension define per-object light affection matrix (which lights affects which objects)?

Can be done later (1.1+) via something like `node.layer`.

> Should Extension support "Projector" lights with texture instead of a simple color?

A separate extension should be proposed for such case to keep `KHR_materials_common` aligned with COLLADA.

lexaknyazev:
[...]

PR #780, opened 2016-11-26
naming scheme for light properties in KHR_materials_common

xelatihy:
In the schema for KHR_materials_common extension, light filenames names where defined as 

```
light.XXX.schema.json
```

For consistency with the other schemas, I suggest to rename them with

```
light.XXX. KHR_materials_common.schema.json
```

This naming convention has the following benefits:

1.  automatic code generation from the schema have the same names for properties in both lights and materials
2. should be backward compatible in validation


pjcozzi:
Thanks @xelatihy, sorry for the slow response.

@lexaknyazev this change is OK with me but is it OBE given https://github.com/KhronosGroup/glTF/issues/632#issuecomment-268303282?

PR #787, opened 2016-12-01
Added yocto-gltf to the C++ library list.

xelatihy:
I would like to add yocto-gltf to the C++ library list.


pjcozzi:
Very cool, thanks @xelatihy!

PR #793, opened 2016-12-08
Add the link to Japanese introduction post of glTF

emadurandal:
Finally completed the article, so I will pull the request.
I will maintain this article for a long time. It will be the best resource for Japanese-speaking people to learn glTF.

(I will also write articles about glTF skinning animation (using raw WebGL) in the future as well.)

If there is no problem, please merge.


pjcozzi:
Thanks @emadurandal!

PR #804, opened 2017-01-03
KHR_skin_culling

mre4ce:


javagl:
Although I can't dare to suggest having understood this in all depth, I'm a bit confused that all this does not incorporate the *weights*. So this computes some sort of "worst case bounds", based on the original bounds of the geometry, all joint matrices, and the *assumption* that the weights will be at most 1.0 (and not larger), right?

mre4ce:
You can look at it as if any vertex that is influenced by a joint is rigidly skinned to that joint. In other words, for every vertex that is influenced by a joint it assumes the weight is 1.0 (and not larger). If a vertex is influenced by multiple joints, then the vertex is assumed to be rigidly skinned to all joints that influence the vertex.

So yes, these are worst case bounds in that regard, but for any normal animation there is more slop in using bounding volumes (AABBs in this case). With that said, if you run through various animations you'll notice that the bounds are surprisingly tight.

The advantages of this approach are that it is fast, requires minimal data, and works with arbitrary animation blending (between key frames, and future animation blending).

pjcozzi:

Issue #587, opened 2016-04-30
Version numbers and profile specifiers in shaders


MarcoHutter:
(This might be related to https://github.com/KhronosGroup/glTF/issues/405 )

Compiling the current shaders (e.g. from the "duck" example) on desktop (non-GLES) systems results in 

```
0(1) : warning C7022: unrecognized profile specifier "highp"
0(1) : warning C7022: unrecognized profile specifier "precision"
0(18) : warning C7536: OpenGL does not allow matrix casts without #version 120 or later
```

The concerning thing is that the "matrix cast" warning may be considered as an **error** in some cases, and cause the compilation to fail. I'm not sooo familiar with the details, but know that this depends on the graphics card vendor / driver and potentially other factors (e.g. these "compatibility" and "core" settings...). At least, according to the GLSL ES Spec https://www.khronos.org/registry/gles/specs/3.2/GLSL_ES_Specification_3.20.pdf , Section 3.3, Version Declaration:

> Shaders must declare the version of the language they are written to. The version is specified in the first line of a shader by a character string:
> # version number es
> 
> where number must be a version of the language, following the same convention as **VERSION** above. The directive “#version 320 es” is required in any shader that uses version 3.20 of the language. Any number representing a version of the language a compiler does not support will cause an error to be generated. Version 1.00 of the language does not require shaders to include this directive, and shaders that do not include a #version directive will be treated as targeting version 1.00.

So I guess that the version number **must** be included (in the RFC2119 sense), unless the shader really targets GLSL 1.00. 

I fixed this for my local tests, by adding

```
#version 120
#ifdef GL_ES
    precision highp float;
#endif
```

at the beginning of the shaders, but am not sure whether this an appropriate or generally applicable solution.

pjcozzi:
Thanks for pointing this out, @MarcoHutter!

I think there's a few components here:
* The sample models only have WebGL/ES-compatible shaders
* COLLADA2GLTF only outputs WebGL/ES-compatible shaders
* The glTF spec doesn't explicitly specify the GLSL version

This is potentially related to #236.

Labeling this `1.0.1` so we consider this in the near-term.

pjcozzi:
@lexaknyazev @MarcoHutter I think we have a few options here:
* Push for after 1.0.1 since SIGGRAPH is less than three weeks away and this issue hasn't come up as much as others
[...]

PR #807, opened 2017-01-03
KHR_glsl_shader_versions

mre4ce:


javagl:
If I understood this correctly, this could eventually make https://github.com/KhronosGroup/glTF/issues/587 obsolete. I'm not sure whether one could make this even more generic, as in

    "KHR_glsl_shader_versions" : {
        "shaderRef0" : {
             "api" : "gl"
             "version" : "140 es";
             "uri" : "shader140es.vert";
        },
        "shaderRef1" : {
             "api" : "vulkan"
             "version" : "100";
             "uri" : "vulkan.spirv";
        }
    }

but this is a wild guess - as pointed out in the linked issue, this is a difficult topic (that I'm not really familiar with), and requires careful thoughts by the respective experts.

mre4ce:
That generally sounds like a good idea. Specifying the GLSL version would allow an application to select an appropriate version for the platform.

However, I would make it an array and make the distinction between OpenGL and OpenGL ES. Also note that this extension is specifically for GLSL. I have a separate extension for SPIRV, which will also be supported by OpenGL, OpenGL ES and Vulkan.

"KHR_glsl_shader_versions" : [
    {
         "api" : "opengl"
         "version" : "140";
         "uri" : "opengl_glsl140.vert";
    },
    {
         "api" : "opengles"
         "version" : "130 es";
         "uri" : "opengles_glsl130es.vert";
    },
    {
         "api" : "vulkan"
         "version" : "450";
         "uri" : "vulkan_glsl450.vert";
    }
]

javagl:
Again, the idea is very vague. Whether it should/could be

    "api" : "opengl"
    "version" : "130 es"

or

    "api" : "opengles"
    "version" : "130"

are things that have to be sorted out by the experts, because I can't say which level of detail or structure would be most appropriate here.  

My main point was about the fact that the strings `"uriGlslOpenGLES"` etc. haven't been specified in any form. They could only be matched (verbatim) by the client, and did not contain more detailed version infos. Considering the differences between the shader versions (that you pointed out in the other PRs), an additional version information may be reasonable. 

A basic guideline of how the client is supposed to *use* this information could also be helpful - something like "Pick the shader based on the `"api"`, then the highest version that you support. If you don't find one, then...". But this may be a next step.

mre4ce:
OpenGL and OpenGL ES are different APIs with different features and extensions so I do think it is important to make the distinction.

Valid point about adding guidelines for how a glTF loader is supposed to use this data.

mre4ce:
We should probably also list the extensions used per shader so a glTF loader can check whether those extensions are support on the platform:

```
"KHR_glsl_shader_versions" : [
    {
         "api" : "opengl",
         "version" : "140",
         "uri" : "opengl_glsl140.vert"
    },
    {
         "api" : "opengles",
         "version" : "130 es",
         "extensinos" : [
             GL_OVR_multiview2
         ],
         "uri" : "opengles_glsl130es.vert"
    },
    {
         "api" : "vulkan",
         "version" : "450",
         "uri" : "vulkan_glsl450.vert"
    }
]
```

Thanks Marco, this has been some really good feedback!

javagl:
Regarding the extensions: I'm not sure about the best way to differentiate/specify this, but note that this may be strongly related to the `glExtensionsUsed` that is introduced in glTF 1.1: https://github.com/lexaknyazev/glTF/tree/ecfc8f69bb5c6e5d92aaf995831c7287d8083b50/specification#specifying-gl-extensions

mre4ce:
Thanks, I had not see that yet. That would lock a glTF into a specific platform which I think is wrong. Specifying the GL extensions per shader allows an one glTF to support various different platforms, in particular in combination with KHR_spirv_shader_versions, KHR_hlsl_shader_versions and KHR_metalsl_shader_versions.

javagl:
This point about the possible side effects of the `glExtensionsUsed` may be important. As glTF 1.1 is not completely finalized, maybe @lexaknyazev and @pjcozzi should have another look at this...

pjcozzi:

I suppose this would need to be reworked given that shaders were moved to [KHR_techniques_webgl](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_techniques_webgl/README.md).

PR #808, opened 2017-01-03
KHR_glsl_joint_buffer

mre4ce:


lexaknyazev:
1. According to OpenGL ES 3.0 specs, GLSL ES version string is `#version 300 es`.
2. All platforms that support uniform buffers, also support integer attributes. 
So, can we use `in ivec4 a_joint;` for joint indices?

pjcozzi:

PR #809, opened 2017-01-03
KHR_glsl_multi_view

mre4ce:


pjcozzi:
Wow @mre4ce, thanks for all these extensions!  We'll focused on glTF 1.1 right now, but will try to find someone to review these soon.  Anyone is welcome!

mre4ce:
More extensions are coming. Maybe these will all translate directly to glTF 1.1 or at least they will provide a reference for what we need for high performance rendering.

pjcozzi:

PR #811, opened 2017-01-04
Link to current (KHR) binary extension spec.

donmccurdy:
I assume EXT should now be KHR? The current link is broken, if not.


pjcozzi:
Yes, good catch, thanks @donmccurdy!

PR #814, opened 2017-01-05
KHR_glsl_view_projection_buffer

mre4ce:


pjcozzi:

PR #633, opened 2016-06-26
KHR_materials_common

lexaknyazev:

PR #818, opened 2017-01-18
Remove an extra row from Light Types table of KHR_materials_common

takahirox:
This PR removes an extra row from Light Types table of KHR_materials_common

https://github.com/KhronosGroup/glTF/blob/master/extensions/Khronos/KHR_materials_common/README.md#light-types

It seems like been wrongly added by #478


pjcozzi:
Thanks @takahirox.

@lexaknyazev what do you think?

lexaknyazev:
That was fixed in https://github.com/KhronosGroup/glTF/pull/633/commits/e047fa661df34fa1b74f09ca384aba6f8f578ad3.

Issue #742, opened 2016-10-10
Multi-uv support ?


AurL:
Hi, 

There are some use-cases that would require multi-uv related data in glTF assets and I wanted to ask here and have your feedbacks about it:

Let's say I have a scene with several objects, materials and textures, and I bake a single AO map that covers the whole scene. I will need to define a new UV channel for each object.
I export my scene into glTF in order to render it in a WebGL engine. The glTF specification allows meshes to have several TEXCOORDs, but if I am right, there is no property that allow to know which UV channel is used by a texture (or a material channel).

Regarding the glTF specification, what would be the most relevant place to put this data ? (We obviously don't want to read it from the shader, and it makes sense outside of a specific implementation)

Thanks for your time!

lexaknyazev:
> The glTF specification allows meshes to have several TEXCOORDs, but if I am right, there is no property that allow to know which UV channel is used by a texture (or a material channel).

You're right: there's no generic way to bind TEXCOORD_n to a specific texture, since all technique/material stuff is based on GLSL.

---

@pjcozzi @mre4ce @mlimper 

Considering [ongoing efforts](https://github.com/KhronosGroup/glTF/issues?q=is%3Aissue+is%3Aopen+label%3APBR) to decouple glTF from specific shaders and provide shaderless PBR materials, do we want to address this use case in the core spec?

lexaknyazev:
One possible way to handle multi-UVs is to change texture-related `material.values` properties from a plain ID to an object with two fields (example with PBR, other material extensions could be changed alike):
```JavaScript
{
  "meshes" : {
    "mesh_id": {
      "primitives": [
        {
          "attributes": {
            "NORMAL": "accessor_id0",
            "POSITION": "accessor_id1",
            "TEXCOORD_0": "accessor_id2",
            "TEXCOORD_1": "accessor_id3"
          },
          "material": "material_id"
        }
      ]
    }
  },
  "materials" : {
    "material_id": {
      "extensions": {
        "FRAUNHOFER_materials_pbr" : {
          "materialModel" : "PBR_metal_roughness",
          "values": {
            "baseColorFactor": [ 0.5, 0.5, 0.5, 1 ],
            "metallicFactor": 0.0,
            "roughnessFactor": 0.2,
            "metallicRoughnessTexture": {
              "texture": "texture_id",
              "semantic": "TEXCOORD_1" // defaults to "TEXCOORD_0"
            }
          }
        }
      }
    }
  }
}
```
[...]

Issue #736, opened 2016-10-04
FlipY flag in gltf ? 


AurL:
Hi,

I have a question regarding this point in the glTF spec:

> **Implementation Note**: With WebGL API, the first pixel transferred from the `TexImageSource` (i.e., HTML Image object) to the WebGL implementation corresponds to the upper left corner of the source. Non-WebGL runtimes may need to flip Y axis to achieve correct texture rendering.

Do we really expect _almost_ all the 3D softwares to flip all their textures to fit this model so that referenced images are correctly oriented in the output glTF file?
Is there any plan on adding a flag to know if a texture actually needs to have the Flip-Y flag enabled?

I understand that glTF is built closely to OpenGL spec (that uses top/left corner as image origin), but when we think about it's use as transport format by 3d softwares (and also as a replacement of the proprietary FBX format), this point seems to bring a limit that could be seen as a real drawback.

What's the actual reason doing this ? Maybe I am missing something?

Any detail is welcome, thanks for your time!

lexaknyazev:
glTF 1.0 targets mainly WebGL 1.0. Next version of the format should have pixel transfer flags.

lexaknyazev:
Also see #674.

donmccurdy:
I'm working on PBR support in THREE.GLTFLoader, and noticing that the Sketchfab/Unity glTF exporter [includes a `flipY` flag](https://github.com/sketchfab/Unity-glTF-Exporter/blob/master/GlTF_Texture.cs#L37).

Is this intended for the spec, or should I file a bug with the exporter?

AurL:
It not intended for the spec or a bug actually. At the moment of the development, we added this flag in order to know what was the orientation of the textute. We added this flag as a fallback as we were generating files from Unity that has it set to true (contrary to glTF that has it false) 
It's not official, as mentionned in the README.

Moreover, materials doesn't perfectly fit the specification of the PBR extension. For example, we split metalness and roughness into two distinct textures. 

This code is being updated to fit with the progress of glTF 2.0, but for now you might find some out-of spec contents in our files

donmccurdy:
Ok great, thanks!

pjcozzi:
Thanks for the quick response, @AurL.  @donmccurdy we're really looking forward to glTF 2.0 support in three.js!

@lexaknyazev OK to close this?

lexaknyazev:
Yes, this could be closed and texture origin should be defined as a part of #835.

Issue #699, opened 2016-08-26
[PBR extension] Additional maps (normal maps, displacement, ...?)


mlimper:
We need to check if / how bump maps / normal maps could be integrated.

cedricpinson:
Yeah it's clearly missing, why not adding a normal and bump channel ? I guess we only need to decide if we add extra information like normal map coordinate (DX/GL) or if the coordinates must be fixed.

mlimper:
> I guess we only need to decide if we add extra information like normal map coordinate (DX/GL) or if the coordinates must be fixed.

Agreed - maybe also information about the space of the normals... could make sense to allow for object space and tangent space?

cedricpinson:
yeah maybe, I dont a lot of usage or request for this on our plateform, but it's maybe safer to add this information

erich666:
As usual, I'd say "do as three.js does", and they have [these](https://threejs.org/docs/index.html#Reference/Materials/MeshPhongMaterial) - normal maps seem like a minimum since they've been in use well over a decade in interactive apps. Heck, I get requests to add normal maps and specular maps to Mineways, of all things - people do use these. Firms such as Simplygon use these in order to make content work on the Hololens, for example, see [this video](https://youtu.be/_RGGV4KUsrM?t=68), they use a normal map to capture fine detail instead of a zillion polys.

Three.js goes wild in this area: normal/bump, displacement (vertex shifts), roughness, metalness, alpha, and the IBL (environment map). I do hope you're planning on IBL, that's a must (@cedricpinson I suspect will agree, as Sketchfab uses these a lot). Lighting doesn't get covered in the PBR spec, and unfortunately lighting matters to materials. Again, I'd just steal the three.js model, maybe lopping off some of the more obscure textures for now.

> yeah maybe, I dont see a lot of usage or request for this on our plateform, but it's maybe safer to add this information

I agree, though have limited knowledge in this area. By the "WW3JS do?" principle, I don't see it there, so let's not include it here.

cedricpinson:
Ok here some stats on sketchfab:

**PBR workflow usage:**
80% metalness
20% specular

**Channels usage:**
normal/bump 100%
emit     	        30%
f0       	        30% only used on metalness workflow
opacity  	        15%
ao       	        60%
cavity   	        12%
roughness	76%
glossiness       24%

mlimper:
[...]

Issue #697, opened 2016-08-26
[PBR extension] Provide example implementation and specify (suggested) shading model


mlimper:
We need and simple, yet expressive reference application, and especially example shader code for reference.
Also, the shading model we use must be very concretely defined in either the spec or in the appendix, being a suggested shading model.

This also involves to create corresponding, extensive documentation in the spec, especially about the concrete shading model and exact behaviour / semantics of parameters.

cedricpinson:
There are references implementation well documented:
- ue4 Real Shading in Unreal Engine 4 (Brian Karis) siggraph 2013 http://blog.selfshadow.com/publications/s2013-shading-course/
- frostbite Moving Frostbite to PBR (Sébastien Lagarde & Charles de Rousiers)  http://blog.selfshadow.com/publications/s2014-shading-course/

There is no only one way to write PBR shader so if we provide an example / code sample we will need to choose one implementation.

erich666:
There's "documented" and there's "coded" - I massively prefer coded, as code is clear. One problem with the ancient VRML specification was that the shader (and other) behavior wasn't specified, and 5 browser plugins and viewers implemented 5 different models (gamma correction "per surface" on and off, specular computed with the light at infinity vs. locally (back when we used to carefully count instructions), etc.). Having a consistent look among viewers of glTF PBR is critical - done right, you'll see on your screen what I'm seeing on my screen. We don't have to go so crazy as to implement ICC profiles, assume sRGB, but there's no reason that we shouldn't get everything else right.

An example, in talking with Brian Karis about his SIGGRAPH 2013 paper and how it bears on this proposal, here's his reply to a long email I had with questions.

> You wrote [here](https://de45xmedrsdbp.cloudfront.net/Resources/files/2013SiggraphPresentationsNotes-26915738.pdf): "We found variable index of refraction (IOR) to be fairly unimportant for nonmetals, so we have recently replaced Specular with the easier to understand Cavity parameter."

BK: That was dumb. We didn't end up doing that. We still have Specular and use it now more than we did.

----

So going by what's in an article can be misleading at best. There are also times when an article says one thing, but the actual code does something different, for any number of reasons.

I thought I'd pass on some other comments from Brian, since they may be of interest.

> Also, to me a separate specularFactor and glossinessFactor is a tad odd, it definitely implies that you’re using two different reflection maps (one sharp, one blurred to some fixed amount), period, and you want to control each separately.

[...]

PR #830, opened 2017-02-04
Merge PBR Metallic-Roughness into glTF 2.0

sbtron:
Removed all references to techniques, shaders, programs.
Added Metal Roughness to Materials along with additional maps, multi uv etc.

Still a few TODOs called out in the markdown but enough for people to start reviewing.
https://github.com/sbtron/glTF/tree/2.0/specification/2.0


specification/2.0/README.md
```diff
@@ -2183,38 +2150,65 @@ Application-specific data.
 <a name="reference-material"></a>
 ## material
 
-The material appearance of a primitive.
+glTF 2.0 uses the metal-roughness material model from Physically Based Renderring. The metal-roughness material model is defined by the following properties:
```
lexaknyazev: Property reference section will be auto-generated from JSON-Schemas. This brief description should go into Materials (above) part of README, while exact equations and implementation details should be in a new Appendix.

lexaknyazev:
Why were all material properties moved out of `material.values` object? This could make future material updates more complicated.

Why not keep existing layout (with `model` instead of `technique`):
```json
{
    "materials" : [
        {
            "model" : "METALLIC_ROUGHNESS",
            "values" : {
                "baseColorFactor": [0.5, 0.5, 0.5, 1],
                "mettalicRoughnessTexture": 1,
                "normalTexture": 1,
                "normalTexCoord": 1,
                "emissionFactor": [0.2, 0.1, 0.0]
            }            
        }
    ]
}
```

Or use the same approach as cameras:
```json
{
    "materials" : [
        {
            "model" : "metallicRoughness",
            "metallicRoughness" : {
                "baseColorFactor": [0.5, 0.5, 0.5, 1],
                "mettalicRoughnessTexture": 1,
                "normalTexture": 1,
                "normalTexCoord": 1,
                "emissionFactor": [0.2, 0.1, 0.0]
            }
        }
    ]
}
```

specification/2.0/README.md
```diff
@@ -2183,38 +2150,64 @@ Application-specific data.
...
-
-Additional properties are not allowed.
-
+glTF 2.0 uses the metal-roughness material model from Physically Based Renderring. The metal-roughness material model is defined by the following properties:
+* Base Color: The base color of the material
+* Metallic-ness: A value between 0 and 1, indicating if the material is metallic (1) or not (0)
+* Roughness: Roughness of the material surface
+
+Todo: Need a good figure illustrating this
+
```
javagl: Renderring -> Rendering (also appeared above, search for "Renderring")

As for the figure: https://github.com/tsturm/glTF/tree/master/extensions/Vendor/FRAUNHOFER_materials_pbr#metal---roughness could be a start (but maybe a simpler one, or one involving a Khronos/glTF logo would be nicer)

specification/2.0/README.md
```diff
@@ -2183,38 +2150,64 @@ Application-specific data.
...
+The `base` color has two different interpretations depending on the value of `metallic`, which is defined as `0` for dielectrics and `1` for metals.
+For `metallic = 1`, `base` is the specific measured reflectance value (F0).
+For `metallic = 0`, `base` represents the reflected diffuse color of the material.
+
+In this model it is not possible to specify a reflectance value for non-metals, where a reflectance value of 4% (0.04) is often used.
+The `roughness` property is related with the `glossiness` parameter in the specular-glossiness model and is defined as `roughness = 1 - glossiness`. 
+
```
javagl: Although it is shortly *mentioned* in the "meshes" section, the term **"set index"** is used several times here, but it might not be clear for everybody what this refers to. Elaborating it here (or detailing it at another point and adding a link) may be helpful.

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
-            "type" : "object",
-            "description" : "A dictionary object of parameter values.",
-            "properties" : {
-            },
-            "additionalProperties" : {
-                "$ref" : "material.values.schema.json"
-            },
-            "default" : {},
-            "gltf_detailedDescription" : "A dictionary object of parameter values.  Parameters with the same name as the technique's parameter override the technique's parameter value."
+        "baseColorFactor" : {
```
javagl: I assume that the valid ranges are still being defined? (Just wondering, because in the README, the ranges for `metallicFactor` and `roughnessFactor` are already given, but here, only  the `roughnessFactor` has a `minimum` and `maximum`...)

sbtron:
Good point I dropped the materialModel thinking core spec will always be Metal Roughness and the extension will be Spec Gloss but future versions of core could maybe include spec goss or other material models. Will update to have a materialModel which is a string and only allowed value for 2.0 will be "Metal-Roughness". Future we could add additional string values. Does that work?

lexaknyazev:
> Will update to have a materialModel which is a string and only allowed value for 2.0 will be "Metal-Roughness". Future we could add additional string values. Does that work?

Yes. Maybe exact string enums should be refined to be more consistent across whole spec - @pjcozzi.

pjcozzi:
@sbtron congrats, this is a **BIG** pull request for glTF 2.0!

@mlimper can you please review?

@moneimne can you please carefully review this based on your implementation experience and verify that all the right details are provided?

pjcozzi:
>> Will update to have a materialModel which is a string and only allowed value for 2.0 will be "Metal-Roughness". Future we could add additional string values. Does that work?

> Yes. Maybe exact string enums should be refined to be more consistent across whole spec - @pjcozzi.

Consider renaming "Metal-Roughness" to "METAL_ROUGHNESS" so it is consistent with uniform semantics, attribute semantics, animation interpolation, etc.

@lexaknyazev what exactly are you asking?  What is not consistent in the existing spec?

lexaknyazev:
> @lexaknyazev what exactly are you asking? What is not consistent in the existing spec?

Exactly this:
> Consider renaming "Metal-Roughness" to "METAL_ROUGHNESS"

Also, we usually don't repeat names of parents, i.e., `model` instead of `materialModel`.

pjcozzi:
> Also, we usually don't repeat names of parents, i.e., model instead of materialModel.

Ah, yes, good eye!

javagl:
PBR is somewhat complex - at least for me, and I'm still trying to wrap my head around it, and do the step from the `technique`-based rendering to the PBR rendering. I'm not sure where to ask for clarifiction, but hope it is OK to do this here. Otherwise, just give me a pointer.

Right now, it seems that it should be possible to have a material like this:

    "materials" : [
        {
            "model" : "METALLIC_ROUGHNESS",
            "values" : {
                "baseColorTexture": 12,
                "baseColorTexCoord": 0,
                ...
            }            
        }
    ]

This material description is *in no way* compatible with or related to a `technique` (as it was used in 1.0, and as it is described in the `KHR_technique_webgl` - is this correct? 

Particularly, the `baseColorTexture` here is a `GL_INT`, which is the *index* of a `texture` object. So there has to be some magic in the renderer to "convert" this material into a material that **could** be used with a (PBR) `technique` - namely, into a `material` where the value for `baseColorTexture` is of type `GL_SAMPLER2D`. Similarly, the `baseColorTexCoord` is a `GL_INT` that is supposed to be appended to a `"TEXCOORD_"` string, to figure out that the `semantic` of this property should be `"TEXCOORD_0"`, which is later to be looked up in the `mesh.primitive.attributes` - right?

sbtron:
@lexaknyazev 
Few more things I was considering: 
- Explicitly calling out that this is a PBR material model. For Metal Roughness this might be obvious but it can get confusing with Spec-Gloss which can also be used for non-PBR. So just consistently calling out its would be good.
- Split apart the additional maps better as they don't necessarily apply to metal roughness and can be used by other workflows including non-pbr ones.

```
"materials" : [
        {
            "type" : "pbrMetalRoughness",
            "pbrMetalRoughness" : {
                "baseColorFactor": [0.5, 0.5, 0.5, 1],
                "metallicRoughnessTexture": 1
            },
            "normalTexture": 1,
            "normalTexCoord": 1,
            "emissionFactor": [0.2, 0.1, 0.0]
        }
    ]
```
Thoughts on the casing for pbrMetalRoughness ?

lexaknyazev:
> Split apart the additional maps better as they don't necessarily apply to metal roughness and can be used by other workflows including non-pbr ones.

Having some maps/factors outside of a particular material model makes them mandatory for all other material models, right? I don't have a strong opinion here. Could you give two full-featured (with all possible maps) examples with different models (metal-rough & spec-gloss)?

> Explicitly calling out that this is a PBR material model. For Metal Roughness this might be obvious but it can get confusing with Spec-Gloss which can also be used for non-PBR. So just consistently calling out its would be good.

Is it a question about applying some sort of Hungarian notation to material types? In case of Spec-Gloss, are we expecting two flavors of it (PBR and non-PBR)?

sbtron:
> Having some maps/factors outside of a particular material model makes them mandatory for all other material models, right?

Sort of, assets choose to use what they need of course so the maps are not always required to be present in the asset. Adding it to the core means all runtime implementations should support them. But there will likely be some resource limitations across platforms and if a runtime can't support all of them due to resource restrictions we provide an order of preference on how it should drop support for additional maps so the visual quality can degrade gracefully. This should address @mlimper 's [comment](https://github.com/KhronosGroup/glTF/pull/643#issuecomment-273470880).
Looking at the common materials extension - the ambient and emission maps defined in the extension right now could just be these additional maps in the core. So no need of defining them in the extension again as they should really be the same thing and can address this [comment](https://github.com/KhronosGroup/glTF/issues/699#issuecomment-270061461)


> Could you give two full-featured (with all possible maps) examples with different models (metal-rough & spec-gloss)?

Some examples using the older extension model that required you to define the maps for each material type: 
https://github.com/sbtron/BabylonJS-glTFLoader/blob/master/Models/Telephone/PBR-MetalRoughness/Telephone.gltf#L149-L150
https://github.com/sbtron/BabylonJS-glTFLoader/blob/master/Models/Telephone/PBR-SpecGloss/Telephone.gltf#L150-L151

> Is it a question about applying some sort of Hungarian notation to material types? In case of Spec-Gloss, are we expecting two flavors of it (PBR and non-PBR)?

We aren't really expecting two flavors but just wanted to make it explicitly clear that this is referring to the PBR workflow. I believe there were a few discussions earlier where certain texture names also overlap with "classical" shading approaches like those in common materials and its good to clarify that the specular/shininess used with common materials is not quite the same as the specular/glossiness used in PBR materials. This can be made explicit in the spec doc itself but thought the json should also call it out.

javagl:
> Could you give two full-featured (with all possible maps) examples with different models (metal-rough & spec-gloss)?

That would be great. The telephone may be a start. A while ago I tried to prepare https://github.com/javagl/gltfTestModels/tree/master/DamagedHelmetModified for tests. It's basically 1.1, thus, still using techniques, but thought it might be helpful to have something to compare PBR implementations against. (The shaders do not yet handle all the texture types, though...)

specification/2.0/schema/examples/materials.json
```diff
@@ -1,26 +1,13 @@
...
-            },
-            "extras" : {
-                "Application specific" : "The extra object can contain any properties."
-            }     
+    "materials" : 
+        [
+        {
+        "name": "Material0",
+        "baseColorFactor": [ 0.5, 0.5, 0.5, 1 ],
+        "mettalicRoughnessTexture": 1,
```
mlimper: small typo (-> "metallicRoughnessTexture")

specification/2.0/schema/examples/materials.json
```diff
@@ -1,26 +1,13 @@
...
+        "normalTexture": 1,
+        "normalTexCoord": 1,
+	    "emissionFactor": [0.2,0.1,0.0]
```
mlimper: Ah, I recall we talked about that - guess that's still missing in our PBR spec draft - thanks for adding it here.
I would personally have called it "emissiveFactor", would match with "diffuse" / "specular"

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+        },
+        "occlusionTexture" : {
+            "extends" : "number",
+            "description" : "The index the texture containing the occlusion map."
+        },
+        "occlusionTexCoord" : {
+            "extends" : "number",
+            "description" : "The set index of occlusionTexture's TEXCOORD attribute to be used for texture coordinate mapping"
+        },
+        "occlusionStrength" : {
```
mlimper: could this also be "occlusionFactor", to match with the other "...factors"?

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+        },
+        "emmissionTexture" : {
+            "extends" : "number",
+            "description" : "The index the texture containing the emission map."
+        },
+        "emmissionTexCoord" : {
+            "extends" : "number",
+            "description" : "The set index of emmissionTexture's TEXCOORD attribute to be used for texture coordinate mapping"
+        },
+        "emmissionStrength" : {
```
mlimper: same here (-> call it emissionFactor / emissiveFactor ?)
typo (-> emission)

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
-            "gltf_detailedDescription" : "A dictionary object of parameter values.  Parameters with the same name as the technique's parameter override the technique's parameter value."
+        "baseColorFactor" : {
+            "type" :"array",
+            "description" : "The material's base color factor.",
+            "default" : [1,1,1,1],
+            "gltf_detailedDescription" : "The RGBA components of the base color (RGB) of the material. The fourth component (A) is the opacity of the material."
+        },
+         "baseColorTexture" : {
+            "extends" : "number",
+            "description" : "The index of the base color texture."
```
mlimper: In the original proposal this was a string, referring to a glTF texture (a glTF texture id) - did this become a numerical id / index for 2.0?

sbtron: Yup this is to account for https://github.com/KhronosGroup/glTF/issues/831

specification/2.0/schema/material.schema.json
sbtron: Yup went back and forth between using factor for consistency vs trying to use something that explains what it is :). Just thought I will propose an alternative in case people like it. Since all of them are multipliers factor works well to and will update it to that,

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+        },
+        "baseColorTexCoord" : {
+            "extends" : "number",
+            "description" : "The set index of the base color texture's TEXCOORD attribute to be used for texture coordinate mapping."
+        },
+        "metallicFactor" : {
+            "type" : "number",
+            "description" : "The metallic-ness the material (1 for metals, 0 for non-metals).",
```
moneimne: The metallic-ness the material -> The metallic-ness **of** the material

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+            "default" : 1
+        },
```
moneimne: Is there a reason that metallicFactor doesn't have a minimum and maximum value like roughnessFactor?

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+        "roughnessFactor" : {
+            "type" : "number",
+            "description" : "The perceptual roughness of the material surface (1 for rough, 0 for smooth)..",
```
moneimne: 0 for smooth).." -> 0 for smooth)." (Double period)

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+            "extends" : "number",
+            "description" : "The index of the metallic-roughness texture."
+        },
+        "metallicRoughnessTexCoord" : {
+            "extends" : "number",
+            "description" : "The set index of the metallic-roughness texture's TEXCOORD attribute to be used for texture coordinate mapping."
+        },
+        "normalTexture" : {
+            "extends" : "number",
+            "description" : "The index the texture containing the normal map."
```
moneimne: The index the texture -> The index **of** the texture

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+            "extends" : "number",
+            "description" : "The set index of normalTexture's TEXCOORD attribute to be used for texture coordinate mapping"
+        },
+        "normalScale" : {
+            "extends" : "number",
+            "description" : "The scalar multiplier applied to each normal vector of the texture."
+        },
+        "occlusionTexture" : {
+            "extends" : "number",
+            "description" : "The index the texture containing the occlusion map."
```
moneimne: The index the texture -> The index **of** the texture

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+            "extends" : "number",
+            "description" : "The set index of the metallic-roughness texture's TEXCOORD attribute to be used for texture coordinate mapping."
+        },
+        "normalTexture" : {
+            "extends" : "number",
+            "description" : "The index the texture containing the normal map."
+        },
+        "normalTexCoord" : {
+            "extends" : "number",
+            "description" : "The set index of normalTexture's TEXCOORD attribute to be used for texture coordinate mapping"
```
moneimne: coordinate mapping" -> coordinate mapping." (Missing period)

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+            "extends" : "number",
+            "description" : "The scalar multiplier applied to each normal vector of the texture."
+        },
+        "occlusionTexture" : {
+            "extends" : "number",
+            "description" : "The index the texture containing the occlusion map."
+        },
+        "occlusionTexCoord" : {
+            "extends" : "number",
+            "description" : "The set index of occlusionTexture's TEXCOORD attribute to be used for texture coordinate mapping"
```
moneimne: coordinate mapping" -> coordinate mapping." (Missing period)

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+        },
+        "occlusionStrength" : {
+            "extends" : "number",
+            "description" : "A scalar multiplier controlling the amount of occlusion applied. A value of 0 means no occlusion; 1.0 means full occlusion."
```
moneimne: A value of 0 means -> A value of 0.0 means (Consistency between 0.0 and 1.0 in next phrase?)

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+        },
+        "emmissionTexture" : {
+            "extends" : "number",
+            "description" : "The index the texture containing the emission map."
+        },
+        "emmissionTexCoord" : {
+            "extends" : "number",
+            "description" : "The set index of emmissionTexture's TEXCOORD attribute to be used for texture coordinate mapping"
```
moneimne: coordinate mapping" -> coordinate mapping." (Missing period)

specification/2.0/schema/material.schema.json
```diff
@@ -5,28 +5,79 @@
...
+            "extends" : "number",
+            "description" : "The set index of occlusionTexture's TEXCOORD attribute to be used for texture coordinate mapping"
+        },
+        "occlusionStrength" : {
+            "extends" : "number",
+            "description" : "A scalar multiplier controlling the amount of occlusion applied. A value of 0 means no occlusion; 1.0 means full occlusion."
+        },
+        "emmissionTexture" : {
+            "extends" : "number",
+            "description" : "The index the texture containing the emission map."
```
moneimne: The index the texture -> The index **of** the texture

specification/2.0/README.md
```diff
@@ -2183,38 +2150,64 @@ Application-specific data.
...
-|**values**|`object`|A dictionary object of parameter values.|No, default: `{}`|
-|**name**|`string`|The user-defined name of this object.|No|
-|**extensions**|`object`|Dictionary object with extension-specific objects.|No|
-|**extras**|`any`|Application-specific data.|No|
-
-Additional properties are not allowed.
-
+glTF 2.0 uses the metal-roughness material model from Physically Based Renderring. The metal-roughness material model is defined by the following properties:
+* Base Color: The base color of the material
+* Metallic-ness: A value between 0 and 1, indicating if the material is metallic (1) or not (0)
```
moneimne: The [Sketchfab models](https://sketchfab.com/features/gltf) include metallic-ness values that are between 0 and 1, not just 0 or 1. Is this an exception to the spec?

specification/2.0/schema/material.schema.json
bghgary: I'm okay either way, but I want to point out that occlusionStrength and normalScale are a bit different than the other factors in that they are invalid without the corresponding textures.

mlimper: I'm okay either way, but I want to point out that occlusionStrength and normalScale are a bit different than the other factors in that they are invalid without the corresponding textures.

Oh, that is a good point. Better leave it like that then.

javagl:
Although the specs and extensions are still WIP: Assuming one uses `KHR_techniqe_webgl` and wants to define a material there - is this supposed to happen via an extension object in the `material` that contains the index of the `technique`?

lexaknyazev:
At the moment, I'm leaning towards this approach:
```javascript
{
  "materials": [
    {
      "extensions": {
        "KHR_technique_webgl": { 
          "technique": 0 
        }
      }
    }
  ],
  "extensions": {
    "KHR_technique_webgl": {
      "techniques": [
        // technique object
      ]
    }
  }
}
```

We need to evaluate compatibility of such scheme with possibility of several techniques (targeting different APIs, #807). So, maybe (just saying) all techniques should share same array or each technique could have different "implementations".

javagl:
OK, so there likely will be some extension object in the material, and this will somehow establish the link to the technique.  

---

(For a moment, I thought that there might be a material like
    
    "mode" : "KHR_TECHNIQUE_WEBGL",
    "values" : {
        "techniqueId" : 42
    }

but that likely doesn't make sense at all: The core should not be aware of the extension)

---
    
The technique "implementations" point sounds reasonable. But I think that it still has to be sorted out at which point the version-specific differentiation should take place - i.e. whether it happens by using different `technique` objects, or a single `technique` that refers to different, version-specific `program` objects. (This PR may not be the best place to discuss this. If necessary, this could be moved to the PR that you linked to, or into a dedicated issue....)

sbtron:
@lexaknyazev  could you take a look at the latest sample and schemas. It should have the texture objects and other changes discussed.
I am still not convinced we should add a type or model enum in there. Unlike the camera example its not an either or case. You could have a material that has both MetallicRoughness as well as SpecGloss values defined. Spec Gloss could be sort of an addition to MR which you can load if your engine supports it or if not you can fall back on MR. 

```
    "materials": [
        {
            "pbrMetallicRoughness": {
                "baseColorfactor": [ 0.5, 0.5, 0.5, 1.0 ],
                "baseColorTexture": {
                    "index": 1,
                    "texCoord": 1
                },
                "metallicFactor": 1,
                "roughnessFactor": 1,
                "metallicRoughnessTexture": {
                    "index": 2,
                    "texCoord": 1
                }
            },
            "extensions": {
                "KHR_materials_pbrSpecularGlossiness": {
                    "diffuseTexture": {
                        "index": 0
                    },
                    "glossinessFactor": 0.5,
                    "specularGlossinessTexture": {
                        "index": 0,
                        "texCoord": 0
                    }
                }
            },
            "normalTexture": {
                "index": 3,
                "texCoord": 1
            },
            "emmissionTexture": {
                "index": 4,
                "texCoord": 1
            },
            "occlusionTexture": {
                "index": 5,
                "texCoord": 1
            }
        }
    ]
}
```
The above shows example with SpecGloss extension but similar scenarios would apply to Common_Materials or GLSL extension. If you wanted to something specific with GLSL then say you could add the GLSL extension but if you still wanted to maintain portability you would include the MR properties in there for engines that won't load GLSL code. 
Type could be added later if there was more than one material model that you could choose from in the core spec but I don't think its needed at the moment?

lexaknyazev:
> Spec Gloss could be sort of an addition to MR which you can load if your engine supports it

Just to confirm: in case when both are present, does SpecGloss extend MR or supersede it in terms of implementation? I also see, that they use different maps. All of them must be provided in the asset, correct? Do we want to support such redundancy (just asking)? This should be fine for online usage when client dynamically decides what to download, not sure about offline clients.

> The above shows example with SpecGloss extension but similar scenarios would apply to Common_Materials or GLSL extension.

So in the worst case a material could contain at the same time: MR, SpecGloss, CommMat, GLSL. And it would be up to engine to decide what to use. Maybe we should suggest priority there.

> Type could be added later if there was more than one material model that you could choose from in the core spec but I don't think its needed at the moment?

So the proposal is to keep `material.pbrMetallRoughness`, but remove `material.model`, right? This could work since it's the only model for 2.0 core. I'm still not quite convinced with `pbr` prefix in its name, since there's no non-PBR MR. But as long as PBR team and @pjcozzi agree on that, let's keep it.

javagl:
The spec currently says 

>  If a texture is not given, all respective texture components are assumed to have a value of 1.0. 

Should this also apply to the *normal* texture?

mlimper:
> Should this also apply to the normal texture?

No, and the normal texture should not be considered part of the PBR part of the spec.
There will be a separate part for this, so it can also be shared between PBR and materials_common

vorg:
About the `normalTexture` currently the spec from this pull request says:

> The index of the texture containing the normal map (RGB) of the material. Each texel represents the XYZ components of a normal vector in tangent space.

does this assume R=X, G=Y, B=Z? and additionally that `tangentNormal = sampledNormal * 2 - 1` will that clarification be in the spec or extension?

bghgary:
Yes, we are planning on adding that information. See the schema where we already have a TODO :)

https://github.com/sbtron/glTF/blob/077114e95302a603f76d3ee375d96c2d407f0331/specification/2.0/schema/material.schema.json#L28)

mlimper:
> A lot of these comments are just small typos. I also asked if the spec should expect metallic-ness values of 0 or 1 when models such as the ones on Sketchfab include values in between. What are your thoughts on this?

We would need more feedback from true PBR experts on this... my understanding was that it will not be "physically-based" any more if you allow to freely mix between metallic and dielectric.

sbtron:
>  I also asked if the spec should expect metallic-ness values of 0 or 1 when models such as the ones on Sketchfab include values in between. What are your thoughts on this?

Yes values between 0 and 1 are valid. Already cleaned up the spec to explain better. lmk if any further clarifications are needed -  https://github.com/sbtron/glTF/blob/2.0/specification/2.0/README.md#pbrmetallicroughnessmetallicfactor

>We would need more feedback from true PBR experts on this... my understanding was that it will not be "physically-based" any more if you allow to freely mix between metallic and dielectric.

@mlimper not sure what you were referring to here. Values between metallic and dielectric are fairly common in the metallic-roughness model? Were you thinking about F0 values?

mlimper:
> @mlimper not sure what you were referring to here. Values between metallic and dielectric are fairly common in the metallic-roughness model? Were you thinking about F0 values?

Yeah, for F0 we already make a binary classification "metal or non-metal", right? As said, I'm not a PBR expert - my current impression was that the "correct" way would be to only allow for either the one or the other, while there could be a "not exactly physically-based" way of mixing them in the shader which would allow artists a bit more freedom... but that's just an educated guess... @jeffdr probably knows better?

javagl:
Wouldn't this also apply to the "metalness" *texture*? (This would then have to be a binary one - unlikely). I thought that the whole point of these factors would be this "interpolation" that was also shown in https://github.com/KhronosGroup/glTF/issues/697#issuecomment-257182909 ...

jeffdr:
Metalness is not binary (nor is the derived F0 value). Things usually look something like this:

    f0 = mix( 0.04, albedo.rgb, metalness )

where metallness is [0,1], and .04 is your chosen dieletric f0. The notion that metalness should be binary is more like good advice for artists starting out with the system (intermediate values being harder to reason about), but it is not a technical aspect.

One could, I suppose, encode the metalness value as a single bit (DXT1 alpha or something) and get away with it in many cases, but I wouldn't recommend doing that here. And textures aside, artists often want a 'metalness' scalar value for the material as well.

sbtron:
@jeffdr it would be great if you could review - https://github.com/sbtron/glTF/blob/2.0/specification/2.0/README.md#metallic-roughness-material

mlimper:
> The notion that metalness should be binary is more like good advice for artists starting out with the system (intermediate values being harder to reason about), but it is not a technical aspect.

Great, thanks a lot - this was really helpful to me.

jeffdr:
@sbtron The 'Metallic' section of that document reads fine to me.

javagl:
Please add information about what 
- *c<sub>diff</sub>*
- *F<sub>0</sub>*
- *&alpha;* and
- `epsilon`

are (unless this becomes unambiguously clear when the appendix is filled, together with references to the original papers)

bghgary:
The first 3 should be clear once the appendix is filled. See the original appendix for how it might look: https://github.com/tsturm/glTF/blob/master/extensions/Vendor/FRAUNHOFER_materials_pbr/Appendix.md

Epsilon is a small number to avoid edge cases (mostly division by zero) in the maths. Depending on the approximations used, it might need to be bigger or smaller to avoid precision issues. We can add something like this to the equations section.

javagl:
@bghgary Yes, such definitions should be added there as well. Unfortunately, much of the literature about PBR seems to be a bit lax about definitions. Moreover, the same functions seem to be implemented differently everywhere. The implementations may still be equivalent for other, non-obvious reasons - but there's just too much math behind that to manually derive each and every formula, starting with the first papers that have been written about that in the late 60's. For example, some seem to do the division by `4*NdotL*NdotV` in the BRDF, and others simply seem to omit it. So although they all refer to the same paper (Cook), they seem to use slightly different derivations or implementations - and those who *do* the division don't seem to mention anything about the case of a divide by zero. This, in turn, is related to the `epsilon`, which already caused [some debugging headaches](https://github.com/javagl/gltfTestModels/commit/2e6c0812f7f2597315299f07891e9a939ede7532#diff-57c787dcf5b07a7d9666c613542c5d2aR171) for me today...

However, hopefully, the efforts here will achieve their "implicit" goal - namely, to define something like *"A Standard Formulation and (CC0!) Reference Implementation for Metal-Roughnes-based PBR"* ...

AurL:
It's really cool to see that PBR is going to be in glTF core.
Side question: for now we are still using split textures (metallicTexture + roughnessTexture) for gltf PBR materials in our pipeline, and I wanted to know if texture packing was definitively decided ? 

Also, is it safe rely on this issue for our material update to target glTF 2.0 ? (or changes in the structure are still possible?)


Thanks for all the work done

javagl:
@AurL According to the discussion starting at https://github.com/KhronosGroup/glTF/pull/643#issuecomment-274874421 , the information will be stored in a combined metallic+roughness texture, and I did not see any other statements. (Something like https://javagl.github.io/MetallicRoughnessCreator/MetallicRoughnessCreator.html will hopefully become obsolte, and be part of the glTF pipeline tools, implemented by someone who knows this stuff)

sbtron:
> Also, is it safe rely on this issue for our material update to target glTF 2.0 ? (or changes in the structure are still possible?)

@AurL the changes in there already should be very stable. We also want to incorporate all your feedback in and make sure 2.0 can meet your needs.

Here's the summary of all the things in your unity exporter materials and how they match up with the spec 2.0 spec:
1) Texture Packing 
As per comment above we do want to keep the texture packing for runtime efficiency so exporters will have to create the correctly packed textures

2) Additional Maps (#699) 
We have added normal, occlusion, emissive https://github.com/sbtron/glTF/tree/2.0/specification/2.0#additional-maps
You also seemed to have bump maps in the unity exporter which isn't currently included. Do you need this or are normal maps good enough?

3) Multi-UV support (#742)
Added - https://github.com/sbtron/glTF/tree/2.0/specification/2.0#reference-textureInfo 

4) FlipY flag (#736)
Overall we still want to stay away from adding it to the spec. The exporter will have to flip the textures when exporting if they don't match. Similarly with yUp flag. (still need to add detail to the spec on which way it needs to be)

5) Transparency (#183)
This is the last major area we want to include. Will be updating that issue with a proposal shortly with details. 

So hoping the need for you to go off spec is minimized and everything is supported in 2.0. 

Also specular-glossiness material model will continue to be an extension but it has been updated to match the 2.0 changes. Would be good if you could review and provide feedback on that too - #842 

We have some sample models which show the 2.0 spec with pbr materials in action here - 
https://sbtron.github.io/BabylonJS-glTFLoader/
Models - https://github.com/lasalvavida/glTF-Sample-Models/pull/1

AurL:
Thanks @sbtron for these details, and sorry for the late reply.

We started writing glTF files with some "out of spec" parts but we are aiming to completely avoid them.
We are currently updating the exporter to make it compliant with the points you mentionned.

The only part that we would eventually need to discuss is about bump maps. We don't really want to alter source data so we will need to keep the bump map channel. Would it be possible and make sense to have them in the core spec? 

Specular-glossiness update looks fine for me, thanks for this too.
I will take a look at the samples from your link.
Also note that we should be able to generate clean versions of our samples in 2.0 soon as well, I'll keep you in touch.

Thanks

bghgary:
@cedricpinson posted these numbers from here https://github.com/KhronosGroup/glTF/issues/699#issuecomment-253219927 
> Channels usage:
> normal/bump 100%

@AurL @cedricpinson Do you have numbers on how often bump maps are used compared to normal maps?

See https://github.com/KhronosGroup/glTF/issues/699#issuecomment-270069246 for @mlimper's comment on bump maps.

jeffdr:
Some possible confusion in this thread. Terminology as I understand it:

- "Normal Maps" - Contain surface normal vectors
- "Height Maps" - Grayscale surface height
- "Bump Maps" - Refers either to Normal Maps (confusingly), or grayscale height. This term should be avoided IMO.

I think height maps are generally only used for parallax and/or displacement shaders, so unless that effect is to be included then I think this map could be omitted. Normal maps are used essentially always (which may be what @cedricpinson meant, since "bump" can mean normal maps to some people). Here are my rough estimates from Marmoset's user base:

- Normal Maps: Always
- Height Maps: Occasionally, where displacement or parallax are in use
- "Bump" Maps: Never

lexaknyazev:
PNG and JPG files could contain ICC profile/intent flags. By default, web-browsers apply such transformations and return sRGB data (or other device-default colorspace; WebGL spec isn't very precise on that topic).
To achieve consistency across different platforms, we could add a note that any ICC profile/intent information in PNG or JPG files must be ignored. In WebGL, such conversion could be disabled with:
```javascript
gl.pixelStorei(gl.UNPACK_COLORSPACE_CONVERSION_WEBGL, gl.NONE);
```

bghgary:
@jeffdr Thanks for the clarification! I had assumed that people are talking about height maps when they said bump maps due to the comments being made about displacement, but maybe that's a bad assumption.

@AurL Were you asking for support for height maps in the core spec?

specification/2.0/README.md
```diff
@@ -2470,14 +2470,14 @@ A set of parameter values that are used to define the metallic-roughness materia
    
 ### material.normalTexture
 
-A tangent space normal map. Each texel represents the XYZ components of a normal vector in tangent space. <TODO: Add more explaination on how XYZ is calculated>.
+A tangent space normal map. The texture contains RGB components in linear space. Each texel represents the XYZ components of a normal vector in tangent space. Red [0 to 255] maps to X [-1 to 1]. Green [0 to 255] maps to Y [-1 to 1]. Blue [128 to 255] maps to Z [0 to 1]. The normal vectors use OpenGL conventions where +X is right and +Y is up. +Z points toward the viewer.
```
lexaknyazev: > [0 to 255] maps to X [-1 to 1]...

I assume that we expecting something like this in GLSL:
```glsl
vec3 norm = tex_norm.rgb * 2.0 - 1.0;
```

So X, Y = 0 aren't representable, and Blue [128, 255] maps to Z [1/255 to 1] right?
 - `127 -> -1/255`
 - `128 -> +1/255`

specification/2.0/README.md
```diff
@@ -2470,14 +2470,14 @@ A set of parameter values that are used to define the metallic-roughness materia
...
 
 * **Type**: [`material.normalTextureInfo`](#reference-material.normalTextureInfo)
 * **Required**: No
 
 ### material.occlusionTexture
 
-The occlusion map is a greyscale texture, with white indicating areas that should receive full indirect lighting and black indicating no indirect lighting. <Todo: sRGB ? Color Space>.
+The occlusion map is a greyscale texture in sRGB color space, with white indicating areas that should receive full indirect lighting and black indicating no indirect lighting. If an alpha channel is present, it is ignored.
```
lexaknyazev: > greyscale texture in sRGB color space

Only some GPU (e.g., PowerVR 6 or Adreno 400) support hardware sRGB sampling of one-channel textures. 
While client could unpack grey-scale PNG to three-component texels, it's not very GPU memory efficient.

To get accurate values, client without such h/w support would need to sample, filter and un-sRGB texels manually in shaders (e.g., via `texelFetch` in GLSL ES 3.0). Other option is to use linear hardware filtering on sRGB data which could produce slightly incorrect result.

> If an alpha channel is present, it is ignored.

In general case, texture could have from one to four components:
- `LUMINANCE`: `(L, L, L, 1)`
- `LUMINANCE_ALPHA`: `(L, L, L, A)`
- `RED`: `(R, 0, 0, 1)`
- `RG`: `(R, G, 0, 1)`
- `RGB`: `(R, G, B, 1)`
- `RGBA`: `(R, G, B, A)`

What channel should shaders use (assume RED since it's always populated)?

bghgary: Is this all because it's a grayscale texture and wanting to store the texture efficiently in memory? I don't know the details of the support for this on different platforms. From a data perspective, occlusion should be sRGB to avoid banding (just like color textures).

specification/2.0/README.md
bghgary: Yes, you are right. Practically speaking it won't make much of a difference since the normal vector needs to be normalized anyways. That said, I'll make the change.

specification/2.0/README.md
lexaknyazev: The main question here:

How much of a problem it is to perform gamma-correction before filtering on this texture?
In other words, must clients sacrifice memory (or perform manual filtering) to get accurate values, or they can tolerate error from h/w filtering in non-linear space?

bghgary: I don't know the answer to this.

For reference, BabylonJS seems to load the texture into an RGB texture and multiplies all the channels against the diffuse color (in linear space!), so I may have a bug here.

cedricpinson:
So when we talk about bump map, it is greyscale texture used instead of normal map. So If the name is confusing, maybe we should use a better one. I am completely open to that.

sbtron:
Need to get this PR merged so that majority of the other changes are part of the main 2.0 branch but lets continue the discussion on bump maps...

sbtron:
Pinging this again to continue discussion on "bump maps"

@cedricpinson mentioned: 
> it is greyscale texture used instead of normal map. So If the name is confusing, maybe we should use a better one. I am completely open to that. 

Is this what @jeffdr referred to [here](https://github.com/KhronosGroup/glTF/pull/830#issuecomment-281915909) as 'Height Map' ?
Could you elaborate more on what its used for and how this is being rendered?

(Sorry continuing discussion on a merged PR but lots of context on the discussion so didn't want to open up a new issue)

moneimne:
Height map on left vs. normal map on right for a brick texture. Taken from [Unity documentation](https://docs.unity3d.com/Manual/StandardShaderMaterialParameterNormalMap.html).


As far as my understanding goes, there are two methods to manipulate normals to fake surface details on a surface without actually changing the geometry. The first is height mapping, which is a grayscale texture that contains height displacements of a surface where white typicalls means a raised surface and black means a depressed surface. With height maps, you can get the partial derivative between adjacent samples in the texture to determine the change in height, affecting the normals accordingly. The second is normal mapping, which is a texture that contains displacements of normal data using the RGB channels as the XYZ components of the normal. With normal maps, you need to transform the normals within the texture from tangent space to world space using.

They both generally do the same thing and can achieve the same effects, fake surface detail, but people tend to prefer normal maps because they can be painted more easily and are a little more intuitive to implement and understand.

xelatihy:
Folks,

there seems to be a lot of confusion on bump/normal/displacement mapping. Let me chime in on just what they are.

1. Normal maps: the colored bluish image above. Stores normals directly, specified in the tangent frame. You cannot have a normal map if tangent space is not present (or you need to define how to compute one) . Normal maps are not directly edited. They are either generated via polygon simplification or from bump map.
2. Bump map: gray scale image above. Used to specify the normals on the surface by specify the movement along the tangent normal space normal of a fake surface. Formally, normal maps are the derivatives with respect to u and v of the bump map. You need to define tangent space to apply bump mapping. Bump maps are sometimes edited manually or procedurally.
3. Displacement mapping is controlled by the *same* image as bump map but is rendered differenty. In offline renderers you literally deform the surface by moving the vertices along the normals by the amount specified in the image. In real-time renderers you sometimes can fake this a bit on the interior of the triangles (silhouttes still bad).

If you really care, take a look at the slides 
http://pellacini.di.uniroma1.it/teaching/graphics16/lectures/14_textures.pdf
starting at slide 35.

Hope this helps the discussion on the format.

emackey:
@xelatihy That's a good description except "Bump map" should be replaced by "Height map" in both `2` and `3`.  As [has been explained](https://github.com/KhronosGroup/glTF/pull/830#issuecomment-281915909), the term "Bump" is not exclusive to greyscale height maps, and is sometimes used on normal maps, leading to confusion.  For example, [Unity's documentation](https://docs.unity3d.com/Manual/StandardShaderMaterialParameterNormalMap.html) states:

> Normal maps are a type of Bump map.

This is just below the first image.  Think of "Bump mapping" as a general concept, rather than any specific format of texture.  Normal maps and height maps are both very specific forms of textures that supply needed data for the bump mapping effect.

@sbtron To get back to your question, I think Normal maps are the specific type of bump mapping we expect to find in glTF 2.0 files.  Using the supplied vectors from a Normal map is likely less burden on the renderer than attempting to calculate screen-space derivatives from a greyscale height map.

Still, height maps can do things unavailable to normal maps, like parallax rendering, which I personally think is unbelievably cool, but I don't think that should be enforced that as a part of core glTF 2.0.  Perhaps a height map extension might show up at some point.

sbtron:
Thanks everyone for the explanations. To summarize my understanding the request is for "Height Maps" which are used for bump mapping effect just like normal maps. While height maps can be additional things like parallax rendering there is a additional complexity associated with implementing them which we shouldn't require in the core spec. 
Seems like a good candidate to create optional extensions which some runtimes could opt in to implement. 
@cedricpinson @AurL This might be one of those things which you do need to use an extension for. Does that work?

AurL:
@sbtron I think that defining an extension for a single channel would be overkill. Enclosing `height map` channel into an `extra{ }` seems to be a better solution but I'm not really clear about the purpose of `extra` field. 
Would it make sense ?

emackey:
The `extra` field is for application-specific data.  Certainly a particular app could jam an extra channel or two in extra, but a height map is something that we might want to share between apps.  So, a dedicated extension for the one channel may be warranted.

But the main question is, do greyscale height maps belong in core 2.0?  I don't think they do.

AurL:
Ok thanks, so it doesn't really make sense to use `extra` field. 
I don't think grayscale height maps belong in core 2.0 either, so we can't expect to have an optional channel for them without using an extension?

If using an extension is the most convenient way to handle them, we will go for it.
Thanks

emackey:
Ah I hadn't considered putting an optional height channel in core 2.0.  I wonder what @lexaknyazev thinks of that?  I'd be in favor except I wouldn't want the resulting pool of glTFs to be split between normal maps and height maps.  Actually, come to think of it, the normal maps are themselves optional, since not all models use them.  I wouldn't want authors to get in the habit of supplying a height map without a normal map and expecting their glTF to work everywhere.  It would be much better for the authoring toolchain to convert everything to normal maps on the way out to glTF.  Maybe it's still best to leave height maps out of core 2.0.

jeffdr:
I would strongly urge omission of height maps for surface orientation purposes in glTF. Leaving that in would I think be very incongruent with the other decisions this group has already made, namely supporting a single format for most types of data. Content authors should convert any surface detail to normal maps. And people rendering glTF content then only need to worry about normal maps and not height content.

If height maps were to be used for displacement, parallax, etc. then that seems like good extension territory to me.

lexaknyazev:
> I wouldn't want authors to get in the habit of supplying a height map without a normal map and expecting their glTF to work everywhere. It would be much better for the authoring toolchain to convert everything to normal maps on the way out to glTF.

+1.

> If height maps were to be used for displacement, parallax, etc. then that seems like good extension territory to me.

+1.

Issue #820, opened 2017-01-21
Sparse Array storage.


emilian0:
**Rationale** 
Sparse arrays are sometimes more efficient than dense arrays when describing incremental changes. 

This scenario is common when encoding morph targets ( #210 ) that often display localized deformations (for instance, for a facial morph, a "blink-left" morph target will only displace vertices around the left eye). It is inefficient to re-transmit an entire full set of vertices positions when only a few of them were displaced. It is, in general, more efficient to just encode the indices and values of those vertices that were displaced in the morph target.

Following the COLLADA extension presented [here](https://www.khronos.org/collada/wiki/Sparse_Array_KHR_extension), I suggest to include sparse array storage in the core glTF specs. 

**Possible implementation**
glTF uses buffers to store data and accessors to retrieve it. I believe accessors are the structures to extend to accommodate for sparse storage. Here below a draft of what a sparse accessor might look like:

A sparse accessor has 3 components:
- an initialization array (used to initialize the accessor). Note: the bufferView entry is optional, in its absence the sparse accessor is initialized with zeroes.
- an index array used to point to those values that deviate from the initialization value.
- a value array encoding the values of the entries corresponding to the indices in the index array.

 
```
"mySparseAccessor": {
//---Initialization array, initializes the accessor with a syntax similar to a standard accessor---//
  "bufferView":"bufferViewInitializationList",
  "byteOffset":,
  "byteStride":,
  "count":,
 
  //--sparse encoding of the values that changed with respect to the initialization array--//
  "sparseCount": ,//<=number of "displaced" entries stored in the sparse array

  "sparseValueBufferView": "bufferViewDisplaceValuesArray",//<=only displaced entries are stored
  "sparseValueByteOffset": ,
  "sparseValueByteStride": ,

  "sparseIndexBufferView": "bufferViewDisplacedIndicesArray",//<=buffer containing the indices of the displaced entries.
  "sparseIndexByteOffset": ,
  "sparseIndexByteStride": ,
 
  //---type and component type have the same semantic as in a standard accessor--//
  "componentType": ,
  "type": ""
  }
```

lexaknyazev:
Please, confirm or correct me if I'm wrong.

Unpacking such accessor into regular one would be like (roughly):
1. Load init data (or zeros) into temporary buffer.
2. `for (var i = 0; i < sparseCount; i++) myData[sparseIndex[i]] = sparseValue[i];`

[...]

Issue #827, opened 2017-02-03
Move byteStride from accessor to bufferView?


kainino0x:
According to #817 glTF 2.0 will target other graphics APIs. Apologies if this issue has been discussed, I couldn't find it.

In the newer graphics APIs (Metal, Vulkan, D3D12, D3D11, GL 4.3+, GL ES 3.1+), I think we need stride to be constant across multiple `accessor`s pointing at the same `bufferView`. This is because stride is a per-input (per-buffer-binding) property while offsets are a per-attribute property.

The current structure of glTF is difficult (and slow) to shoehorn into these new designs. Is there anything preventing byteStride from being moved to bufferView?

Here's how I think the glTF concepts map to API concepts:

Vulkan: [stride](https://www.khronos.org/registry/vulkan/specs/1.0/man/html/VkVertexInputBindingDescription.html); [accessor.offset](https://www.khronos.org/registry/vulkan/specs/1.0/man/html/VkVertexInputAttributeDescription.html)

Metal: [stride](https://developer.apple.com/reference/metal/mtlvertexbufferlayoutdescriptor); [accessor.offset](https://developer.apple.com/reference/metal/mtlvertexattributedescriptor)

D3D12: [stride](https://msdn.microsoft.com/en-us/library/windows/desktop/dn986883(v=vs.85).aspx); [accessor.offset](https://msdn.microsoft.com/en-us/library/windows/desktop/dn770377(v=vs.85).aspx)

D3D11: [offset+stride](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476456(v=vs.85).aspx); [accessor.offset](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476180(v=vs.85).aspx)

OpenGL ES 3.1+, OpenGL 4.3+: [offset+stride](https://www.khronos.org/opengles/sdk/docs/man31/html/glBindVertexBuffer.xhtml); [accessor.offset](https://www.khronos.org/opengles/sdk/docs/man31/html/glVertexAttribFormat.xhtml)

lexaknyazev:
Related: #819

pjcozzi:
@sbtron do you have any input here?

@kainino0x we can think through moving `byteStride` to `bufferView`, but it would need to happen quickly so we can get it into glTF 2.0 so we minimize future breaking changes.

[...]

PR #826, opened 2017-02-02
glTF 2.0 draft

lexaknyazev:
Previous discussion was in #784.

* Overview of spec changes: #605
* Direct link to 2.0 spec: https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0
* Direct link to `KHR_technique_webgl` extension stub: https://github.com/KhronosGroup/glTF/tree/2.0/extensions/Khronos/KHR_technique_webgl
* Direct link to branch: https://github.com/lexaknyazev/glTF/tree/2.0

---
**Story so far:**
- rewritten git history of `1.0.1` and `lexaknyazev/master-test` branches with changed paths to `specification/2.0` directory;
- objects-to-arrays syntax change update;
- techniques, shaders, programs were moved to `KHR_technique_webgl` extension;
- updates on JSON encoding and binary data layout
---

**Specification TODOs**
General 2.0 stuff:
- [x] Update wetzel to v4 schema and regenerate property reference, @lasalvavida
- [ ] Update Overview / Reference Card #761, @javagl 
- [x] Make sure everyone is properly acknowledged in the spec, @pjcozzi
- [x] Add notes about scenes #815, @lexaknyazev 
- [x] Publish reasons on syntax and encoding changes #831, @lexaknyazev 
- [x] Elaborate on asset reqs / platform caps #829
- [x] Update date in spec for "Last Updated"

Binary GLB format integration:
- [x] Integrate `.glb` file format as a specification appendix, @lexaknyazev
- [x] Elaborate on referencing `.glb` buffer #828, @lexaknyazev

PBR:
- [x] Integrate PBR, #830, @sbtron, @bghgary
- [ ] Elaborate on PBR conformance #785
- [x] Evaluate multi-UV #742
- [x] Tangent-space storage #812

Skinning & Morphs:
- [x] Skinning update https://github.com/KhronosGroup/glTF/issues/723#issuecomment-254674004
- [x] Sparse Arrays #820, @emilian0
- [x] Integrate morph targets, #164, #210, @emilian0

Other:
- [ ] Close issues labeled [2.0](https://github.com/KhronosGroup/glTF/labels/2.0), all of which should also be labeled `resolved`, @lexaknyazev / @pjcozzi

glTF 2.0 project: #867

PR #833, opened 2017-02-06
Issue 820

emilian0:
Work in progress branch (Issue #820 ) to be merged to glTF 2.0 specs branch.
TODO:
- [x] Verify with the team that UNSIGNED_INT should be supported as indices type. 
- [x] Decide, together with the team, if the index type should be defined in "sparse.indices.componentType" or  "sparse.indices.type"

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -500,6 +500,7 @@ The following fragment shows an example of `sparse` accessor with 10 attributes 
 }
 ```
 A sparse accessor differs from a regular one in that `bufferView` and `byteOffset` properties aren't required. When these properties are omitted, the sparse accessor is initialized as an array of zeros of size `(size of the accessor attribute type) * (accessor count)` bytes.
+A sparse accessor `min` and `max` properties correspond, respectively, to the minimum and maximum component values once the sparse substitution is applied.
 
 #### BufferView and Accessor Byte Alignment
 
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -488,13 +488,13 @@ The following fragment shows an example of `sparse` accessor with 10 attributes 
   "sparse": {
     "count": 10,
     "indices": {
-      "bufferView": 0,
-      "byteOffset": 75876,
-      "type": 5123,
+      "bufferView": 1,
+      "byteOffset": 0,
+      "componentType": 5123,
     },
     "values": {
-      "bufferView": 0,
-      "byteOffset": 75816
+      "bufferView": 2,
+      "byteOffset": 0
     }
   }
 }
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -465,6 +465,45 @@ For example:
 
 In this accessor, the `componentType` is `5126` (FLOAT), so each component is four bytes.  The `type` is `"VEC3"`, so there are three components.  The size of the attribute type is 12 bytes (`4 * 3`).
 
+#### Sparse accessors
+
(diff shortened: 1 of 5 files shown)


pjcozzi:
@emilian0 since this is WIP, can you add a tasklist (like the TODOs in #826) with any remaining items?

@tparisi @lexaknyazev do you guys want to review this?

lexaknyazev:
Tightening up:
- With #827, we've added new restriction on `accessor`/`bufferView` relations (all `accessor`s using the same `bufferView` must have the same `byteStride`). This should be reflected in `indices` and `values` sub-objects (i.e., remove `byteStride` from there). Such change ensures proper alignment when indices and values use the same `bufferView`. 
- `bufferView`s used for sparse data must have `bufferView.target` undefined.
- For more predictable unpacking process, sparse indices buffer must be sorted and it can't contain duplicate elements.

lexaknyazev:
When `accessor.type` is not `"SCALAR"`, e.g., `"VEC3"`, how are indices stored: per component (e.g., 3 indices for each vec3) or per "value" (e.g., 1 index for each vec3)? The latter seems more effective.

emilian0:
@lexaknyazev Indices are intended per value right now (one per VEC3 entry). What use case do you have in mind when you say it will make more sense with the index being per component? (skinning?)
Having it per component will be more granular:
- we will have to store ~*3 # indices. 
- we will have more control over the entries that are zeroes. 
I see value in both, shall we have both versions, what do you think?

lexaknyazev:
I meant per value.

emilian0:
So @lexaknyazev it sounds like we both want indices to be per value. Thank you for clarifying that. I will re-iterate that in the specs. Will make the changes and ping you.

emilian0:
@lexaknyazev I believe I addressed your concerns.
- I removed byte Stride from the accessor, sparse and indices arrays.
- the new language specifies that `indices` is a strictly increasing array of integer pointing to attributes (rather than components):
`indices`: strictly increasing array of integers of size `count` and specific `componentType` that stores the indices of those attributes that deviate from the initialization value.
I haven't address the note on "bufferView.target". I suggest to do this in a second pass, after the merge to 2.0 (when we will have a first stab at both sparse array and new bufferViews specs).
Unless you have further comments this is ready to merge for me. 
Thanks!

lexaknyazev:
Looks good. Couple more tweaks:
- when accessor is sparse, `accessor.min` and `accessor.max` must reflect decoded data, not initial;
- for example to be correct,
  - it must contain `min` and `max`;
  - `sparse` object shouldn't use the same `bufferView` as base accessor (it can for animation data, but not for vertex arrays because of `bufferView.target`);
- example uses `indices.type`, while in all other places it's `indices.componentType`.

emilian0:
@lexaknyazev accessor range should be a union between the range in the initialization array and the range in the final (sparse) array. Does this sound reasonable to you? Rationale: a glTF implementation not supporting sparse arrays will fall back to the standard accessor implementation, we want the range (min/max) to work for both. Sounds good? (the other two were typos, thank you for catching them!)

lexaknyazev:
Not sure about union of ranges. Why reduce bounding box efficiency for fully conformant implementation?

emilian0:
To avoid things not rendering at all on non-conformant implementations! It is a trade off between being efficient in the fully conformant implementations vs being conservative enough so that things work in the old version. I am fine either way, it sounds like you value efficiency in conformat over fall-back on non conformant, I am happy to make the change.

Issue #835, opened 2017-02-11
glTF 2.0 Texture Formats


lexaknyazev:
glTF 2.0 being runtime API-agnostic format must have a robust and extendable texture handling framework.

Right now, glTF has a very simple `image` object for referencing images of Web-browser compatible codecs (such as JPG or PNG) and a `texture` object for specifying desired GPU representation. glTF 1.0 doesn't provide any standard way to use native GPU formats (both compressed and uncompressed). Also it lacks support of more advanced texture usage patterns, such as cubemaps or MIP-levels.

There was an attempt (#739) to refactor `image` and `texture` object in a way that would enable aforementioned use cases, however that proposal came way before glTF 2.0 API-neutrality strategy change.

Remember, we don't want to introduce any breaking changes after glTF 2.0 release, so such core functionality must be as future-proof as possible. While we don't need to enable every possible format and feature now, we must be sure that future changes could be done in a non-destructive way.

---

# State of cross-API formats support

## GPU formats

Here's an overview of GPU texture formats that could be supported across different APIs (OpenGL ES 2.0/3.0, OpenGL 4.5, Vulkan, Metal, D3D11/12). Keep in mind, that in some cases API support doesn't guarantee hardware support. OpenGL ES 2.0 supports very few of them, so it's present there only for reference.

### Uncompressed (byte-aligned) formats

The following table contains a matched list of byte-aligned GPU formats. Formats, exclusive to only one API are not mentioned.

- __Green:__ All modern APIs support.
- __Yellow:__ Most modern APIs support natively (OpenGL ES 3.0 extensions could be required for some).
- __Red:__ Support is non-uniform.

![image](https://cloud.githubusercontent.com/assets/3479527/22855489/b20372f0-f09a-11e6-9e14-058413e397c9.png)

### Packed formats

The following table contains a matched list of packed GPU formats.

- __Green:__ All modern APIs support (there's exception with D3D for formats available in WebGL 1.0).
- __Yellow:__ Most modern APIs support (no OpenGL ES 3.0).
- __Red:__ Native support is non-uniform.

![image](https://cloud.githubusercontent.com/assets/3479527/22855520/e869c47a-f09a-11e6-8722-e13a86d45d90.png)

### Compressed formats

The following table contains a matched list of compressed GPU formats. Actual support varies both by API and hardware. Note, that ETC1 format isn't mentioned there, however ETC2-enabled systems support it.

- __Yellow:__ BC1 RGB
[...]

PR #842, opened 2017-02-15
KHR_materials_pbrSpecularGlossiness extension 

sbtron:
Updated PBR spec-gloss extension written against glTF 2.0

spec link - https://github.com/sbtron/glTF/tree/KHRpbrSpecGloss/extensions/Khronos/KHR_materials_pbrSpecularGlossiness


extensions/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The specular property from specular-glossiness material model is the same as the base color value from the metallic-roughness material model for metals. The glossiness property from specular-glossiness material model is related with the roughness property from the metallic-roughness material model and is defined as `glossiness = 1 - roughness`. See [appendix](#appendix) for more details on how you can convert between these two material models.
+
+The value for each property (`diffuse`, `specular`, `glossiness`) can be defined using factors or textures. The `specular` and `glossiness` properties are packed together in a single texture called `specularGlossinessTexture`.  If a texture is not given, all respective texture components within this material model are assumed to have a value of `1.0`. The factors (`diffuseFactor`, `specularFactor`, `glossinessFactor`) scale, in linear space, the components given in the respective textures (`diffuseTexture`, `specularGlossinessTexture`). Texture content must be converted to linear space before it is used for any lighting computations.
```
lexaknyazev: > Texture content must be converted to linear space

Does it imply that all textures should be stored in sRGB? If so, we must mention that.
Should it be possible to directly store linear data (e.g., via floats in future) in textures? Also, see #835.

extensions/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+### specularGlossinessTexture
+
+The specular-glossiness texture is a RGBA texture, containing the specular color (RGB) in sRGB space and the glossiness value (A) in linear space.
+
+* **Type**: [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) 
+* **Required**: No
+
+
+#### Additional maps
```
lexaknyazev: Can we be consistent on maps/textures term?

extensions/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+A glTF asset may also choose to use the pbrSpecularGlossiness extension without including any metallic-roughness properties for its material definition. If such an asset chooses to include the extension using the `extensionUsed` property then a runtime that doesn't support the pbrSpecularGlossiness extension may choose to render the asset by converting the specular-glossiness values to metallic-roughness using the conversion mechanism described in the [appendix](#appendix). Such a conversion during load can be slow and potentially lossy so the runtime could also choose to ignore the material parameters and render the asset using 50% gray emissive color as described in the [base specification](/specification/2.0/README.md#appendix-a).
+
+If an asset includes the pbrSpecularGlossiness extension using the `extensionsRequired` property and does not specify any metallic-roughness properties then only a runtime that supports the pbrSpecularGlossines extension can render it. With this approach you trade off portability of the asset to ensure quality.  
+
+The following table describes the expected rendering behavior based on the material definitions included in the asset:
+
```
lexaknyazev: The last column implies that MR model isn't supported, right? Doesn't that contradict base 2.0 spec?

extensions/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
bghgary: I don't think it's required that all textures are stored in sRGB. The two textures in this spec will be interpreted as sRGB (as indicated at the bottom), but other textures in extensions or future additions may or may not be depending on the situation. The point of this statement is to ensure that all the math must be done in linear space regardless of how it is stored.

extensions/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
bghgary: My opinion is that texture is a type where map is a concept (i.e. maps are stored as textures). That said, maybe we can call this "common properties" to avoid the problem all together.

extensions/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
bghgary: To me, this is a general issue with extensionRequired. If we say an extension is required, then it basically means the asset isn't following the core spec anymore. This can happen with any extension, right? What happens if someone puts KHR_technique_webgl as required?

lexaknyazev: > the asset isn't following the core spec anymore

Right. But a runtime should support core spec anyway. So in what case does last column apply? Are we basically allowing to implement only SpecGloss and always convert MR?

lexaknyazev:
Is it possible to provide a sound example (with screenshots) of SpecGloss superiority over MR?
It would help implementors to understand implications.

PR #843, opened 2017-02-15
Add license field

### LINKED ISSUE #839: Add optional asset.license field
The field `asset.copyright` defines the copyright owner of the content, but it does not say anything about licenses. Some content is licensed under CC or even [GPL](https://github.com/kalmykov/fr24-3d-models). If there's no clear indication what the license of the content is, it must be considered to be proprietary. A `license` field would avoid any possible confusion.

Would it be an idea to specify a list of popular licenses that people can choose from when exporting? The spec should make it clear that the version of the license should be included. For example:

- `CC-BY-SA-4.0` equals Creative Commons Attribution-ShareAlike 4.0 International License
- `GPLv2+`  equals GNU General Public License version 2 or any later version
- `GPLv3` equals GNU General Public License version 3

The list should be extended with the other Creative Commons licenses and probably a bunch of other licenses as well. Note that in the case of the GPL, the notice (in source files) sometimes contains "or (at your option) any later version", but sometimes it does not. I've used the `+` here to make that distinction.

Having a predefined list would make it easy for 3D editing programs to allow the user to choose one of those when exporting the model to glTF.

The field should be optional (in which case the model is proprietary). Perhaps it is a good idea to allow the field to contain a custom / arbitrary license as well (not just one from the predefined list above).

emackey:
Fixes #839.


sbtron:
@pjcozzi if its ok with you we can take this in. Looks like a small enough change.

pjcozzi:
I still think this is being done in haste without the proper discussions that would go into licensing, which we don't have time to scope for 2.0.

* Is Khronos on board with this?  I would not add this without asking.
* Are textures ever licensed separately?
* Do renderers have implicit requirements for some licenses such as displaying attribution?
* Has anyone evaluated all the common licenses used for 3D models?

I know some of this was discussed in #839, but I don't think the depth is there.  If someone else is able to fully explore this topic, please go for it.

emackey:
Closing this, since it targeted the `2.0` branch and didn't make 2.0.  Discussion can continue in #839.

Issue #210, opened 2013-12-20
Basic Morph Syntax


tparisi:
Here is an example of the basic morph syntax I am proposing.

First, the animation, consisting of a channel to drive the morph from TIME input and MORPH output parameters.

```
"animation_1": {
        "channels": [
            {
                "sampler": "animation_1_morph_sampler",
                “target”: “BallUV3-morph”
            },

        "parameters": {
            "TIME": "accessor_0", // float buffer w times e.g. 0 0.466667 0.966667
            “MORPH”: “accessor_1” // float buffer w output weight values e.g. 0 1 0
        },

        "samplers": {
            "animation_1_morph_sampler": {
                "input": "TIME",
                "interpolation": "LINEAR",
                "output": "MORPH"
            },
        }
```

Now, the morph controller. NORMALIZED and RELATIVE methods are taken directly from the COLLADA spec. Do we need these? Default is NORMALIZED. In this example the single morph target has zero weight i.e. at resting position the geometry is un-morphed.

"morphs": {
        "BallUV3-morph": {
        “method” : “NORMALIZED”, // one of “NORMALIZED” or “RELATIVE”
        “source”: "BallUV3_geometry",
                "targets": [
                    "target-BallUV3Custom_Morph"
                ],
         “weights”: [
                    0
         ]
        }
    },

Finally, the instance of the morph (will appear inside a node):

```
        "instanceMorph": {
                       "morph": "BallUV3-morph",
},
```

tparisi:
@pjcozzi I assume morphs and non-linear interpolation are still after v1.0? the schema implies that; so, wondering why you removed the "post 1.0" tag....

pjcozzi:
Yes, still post 1.0.  There is no post 1.0 tag anymore.  Everything not 1.0 is post 1.0.  We'll prioritize after we get the spec out.

tparisi:
great. thanks

On Tue, Sep 22, 2015 at 11:21 AM, Patrick Cozzi <notifications@github.com>
wrote:

> Yes, still post 1.0. There is no post 1.0 tag anymore. Everything not 1.0
> is post 1.0. We'll prioritize after we get the spec out.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/KhronosGroup/glTF/issues/210#issuecomment-142371051>.
>



-- 


Tony Parisi                             tparisi@gmail.com
Follow me on Twitter!             http://twitter.com/auradeluxe
[...]

PR #852, opened 2017-02-24
Emiliano/210

emilian0:
Implementing Morph Targets as discussed in #210. 
@lexaknyazev all yours!


specification/2.0/README.md
```diff
@@ -594,6 +594,41 @@ Valid attribute semantic property names include `POSITION`, `NORMAL`, `TEXCOORD`
 
 > **Implementation note:** Each primitive corresponds to one WebGL draw call (engines are, of course, free to batch draw calls). When a primitive's `indices` property is defined, it references the accessor to use for index data, and GL's `drawElements` function should be used. When the `indices` property is not defined, GL's `drawArrays` function should be used with a count equal to the count property of any of the accessors referenced by the `attributes` property (they are all equal for a given primitive).
 
+#### Morph Targets
+
+Morph Targets are defined by extending the Mesh concept.
+A Morph Target is a morphable Mesh where primitives' attributes are obtained by adding the original attributes to a weighted sum of targets attributes (this operation corresponds to COLLADA's `RELATIVE` blending method).
```
lexaknyazev: It would be better to show equation than use external COLLADA reference.

specification/2.0/README.md
```diff
@@ -594,6 +594,41 @@ Valid attribute semantic property names include `POSITION`, `NORMAL`, `TEXCOORD`
...
+Morph Targets are implemented via the `targets` property defined in the Mesh `primitives`. Each target in the `targets` array is a dictionary mapping a primitive attribute to a Morph Target displacement, currently only two attributes ('POSITION' and 'NORMAL') are supported. All primitives are required to list the same number of `targets` in the same order.
```
lexaknyazev: > Each target ... mapping a primitive attribute to a Morph Target displacement

... mapping a primitive attribute to an accessor containing Morph Target displacement data

specification/2.0/README.md
```diff
@@ -594,6 +594,41 @@ Valid attribute semantic property names include `POSITION`, `NORMAL`, `TEXCOORD`
...
+                    "POSITION": 33,
+                },
+                {
+                    "NORMAL": 45,
+                    "POSITION": 43,
+                },
+            ]
+        }
+    ],
+    "weights": [0, 0.5]
```
lexaknyazev: It's still not quite obvious to me, why do we have two ways of defining initial morph state: `mesh.weights` and `node.weights`. Just asking.

specification/2.0/schema/mesh.primitive.schema.json
```diff
@@ -31,6 +31,14 @@
             "default" : 4,
             "gltf_detailedDescription" : "The type of primitives to render. All valid values correspond to WebGL enums."
         },
+        "targets" : {
+            "type" : "array",
+            "description" : "An array of Morph Targets, each  Morph Target is a dictionary mapping attributes (only "POSITION" and "NORMAL" supported) to their deviations in the Morph Target.",
```
lexaknyazev: Double-quotes must be escaped here.

emilian0: fixed, thanks

specification/2.0/README.md
emilian0: `mesh.weights` are the Morph Targets default weights. 
`node.weights` are the instantiated Morph Targets weights (that override the weights of the mesh). 
Say that you have a Morph Target, it is a character face. You want to instantiate it multiple times in your scene, and you want to instantiate it as a smiling face. Then just set the default `mesh.weights` accordingly (make it smile!) and don't set the instantiation weights `node.weights` when you instantiate it within a node. 
Makes sense?

specification/2.0/README.md
emilian0: fixed, thanks

specification/2.0/README.md
emilian0: addressed. Thanks

emilian0:
@lexaknyazev I addressed your concerns, back to you. Thanks

Issue #828, opened 2017-02-03
Referencing glb-stored buffer


lexaknyazev:
With deprecating `KHR_binary_glTF` and removing string-based IDs from the core spec, we need to find a proper way to refer to buffer, stored in `.glb` file after JSON.

`.glb` layout:
![image](https://cloud.githubusercontent.com/assets/3479527/22600725/36b87122-ea55-11e6-9d40-6fd42819fcab.png)

glTF 1.0 way to reference "binary data":
```json
{
  "buffers": {
    "binary_glTF": {
      "uri": "data:,",
      "byteLength": 123456
    }
  },
  "bufferViews": {
    "bufferView_0": {
      "buffer": "binary_glTF",
      "byteLength": 123
    }
  }
}
```

With glTF 2.0 we can't use magic `"binary_glTF"` ID anymore. Also we can't assume that it's the only buffer used.

Also, previous usage of DataURI isn't quite correct: `"data:,"` means "empty text/plain string" according to RFC.

As one option, we could define a custom URI scheme like this (just possible example):
```json
{
  "buffers": [
    {
      "uri": "glb://#buffer",
      "byteLength": 123456
    }
  ],
  "bufferViews": [
    {
      "buffer": 0,
      "byteLength": 123
    }
  ]
}
```

As another option, we could refine schema of `buffer` object.

Make URI optional:
```json
{
  "buffers": [
    {
      "byteLength": 123456
    }
  ]
}
```

Or make URI nullable:
```json
{
  "buffers": [
    {
      "uri": null,
      "byteLength": 123456
    }
  ]
}
```

Or add new boolean field:
```json
{
  "buffers": [
    {
      "useGlb": true,
      "byteLength": 123456
    }
  ]
}
```

---

@pjcozzi @javagl 
What do you think?

bghgary:
> Also we can't assume that it's the only buffer used.

I would prefer if a glb file is completely self-contained and cannot reference other URIs. This would make it much simpler for native clients to handle files on disk (just like a JPEG file) since they will not have to deal with the network stack or other file references. There are security implications with this as well since sandboxed applications will only be able to access the single file that is chosen by the user and not all the references. Sandboxed applications may not have access to the internet at all, depending on application permissions.

In short, GLB will be the on-disk self-contained file while glTF will be for transmission across the web or other services.

If we agree to that, then you can assume there is only one buffer.

lexaknyazev:
> there is only one buffer

[...]

PR #853, opened 2017-02-25
GLB v.2 Specification

lexaknyazev:
From #828.
Before merging, we must give this appendix a number and put a link to it in the main spec.

CC @pjcozzi @sbtron @bghgary @javagl


pjcozzi:
Looks great, nice job!

For searchability inside the spec, do you think this should be in the main README.md?

lexaknyazev:
> For searchability inside the spec, do you think this should be in the main README.md?

I'd wait until we have all appendices (PBR equations, tangent-space reconstruction for morph targets) and evaluate what's better: one giant file or several interlinked.

pjcozzi:
Sounds good.

bghgary:
Looks good to me.

bghgary:
Is it guaranteed that the JSON chunk is the first chunk in the file?

lexaknyazev:
> Is it guaranteed that the JSON chunk is the first chunk in the file?

Yes. I've added a note, please see if it's sufficient.

PR #856, opened 2017-02-27
Fix minor wording

bghgary:
- Emission to Emissive
- Removed epsilon from BRDF input equations
- Fixed capitalization / formatting


pjcozzi:
Does this need a review for the equation change or was that previously wrong / already decided to change?

bghgary:
Epsilon may or may not be the same value for every implementation. I was going to move it to the appendix.

pjcozzi:
Sounds good

PR #858, opened 2017-02-28
Refine accessor definition

lexaknyazev:
Accessors are used not only for attributes, so language was changed to not use the term "attribute" in accessor definition. This also allows listing UNSIGNED_INT in a table with accessor component sizes.

CC @pjcozzi


pjcozzi:
Somewhere, perhaps this sentence:

> The latter specifies the number of elements within the `bufferView`, *not* the number of bytes.

The spec should give concrete examples of what "elements" could be, e.g., attributes.

lexaknyazev:
@pjcozzi 
I've added couple examples.

pjcozzi:
Looks great!

PR #859, opened 2017-02-28
Generalize textureInfoBase.index and restrict textureInfoBase.texCoord

lexaknyazev:
CC @sbtron


bghgary:
+1, good catch on both changes

PR #860, opened 2017-03-01
Textures update for glTF 2.0

lexaknyazev:
This PR contains updated schemas and description of texture data handling for glTF 2.0.

#### Why move `internalformat` to `image`?
Because this property is deeply bound with supplied data. One could switch different images while keeping the same texture object.

#### Why remove `type` and `format`?
These two properties can be precisely derived from one `internalformat` value. No need to complicate asset validation with checking all possibly wrong combinations.

#### Why use OpenGL ES 3.0 enums instead of ES 2.0?
They provide distinct values for each `format`/`type` combination.

#### Why `source.uris` and `source.bufferViews` are defined as arrays?
To support PNG/JPEG-based cubemaps and texture arrays in future 2.x releases without breaking JSON syntax.

#### Why `texture.sampler` is optional now?
Because all fields of `sampler` object were already optional. So exporters won't need to create an empty sampler object as before:
```json
{
  "textures": [
    {
      "sampler": 0, ...
    }
  ],
  "samplers": [ { } ]
}
```

#### What do "expand" and "repack" mean in the second table?

**Expand** means populating non-provided texture channels by some rule. For example, to expand LUMINANCE into RGBA, client should emit `(L, L, L, 1.0)` for each texel. See more on that topic in OpenGL specs.

**Repack** means changing the order of packed components to adapt to another API's convention.

Note, that ANGLE does such transformations when running on D3D backend.

---

CC @pjcozzi @sbtron

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -57,6 +57,7 @@ Copyright (C) 2013-2017 The Khronos Group Inc. All Rights Reserved. glTF is a tr
   * [Texture Data](#texture-data)
     * [Textures](#textures)
     * [Images](#images)
+      * [Image Formats](#image-formats)
     * [Samplers](#samplers)
   * [Materials](#materials)
     * [Metallic-Roughness Material](#metallic-roughness-material)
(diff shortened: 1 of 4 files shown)


pjcozzi:
For #835

pjcozzi:
@lilleyse @lasalvavida can you both please review this to confirm that there is a reasonable update path from our current implementations (pipeline and runtime) to this?  This is a significant, but hopefully isolated change.

specification/2.0/README.md
```diff
@@ -831,57 +832,108 @@ A skin is instanced within a node using a combination of the node's `mesh` and `
...
+ 
+Each `image` defines an `image.sources` set of image sources. Type and format of image source are defined by `mimeType` and `internalFormat` specifiers.
+
+An image source can refer to image data either by array of URIs (external or base64-encoded), or by array of Buffer Views.
+
+> **Implementation Note** glTF 2.0 requires these arrays to contain no more than one element. This restriction could be lifted in future to support more texture targets.  
+ 
+First image pixel corresponds to the lower left corner of the image.
+
+> **Implementation Note**: With WebGL API, the first pixel transferred from the `TexImageSource` (i.e., HTML Image object) to the WebGL implementation corresponds to the upper left corner of the source. To achieve correct rendering, WebGL runtimes must flip Y axis by enabling `UNPACK_FLIP_Y_WEBGL` flag. This behavior was changed from glTF 1.0.
```
pjcozzi: Once this is merged, please update https://github.com/KhronosGroup/glTF/issues/605 with changes like this, especially this one since it is breaking but the syntax doesn't change.

specification/2.0/README.md
```diff
@@ -831,57 +832,108 @@ A skin is instanced within a node using a combination of the node's `mesh` and `
...
 
-Images referred to by textures are stored in the `images` array of the asset. Each image contains a URI to an external file (or a reference to a bufferView) in one of the supported images formats. Image data may also be stored within the glTF file as base64-encoded data and referenced via data URI. For example:
+Any colorspace information (such as ICC profiles, intents, etc) from PNG or JPEG containers must be ignored.
+
+> **Implementation Note**: This increases portability of an asset, since not all image decoding libraries fully support custom color conversions. To achieve correct rendering, WebGL runtimes must disable such conversions by setting `UNPACK_COLORSPACE_CONVERSION_WEBGL` flag to `NONE`.
+
+In the following example asset provides two versions of the same image: in sRGB and in linear colorspaces; so clients with hardware sRGB filtering support (or with enough fragment shader processing power) benefit from reduced banding, whilst low-power clients still get correct texel values by using linear-space image data directly.
```
pjcozzi: Sounds like a conformant renderer is free to select any object in the `sources` array.  Can the spec be more explicit about this?

specification/2.0/README.md
```diff
@@ -831,57 +832,108 @@ A skin is instanced within a node using a combination of the node's `mesh` and `
...
         }
     ]
 }
 ```
-> **Implementation Note**: With WebGL API, the first pixel transferred from the `TexImageSource` (i.e., HTML Image object) to the WebGL implementation corresponds to the upper left corner of the source. Non-WebGL runtimes may need to flip Y axis to achieve correct texture rendering.
 
-### Samplers
+#### Image Formats
+
+glTF 2.0 allows a limited set of possible `internalformat` values based on OpenGL ES 3.0 enums.
```
pjcozzi: Isn't it possible that a glTF asset will contain a texture format that a WebGL 1.0 client will not be able to render?

specification/2.0/README.md
```diff
@@ -831,57 +832,108 @@ A skin is instanced within a node using a combination of the node's `mesh` and `
...
+| `source.internalformat` | OpenGL ES 2.0 Format & Internal Format | OpenGL ES 3.0 Format | OpenGL ES 3.0 Internal Format | OpenGL ES 2.0/3.0 Type |
+|-------------------------|----------------------------------------|----------------------|-------------------------------|---------------------------|
+| ALPHA (`6406`) | ALPHA | ALPHA | ALPHA | UNSIGNED_BYTE |
+| LUMINANCE (`6409`) | LUMINANCE | LUMINANCE | LUMINANCE | UNSIGNED_BYTE |
+| LUMINANCE_ALPHA (`6410`) | LUMINANCE_ALPHA | LUMINANCE_ALPHA | LUMINANCE_ALPHA | UNSIGNED_BYTE |
+| RGB8 (`32849`) | RGB | RGB | RGB8 | UNSIGNED_BYTE |
+| RGBA4 (`32854`) | RGBA | RGBA | RGBA4 | UNSIGNED_SHORT_4_4_4_4 |
+| RGB5_A1 (`32855`) | RGBA | RGBA | RGB5_A1 | UNSIGNED_SHORT_5_5_5_1 |
+| RGBA8 (`32856`) | RGBA | RGBA | RGBA8 | UNSIGNED_BYTE |
+| SRGB8 (`35905`) | SRGB_EXT | RGB | SRGB8 | UNSIGNED_BYTE |
```
pjcozzi: Does this require an extension? https://www.khronos.org/registry/webgl/extensions/EXT_sRGB/

The spec should clarify this.

lexaknyazev: The best WebGL 1.0 runtime strategy is to use that extension, yes. Without it, the only correct option is to disable hardware filtering and perform it manually in GLSL (after manual colorspace conversion).

Supplying alternative image source with linear data could help such incapable clients. Quality will suffer from banding, obviously.

specification/2.0/README.md
lexaknyazev: `sources` are different representations of the _same_ image. So choice is ultimately up to engine (based on `mimeType` and `internalformat`).

pjcozzi: Agreed, let's just be super explicit about it in the spec.

lilleyse:
I don't see any problems here, just making the conversions to the correct `internalFormat`, `format`, `type` whether its a webgl1 or 2 context.

lexaknyazev:
Retargeting to post-2.0.

PR #861, opened 2017-03-05
Update link to A-Frame glTF loader.

donmccurdy:
The `gltf-model` component was merged into the A-Frame library as of 0.5.0, and the external component is no longer maintained.


pjcozzi:
Very nice, thanks @donmccurdy!

PR #862, opened 2017-03-06
Added VKTS in list of glTF C++ and Vulkan programs

McNopper:


pjcozzi:
Fantastic!  Thanks @McNopper!

Issue #822, opened 2017-01-23
[PBR extension] Add rendering mode for transparency


bghgary:
@mlimper 

_Also known as alpha mode or blend mode._

@erich666 says this about transparency from https://github.com/KhronosGroup/glTF/issues/697#issuecomment-252343211
> Transparency is another area that is not addressed in the specification. Again, it might be fine to ignore this whole area for now, at best providing a traditional "blend" transparency factor and stress that the model presented is for basic opaque materials and transparency is just basic alpha blending (no Fresnel involved, for example). Similarly, clear coat materials (car paints), anisotropic materials (brushed metal, carbon fiber, etc.) are not included, which I think is fine for now. A basic, solid PBR material will be reasonable for 95% of materials for 95% of users, and can be expanded in future extension proposals.

Here is @cedricpinson's implementation:
https://github.com/sketchfab/Unity-glTF-Exporter#transparency
> In order to differenciate between transparency types in Unity, an extra metadata is added to the material.
It allows to know which blendMode is used and the cutoff value.
For now, blendMode valid values are alphaMask and alphaBlend.
> 
> ````
> "extras": {
>     "blendMode" : "alphaMask",
>     "cutoff" : 0.5
> },
> ````

I believe it is important to specify how to interpret the alpha channel of the PBR diffuse / baseColor texture. At the minimum, we need to document the expected implementation behavior in the spec.

This proposal is now out-of-date, please see https://github.com/KhronosGroup/glTF/issues/822#issuecomment-281847310 for the new proposal.

> I propose we add the following two material parameters to the spec. Note that the rendering mode represents the usage of transparency instead of how the blending works technically, similar to what Unity3D is doing with their standard shaders.
> 
> **alphaRenderingMode**
> 
> | Value | Description
> |:--|:--|
> Opaque | This mode ignores the alpha channel of the texture and renders as if alpha is 1. Use this mode if the texture does not have an alpha channel or if the alpha channel is for something other than transparency.
> Cutout | Also known as Mask, this mode allows for hard edges between the opaque and transparent areas. The rendered result is either fully opaque or fully transparent depending on the alpha channel of the texture and the specified cutoff value. The implementation determines anti-aliasing support if any. Use this mode for materials that simulate geometry such as leaves or wire fences.
[...]

PR #864, opened 2017-03-10
Add alpha coverage details

sbtron:
Adding details on alpha coverage as per #822 
Also added doubleSided property from common materials extension to core spec since it can be used across multiple material models.

Doc: https://github.com/sbtron/glTF/tree/2.0Transparency/specification/2.0#alpha-coverage
Schemas:
https://github.com/sbtron/glTF/blob/2.0Transparency/specification/2.0/schema/material.schema.json
https://github.com/sbtron/glTF/blob/2.0Transparency/specification/2.0/schema/material.pbrMetallicRoughness.schema.json 

@AurL @cedricpinson could you review and ensure this meets your needs so you no longer have to use the extras field.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -62,6 +62,7 @@ Copyright (C) 2013-2017 The Khronos Group Inc. All Rights Reserved. glTF is a tr
     * [Metallic-Roughness Material](#metallic-roughness-material)
     * [Additional Maps](#additional-maps)
     * [Alpha Coverage](#alpha-coverage)
+    * [Double Sided](#double-sided)
   * [Cameras](#cameras)
     * [Projection Matrices](#projection-matrices)
   * [Animations](#animations)
@@ -1015,12 +1016,14 @@ The `alphaMode` property defines how the alpha value of the main factor and text
 
  When `alphaMode` is set to `MASK` the `alphaCutoff` property specifies the cutoff threshold. If the alpha value is greater than or equal to the `alphaCutoff` value then it is rendered as fully opaque, otherwise, it is rendered as fully transparent. `alphaCutoff` value is ignored for other modes.
 
-The `doubleSided` property specified whether the material is double sided.
-
 >**Implementation Note for Real-Time Rasterizers:** Real-time rasterizers typically use depth buffers and mesh sorting to support alpha modes. The following describe the expected behavior for these types of renderers.
 >* `OPAQUE` - A depth value is written for every pixel and mesh sorting is not required for correct output.
 >* `MASK` - A depth value is not written for a pixel that is discarded after the alpha test. A depth value is written for all other pixels. Mesh sorting is not required for correct output.
 >* `BLEND` - Support for this mode varies. There is no perfect and fast solution that works for all cases. Implementations should try to achieve the correct blending output for as many situations as possible. Whether depth value is written or whether to sort is up to the implementation. For example, implementations can discard pixels which have zero or close to zero alpha value to avoid sorting issues.
+
+### Double Sided
+
+The `doubleSided` property specified whether the material is double sided. When this value is false, back-face culling is enabled. When this value is true, back-face culling is disabled and double sided lighting is enabled. The back-face must have its normals reversed before the lighting equation is evaluated.
 
 ## Cameras
 
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -61,6 +61,7 @@ Copyright (C) 2013-2017 The Khronos Group Inc. All Rights Reserved. glTF is a tr
   * [Materials](#materials)
     * [Metallic-Roughness Material](#metallic-roughness-material)
     * [Additional Maps](#additional-maps)
+    * [Alpha Coverage](#alpha-coverage)
   * [Cameras](#cameras)
     * [Projection Matrices](#projection-matrices)
   * [Animations](#animations)
@@ -1002,6 +1003,24 @@ The following examples shows a material that is defined using `pbrMetallicRoughn
 >| Normal    | Geometry will appear less detailed than authored. |
 >| Occlusion | Model will appear brighter in areas that should be darker. |
 >| Emissive  | Model with lights will not be lit. For example, the headlights of a car model will be off instead of on. |
+
+### Alpha Coverage
+
+The `alphaMode` property defines how the alpha value of the main factor and texture should be interpreted. The alpha value is defined in the `baseColor` for metallic-roughness material model. 
+
+`alphaMode` can be one of the following values:
+* `OPAQUE` - The rendered output is fully opaque and any alpha value is ignored.
+* `MASK` - The rendered output is either fully opaque or fully transparent depending on the alpha value and the specified alpha cutoff value. This mode is used to simulate geometry such as tree leaves or wire fences.
+* `BLEND` - The rendered output is combined with the background using the normal painting operation (i.e. the Porter and Duff over operator). This mode is used to simulate geometry such as guaze cloth or animal fur. 
+
+ When `alphaMode` is set to `MASK` the `alphaCutoff` property specifies the cutoff threshold. If the alpha value is greater than or equal to the `alphaCutoff` value then it is rendered as fully opaque, otherwise, it is rendered as fully transparent. `alphaCutoff` value is ignored for other modes.
+
+The `doubleSided` property specified whether the material is double sided.
+
+>**Implementation Note for Real-Time Rasterizers:** Real-time rasterizers typically use depth buffers and mesh sorting to support alpha modes. The following describe the expected behavior for these types of renderers.
+>* `OPAQUE` - A depth value is written for every pixel and mesh sorting is not required for correct output.
+>* `MASK` - A depth value is not written for a pixel that is discarded after the alpha test. A depth value is written for all other pixels. Mesh sorting is not required for correct output.
+>* `BLEND` - Support for this mode varies. There is no perfect and fast solution that works for all cases. Implementations should try to achieve the correct blending output for as many situations as possible. Whether depth value is written or whether to sort is up to the implementation. For example, implementations can discard pixels which have zero or close to zero alpha value to avoid sorting issues.
 
 ## Cameras
 
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -1023,7 +1023,7 @@ The `alphaMode` property defines how the alpha value of the main factor and text
 
 ### Double Sided
 
-The `doubleSided` property specified whether the material is double sided. When this value is false, back-face culling is enabled. When this value is true, back-face culling is disabled and double sided lighting is enabled. The back-face must have its normals reversed before the lighting equation is evaluated.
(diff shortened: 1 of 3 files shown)


specification/2.0/README.md
```diff
@@ -1003,6 +1004,24 @@ The following examples shows a material that is defined using `pbrMetallicRoughn
...
+The `alphaMode` property defines how the alpha value of the main factor and texture should be interpreted. The alpha value is defined in the `baseColor` for metallic-roughness material model. 
+
+`alphaMode` can be one of the following values:
+* `OPAQUE` - The rendered output is fully opaque and any alpha value is ignored.
+* `MASK` - The rendered output is either fully opaque or fully transparent depending on the alpha value and the specified alpha cutoff value. This mode is used to simulate geometry such as tree leaves or wire fences.
+* `BLEND` - The rendered output is combined with the background using the normal painting operation (i.e. the Porter and Duff over operator). This mode is used to simulate geometry such as guaze cloth or animal fur. 
+
+ When `alphaMode` is set to `MASK` the `alphaCutoff` property specifies the cutoff threshold. If the alpha value is greater than or equal to the `alphaCutoff` value then it is rendered as fully opaque, otherwise, it is rendered as fully transparent. `alphaCutoff` value is ignored for other modes.
+
+The `doubleSided` property specified whether the material is double sided.
```
bghgary: This seems a bit out of place. Double sided has nothing to do with alpha coverage?

specification/2.0/schema/material.schema.json
```diff
@@ -39,7 +39,30 @@
...
+            "type": "number",
+            "minimum": 0.0,
+            "default": 0.5,
+            "description": "The alpha cutoff value of the material.",
+            "gltf_detailedDescription": "This value specifies the cutoff threshold when in `MASK` mode. If the alpha value is greater than or equal to this value then it is rendered as fully opaque, otherwise, it is rendered as fully transparent. This value is ignored for other modes."
+        },
+        "doubleSided": {
+            "type": "boolean",
+            "default": false,
+            "description": "Indicates whether a material is double sided.",
```
bghgary: nit: a material --> the material (to be consistent with the rest of the descriptions)

AurL:
@sbtron It looks good, we don't need the extra field anymore, and the `doubleSided` field will be useful for us as well. We will include these changes in the next update.
Thanks a lot for this.

emackey:
Agreed glTF needs `doubleSided` particularly to enable/disable back-face culling.  But we should be careful not to confuse this with the choice of single-sided lighting vs. double-sided lighting (specifically, should the back side of a material, if not culled, should render with the front side's normal & reflection vectors or the opposite of those.  Typically, thin translucent materials will use the front side's vectors, and thicker opaque materials will flip the vectors).

bghgary:
@emackey We should clarify this in the spec. Should this be an option or should we pick one for the spec?

emackey:
I wonder if people would agree to an enum here?  There could be one enum value to cull the back face, another value to render it with single-sided lighting (using the front face's normal vector) and another option to render it with double-sided lighting (flip the normal vector, after applying the normal map etc).

emackey:
glTF 1.0 had backface culling as an option `CULL_FACE` in an enum in [`technique.states`](https://github.com/KhronosGroup/glTF/blob/master/specification/1.0/schema/technique.states.schema.json).  It didn't need to address the single/double-sided lighting question, as that was an implementation detail of the supplied shader.

bghgary:
I don't have data here. How common is singled-sided lighting vs. double-sided lighting?

emackey:
I don't have data from the outside world.  My company does maintain a collection of aerospace-related models that use both lighting models for different circumstances.  But I can't speak for the graphics industry in general.

Ye olde fixed-function OpenGL did this with a parameter named `GL_LIGHT_MODEL_TWO_SIDE`.

https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glLightModel.xml

https://msdn.microsoft.com/en-us/library/windows/desktop/dd373582(v=vs.85).aspx

emackey:
In WebGL we use `gl_FrontFacing` to detect which side is being rendered in the fragment shader.  Any shader that doesn't pay attention to that effectively uses single-sided lighting when backface culling is turned off.  That does keep the shader simpler if we ignore that, so I wouldn't be strongly opposed to omitting the option for two-sided lighting on double-sided materials.

lexaknyazev:
In #822, we agreed on using alpha for _coverage only_. Single-sided lighting with double-sided rendering refers to light transmission scenario which we wanted to avoid.

emackey:
> Single-sided lighting with double-sided rendering refers to light transmission scenario which we wanted to avoid.

True, but I may have opened a can of worms here.  Take a look at the various sample PBR implementations that have been posted here, do any of them use `gl_FrontFace` in the fragment shader?  I haven't found one that does.  ThreeJS.MeshStandardMaterial doesn't support it.  Two-sided lighting may be a concept that's fallen out of use, and would require new code paths to support.

In the name of simplicity for the runtime engine, then, we could just offer a boolean for backface culling, but not add the complexity of two-sided lighting.

bghgary:
> Single-sided lighting with double-sided rendering refers to light transmission scenario which we wanted to avoid.

What about opaque materials? Should be able to support double-sided and flipped normals on the back?

lexaknyazev:
> What about opaque materials? Should be able to support double-sided and flipped normals on the back?

Yes.

bghgary:
Should have been more clear with my previous comment. Is the following statement correct?

"Normals should always be flipped on the backside for lighting calculations when doubleSided property is true."

This is the exact opposite of what @emackey said with https://github.com/KhronosGroup/glTF/pull/864#issuecomment-286429115?

emackey:
> > What about opaque materials? Should be able to support double-sided and flipped normals on the back?
>
> Yes.

No, please, do some research before you jump on this.  My own examples were from the 1990's fixed-function days, and looking around today I have yet to find any shaders doing this in modern times.  I now think this could become a roadblock to writing conformant glTF renderers.

Remember, the goal of glTF is to deliver something that's easy and practical to render -- *not* to uphold some purist notions about alpha being used only for coverage.  Sometimes we need to be practical.

emackey:
I hope I didn't derail this with my old memories of two-sided lighting.  Having an option to enable/disable backface culling is needed, as is the alpha settings that make up the original point of this PR.  Perhaps the lighting issue could be addressed separately.

bghgary:
I'm working with @BeardedGnome to ensure that it is reasonable to implement double-sided lighting in WebGL, using BabylonJS. If it is reasonable, then I think double-sided materials should imply double-sided lighting (i.e. the normals are flipped on the back side). Thoughts?

emackey:
It's probably not a huge burden, I think you just test `gl_FrontFacing` in the fragment shader and then flip the back-facing normals.  I just think it may be uncommon these days.

There will be 3D artists who insist on using alpha as non-coverage translucency, as they have no other means to achieve that effect, and some option for single-sided lighting on non-culled back faces will be quite important to them.  I'm still under the impression that "modern" rendering engines generally only offer single-sided lighting, so they may expect this quite strongly.

lexaknyazev:
> It's probably not a huge burden, I think you just test `gl_FrontFacing` in the fragment shader and then flip the back-facing normals.  I just think it may be uncommon these days.

@emackey Just to summarize, are you OK with `doubleSided` bool flag?

> There will be 3D artists who insist on using alpha as non-coverage translucency

I think that we should limit the scope of 2.0 material model to opaque (with coverage) materials only. Translucency effects are still important, so future glTF updates should address some of them.

emackey:
> are you OK with `doubleSided` bool flag?

I want the flag, but I think it would be clearer to name it `backfaceCulling` or such, although this would change the true/false sense of it.

I don't like the idea of switching both the lighting model and the culling option with a single boolean value, as they are quite separate concerns to the engine.  In OpenGL/WebGL, the choice of culling is a simple state enable/disable, but the lighting model requires an extra `if` block with some vector math in the fragment shader, plus a uniform if you want control of that.

bghgary:
It looks like Unreal 4 supports two-sided materials with two-sided lighting and they don't have an option to turn it off. Here is what it looks like:

From the top:
image

From the bottom:
image

Here is what it takes to implement two-sided lighting in BabylonJS:
https://github.com/BabylonJS/Babylon.js/pull/1930

Here is a live example of what it looks like in BabylonJS:
https://bghgary.github.io/glTF/two-sided-lighting

Given this, I think it's clear that two-sided lighting works and it should be okay for us to assume it is always on since Unreal does it too.

I'm okay with calling it `backfaceCulling` instead of `doubleSided`.

Thoughts?

emackey:
Sounds good.  Thanks for looking into UE4.

bghgary:
We are planning on leaving the name `doubleSided` since it implies backface culling is off and double-sided lighting is on. This is analogous to UE4 where they have a property called *Two Sided* on their materials.

image

emackey:
If you're doing both from the same flag, then yes, that makes sense.

McNopper:
FYI, some tools (e.g. Blender) store a similar flag per mesh and not material.

emackey:
Blender defaults to double-sided, but has a number of options that apply or don't apply to its various rendering engines.  Blender Original, Blender Game, Cycles renderer, and the OpenGL/GLSL preview render, as well as the "Multitexture" preview render, all can behave differently or support options that others don't support inside of Blender itself.  The OpenGL/GLSL one does make a distinction between double-sided lighting and backface culling, but it may be alone in that distinction.

blenderdoublesided

Anyway, I've agreed to having a single `doubleSided` flag in glTF.  I don't think we should continue to hold up progress on 2.0 for this minor issue.  It can always be tweaked with extensions if it turns into a problem later.

sbtron:
Added explanation of doublesided lighting to readme and schema. @emackey can you review?

emackey:
> Added explanation of doublesided lighting to readme and schema

The new explanation looks correct to me.  Thanks!

McNopper:
Good for me as well.
My comment was more about, that the value is assigned to a mesh and not the material. But as I said, I am fine with this.

emackey:
@McNopper Ah I see, but, I think the flag does "belong" on the material more than the mesh.  Lightwave does this IIRC.  Blender does some very odd things like associating its backface culling toggle with particular viewports instead of either meshes or materials.  I think associating it with the material makes the most sense, particularly for engines that are trying to minimize render state changes.

specification/2.0/README.md
```diff
@@ -1015,13 +1016,15 @@ The `alphaMode` property defines how the alpha value of the main factor and text
...
 >* `OPAQUE` - A depth value is written for every pixel and mesh sorting is not required for correct output.
 >* `MASK` - A depth value is not written for a pixel that is discarded after the alpha test. A depth value is written for all other pixels. Mesh sorting is not required for correct output.
 >* `BLEND` - Support for this mode varies. There is no perfect and fast solution that works for all cases. Implementations should try to achieve the correct blending output for as many situations as possible. Whether depth value is written or whether to sort is up to the implementation. For example, implementations can discard pixels which have zero or close to zero alpha value to avoid sorting issues.
 
+### Double Sided
+
+The `doubleSided` property specified whether the material is double sided. When this value is false, back-face culling is enabled. When this value is true, back-face culling is disabled and double sided lighting is enabled. The back-face must have its normals reversed before the lighting equation is evaluated.
```
bghgary: specified should be specifies

Issue #872, opened 2017-03-13
glTF 2.0 buffer.uri should not be a required property


robertlong:
A [GLB Stored Buffer](https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#glb-stored-buffer) requires the uri property to be undefined. However, the spec has it as a required property.

https://github.com/KhronosGroup/glTF/blob/2.0/specification/2.0/schema/buffer.schema.json#L26

We should remove the requirement.

lexaknyazev:
Closed with #873.

pjcozzi:
Good eye, thanks again @robertlong!

PR #873, opened 2017-03-13
glTF 2.0 buffer.uri should not be a required property

robertlong:
Addresses #872


lexaknyazev:
Thanks, that was overlooked.

Issue #868, opened 2017-03-10
KHR_technique_webgl and binary shaders


lasalvavida:
@lexaknyazev correct me if I'm wrong or there is already something in the work for this.

Currently I don't see a defined way to point shaders in KHR_technique_webgl to a binary bufferView. This was done in 1.0 like this:

```javascript
shaders: {
  shader: {
    extensions: {
      KHR_binary_glTF: {
        bufferView: "some buffer view"
      }
    }
  }
}
```

I would propose just allowing shaders to point into the bufferView for 2.0, which seems in line with what was done for images.

```javascript
shaders: [
  {
    bufferView: 0
  }
]
```

lasalvavida:
I can make the PR for this if necessary

pjcozzi:
+1

> I can make the PR for this if necessary

@lexaknyazev do you have a preference?

lasalvavida:
Bump

pjcozzi:
@lasalvavida can you make the PR to keep momentum here?  A lot of folks are swamped with finishing other spec work.

pjcozzi:
@donmccurdy should this be considered for KHR_technique_webgl in the short-term?  Seems like yes to stay consistent with glTF 2.0., e.g., textures can be backed by buffers instead of just image files.

donmccurdy:
Yes, i see `bufferView` mentioned in the extension text now (with #894), probably just a matter of finishing the description of the change under the Shader object’s properties.

lexaknyazev:
This feedback has been included into `KHR_techniques_webgl`.

PR #894, opened 2017-03-26
Shader bufferView README changes

lasalvavida:
PR for #868,

it looks like the schema already allows this


pjcozzi:
+1 from me.

@lexaknyazev do you want to look at this before I merge?  This extension is still draft, but we need it to keep moving forward gltf-pipeline work.

pjcozzi:
Merging since this extension is still in draft and we want to keep it consistent with current implementations.

PR #895, opened 2017-03-31
Fix inconsistent schema JSON space formatting

bghgary:


emackey:
Is there a script or something that does this?  (I ask because I've run this schema through homegrown converters and attempted to `JSON.stringify` it, which really highlights inconsistencies).

bghgary:
I did some searching and there are a bunch of tools that do this, but none of them leave the expand/collapse of objects/arrays alone in the file. The best one I've found so far is https://github.com/MaxArt2501/json-fmt which at least allows specifying what to do with the spacing.

I did this one file at a time with Visual Studio proper (not VSCode) which doesn't change the expand/collapse of the objects/arrays in the file, but not everyone will have access to this.

Perhaps, we should just format all the files using an open source tool like json-fmt with specific settings. A script can be easily written to do this. Thoughts?

pjcozzi:
Merging, but please continue the tools conversation here or elsewhere if worthwhile.

PR #898, opened 2017-04-07
Yet another glTF 2.0 texture updates

lexaknyazev:
@bghgary @sbtron 
Please review refined schemas. If everything's OK, I'll update readme and merge asap.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -836,34 +836,30 @@ A skin is instanced within a node using a combination of the node's `mesh` and `
 
 ## Texture Data
 
-**TODO: describe separation of concerns: textures / images / samplers**
+glTF separates texture access into three distinct types of objects: Textures, Images, and Samplers.
 
 ### Textures
 
-**TODO: update to latest revisions**
-
-All textures are stored in the asset's `textures` array. A texture is defined by an image file, denoted by the `source` property; `format` and `internalFormat` specifiers, corresponding to the GL texture format types; a `target` type for the sampler; a sampler identifier (`sampler`), and a `type` property defining the internal data format. Refer to the GL definition of `texImage2D()` for more details.
+All textures are stored in the asset's `textures` array. A texture is defined by an image resource, denoted by the `source` property and a sampler index (`sampler`).
 
 ```json
 {
     "textures": [
         {
-            "format": 6408,
-            "internalFormat": 6408,
             "sampler": 0,
-            "source": 2,
-            "target": 3553,
-            "type": 5121
+            "source": 2
         }
     ]
 }
 ```
 
+> **Implementation Note** glTF 2.0 supports only 2D textures.
+
 ### Images
 
-**TODO: update to latest revisions**
+Images referred to by textures are stored in the `images` array of the asset. 
 
-Images referred to by textures are stored in the `images` array of the asset. Each image contains a URI to an external file (or a reference to a bufferView) in one of the supported images formats. Image data may also be stored within the glTF file as base64-encoded data and referenced via data URI. For example:
+Each image contains a URI to an external file (or a reference to a bufferView) in one of the supported images formats. Image data may also be stored within the glTF file as base64-encoded data and referenced via data URI. For example:
 
 ```json
 {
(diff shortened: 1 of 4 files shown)


specification/2.0/schema/image.schema.json
```diff
@@ -29,5 +28,8 @@
     "additionalProperties": false,
     "dependencies": {
         "bufferView": [ "mimeType" ]
+    },
+    "not": {
```
bghgary: I think anyOf here is clearer than using not? See http://stackoverflow.com/questions/31839578/how-to-define-a-json-schema-that-requires-at-least-one-of-many-properties

lexaknyazev: Current approach only disallows presence of both fields at the same time.
"anyOf" with two "required" entries has different meaning (either is required) and doesn't prevent ambiguous case. 
```json
{
  "anyOf": [
    {"required":["uri"]},
    {"required":["bufferView"]}
  ]
}
```

sbtron: Got it. Looks good to me.

lexaknyazev:
@sbtron @bghgary 
Please, review updated README.

specification/2.0/README.md
```diff
@@ -836,34 +836,30 @@ A skin is instanced within a node using a combination of the node's `mesh` and `
...
 
+> **Implementation Note** glTF 2.0 supports only 2D textures.
+
 ### Images
 
-**TODO: update to latest revisions**
+Images referred to by textures are stored in the `images` array of the asset. 
 
-Images referred to by textures are stored in the `images` array of the asset. Each image contains a URI to an external file (or a reference to a bufferView) in one of the supported images formats. Image data may also be stored within the glTF file as base64-encoded data and referenced via data URI. For example:
+Each image contains a URI to an external file (or a reference to a bufferView) in one of the supported images formats. Image data may also be stored within the glTF file as base64-encoded data and referenced via data URI. For example:
```
pjcozzi: Perhaps reorganize this paragraph as:

Each image contains one of
* a URI to an external file in one of the supported images formats, or
* a URI with embedded base64-encoded data, or
* a reference to a `bufferView`.

The following example shows an image pointing to an external .png image file and another image referencing a `bufferView`.

specification/2.0/README.md
```diff
@@ -878,11 +874,15 @@ Images referred to by textures are stored in the `images` array of the asset. Ea
     ]
 }
 ```
-> **Implementation Note**: With WebGL API, the first pixel transferred from the `TexImageSource` (i.e., HTML Image object) to the WebGL implementation corresponds to the upper left corner of the source. Non-WebGL runtimes may need to flip Y axis to achieve correct texture rendering.
+First image pixel corresponds to the lower left corner of the image.
 
-### Samplers
+> **Implementation Note:** With WebGL API, the first pixel transferred from the `TexImageSource` (i.e., HTML Image object) to the WebGL implementation corresponds to the upper left corner of the source. To achieve correct rendering, WebGL runtimes must flip Y axis by enabling `UNPACK_FLIP_Y_WEBGL` flag. This behavior was changed from glTF 1.0.
```
pjcozzi: @lasalvavida heads up on this subtle 1.0 -> 2.0 change.

specification/2.0/README.md
```diff
@@ -878,11 +874,15 @@ Images referred to by textures are stored in the `images` array of the asset. Ea
...
+
+Any colorspace information (such as ICC profiles, intents, etc) from PNG or JPEG containers must be ignored.
 
-**TODO: update to latest revisions**
+> **Implementation Note:** This increases portability of an asset, since not all image decoding libraries fully support custom color conversions. To achieve correct rendering, WebGL runtimes must disable such conversions by setting `UNPACK_COLORSPACE_CONVERSION_WEBGL` flag to `NONE`.
```
pjcozzi: @lasalvavida also note this update.

specification/2.0/schema/image.schema.json
lexaknyazev: @bghgary @sbtron 
Given that the spec doesn't define a case when both fields aren't present, how about changing this to `"oneOf"`? (I.e. one of these fields must be provided.)
```json
{
  "oneOf": [
    {"required":["uri"]},
    {"required":["bufferView"]}
  ]
}
```

Otherwise we can add a note on how to handle a case when neither of them is provided.

pjcozzi: +1 for `oneOf` so there aren't non-obvious cases.

bghgary: Only concern for me is code generators. Let me try it out and see what they do with this.

specification/2.0/schema/image.schema.json
```diff
@@ -14,8 +14,8 @@
         },
         "mimeType": {
             "type": "string",
-            "description": "The image's MIME type.",
-            "minLength": 1
+            "enum": [ "image/jpeg", "image/png" ],
```
lexaknyazev: @sbtron @bghgary 
Should we change this to forward-compatible style?

specification/2.0/schema/image.schema.json
bghgary: Just tried with NJsonSchema and gltf-typescript-generator and both seems to ignore the oneOf which is probably the correct behavior. +1 for me.

specification/2.0/schema/image.schema.json
bghgary: Yes. Good catch. Missed this from the merge.

lexaknyazev: Fixed with https://github.com/KhronosGroup/glTF/commit/3a18fdba00f16b4f3b358c6d14e755bfa626c81a.

specification/2.0/schema/image.schema.json
lexaknyazev: Fixed with https://github.com/KhronosGroup/glTF/commit/1673ba355385e82f15216918cf6d0a9d51c3f76e.

specification/2.0/schema/texture.schema.json
```diff
@@ -5,55 +5,18 @@
...
-            "enum": [ 6406, 6407, 6408, 6409, 6410 ],
-            "gltf_enumNames": [ "ALPHA", "RGB", "RGBA", "LUMINANCE", "LUMINANCE_ALPHA" ],
-            "default": 6408,
-            "gltf_detailedDescription": "The texture's internal format.  Valid values correspond to WebGL enums: `6406` (ALPHA), `6407` (RGB), `6408` (RGBA), `6409` (LUMINANCE), and `6410` (LUMINANCE_ALPHA).  Defaults to same value as format.",
-            "gltf_webgl": "`texImage2D()` internalFormat parameter"
-        },
         "sampler": {
             "allOf": [ { "$ref": "glTFid.schema.json" } ],
-            "description": "The index of the sampler used by this texture."
+            "description": "The index of the sampler used by this texture. When undefined, a sampler with repeat wrapping and auto filtering should be used."
```
significant-bit: What exactly does "auto filtering" mean? Why were the default min/mag filter values removed from the spec? I'm writing an importer.

Issue #891, opened 2017-03-22
Replace 'gltf_enumNames' with 'oneOf', and other schema cleanup


emackey:
I encountered a couple issues with the structure of the glTF schema while attempting to import it into my [gltf-vscode extension](https://marketplace.visualstudio.com/items?itemName=cesium.gltf-vscode).  I'd like to propose some structural changes (at least to 2.0, although they would be completely non-breaking if applied retroactively to 1.0 as well).

First, and primarily, the use of `gltf_enumNames` is non-standard and not compatible with JSON validators such as the one in vscode.  For example, `texture.schema.json` in both 1.0 and 2.0 (draft) contains the following lines:

```
        "format": {
            "type" : "integer",
            "description" : "The texture's format.",
            "enum" : [6406, 6407, 6408, 6409, 6410],
            "gltf_enumNames" : ["ALPHA", "RGB", "RGBA", "LUMINANCE", "LUMINANCE_ALPHA"],
            "default" : 6408,
            "gltf_detailedDescription" : "The texture's format.  Valid values correspond to WebGL enums: `6406` (ALPHA), `6407` (RGB), `6408` (RGBA), `6409` (LUMINANCE), and `6410` (LUMINANCE_ALPHA).",
            "gltf_webgl" : "`texImage2D()` format parameter"
        },
```

The JSON validator has two problems with this: (1) It can't understand the descriptions given for the individual enum names, and (2) the use of `"type" : "integer"` causes it to suggest `0` when no default value is supplied (as opposed to suggesting the first enum value).  (Well I picked a bad example here since there *is* a default value for this one, but there are other places like [technique parameters](https://github.com/KhronosGroup/glTF/blob/master/specification/1.0/schema/technique.parameters.schema.json#L21-L27) where no default is supplied).

I ended up writing a [node.js schema transformation script](https://github.com/AnalyticalGraphicsInc/gltf-vscode/blob/1.0.3/util/importSchema.js#L57-L86), that takes the glTF schema and transforms its enums to look similar to this:

```
        "format" : {
            "description" : "The texture's format.",
            "default" : 6408,
            "gltf_detailedDescription" : "The texture's format.  Valid values correspond to WebGL enums: `6406` (ALPHA), `6407` (RGB), `6408` (RGBA), `6409` (LUMINANCE), and `6410` (LUMINANCE_ALPHA).",
            "gltf_webgl" : "`texImage2D()` format parameter",
            "oneOf" : [
                {
[...]

PR #899, opened 2017-04-07
Add versioning info

sbtron:
Added explanation of how we envision glTF will version with backwards and forwards compatibility in the 2.x timeframe. 

https://github.com/sbtron/glTF/tree/Versioning/specification/2.0#versioning 

Updated schema to allow for forwards compatibility
- additionalProperties:false is removed so that future versions can add additional properties and a 2.0 loader can still validate a 2.x file using schema.
- All enums now use anyOf with an additional generic type object to allow for additional enums in future versions. This also addresses #891
- Added optional minVersion property to allow asset authors to opt out of forwards compat and require a minimum supported version (similar to extensionsRequired concept).


specification/2.0/schema/accessor.schema.json
```diff
@@ -18,12 +18,38 @@
...
+                {
+                    "enum": [ 5125 ],
+                    "description": "UNSIGNED_INT"
+                },
+                {
+                    "enum": [ 5126 ],
+                    "description": "FLOAT"
+                },
+                {
+                    "type": "integer"
```
emackey: Why is this needed?  Are there integers that are valid here, other than the ones listed above?

specification/2.0/schema/accessor.schema.json
```diff
@@ -39,10 +65,33 @@
...
+                    "enum": [ "MAT2" ]
+                },
+                {
+                    "enum": [ "MAT3" ]
+                },
+                {
+                    "enum": [ "MAT4" ]
+                },
+                {
+                    "type": "string"
```
emackey: Same comment here with `string`.  I'm not a JSON Schema expert, but I tried actually loading this branch into VSCode's JSON validator, and having this `type` at the end of the enum list allows arbitrary strings (or arbitrary ints in the comment above) to validate as correct.  Removing this `type` means it will only validate with an exact match to one of the named enum values.

emackey: For example:

```
            "componentType": 99999,
            "type": "UNKNOWN"
```

This validates as allowed, because it matches `anyOf` `integer` and `anyOf` `string`.

sbtron: There are a couple of ways these schemas will get used;
- Viewers/Loaders can validate a file before trying to display it. The additional objects in anyOf allow loaders to validate a future version and continue to load it while ignoring any new elements.  
- Validators like in vs-code or glTF-validator may want to ensure the file is exactly 2.0 and not 2.1. Loosening up the schema will not be useful for this type of use case. If strict validations are required then these tools will need to use a stricter version of the schema that is specific to a version and doesn't do the forwards compatibility. 

Do we need to consider making a "strict" version of the schema? By default the schema should be flexible because its still valid 2.0 with the forwards compat capabilities. We could do some script to create the stricter version that specifically checks for 2.0 without forwards compat.

emackey: I see.  What behavior should a 2.0-capable loader take if it finds a "future" value that it doesn't understand?  Will it just look for a default value in the schema?

sbtron:
The working group has been discussing this. Pasting it here so everyone can read. The examples are speculations of features that could be added in future versions and are only meant to show potential approaches of adding them.

To support forwards compatibility, whenever a new feature is added the spec needs to allow for the following requirements: 

1. Only specifying old the feature set (no new features used)
2. Specifying the new feature but allowing for a fallback 
   a. The fallback could be allowing older parsers to completely ignore the feature.
   b. If the new feature is not supported then use the old feature. 
3. Allowing authors to require that a runtime supports a new feature in order to load the asset 

Requirement 1. is essentially backwards compatibility- you cannot change existing behavior in a spec update. 

For requirement 2.a a glTF 2.0 parser should still be able to read a 2.1 file and completely ignore any new features.

For example say glTF 2.1 adds lighting capability by: 
* Adding a new light node at root
* Added a new light property under the nodes property

(This is a completely made up example to show how this will work) 
```
"lights": [
    "directional" : {
        "color": [ 1, 1, 1 ]
    },
],
"nodes": [
    {
        "children": [ ],
        "light": 0,
    }
]
```
A 2.0 parser can ignore any additional properties it finds whereas a 2.1 parser can correctly interpret those additional parameters. 

For requirement 2.b any new features are ignored but a fall back behavior to 2.0 should exist.
For example say Draco Mesh compression is added to core glTF 2.1. To stay compatible with 2.0 clients, the compressed mesh should be provided along with the uncompressed mesh to allow for a fallback. 
(Example only, not meant to be exact) 
```
"primitives": [
    {
        "attributes": {
            "POSITION": 11,
            "NORMAL": 12,
            "TEXCOORD_0": 13
        },
        "indices": 10,
        "mode": 4,
        "draco_mesh": {
            "bufferView": 5,
            "indexCount": 1000,
            "vertexCount": 500,
            "attributes": [ "POSITION", "NORMAL", "TEXCOORD_0" ],
            "version": "0.9.1"
        }
    }
]
```
A 2.1 client can read and understand the “draco_mesh” element. A 2.0 client can simply ignore it and use the uncompressed geometry.

Requirement 3. will technically break forwards compatibility, but asset creators may want to choose to explicitly opt out of compatibility and require a specific version.  The draco mesh compression is a good example where someone could choose to require a minimum version. For a client-server type of scenario where a JSON file points to the bin files, you can have both the compressed mesh and uncompressed mesh and a client can just download the one it understands. However, for a self contained binary glTF file scenario, you may only want to keep the compressed version since the whole point was to reduce the file size. In this scenario an author could choose to explicitly require a minimum version of 2.1. The JSON for such a file must still contain the required elements so that it validates, but it could just point to an empty geometry which will never gets used. 

Requirements 2.a and 2.b are satisfied by allowing additional properties in all nodes of the JSON.
Requirement 3. is satisfied by adding a minVersion property to the asset node.

Enums provide an interesting challenge and example below show how to update them in a forwards compatible manner : 
Imagine `SPLINE` interpolation is added in glTF 2.1 (its actually part of 2.0 but just using it as an example). Say SPLINE is added such that you can specify a fallback to Linear interpolation.
To do this in a forwards compatible way a new `additionalSamplers` property will need to be added under the channels node. 
 
```
"samplers" : [
    {
        "input": 2,
        "interpolation": "LINEAR",
        "output": 3
    },
    {
        "input": 4,
        "interpolation": "SPLINE",
        "output": 5
    }
], "channels" : [
    {
        "additionalSamplers": [ 1 ],
        "sampler": 0,
        "target": {
            "node": 0,
            "path": "rotation"
        }
    }
]
```
Only a client that understands glTF 2.1 will understand the additionalsamplers property and will read the sampler which uses spline interpolation. Older samplers should simply ignore the additionalSamplers property and read the 2.0 sampler property to use linear interpolation.

specification/2.0/schema/accessor.schema.json
sbtron: The short answer is - it can be ignored. The long answer is - future updates will have to be designed such that new values can be ignored safely. Just pasted some examples we have been discussing below that goes into more detail.

lexaknyazev:
@sbtron @bghgary @pjcozzi 
What should `asset.version` string look like? There could be two approaches:
1. Arbitrary string, order and values to be defined manually by each subsequent update ("2.0", "2.0a", "2.2.45").
2. Fixed format string value ("2.0", "2.1", "2.2", etc). In this case, a simple regex could be used to retrieve major/minor integer.

Same question about `asset.minVersion`.

pjcozzi:
@lexaknyazev following Khronos versioning, the format for `version` and `minVersion` could be "major.minor.revision"

I don't think we will have other variations other than perhaps `minor` and `revision` being optional.

specification/2.0/README.md
```diff
@@ -131,6 +132,14 @@ Version 2.0 of glTF does not define compression for geometry and other rich data
 
 > The 3D Formats Working Group is developing partnerships to define the codec options for geometry compression.  glTF defines the node hierarchy, materials, animations, and geometry, and will reference the external compression specs.
 
+## Versioning
+
+Any updates made to glTF in a minor version will be backwards and forwards compatible. Backwards compatibility will ensure that any client implementation that supports loading a glTF 2.x file will also be able to load a glTF 2.0 file. Forwards compatibility will allow a client implementation that only supports glTF 2.0 to load glTF 2.x files while gracefully ignoring any new features it does not understand.
```
pjcozzi: Perhaps "minor **or revision** version"

pjcozzi: Throughout this section, replace "glTF 2.0 file" with "glTF 2.0 asset", which is more consistent with the rest of the spec.  (Looks like the spec itself could use a tweak or two here as well if you do a quick search).

specification/2.0/README.md
```diff
@@ -176,6 +185,9 @@ Each glTF asset must have an `asset` property. In fact, it's the only required t
 }
 ```
 
+> **Implementation Note:** Client implementations should first check whether a `minVersion` property is specified and ensure both major and minor versions can be supported. If no `minVersion` is specified, then clients should check the `Version` property and ensure the major version is supported. Clients that load [GLB format](GLB_FORMAT.md) should also check for the `minVersion` and `Version` properties in the JSON chunk as the version specified in the GLB header only refers to the GLB container version.
```
pjcozzi: Throughout, `Version` -> `version`.  Property names should be camelCase.

pjcozzi:
+1 from me.  Just trivial comments.

lexaknyazev:
> @lexaknyazev following Khronos versioning, the format for `version` and `minVersion` could be "major.minor.revision"

@pjcozzi 
I'd make minor part required and maybe drop revision, because I don't think that file format needs it.
We can continue refining language of 2.0 spec without changing version string.

pjcozzi:
OK with me, but let's confirm on the call today about dropping `revision`.

lexaknyazev:
@sbtron @bghgary 
Textures and bufferView conflicts resolved, please merge if it's ready.

specification/2.0/schema/animation.sampler.schema.json
```diff
@@ -13,23 +13,19 @@
...
             "anyOf": [
                 {
-                    "enum": [ "LINEAR" ],
-                    "description": "When an animation targets a node's rotation, and the animation's interpolation is `\"LINEAR\"`, spherical linear interpolation (slerp) should be used to interpolate quaternions."
+                    "enum": [ "LINEAR" ]
                 },
                 {
-                    "enum": [ "STEP" ],
-                    "description": "When interpolation is `\"STEP\"`, animated value remains constant to the value of the first point of the timeframe, until the next timeframe."
+                    "enum": [ "STEP" ]
```
emackey: We should make a note to restore these once wetzel is upgraded to handle them.  The whole point of my lobbying for the change from non-standard to `anyOf` was specifically to get these descriptions per-enum-value.

bghgary: Agreed. Descriptions are handled, but long descriptions make the generated text hard to read. Since these didn't have descriptions initially, I just put them back to the original for now.

emackey: I opened #955 for this.

PR #900, opened 2017-04-08
Refine bufferView.target

lexaknyazev:
As discussed offline.

CC @sbtron @bghgary @pjcozzi

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -373,8 +373,11 @@ The following example defines a buffer. The `byteLength` property specifies the 
 }
 ```
 
-A *bufferView* represents a subset of data in a buffer, defined by an integer offset into the buffer specified in the `byteOffset` property, a `byteLength` property to specify length of the buffer view.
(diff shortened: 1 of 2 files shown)


pjcozzi:
This makes `target` optional, correct?

If it's optional, it is still not possible to write a bottom-up loader since the loader would have to fall back to top-down.  If we are convinced that no one should ever write a bottom-up loader...which still seems like a big **if** to me...then why not completely remove this property since it can't be relied on?

lexaknyazev:
> This makes `target` optional, correct?...

Strictly speaking, it was always optional schema-wise. Its requiredness was implicit from description. An ideal solution would be to require it always, and use `null` (or `GL_NONE`) for non-GPU buffer views, but we can't make such change anymore.

Pure bottom-up loader is still possible as long as an asset follows all recommendations.

pjcozzi:
Sorry, I am still confused by this.  Is correct usage to define this when the data data is vertex/index data and to not define this when the data is skin/keyframes?  Or can this not be defined when the data is vertices/indices.  If it is the former, sounds OK.  If it is the later, should this be removed for the reason stated above?

Issue #812, opened 2017-01-04
Add BINORMAL and TANGENT attribute semantics


pjcozzi:
From https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/203/:

> The glTF spec does not define `NORMAL` and `TANGENT` attributes.  See
>
> https://github.com/KhronosGroup/glTF/tree/master/specification/1.0#parametersemantic
>
> Perhaps we should add them for 1.1 since it is trivial?
>
> Otherwise, the PBR extension could define them, but they would need to be prefixed with `_` here so the glTF asset is still valid before the extension is added.

@lexaknyazev @mlimper

lexaknyazev:
Did you mean `BINORMAL`?

pjcozzi:
Yes, I updated the title.

javagl:
As this is supposed to affect the spec, a nitpick from me, quoting from http://mathworld.wolfram.com/BinormalVector.html :

> In the field of computer graphics, two orthogonal vectors tangent to a surface are frequently referred to as tangent and binormal vectors. However, **for a surface, the two vectors are more properly called tangent and bitangent vectors**. 

Some people have strong opinions about this, because the surface only has *one* normal, but it has infinitely many tangents, and one can only compute a bitangent for any given tangent. 

(I know, "binormal" is *so* common that one might have the choice between "common terminology" and "right terminology" here. But some tutorials, e.g. http://www.opengl-tutorial.org/intermediate-tutorials/tutorial-13-normal-mapping/ are already trying to use the "right" one...)

lexaknyazev:
If we can universally figure out following questions, it should be possible to include that in the core:
- Normals could deliberately have values that are different from actual surface normals (smooth/flat shading). Is similar behavior acceptable with these new vectors? Or are they strictly defined by existing positions/normals?
- Are Cesium equations sufficient for most common usecases?
- Are there any additional limitations on these semantics, such as: 
  - if one is defined, other also must be defined, or
  - they must form only right (or left) hand system, etc?

pjcozzi:
> Normals could deliberately have values that are different from actual surface normals...Is similar behavior acceptable with these new vectors?

Yes, this is similar in the GL world, where, for example, vertex normals are not used by the pipeline for backface culling since they could be anything.

> Are Cesium equations sufficient for most common usecases?

[...]

PR #901, opened 2017-04-10
Add support for stored tangents and add mesh attributes table

bghgary:


specification/2.0/README.md
```diff
@@ -637,13 +638,28 @@ The following example defines a mesh containing one triangle set primitive:
...
+Valid accessor type and component type for each attribute semantic property are defined below.
+
+|Name|Accessor Type(s)|Component Type(s)|Description|
+|----|----------------|-----------------|-----------|
+|`POSITION`|`"VEC3"`|`5126`&nbsp;(FLOAT)|XYZ vertex positions|
+|`NORMAL`|`"VEC3"`|`5126`&nbsp;(FLOAT)|XYZ vertex normals|
+|`TANGENT`|`"VEC4"`|`5126`&nbsp;(FLOAT)|XYZW vertex tangents where the w component is a signed float (-1, +1) indicating handedness of the tangent basis|
+|`TEXCOORD_0`|`"VEC2"`|`5126`&nbsp;(FLOAT)|UV texture coordinates for the first set|
+|`TEXCOORD_1`|`"VEC2"`|`5126`&nbsp;(FLOAT)|UV texture coordinates for the second set|
+|`COLOR_0`|`"VEC3"`<br>`"VEC4"`|`5126`&nbsp;(FLOAT)|RGB or RGBA vertex color|
```
lexaknyazev: Normalized UNSIGNED_BYTE should be also allowed here.

specification/2.0/README.md
```diff
@@ -637,13 +638,28 @@ The following example defines a mesh containing one triangle set primitive:
...
 Each attribute is defined as a property of the `attributes` object. The name of the property corresponds to an enumerated value identifying the vertex attribute, such as `POSITION`. The value of the property is the index of an accessor that contains the data.
 
-Valid attribute semantic property names include `POSITION`, `NORMAL`, `TEXCOORD`, `COLOR`, `JOINT`, and `WEIGHT`.  `TEXCOORD` and `COLOR` attribute semantic property names must be of the form `[semantic]_[set_index]`, e.g., `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_1`, etc.  For forward-compatibility, application-specific semantics must start with an underscore, e.g., `_TEMPERATURE`.
+Valid attribute semantic property names include `POSITION`, `NORMAL`, `TANGENT`, `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_0`, `JOINT`, and `WEIGHT`.  Application-specific semantics must start with an underscore, e.g., `_TEMPERATURE`.
+
+Valid accessor type and component type for each attribute semantic property are defined below.
+
+|Name|Accessor Type(s)|Component Type(s)|Description|
+|----|----------------|-----------------|-----------|
+|`POSITION`|`"VEC3"`|`5126`&nbsp;(FLOAT)|XYZ vertex positions|
```
lexaknyazev: Why do we need to define component type here? 
WEB3D_quantized_accessors requires normalized UNSIGNED_SHORT.

In general, it's not important spec-wise what the data type is as long as it matches actually stored data. Shaders shouldn't be affected anyway.

specification/2.0/README.md
```diff
@@ -637,13 +638,28 @@ The following example defines a mesh containing one triangle set primitive:
...
+|`NORMAL`|`"VEC3"`|`5126`&nbsp;(FLOAT)|XYZ vertex normals|
+|`TANGENT`|`"VEC4"`|`5126`&nbsp;(FLOAT)|XYZW vertex tangents where the w component is a signed float (-1, +1) indicating handedness of the tangent basis|
```
lexaknyazev: Sounds like a good candidate for 10-10-10-2 vertex format (ES 3.0+). Could be considered for the next glTF update.

specification/2.0/README.md
bghgary: Okay, I can add that.

specification/2.0/README.md
```diff
@@ -637,13 +638,28 @@ The following example defines a mesh containing one triangle set primitive:
...
+|`TEXCOORD_0`|`"VEC2"`|`5126`&nbsp;(FLOAT)|UV texture coordinates for the first set|
```
lexaknyazev: Normalized UNSIGNED_SHORT should be allowed here.

specification/2.0/README.md
```diff
@@ -637,13 +638,28 @@ The following example defines a mesh containing one triangle set primitive:
...
+|`TEXCOORD_1`|`"VEC2"`|`5126`&nbsp;(FLOAT)|UV texture coordinates for the second set|
```
lexaknyazev: Do we limit the number of UVs to 2? 
If so, we should state this more explicitly somewhere.

specification/2.0/README.md
bghgary: The issue is that runtime implementations that have an abstraction for vertex data don't directly deal with buffers. BabylonJS, for example, can only accept number[] or Float32Array for positions. Support for normalized UNSIGNED_SHORT can be added, so we can definitely do that, but I don't think we should leave it open ended.

specification/2.0/README.md
bghgary: Agreed.

specification/2.0/README.md
bghgary: I'll add normalized UNSIGNED_SHORT to all relevant attributes.

specification/2.0/README.md
bghgary: Yes. We need to limit them to something to prevent inconsistencies between implementations. 2 seemed reasonable to me. I'll add more text on this.

pjcozzi:
+1

specification/2.0/README.md
lexaknyazev: In general, runtime should be able to read any vertex format using info from accessor's fields. I agree that for the core spec other types for vertex positions make little sense.
Can we say that _required_ extensions (e.g. `WEB3D_quantized_accessors`) may extend this table?

bghgary: I would say for any extension not just required extension may extend this table. This is analogous to the versioning concepts. I will add some text that states this explicitly.

bghgary:
This addresses #812.

PR #907, opened 2017-04-11
Update vertex color calculation for metal-roughness and spec-gloss

sbtron:
Adding explanation on how pbr materials properties interact with vertex colors if they are present.

### DIFF
--- a/extensions/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
+++ b/extensions/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
@@ -66,9 +66,9 @@ The specular property from specular-glossiness material model is the same as the
 
 The value for each property (`diffuse`, `specular`, `glossiness`) can be defined using factors or textures. The `specular` and `glossiness` properties are packed together in a single texture called `specularGlossinessTexture`.  If a texture is not given, all respective texture components within this material model are assumed to have a value of `1.0`. The factors (`diffuseFactor`, `specularFactor`, `glossinessFactor`) scale, in linear space, the components given in the respective textures (`diffuseTexture`, `specularGlossinessTexture`). Texture content must be converted to linear space before it is used for any lighting computations.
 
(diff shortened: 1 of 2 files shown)


extensions/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
```diff
@@ -66,9 +66,9 @@ The specular property from specular-glossiness material model is the same as the
...
 
-The following equations show how to calculate bidirectional reflectance distribution function (BRDF) inputs (*c<sub>diff</sub>*, *F<sub>0</sub>*, *&alpha;*) from the metallic-roughness material properties. 
+The following equations show how to calculate bidirectional reflectance distribution function (BRDF) inputs (*c<sub>diff</sub>*, *F<sub>0</sub>*, *&alpha;*) from the metallic-roughness material properties. In addition to the material properties, if a primitive specifies a vertex color using the attribute semantic property `COLOR_0`, then this value acts as an additional multiplier to *c<sub>diff</sub>*.
 
-*c<sub>diff</sub>* = `diffuse.rgb * (1 - max(specular.r, specular.g, specular.b)`
+*c<sub>diff</sub>* = `diffuse.rgb * (1 - max(specular.r, specular.g, specular.b) * vertexColor`
```
javagl: Missing ) here?

peteroupc:
Can you explain why the vertex color appears in the calculation as given here?  Is this how common physically-based rendering engines treat vertex colors?

bghgary:
The original thought was that we wanted vertex colors to affect metallic-roughness the same as specular-glossiness PBR models. But it looks like at least BabylonJS and three.js both multiply against the baseColor of the metallic-roughness material instead of what is in this PR. We will likely update the PR to reflect this.

pjcozzi:
@sbtron did you make the spec update for this?

sbtron:
@pjcozzi finally updated if you want to review.

pjcozzi:
LGTM

PR #922, opened 2017-04-21
Add optional property mesh.closed specifying that mesh is closed (watertight)

gkv311:
Application (3D viewer) can use closed (watertight) flag of the mesh within rendering techniques where it is important to consider - e.g. when applying clipping planes capped with material or just to enable back-face culling (regardless of doubleSided property of material).

This flag applies to the full list of primitive sets as whole, not for each one individually. E.g. closed mesh can be also declared for several primitive sets with each one defining only an open Shell, but forming a closed mesh as whole through sharing vertex positions (via duplicated vertices).

This property can be easily declared for models exported from CAD application (where it is well-known if mesh forms closed Solid or open Shell), but also can be computed from mesh itself (this operation is however is not cheap). Other applications generating glTF model that don't care might omit this property (considering the mesh is not closed in this case).

### DIFF
--- a/specification/2.0/schema/mesh.schema.json
+++ b/specification/2.0/schema/mesh.schema.json
@@ -23,6 +23,12 @@
                 "type": "number"
             },
             "minItems": 0
+        },
+        "closed": {
+            "type": "boolean",
+            "default": false,
+            "description": "Specifies whether the mesh is closed (watertight).",
+            "gltf_detailedDescription": "Specifies whether the mesh is closed (watertight).  Application can use this property within rendering techniques where it is important to consider - e.g. when applying clipping planes capped with material or just to enable back-face culling (regardless of doubleSided property of material)."
         }
     },
     "required": [ "primitives" ]


pjcozzi:
Thanks for this proposal, @gkv311!

We can't scope this for 2.0, but I labeled this `specification` so we can consider it post 2.0.

In the meantime, in your own applications, you could store this in an `extras` object property:

https://github.com/KhronosGroup/glTF/tree/master/extensions#extensions-vs-extras

donmccurdy:
I would suggest this could also be done as a "boolean" extension, as we've done with KHR_materials_unlit. When the extension is present on a mesh, it implies the mesh is known to be closed. For example, `EXT_mesh_closed`. Or perhaps a larger CAD-related extension, if community support for such a thing comes together.

pjcozzi:
Following up on @donmccurdy's comment above, perhaps this is best done first as a simple extension if it is priority enough for someone to write up or even an app-specific property in an `extras` object?  Over time if it proves broadly useful, it could be candidate for the core spec.

donmccurdy:
Closing this PR, since we cannot change the core 2.0 specification in this way. I'd recommend an extension for CAD features as a better starting point, since a next version of glTF may still be some time away.

PR #928, opened 2017-04-22
Fix primitive.targets, remove unneeded schema indirection

### LINKED ISSUE #927: The encoding of morph targets in the schema
The [current schema for `mesh.primitive.target`](https://github.com/KhronosGroup/glTF/blob/fb8b472701bf54fc2dd80b1ade2dd312f42113ba/specification/2.0/schema/mesh.primitive.schema.json#L64) says that it is an array of `mesh.primitive.target` objects. These objects are supposed to be dictionaries. The [schema for `mesh.primitive.target`](https://github.com/KhronosGroup/glTF/blob/9b131950c0e6a793dbe51a72f6e3ca52e13f2246/specification/2.0/schema/mesh.primitive.target.schema.json) is

    {
        "title": "target",
        "description": "A dictionary object specifying attributes displacements in a Morph Target, where each key corresponds to one of the three supported attribute semantic (`POSITION`, `NORMAL`, or `TANGENT`) and each value is the index of the accessor containing the attribute displacements' data."
    }

I think this means that these objects are just IDs (Integers). 

I'm not a JSON schema expert, but to my understanding, the schema for `mesh.primitive.target` should rather be (roughly!) something like this:

*(EDIT: This may be wrong, see the update below!)*

    {
        "title": "target",
        "type": "object",
        "additionalProperties": {
            "type": { "$ref": "glTFid.schema.json" }
        }
        "description": "A dictionary object specifying attributes displacements in a Morph Target, where each key corresponds to one of the three supported attribute semantic (`POSITION`, `NORMAL`, or `TANGENT`) and each value is the index of the accessor containing the attribute displacements' data."
    }

That is: It is not just an integer, but really an object (dictionary), with arbitrary properties (as indicated by the `additionalProperties`), where the value of each property is an ID. 

Am I overlooking or misinterpreting something here?

--- 

**Update** : I think that `mesh.primitive.target` should be defined as this:

    {
        "title": "target",
        "type": "object",
        "additionalProperties": {
            "allOf": [ { "$ref": "glTFid.schema.json" } ]
        },
        "description": "A dictionary object specifying attributes displacements in a Morph Target, where each key corresponds to one of the three supported attribute semantic (`POSITION`, `NORMAL`, or `TANGENT`) and each value is the index of the accessor containing the attribute displacements' data."
    }

(This way, my code generator creates the targets as `List<Map<String, Integer>> targets` (instead of `List<Integer> targets`), but of course, this does not mean much: The code generator may very well be buggy. However, the original definition for me looks like a `mesh.primitive.target` **is-a** ID, which probably isn't correct either...)

lexaknyazev:
Fix #927, remove indirection.

@bghgary 
Should we require that `attributes` isn't empty? It's possible with `"minProperties": 1` schema rule. Same question about elements of the `targets` array.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -690,7 +690,9 @@ Valid accessor type and component type for each attribute semantic property are 
 #### Morph Targets
 
 Morph Targets are defined by extending the Mesh concept.
+
 A Morph Target is a morphable Mesh where primitives' attributes are obtained by adding the original attributes to a weighted sum of targets attributes.
+
 For instance, the Morph Target vertices `POSITION` for the primitive at index *i* are computed in this way:
 ```
 primitives[i].attributes.POSITION + 
@@ -698,7 +700,8 @@ primitives[i].attributes.POSITION +
   weights[1] * primitives[i].targets[1].POSITION +
   weights[2] * primitives[i].targets[2].POSITION + ...
 ```
-Morph Targets are implemented via the `targets` property defined in the Mesh `primitives`. Each target in the `targets` array is a dictionary mapping a primitive attribute to an accessor containing Morph Target displacement data, currently only two attributes ('POSITION' and 'NORMAL') are supported. All primitives are required to list the same number of `targets` in the same order.
+Morph Targets are implemented via the `targets` property defined in the Mesh `primitives`. Each target in the `targets` array is a dictionary mapping a primitive attribute to an accessor containing Morph Target displacement data, currently only three attributes (`POSITION`, `NORMAL`, and `TANGENT`) are supported. All primitives are required to list the same number of `targets` in the same order.
+
 The `weights` array is optional, it stores the default targets weights, in the absence of `node.weights` the primitives attributes are resolved using these weights. When this property is missing the default targets weights are assumed to be zero.
 
 The following example extends the Mesh defined in the previous example to a morphable one by adding two Morph Targets:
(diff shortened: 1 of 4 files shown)


bghgary:
> Should we require that `attributes` isn't empty?

I can't think of any reason why someone would want an empty attributes property. Attributes should at least include positions for it to be useful. Whether it's actually the `POSITION` attribute depends on extensions. I think requiring non-empty is okay to me.

>  elements of the `targets` array

Seems reasonable to me to require non-empty considering `weights` is already this way.

HowardWolosky:
This now indicates that tangents should be stored here as well.  The [documentation](https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#morph-targets) still indicates that only `NORMAL` and `POSITION` are supported.  So, either this schema needs to remove the reference to `TANGENT`, or the documentation should be updated with this change.

lexaknyazev:
> documentation should be updated with this change

Sure. I just wanted to keep this PR atomic. Tangents were added quite recently and the main spec text hasn't been updated yet.

bghgary:
> Seems reasonable to me to require non-empty considering weights is already this way.

Replying to my own comment here. Missed the fact that targets is an array of a dictionary, not just a dictionary. So this comment doesn't make much sense. But regardless, I still agree with the change.

lexaknyazev:
It's about both: non-empty array of non-empty dictionaries.

PR #930, opened 2017-04-23
Clarify byteStride restrictions

lexaknyazev:
No syntax or semantic changes, just more verbose description.

### DIFF
--- a/specification/2.0/schema/mesh.primitive.schema.json
+++ b/specification/2.0/schema/mesh.primitive.schema.json
@@ -17,7 +17,7 @@
         "indices": {
             "allOf": [ { "$ref": "glTFid.schema.json" } ],
             "description": "The index of the accessor that contains the indices.",
-            "gltf_detailedDescription": "The index of the accessor that contains mesh indices.  When this is not defined, the primitives should be rendered without indices using `drawArrays()`.  When defined, the accessor must contain indices: the `bufferView` referenced by the accessor must have a `target` equal to 34963 (ELEMENT_ARRAY_BUFFER); a `byteStride` that is tightly packed, i.e., 0 or the byte size of `componentType` in bytes; `componentType` must be 5121 (UNSIGNED_BYTE), 5123 (UNSIGNED_SHORT) or 5125 (UNSIGNED_INT), the latter is only allowed when `OES_element_index_uint` extension is used; `type` must be `\"SCALAR\"`."
+            "gltf_detailedDescription": "The index of the accessor that contains mesh indices.  When this is not defined, the primitives should be rendered without indices using `drawArrays()`.  When defined, the accessor must contain indices: the `bufferView` referenced by the accessor should have a `target` equal to 34963 (ELEMENT_ARRAY_BUFFER); `componentType` must be 5121 (UNSIGNED_BYTE), 5123 (UNSIGNED_SHORT) or 5125 (UNSIGNED_INT), the latter may require enabling additional hardware support; `type` must be `\"SCALAR\"`."
         },
         "material": {
             "allOf": [ { "$ref": "glTFid.schema.json" } ],
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -584,7 +584,7 @@ When nor `sparse`, neither `bufferView` is defined, `min` and `max` properties c
 
 The offset of an `accessor` into a `bufferView` (i.e., `accessor.byteOffset`) and the offset of an `accessor` into a `buffer` (i.e., `accessor.byteOffset + bufferView.byteOffset`) must be a multiple of the size of the accessor's component type.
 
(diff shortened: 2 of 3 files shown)


specification/2.0/schema/mesh.primitive.schema.json
```diff
@@ -17,7 +17,7 @@
...
-            "gltf_detailedDescription": "The index of the accessor that contains mesh indices.  When this is not defined, the primitives should be rendered without indices using `drawArrays()`.  When defined, the accessor must contain indices: the `bufferView` referenced by the accessor must have a `target` equal to 34963 (ELEMENT_ARRAY_BUFFER); a `byteStride` that is tightly packed, i.e., 0 or the byte size of `componentType` in bytes; `componentType` must be 5121 (UNSIGNED_BYTE), 5123 (UNSIGNED_SHORT) or 5125 (UNSIGNED_INT), the latter is only allowed when `OES_element_index_uint` extension is used; `type` must be `\"SCALAR\"`."
+            "gltf_detailedDescription": "The index of the accessor that contains mesh indices.  When this is not defined, the primitives should be rendered without indices using `drawArrays()`.  When defined, the accessor must contain indices: the `bufferView` referenced by the accessor should have a `target` equal to 34963 (ELEMENT_ARRAY_BUFFER); `componentType` must be 5121 (UNSIGNED_BYTE), 5123 (UNSIGNED_SHORT) or 5125 (UNSIGNED_INT), the latter may require enabling additional hardware support; `type` must be `\"SCALAR\"`."
```
bghgary: > a `byteStride` that is tightly packed, i.e., 0 or the byte size of `componentType` in bytes

Why is this no longer a requirement?

lexaknyazev: It never was _a requirement_. I don't think that there's any use case for this field except interleaved vertex attribs. So it would be safer to limit its usage explicitly.

lexaknyazev: The description of `bufferView.byteStride` was also refined to reflect this:
https://github.com/KhronosGroup/glTF/blob/5e387210ab4cebc6427e1fca609b1919c7a7495e/specification/2.0/schema/bufferView.schema.json#L29

bghgary: Are we saying byteStride is only for vertex attributes which means it will be ignored (or not allowed) for other data? If so, maybe this needs some description in the data alignment section?

lexaknyazev: From [Buffers and Buffer Views](https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#buffers-and-buffer-views):
> Buffer view could have `byteStride` property. It means byte-distance between consequential elements. This field is defined only for buffer views that contain vertex attributes.

lexaknyazev: It was discussed in #827.

bghgary: Okay, sounds good.

lexaknyazev: I tried to evaluate other usages, but I don't think they are useful:

- Interleaving index buffer data isn't supported by GPUs.
- Interleaving images makes no sense at all.
- Interleaving animation inputs/outputs or sparse indices with sparse values could work, but complicates implementation w/o any clear benefit (or maybe I'm wrong and it exists?)
- Interleaving skin's IBM matrices doesn't seem to be reasonable, too.

The main intent of #827 changeset was to simplify Vulkan vertex bindings by requiring the same stride value across all accessors that use the same buffer view.

Is it clear, that in case of omitted (zero) `bufferView.byteStride` that restriction still applies? Otherwise, I can elaborate a bit more on it in the README.

bghgary: > Is it clear, that in case of omitted (zero) bufferView.byteStride that restriction still applies? 

My understanding from the text is that if byteStride is zero, then effectively byteStride is deduced from the type/componentType of the accessor that references it. If you are saying that accessors using the same bufferView must still use the same byteStride even when byteStride is zero, then no, I don't think that's clear, but I also don't see why anyone would do this.

lexaknyazev: If different accessors have different "deduced" stride, than "byte-distance between consequential elements" isn't the same accross buffer view which undermines its purpose.

More robust approach would be to require `byteStride` when more than one vertex attribute accessor use the same buffer view.

bghgary: > require `byteStride` when more than one vertex attribute accessor use the same buffer view

Sounds good to me. My guess is that, in practice, none of this will matter since I don't see why anyone would point multiple accessors to the same bufferView with different "deduced" strides.

lexaknyazev: > I don't see why anyone would point multiple accessors to the same bufferView with different "deduced" strides

My point is that such layout must be forbidden, so implementations won't need to "fix" it before uploading to GPU.

bghgary: Fair point. I was thinking too much about exporting.

PR #934, opened 2017-04-28
Add note to the pbrMetallicRoughness schema

lilleyse:
Copied this note from the `emissiveFactor` to the other factors:

>  If an emissiveTexture is specified, this value is multiplied with the texel values.

Feel free to edit the wording.


bghgary:
Thanks @lilleyse!

While reviewing your changes, I also noticed we missed specifying the color space for emissiveFactor. Would you mind adding that to your PR?

```
The RGB components of the emissive color of the material. These values are linear. If an emissiveTexture is specified, this value is multiplied with the texel values.
```

lilleyse:
Updated.

Issue #931, opened 2017-04-23
Numeric limitations on morph target's weights


lexaknyazev:
@emilian0 
Spec is nor consistent, neither explicit on valid values for morph target's weights (all cases: default state,  static node, animated node).
1. Must they be non-negative? Why?
2. Must they be less than or equal to `1.0`? Why?

lexaknyazev:
Conclusion after offline discussion: no restrictions on weights.

PR #935, opened 2017-04-28
Refine morph.weights schema, allow normalized values

lexaknyazev:
See #931.


lexaknyazev:
There are two ways of normalizing signed fixed-length values (example for bytes, similar for shorts):

1. `-128 -> -1.0`, `127 -> 1.0`, `0.0` isn't representable. 
2. `-127 -> -1.0`, `127 -> 1.0`, `0 -> 0.0`. 

Which one should clients use?

lexaknyazev:
Representation of `0.0` is important for active targets selection, so clients must use the second approach.

PR #957, opened 2017-05-17
Future-proof animation.channel.target

lexaknyazev:
To maintain forward-compatibility.

### DIFF
--- a/specification/2.0/schema/animation.channel.target.schema.json
+++ b/specification/2.0/schema/animation.channel.target.schema.json
@@ -32,5 +32,5 @@
         "extensions": { },
         "extras": { }
     },
-    "required": [ "node", "path" ]
+    "required": [ "path" ]
 }


pjcozzi:
Can the doc explicitly state that the channel can be ignored if `node` or `path` are not defined?

lexaknyazev:
> Can the doc explicitly state that the channel can be ignored if `node` or `path` are not defined?

Sure. 
How to define validity for these cases?
- `node` without `path`
- `path` without `node` 

Usually, something like `path` is always present when addressing animatable target:
```javascript
[
  {
    // glTF 2.0 Core
    "node": 1,
    "path": "rotation"
  },
  {
    // possible 2.x update, animating all instances of camera 3
    "camera": 3,
    "path": "fovY"
  },
  {
    // possible 2.x update, animating an instance of camera that is used by node 2
    "node": 2,
    "path": "camera.fovY"
  },
  {
    // possible extension, animating multiple objects with one channel
    "path": "rotation",
    "extensions": {
      "EXT_multi_animation_channel": {
        "nodes": [1, 3, 6]
      }
    }
  }
]
```

The core assumption here is that one animation channel corresponds to one property. If it's true, we should keep `path` required, and allow extending only "receiver(s)".

pjcozzi:
> The core assumption here is that one animation channel corresponds to one property. If it's true, we should keep path required, and allow extending only "receiver(s)".

Sounds reasonable.

PR #958, opened 2017-05-17
Update behavior when no default scene is defined

lexaknyazev:


specification/2.0/README.md
```diff
@@ -239,7 +239,7 @@ Positive rotation is counterclockwise.
 
 The glTF asset contains zero or more *scenes*, the set of visual objects to render. Scenes are defined in a `scenes` array. An additional property, `scene` (note singular), identifies which of the scenes in the array is to be displayed at load time.
 
-When `scene` is undefined, scene of index 0 should be displayed. When `scenes` array is undefined, nothing is rendered at load time. 
+When `scene` is undefined, client shouldn't make any assumptions about default scene.
```
pjcozzi: Can this instead state that the runtime is not required to render anything and also include a non-normative sentence or two about how this is useful for treating the materials, etc. as an accessibile library to the engine as @javagl mentioned offline?

pjcozzi:
Great, thanks!

Issue #956, opened 2017-05-16
glTF 2.0: Clarify, that camera default direction is -z


McNopper:
As glTF 2.0 is not bound to WebGL anymore, we should define the default direction of the camera.
Assuming, it is -z,

pjcozzi:
OK, feel free to make a pull request to update the spec and schema into the `2.0` branch.

McNopper:
Will do so. Talk to u later.

PR #959, opened 2017-05-18
glTF 2.0: Added default setting for camera.

McNopper:
As mentioned in the telco, added the negative z-axis as default.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -1093,7 +1093,7 @@ The default material, used when a mesh does not specify a material, is defined t
 
 ## Cameras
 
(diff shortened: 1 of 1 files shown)


recp:
I think if spec specifies the lens direction (-z) then it must specify right and top too like COLLADA did:

> The camera is defined such that the local +X axis is to the right, the lens looks towards the local -Z axis, and the top of the camera is aligned with the local +Y axis...

McNopper:
Agree. I will just copy and paste the wording.

emackey:
Fixes #956

pjcozzi:
Thanks all!

Issue #456, opened 2015-10-19
Ways to contribute to glTF


pjcozzi:
We believe the true usefulness of glTF goes beyond the spec itself; it is an ecosystem of tools, documentation, and extensions contributed by the community.

We encourage everyone to contribute to the glTF community.  We are happy to list your work in this repo's main [README.md](https://github.com/KhronosGroup/glTF/blob/master/README.md).

Way to contribute include:

**Content Pipeline Tools**

* Export glTF from Blender, Maya, etc.
* Convert popular formats to glTF.
* Optimize glTF assets, e.g., texture atlas, batching, etc.
* Add glTF support to [Assimp](http://assimp.sourceforge.net/).

**Runtime**

* Simple minimal loader for WebGL, OpenGL ES, or OpenGL.
* Helper libraries for loading glTF, e.g., decompression extension.

**Documentation**

* Getting started with glTF.
* Tutorial on your experience implementing glTF.
* Size/performance of glTF compared to other formats.

**Extensions**

* Implement and propose [extensions](https://github.com/KhronosGroup/glTF/blob/master/extensions/README.md) to the core spec.

**Models**

* Contribute [sample models](https://github.com/KhronosGroup/glTF/tree/master/sampleModels) for testing.

If you want us to spread the word about your glTF work, leave a comment here or email Patrick Cozzi, pjcozzi@siggraph.org.

tparisi:
How are people going to find this document?

pjcozzi:
It is linked from the main README.md and from https://www.khronos.org/gltf

tparisi:
Excellent



On Mon, Oct 19, 2015 at 9:37 AM, Patrick Cozzi <notifications@github.com>
wrote:

> It is linked from the main README.md and from https://www.khronos.org/gltf
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/KhronosGroup/glTF/issues/456#issuecomment-149274724>.
>



-- 


Tony Parisi                             tparisi@gmail.com
Follow me on Twitter!             http://twitter.com/auradeluxe
Read my blog at                     http://www.tonyparisi.com/
Learn WebGL                         http://learningwebgl.com/
Mobile                                    415.902.8002
Skype                                     auradeluxe

Read my books!
*Learning Virtual Reality*
http://www.amazon.com/Learning-Virtual-Reality-Experiences-Applications/dp/1491922834


*Programming 3D Applications in HTML5 and
[...]

PR #960, opened 2017-05-18
Adds <gltf-viewer> to the list of viewers

AVGP:
As requested in https://github.com/KhronosGroup/glTF/issues/456#issuecomment-302443243


pjcozzi:
Thanks @AVGP!

PR #962, opened 2017-05-19
Update README with spline interpolation appendix

bghgary:


lexaknyazev:
Could you confirm accessor layouts for cubic splines?

My understanding is (for VEC3):
```
{
    vec3 in_tangent;
    vec3 value;
    vec3 out_tangent;
}
```

So, for SCALAR it would be like this:
```
{
    float in_tangent;
    float value;
    float out_tangent;
}
```

That implies that animated morph weights should be stored in the following fashion:
```
[in0, w0, out0, in1, w1, out1, in2, w2, out2, ...] x number_of_frames
```

Couple notes:
1. `accessor.min/max` are meaningless, since these fields will consider both values and tangents.
2. Tangents and values must have the same data type, so their storage could be suboptimal in some cases.

---

> Input *t*<sub>*k*</sub> and output in-tangent ***a***<sub>k</sub>, vertex ***v***<sub>*k*</sub>, and out-tangent ***b***<sub>k</sub> for *k* = 1,...,*n*

Assuming that we're using 1-based indexing here, ***a***<sub>1</sub> and ***b***<sub>n</sub> are ignored. Could we explicitly say that these two tangents should be zeros?

lexaknyazev:
On Catmull-Rom splines:

> Input *t*<sub>*k*</sub> and output vertex ***v***<sub>*k*</sub> for *k* = 1,...,*n* where the spline starts at *t*<sub>2</sub> and ends at *t*<sub>*n*-1</sub>

I would strongly prefer to keep animation input definition, and say instead that `output` accessor has two more elements than `input`.

1. Input's data must be monotonic, so current definition prevents using `0.0` as first frame's time.
2. All other interpolation methods treat number of `input` elements as the number of keyframes.

bghgary:
Cubic Splines

> `accessor.min/max` are meaningless, since these fields will consider both values and tangents.

Yes, very true. Min/max are also weird for matrices. Do you have a suggestion here?

> Assuming that we're using 1-based indexing here, a1 and bn are ignored. Could we explicitly say that these two tangents should be zeros?

We can do that. We can also not specify them. It's slightly harder to describe and slightly harder to implement, but should be okay. What do you think?

Catmull-Rom Splines

> I would strongly prefer to keep animation input definition, and say instead that output accessor has two more elements than input.

I started writing it that way, but then realized the math requires *t*<sub>1</sub> and *t*<sub>*n*</sub>:
> ***m***<sub>*k*</sub> = (***v***<sub>*k*+1</sub> - ***v***<sub>*k*-1</sub>) / (*t*<sub>*k*+1</sub> - *t*<sub>*k*-1</sub>) for *k* = 2,...,*n*-1

I think we can store/auto-calculate the start and end tangents instead to avoid this. What do you think?

lexaknyazev:
> Yes, very true. Min/max are also weird for matrices. Do you have a suggestion here?

They were optional in 1.0, then they became mandatory in 1.1. 
@pjcozzi @lasalvavida do you remember reasons behind that decision?

>> Assuming that we're using 1-based indexing here, a1 and bn are ignored. Could we explicitly say that these two tangents should be zeros?

> We can do that. We can also not specify them. It's slightly harder to describe and slightly harder to implement, but should be okay.

I'd keep them in case of loops support in future. Setting them to zero for now seems to be cleaner than making reading more complex.

> I think we can store/auto-calculate the start and end tangents instead to avoid this. What do you think?

My only concern is about the number of `input` accessor elements. Equations could be tweaked to accommodate this. Otherwise, will auto-calculating tangents sacrifice flexibility?

bghgary:
> Equations could be tweaked to accommodate this.

Do you have a proposal?

> Will auto-calculating tangents sacrifice flexibility?

Yes, but you can always fall back to stored tangents.

pjcozzi:
>> Yes, very true. Min/max are also weird for matrices. Do you have a suggestion here?

> They were optional in 1.0, then they became mandatory in 1.1.
@pjcozzi @lasalvavida do you remember reasons behind that decision?

It could have been because most engines wanted to use them for bounding volumes, and making them required avoids having to have a fallback path.

We might be able to lessen the requirement so that, for example, they are only required for positions.

lasalvavida:
> They were optional in 1.0, then they became mandatory in 1.1.
@pjcozzi @lasalvavida do you remember reasons behind that decision?

We were going to make them required only for accessors corresponding to a `POSITION` semantic for bounding volumes, but it was decided that this was a bit of a complex requirement and that it was easier to just require them for all accessors.

As a side note, it also makes it easier to compute `WEB3D_quantized_attributes` when the mins and maxes are present since you don't need to compute them.

lexaknyazev:
All accessors usages in glTF 2.0 and reasons for `min/max` at runtime:
- Vertex Attributes
  - `POSITION` - local axis-aligned bounding box
  - `NORMAL`, `TANGENT` - ?
  - `TEXCOORD_` - ?
  - `COLOR_` - ?
  - `JOINTS_` - `max` could be used in some cases to limit the number of loaded joint transforms (in rare case when two different meshes share the same IBM data).
  - `WEIGHTS_` - ?
- Mesh Primitive Indices - `max` could be used to check against hardware caps.
- Relative Morph Attributes
  - `POSITION` - morphed local axis-aligned bounding box
  - `NORMAL`, `TANGENT` - ?
- Inverse Bind Matrices - ?
- Animation Sampler
  - `input` - `min` and `max` could be used to sort animation segments
  - `output`
    - with `LINEAR` or `STEP` interpolations, `min` and `max` (combined with meshes bounds) could be used to estimate bounding box of animation segment
    - with `CUBICSPLINE` or `CATMULLROMSPLINE` it's no longer the case because accessor's data is heterogeneous.

---

I'm not sure what can we change at this point, but I see two possible optimizations:
- Require `min/max` only for positions (both static and morphed) and animation inputs. 
- Don't store tangents in the same accessor as values (there're more than one reason for it: `min`/`max` values, storage format, Catmull-Rom indexing). If so, `min` and `max` should be required for animation outputs as well.

---

From validation point-of-view there could be different perspectives on what is valid. E.g. look at `NORMALS` accessor:
- `min` and `max` lie within `[-1.0; 1.0]`, binary data is fine, asset is valid.

- `min` or `max` have invalid values and binary data is wrong. asset is invalid.

- `min` and `max` lie within `[-1.0; 1.0]`, however actual data doesn't. That means that `accessor` JSON object itself is fine, but an asset is invalid, because it will likely have rendering issues.

- `min` or `max` have invalid values, but binary data is fine. These limits have no use for normals, so invalid properties will be ignored and rendering will be okay. So, strictly speaking, such asset may have a validation warning, but it's a bit misleading to name an asset invalid.

bghgary:
@lexaknyazev Nice summary!

> Don't store tangents in the same accessor as values (there're more than one reason for it: min/max values, storage format, Catmull-Rom indexing). If so, min and max should be required for animation outputs as well.

The min/max values for splines keyframe outputs might not be as useful as it sounds. Splines can easily go outside of the bounds of just the keyframes, especially when tangents are specified.

lexaknyazev:
> Splines can easily go outside of the bounds of just the keyframes, especially when tangents are specified.

So accessor's `min`/`max` make no sense for splines at all. That seems to be a valid reason to make them optional for animation outputs.

From the list above, it looks like these fields have no runtime use in most cases. Exporters could write them anyway if they want.

bghgary:
@lexaknyazev I've updated the PR with the changes as discussed:
* Added implementation note that first in-tangent and last out-tangent should be zeros.
* Changed Catmull-Rom splines to use start and end tangents in the outputs without additional inputs.
* Rephrased how the required number of outputs in the schema description.

Issue #945, opened 2017-05-07
glTF 2.0: New KHR_lights extension


McNopper:
As the lights have to be extracted from the KHR_materials_common extension, here is a first try by example:

```
{
  "lights": [
    {
      "ambient": {
        "color": [
          1,
          1,
          1
        ]
      }
    },
    {
      "directional": {
        "color": [
          1,
          1,
          1
        ]
      }
    },
    {
      "point": {
        "color": [
          1,
          1,
          1
        ],
        "constantAttenuation": 1,
        "linearAttenuation": 0.5,
        "quadraticAttenuation": 0.25
      }
    },
    {
      "spot": {
        "color": [
          1,
          1,
          1
        ],
        "constantAttenuation": 1,
        "fallOffAngle": 40,
        "fallOffExponent": 0,
        "linearAttenuation": 0.5,
        "quadraticAttenuation": 0.25
      }
    }
  ]
}
```

Directional, point and spot are obvious for PBR and non-PBR materials.

Regarding ambient and for PBR, this needs to be further discussed and explained. But I suggest to treat the value as if an environment map is used, which does have just this color.

McNopper:
In general, a node will have no or one entry of a light with its index.
Exception is, that no ambient light is allowed to be assigned to a node.

An ambient light can only be assigned to a scene.
In opposite, for a scene, only the ambient light can be assigend.

Is this approach accepted?

lexaknyazev:
> An ambient light can only be assigned to a scene.

Lgtm. Ambient light has no transform to apply.
Also, add required `light.type` string property (as [`camera.type`](https://github.com/KhronosGroup/glTF/blob/2.0/specification/2.0/schema/camera.schema.json#L16-L30)).

McNopper:
I will implement this extension. If it works out fine, I will create a schema and bring it back to the glTF working group.

McNopper:
```
"extensions" : {
        "KHR_lights" : {
            "lights" : [
                {
                    "color" : [
                        1.0, 
                        1.0, 
                        1.0
                    ], 
                    "name" : "Directional", 
                    "type" : "directional"
                }, 
                {
                    "color" : [
                        1.0, 
                        1.0, 
                        1.0
                    ], 
                    "name" : "Point", 
                    "point" : {
                        "quadraticAttenuation" : 1.0
                    }, 
                    "type" : "point"
                }, 
                {
                    "color" : [
                        1.0, 
                        1.0, 
                        1.0
[...]

Issue #947, opened 2017-05-07
glTF 2.0: New/modified KHR_materials_common extension


McNopper:
Using the pbrMetallicRoughness and pbrSpecularGlossiness material "syntax", I have adapted the common materials. Again here by example:

```
{
  "materials": [
    {
      "commonConstant": {
        "ambientFactor": [
          0.2,
          0.2,
          0.2
        ]
      }
    },
    {
      "commonLambert": {
        "ambientFactor": [
          0.2,
          0.2,
          0.2
        ],
        "diffuseFactor": [
          1,
          1,
          1,
          1
        ],
        "diffuseTexture": 0
      }
    },
    {
      "commonPhong": {
        "ambientFactor": [
          0.2,
          0.2,
          0.2
        ],
        "diffuseFactor": [
          1,
          1,
          1,
          1
        ],
        "diffuseTexture": 0,
        "specularFactor": [
          1,
          1,
          1,
          1
        ],
        "specularTexture": 1,
        "shininessFactor": 0.5,
        "shininessTexture": 2
      }
    },
    {
      "commonBlinn": {
        "ambientFactor": [
          0.2,
          0.2,
          0.2
        ],
        "diffuseFactor": [
          1,
          1,
          1,
          1
        ],
        "diffuseTexture": 0,
        "specularFactor": [
          1,
          1,
          1,
          1
        ],
        "specularTexture": 1,
        "shininessFactor": 0.5,
        "shininessTexture": 2
      }
    }
  ]
}
```

Like the current PBR materials, they extend the "base" materials. So e.g. emissive would be inherited.

stevenvergenz:
@McNopper Is this update work happening in public anywhere? I'm looking to implement these types in the UnityGLTF project, as they seem easier to work with on mobile than pbrMetallicRoughness.

McNopper:
It is happening in this thread. As soon as glTF 2.0 is final, the working group will get back to extensions and issues.I am currently implementing the materials, lights in an experimental way, that we do have a base to discuss about.Regarding the lightmap, I suggest u define an extension and provide an example and/or schema. This could also be used as a start to discuss about.
Von meinem Samsung Galaxy Smartphone gesendet.
[...]

Issue #824, opened 2017-01-26
Spotlight falloff_angle


recp:
I'm sorry to bringing a COLLADA issue to here but since glTF has same attribute it's good to hear from glTF team. Because I really tried to get an answer from COLLADA side and on resources (books, online resources, collada forum, open sources...) but there is no answer :/

Actually it seems my renderer renders fine as others (Blender, FBXReview...) with an assumption which is that falloff_angle is full angle of spot shape.

This is my spotlight GLSL func (returns attenuation) 
```glsl
...
 
float
spot(Light light, inout vec3 l) {
  float dist;
  float spotCos;

  l    = l - vPosition;
  dist = length(l);
  l    = l / dist;

  spotCos = dot(light.direction, -l);
  if (spotCos < light.cutoffCos)
    return 0.0;

  return pow(spotCos, light.cutoffExp) / (light.constAttn
                                           + light.linAttn * dist
                                           + light.quadAttn * dist * dist);
}
...
```

Currently I'm sending cosf(rad(falloff_angle/2.0f)) as cutoffCos then everythings seems perfect! COLLADA specs say nothing about this param is full angle or not or how to process in shader I mean a formula like this (https://msdn.microsoft.com/en-us/library/windows/desktop/bb172279(v=vs.85).aspx): 

<img src="https://i-msdn.sec.s-msft.com/dynimg/IC511429.png" />

Blender:
```C
  light.spot_cutoff = RAD2DEGF(la->spotsize * 0.5f);
  light.spot_exponent = 128.0f * la->spotblend;
```
Also I check cesium sources and it seems cesium also uses falloff_angle/2.0 in shader:
[cesium modelMaterialsCommon.js#L401](https://github.com/AnalyticalGraphicsInc/cesium/blob/8328880c6a50e4fb735f68f8b87338eb477efb1f/Source/Scene/modelMaterialsCommon.js#L401)

My shader works fine with comparing falloff_angle/2.0 with dot product but it would be nice to get   confirmation

Thanks

javagl:
Parts of the sketched code are not entirely clear (for me), regarding the inputs, and how the output will be used. 

However, I think that the definitions go back to the ancient definitions of OpenGL 1.x. And looking at the [Spotlights section in the redbook](http://www.glprogramming.com/red/chapter05.html#name11), the cutoff is indeed supposed to be *half* of the total opening angle of the cone:

![image](http://www.glprogramming.com/red/images/Image79.gif)

[...]

PR #965, opened 2017-05-19
Proposed 2.0 materials common

lasalvavida:
This is  my perspective on how to update the materials common extension. 

* Drop lights out (it should be a separate KHR_lights extension as in #945)
* Use the existing pbr and pbrSpecularGlossiness properties to compute the specular lighting parameters in the context of the lighting models
* Very simple extension, just define a technique, and use the existing material to do the rest
* enum for techniques to be more in line with the rest of the spec

```javascript
{
    "materials" : [{
        ...
        "extensions" : {
            "KHR_materials_common" : {
                "technique": 0 // Blinn
            }
        }
    }]
}
```

Feedback appreciated

### DIFF
--- a/extensions/Khronos/KHR_materials_common/README.md
+++ b/extensions/Khronos/KHR_materials_common/README.md
@@ -179,8 +179,7 @@ Constant shading does not reflect light sources in the scene; therefore the comm
 
 ### Properties Reference
 
-* [`KHR_materials_common glTF extension`](#reference-khr_materials_common-gltf-e
-xtension) (root object)
+* [`KHR_materials_common glTF extension`](#reference-khr_materials_common-gltf-extension) (root object)
 
 
 ---------------------------------------


pjcozzi:
Thanks @lasalvavida!

I think @McNopper would be a good person to review and collaborate with here.

McNopper:
Please let us optimze/finalize https://github.com/KhronosGroup/glTF/issues/947, which will become the offical KHR_materials_common extension sooner or later. This will be a "real" Phong, Blinn etc. material model, with the well known parameters which already do exist.

As far as I know, @bghgary is working on some math, where you can make out of Phong etc. a PBR material. As far as I understood your suggestion, you are doing the opposite way: Making out of a PBR material a Phong etc. material.

I think both directions do make sense in a manner of compatibility e.g. an engine does not support a specific renderer.
But a quality loss (PBR -> Phong) is to be expected and the impact should also be discussed in the glTF working group.

If we follow this path, I suggest we define an extension e.g. KHR_materials_conversion, which allows to convert between materials.

pjcozzi:
Also see #824 for discussion on spotlight falloff angle.

pjcozzi:
@McNopper OK to close as duplicate with #947 and #945?

PR #966, opened 2017-05-19
Remove schema examples

bghgary:
These are outdated. We can also add them back if necessary.


lexaknyazev:
Agree with removal.

pjcozzi:
Also agreed.  I don't know if anyone found these useful.  They just helped me validate the 1.0 schema when I was first creating it.

PR #969, opened 2017-05-21
add redcube.js to viewers

Reon90:
Could you add Redcube.js to "Loaders and Viewers" section?
Thank you.


pjcozzi:
Very nice, thanks @Reon90!

PR #971, opened 2017-05-21
Soft enforcement of unit quaternions for node.rotation

lexaknyazev:
Sadly, we can't check quaternion's length with schema, but it's still better than nothing.

### DIFF
--- a/specification/2.0/schema/node.schema.json
+++ b/specification/2.0/schema/node.schema.json
@@ -42,7 +42,9 @@
             "type": "array",
             "description": "The node's unit quaternion rotation in the order (x, y, z, w), where w is the scalar.",
             "items": {
-                "type": "number"
+                "type": "number",
+                "minimum": -1.0,
+                "maximum": 1.0
             },
             "minItems": 4,
             "maxItems": 4,


lexaknyazev:
Looks like we should also restrict values of `node.scale` so that it can't contain zeros.

pjcozzi:
> Looks like we should also restrict values of node.scale so that it can't contain zeros.

Although it isn't typical, that is still a valid scale and will create a valid transform matrix, right?

lexaknyazev:
> will create a valid transform matrix, right?

That depends on how we define "valid transform matrix". Scale of `0.0` will produce transform matrix with zero determinant.

bghgary:
> Scale of `0.0` will produce transform matrix with zero determinant.

Is this going to cause an issue? I can easily imagine animations changing the scale to zero for a number of scenarios. They might not have the correct result if it's not allowed. Even if it did have the correct result, it would be a pain to have to tweak the animation to scale to something close to zero.

pjcozzi:
We could add a non-normative/implementation note that states transform matrices may have a zero scale and therefore may will not be invertible in these cases.

lexaknyazev:
> Is this going to cause an issue?

Strictly speaking, such matrix will not be decomposable to TRS (btw, we require this from `node.matrix`). This will also apply to all children's transformation matrices of a node with zero scale.

> ... and therefore may will not be invertible in these cases.

So engines won't be able to get normal matrix, right?

bghgary:
> Strictly speaking, such matrix will not be decomposable to TRS (btw, we require this from node.matrix). This will also apply to all children's transformation matrices of a node with zero scale.

True, but is this a problem? Animations can only target TRS anyways. The matrix form will be unable to specify a scale of zero.

pjcozzi:
>> ... and therefore may will not be invertible in these cases.
> So engines won't be able to get normal matrix, right?

In that case, right.

lexaknyazev:
In that case, we should properly describe expected behavior.

Let's say that before applying scale, model looks like this (just a tetrahedron):
image

After scaling Y positions to zero, proper normals still exist but getting them without normal matrix could be difficult and likely not all engines will bother.
image

lexaknyazev:
Since `scale` can be negative, we can't definitely say what orientations should normals have in case of zero scale. That's not a big deal for animations, because engines can determine orientation from adjacent keyframes. 

But static transforms are ambiguous.

emackey:
Is there any reasonable way to allow scaling to `[0, 0, 0]` but forbid scaling individual axes to 0?

bghgary:
This is what Unity does when only one of the scale component is zero.

Original scale (1, 1, 1)
image

X scale is zero (0, 1, 1)
image

lexaknyazev:
Should this be expected behavior for other renderers? 

Other options include:
- rendering is undefined;
- mesh is not rendered;
- engine uses very small value instead of `0.0`, tries to deduce its sign.

emackey:
> * rendering is undefined

Sounds OK.  Remember it's not just normal vectors that are undefined:  All of the polygons become co-planar, so depth buffering may be undefined as well.

pjcozzi:
Thanks for looking into Unity, @bghgary!  I suspect this was not designed intentionally in Unity.

> rendering is undefined;

Could be OK, but we try to minimize undefined behavior, so why not go with

> mesh is not rendered;

unless we have a good use case for this?  Paper Mario? 😄

bghgary:
BabylonJS does this if I force the x-scale to zero:

image

Note the z-fighting with the triangles.

I think whatever we do here, it will be an implementation note. I don't think we should restrict this in the spec. I would probably leave it as undefined. Practically, I doubt anyone will do this.

@snagy and I just noticed another thing we probably should specify in the spec. When an odd number of scale components are negative, the faces of the triangles should be flipped. This works correctly in Unity but not currently in BabylonJS. Making the scale negative is useful for mirroring.

lexaknyazev:
> When an odd number of scale components are negative, the faces of the triangles should be flipped. 

We assume here, that mirroring is more usable, than flipping object inside-out, right?
Also, I'm wondering how should it work for nested transforms, which could be defined by both `matrix` and `scale` (in different nodes, of course).

bghgary:
> We assume here, that mirroring is more usable, than flipping object inside-out, right?

Yes. I suppose there might be a use case for double-sided materials, but it will be uncommon.

> nested transforms

My understanding is that the determinant of the matrix can be used to decide whether to flip the faces or not. Chapter 3.1.2 of [Mathematics for 3D Game Programming and Computer Graphics by Eric Lengyel](https://www.amazon.com/dp/1435458869/?tag=terathon-20) describes this. The gist of it is that if the determinant is negative, then the handedness is reversed, which means the faces should be flipped. @snagy checked [Unreal's implementation](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Private/SkeletalMesh.cpp#L5497) and that's what it does.

lexaknyazev:
Relying on determinant sounds good (I was a bit confused by counting the number of negative scale components).

Could you please update this PR with a note on these topics (zero/negative determinant)?

bghgary:
I created a separate PR. Hope that's okay.

PR #975, opened 2017-05-24
Generalize accessor.min/max data type restrictions.

lexaknyazev:
From https://github.com/KhronosGroup/glTF-Sample-Models/issues/63.

CC @javagl

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -52,6 +52,7 @@ Copyright (C) 2013-2017 The Khronos Group Inc. All Rights Reserved. glTF is a tr
     * [Accessors](#accessors)
         * [Floating-Point Data](#floating-point-data)
         * [Accessor Element Size](#accessor-element-size)
+        * [Accessors Bounds](#accessors-bounds)
         * [Sparse Accessors](#sparse-accessors)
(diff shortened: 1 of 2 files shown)


javagl:
As discussed offline: The statement

> Array elements must be treated as having the same data type as accessor's `componentType`

is IMHO clearer than the original one (about the special `float`/`double` case). 

The *implications* and the meaning of the sentence may not be immediately clear for everyone (and the differences may be as subtle as a 0.00000000321 difference between two values...). I think you somewhere mentioned this special JavaScript function that can be used to "truncate" the `double` value that JavaScript originally reads into a proper `float` value. Will this still be mentioned in the spec text, maybe as an "Implementation Note" in https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#accessors ?

( The part in https://github.com/KhronosGroup/glTF/tree/2.0/specification/2.0#accessormax-white_check_mark will probably be updated automatically ...)

lexaknyazev:
Integrated all feedback on this topic.
Also added https://github.com/KhronosGroup/glTF/pull/957#issuecomment-302236138.

@pjcozzi Please merge, if language is OK.

PR #977, opened 2017-05-24
Added implementation note about morph target counts

javagl:
@lexaknyazev I hope this is right: The limit in GL 1.0 implies 

- 8 morph targets with *one* attribute **OR**
- 4 morph targets with *two* attributes **OR**
- 2 morph targets with *three* attributes

Also, someone please drop a note to the one who advocated for an unlimited number of morph targets (I'm not sure who it was)

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -754,7 +754,7 @@ The following example extends the Mesh defined in the previous example to a morp
 
 After applying morph targets to vertex positions and normals, tangent space may need to be recalculated. See [Appendix B](#appendix-b-tangent-space-recalculation) for details.
 
-> **Implementation note:** The number of morph targets is not limited in glTF. A conformant renderer must support at least eight morphed attributes. This means that it has to support at least eight morph targets that contain a `POSITION` attribute, or four morph targets that contain a `POSITION` and a `NORMAL` attribute, or two morph targets that contain `POSITION`, `NORMAL` and `TANGENT` attributes. For assets that contain a higher number of morphed attributes, renderers may choose to either fully support them (for example, by performing the morph computations in software), or to only use the eight attributes of the morph targets with the highest weights. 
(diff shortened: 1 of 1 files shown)


pjcozzi:
Thanks @javagl!

@donmccurdy @bghgary as devs who have actually implemented morph targets, are you good with this?

@emilian0 as the spec creator for morph targets, what do you think?

specification/2.0/README.md
```diff
@@ -754,6 +754,9 @@ The following example extends the Mesh defined in the previous example to a morp
 
 After applying morph targets to vertex positions and normals, tangent space may need to be recalculated. See [Appendix B](#appendix-b-tangent-space-recalculation) for details.
 
+> **Implementation note:** The number of morph targets is not limited in glTF. A conformant renderer must support at least eight morphed attributes. This means that it has to support at least eight morph targets that contain a `POSITION` attribute, or four morph targets that contain a `POSITION` and a `NORMAL` attribute, or two morph targets that contain `POSITION`, `NORMAL` and `TANGENT` attributes. For assets that contain a higher number of morphed attributes, renderers may choose to either fully support them (for example, by performing the morph computations in software), or to only use the eight attributes of the morph targets with the highest weights.
```
bghgary: I think we should change "renderer" to "client" since it might not just be renderers that have this issue. Everywhere else uses "client" or "implementation" except for one specific case where we do mean specifically renderers.

donmccurdy:
Looks good to me. 

/cc @takahirox

takahirox:
I'm the dev who have implemented Three.js glTF loader morph target support.

Yup, this limitation is good to me.
Indeed Three.js has similar limitation.
- 8 morph targets with position attribute OR
- 4 morph targets with position and normal attributes

(We don't support tangent now)

pjcozzi:
Thanks @takahirox!

PR #980, opened 2017-05-29
Add couple notes on image.mimeType

lexaknyazev:
No schema/conformance/validation changes. Just a clarification on expected client behavior.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -939,7 +939,7 @@ Images referred to by textures are stored in the `images` array of the asset.
 Each image contains one of
 - a URI to an external file in one of the supported images formats, or
 - a URI with embedded base64-encoded data, or
-- a reference to a `bufferView`.
+- a reference to a `bufferView`; in that case `mimeType` must be defined.
 
 The following example shows an image pointing to an external PNG image file and another image referencing a `bufferView` with JPEG data.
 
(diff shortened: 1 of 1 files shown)


specification/2.0/README.md
```diff
@@ -956,6 +956,8 @@ The following example shows an image pointing to an external PNG image file and
     ]
 }
 ```
+> **Implementation Note:** When image data is provided by `uri` and `mimeType` is defined, clients should prefer JSON-defined MIME Type over one provided by transport layer.
```
bghgary: Is this easy to implement? Do we consider this to be valid glTF? I would have expected mimeType to be ignored if both `uri` and `mimeType` are specified.

lexaknyazev: > Is this easy to implement?

I don't see any problem with implementing that.

> Do we consider this to be valid glTF? 

Currently, yes. Otherwise, we should change the schema.

>  I would have expected `mimeType` to be ignored if both `uri` and `mimeType` are specified.

There could be a rare case, when `uri` doesn't contain any file extension and server doesn't give meaningful `Content-Type` header value. So JSON-provided `mimeType` could be used instead of guessing image format from magic numbers.

bghgary: Okay. I only asked because I haven't seen any implementations that take `mimeType` into account when `uri` is specified.

PR #985, opened 2017-05-30
Add implementation note for negative transform determinant

bghgary:
As requested in https://github.com/KhronosGroup/glTF/pull/971#issuecomment-304392357

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -315,6 +315,8 @@ TRS properties are converted to matrices and postmultiplied in the `T * R * S` o
 
 When a node is targeted for animation (referenced by an `animation.channel.target`), only TRS properties may be present; `matrix` will not be present. 
 
+> **Implementation Note:** If the determinant of the transform is a negative value, the winding order of the mesh triangle faces should be reversed. This supports negative scales for mirroring geometry.
+
 In the example below, node named `Box` defines non-default rotation and translation.
 
 ```json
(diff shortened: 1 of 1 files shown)


lexaknyazev:
What if animated `scale` goes from positive to negative value? Matrix transform could have (near) zero determinant in a frame or two. In that case, Babylon's approach (z-fighting) may be more preferable than Unity's (black geometry).

pjcozzi:
> Babylon's approach (z-fighting) may be more preferable than Unity's (black geometry).

Seems like a hard thing to get engines to agree on.  I wonder how much of this engine behavior is by accident; I don't know if we should copy it.  Perhaps leave unspecified. :fearful:

lexaknyazev:
@bghgary 
I've added a note on non-invertible transformations. Please merge if everything is fine.

zellski:
Just for clarification -- by spec, generated glTF should, without question, reverse winding order whenever it detects negative scaling either through static transform or an animation path? I was a bit taken aback by this, because I figured it's so common for applications to dynamically mutate these transforms later... that whichever transform is embedded in the glTF is not necessarily so intrinsic that it should affect a geometry rearrangement. It seems like it could foist some unwanted astonishment onto the user. But maybe negative scale is such a specific thing that it's warranted. So it would make sense for a FBX->glTF converter to detect this situation and compensate as suggested?

bghgary:
> by spec, generated glTF should, without question, reverse winding order whenever it detects negative scaling either through static transform or an animation path?

Hmm, perhaps the implementation note should be clarified. The intent is that this is the runtime behavior of a client implementation (e.g. a rendering pipeline).

> So it would make sense for a FBX->glTF converter to detect this situation and compensate as suggested?

I would not expect converters to do this.

zellski:
Thanks @bghgary. Chances are I'm just easily confused and the spec language is fine. I'm relieved to hear you say this, though. It felt like a strange operation for a static converter to perform. If I add it I'll hide it behind a flag.

PR #990, opened 2017-05-31
Small edits of URIs section

lexaknyazev:
Split normative/non-normative parts of URIs section, disallow empty URIs in schema.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -168,11 +168,13 @@ To simplify client-side implementation, glTF has following restrictions on JSON 
 
 ## URIs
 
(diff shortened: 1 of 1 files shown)


specification/2.0/schema/buffer.schema.json
```diff
@@ -9,6 +9,7 @@
             "type": "string",
             "description": "The uri of the buffer.",
             "format": "uriref",
+            "minLength": 1,
```
bghgary: This is a schema change. I think we wanted to make sure the schema is locked?

lexaknyazev: I think that such restriction was always implicit:
- `uri` is an _external_ reference;
- empty string URI points to the glTF file itself.

Following this logic, a valid glTF asset can't have empty string URIs.

If making such schema adjustment is impossible, we should document in the readme how to treat empty URIs.

bghgary: CC @sbtron 

Our opinion is that this change isn't necessary. As you say an empty string URI points to the glTF file itself (which isn't valid), but you can also put a URI to some other random file (whether glTF or not) and that's also not valid. Clients must deal with this whether or not we allow empty URIs.

lexaknyazev:
@bghgary 
I've reverted schema changes. Are you OK with readme edits?

bghgary:
Yes.

PR #997, opened 2017-06-04
Regenerated property reference

sbtron:
Regenerated reference based on latest Wetzel and schema

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -1427,32 +1427,31 @@ For more information on glTF extensions, consult the [extensions registry specif
 
 # Properties Reference
 
-# Objects
+## Objects
 * [`accessor`](#reference-accessor)
-  * [`sparse`](#reference-sparse)
-    * [`values`](#reference-values)
-    * [`indices`](#reference-indices)
+   * [`sparse`](#reference-sparse)
+      * [`indices`](#reference-indices)
+      * [`values`](#reference-values)
 * [`animation`](#reference-animation)
-  * [`channel`](#reference-channel)
-    * [`target`](#reference-target)
-  * [`animation sampler`](#reference-animation-sampler)
+   * [`animation sampler`](#reference-animation-sampler)
+   * [`channel`](#reference-channel)
+      * [`target`](#reference-target)
 * [`asset`](#reference-asset)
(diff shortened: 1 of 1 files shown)


lexaknyazev:
Couple notes:
- Reference sections have different order than ToC list (e.g. `sparse.indices` are located after `images`). Was it intentional?
- [`animation.sampler`](https://github.com/KhronosGroup/glTF/blob/2.0/specification/2.0/schema/animation.sampler.schema.json#L3) has wrong title (compare to [`animation.channel`](https://github.com/KhronosGroup/glTF/blob/2.0/specification/2.0/schema/animation.channel.schema.json#L3)).

sbtron:
The generator isn't smart enough to lay out the paragraphs correctly. I believe it was the same earlier we will have to tweak it manually. 
animation sampler title seems like a json schema issue?

emackey:
> The generator isn't smart enough to lay out the paragraphs correctly

Not sure I follow.  What needs to happen to the generator?

lexaknyazev:
> animation sampler title seems like a json schema issue

Yes, schema file should be tweaked.

pjcozzi:
> Reference sections have different order than ToC list (e.g. sparse.indices are located after images). Was it intentional?

Probably an oversight with manual edits.

pjcozzi:
> I believe it was the same earlier we will have to tweak it manually.

Yes, I did a lot of hand editing for the 1.0 spec.

emackey:
Also, next time this is re-run, try adding the `-a` (`--autoLink`) option, such that intra-schema references are linked together.

lexaknyazev:
@sbtron 
JSON title fixed with cad4e6f1642fe9348ababfd7327b4245fb96e5e8.

lexaknyazev:
@sbtron @emackey 

Could Wetzel generate full names for nested objects? This would fix both ordering issue as well as having two `samplers`:

- `accessor`
  - `accessor.sparse`
     - `accessor.sparse.indices`
     - `accessor.sparse.values`
- `animation`
  - `animation.sampler`
  - `animation.channel`
      - `animation.channel.target`

Otherwise, we could manually change all titles in schema files.

emackey:
This could eventually happen, it won't be immediate though.

PR #1001, opened 2017-06-05
Cleanup links for gltf 2.0

sbtron:
1) Fix link to 2.0 spec on landing page
2) Move overview image from landing page to spec links page since its 1.0 specific
3) Update extension registry and fix links to individual extension specs
4) Mark specgloss extension as complete
5) Add note to binary extension that it is only needed in gltf 1.0


pjcozzi:
Great, thanks @sbtron!

Issue #995, opened 2017-06-03
Storage of vertex attributes in sparse accessors


lexaknyazev:
Should we restrict usage of sparse accessors to morph targets and animation data only? E.g., should it be possible to use them as a source of static vertex data?

Consider this:
```
mesh 0
  POSITIONS -> accessor 0
  NORMALS -> accessor 1
mesh 1
  POSITIONS -> accessor 2
  NORMALS -> accessor 3

accessor 0 // positions for the first mesh
  bufferView -> 0
accessor 1 // normals for the first mesh
  bufferView -> 0
accessor 2 // positions for the second mesh
  bufferView -> 0
  sparse
    indices -> bufferView 1
    values -> bufferView 1
accessor 3 // normals for the second mesh
  bufferView -> 0
  sparse
    indices -> bufferView 1
    values -> bufferView 1

bufferView 0 // positions and normals for the first mesh
  byteStride // stride is defined because it's raw vertex data used by more than one accessor; vertex attributes could be interleaved
bufferView 1 // sparse data for the second mesh
```

The only drawback is that implementations will be required to support software (non-GPU) reading of (possibly interleaved) vertex data.

At the moment, spec doesn't define valid use cases for sparse accessors.

CC @pjcozzi @bghgary

pjcozzi:
To keep runtimes simple, we could restrict in 2.0 and loosen in a future version if use cases beyond morph targets and animation come up.

lexaknyazev:
> To keep runtimes simple

This part of implementation doesn't seem to be very difficult (compared to other 2.0 stuff).

pjcozzi:
If there's no use case, I still wouldn't require it even if it is easy.

Your call.

lexaknyazev:
I still think that we should keep this functionality, but we need to make docs a bit more explicit.

---

Let's start with vectors:
**VEC3, UNSIGNED_BYTE, 4 elements**
Base accessor (each value is aligned to 4-byte boundaries, i.e. `bufferView.byteStride` equals 4):
```
|00|01|02|03|04|05|06|07|08|09|0A|0B|0C|0D|0E|0F|
|==|==|==|==|==|==|==|==|==|==|==|==|==|==|==|==|
|vx|vy|vz|--|vx|vy|vz|--|vx|vy|vz|--|vx|vy|vz|--|
```
[...]

PR #1006, opened 2017-06-09
Elaborate on matrix data alignment

lexaknyazev:
See https://github.com/KhronosGroup/glTF/issues/995#issuecomment-306928016.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -624,7 +624,32 @@ When `byteStride` of referenced `bufferView` is not defined, it means that acces
 
 Each `accessor` must fit its `bufferView`, i.e., `accessor.byteOffset + STRIDE * (accessor.count - 1) + SIZE_OF_ELEMENT` must be less than or equal to `bufferView.length`.
 
-For performance and compatibility reasons, vertex attributes must be aligned to 4-byte boundaries inside `bufferView` (i.e., `accessor.byteOffset` and `bufferView.byteStride` must be multiples of 4). Accessors of matrix type have data stored in column-major order; start of each column must be aligned to 4-byte boundaries.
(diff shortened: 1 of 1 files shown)


specification/2.0/README.md
```diff
@@ -624,7 +624,32 @@ When `byteStride` of referenced `bufferView` is not defined, it means that acces
...
+```
+
+**MAT3, 2-byte components**
+```
+| 00| 01| 02| 03| 04| 05| 06| 07| 08| 09| 0A| 0B| 0C| 0D| 0E| 0F| 10| 11| 12| 13| 14| 15| 16| 17|
+|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|
+|m00|m00|m10|m10|m20|m20|---|---|m01|m01|m11|m11|m21|m21|---|---|m02|m02|m12|m12|m22|m22|---|---|
+```
+
+Alignment requirements apply only to start of each column, so trailing zeros could be omitted if there's no further data.
```
bghgary: > trailing zeros

This is the first mention of padding with zeros. Is there some text somewhere that indicates they have to be zero padded?

lexaknyazev: I don't think so. We could add this or change "zeros" to "bytes". Anyway, exact padding values don't really matter.

That sentence means that, e.g., last matrix within buffer view could be written without last two bytes:
```
| 00| 01| 02| 03| 04| 05| 06| 07| 08| 09| 0A| 0B| 0C| 0D| 0E| 0F| 10| 11| 12| 13| 14| 15|
|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|===|
|m00|m00|m10|m10|m20|m20|---|---|m01|m01|m11|m11|m21|m21|---|---|m02|m02|m12|m12|m22|m22|
```

bghgary: I understood what you meant. I would go with this:

> We could add this or change "zeros" to "bytes".

PR #1007, opened 2017-06-10
Add note about zero xmag/ymag

lexaknyazev:
We can't disallow such case with schema, but it's worth noting.

### DIFF
--- a/specification/2.0/schema/camera.orthographic.schema.json
+++ b/specification/2.0/schema/camera.orthographic.schema.json
@@ -7,11 +7,11 @@
     "properties": {
         "xmag": {
             "type": "number",
-            "description": "The floating-point horizontal magnification of the view."
+            "description": "The floating-point horizontal magnification of the view. Must not be zero."
         },
         "ymag": {
             "type": "number",
-            "description": "The floating-point vertical magnification of the view."
+            "description": "The floating-point vertical magnification of the view. Must not be zero."
         },
         "zfar": {
             "type": "number",


bghgary:
What about the other cases that result in a division by zero (e.g. n == f or y == 0)? I'm wondering if this note is necessary.

lexaknyazev:
Corner cases for perspective camera are covered by schema.

bghgary:
I see y == 0 is restricted by schema, but where is n == f restricted?

bghgary:
Oh, never mind. I see it in the description.

bghgary:
To be consistent, perhaps, the note should go in the schema description?

pjcozzi:
@lexaknyazev do you have a reference to the discussion when we decided to allow this to be undefined instead of just not allow these cases?

lexaknyazev:
@pjcozzi 
> we decided to allow this to be undefined 

I think that such corner case was overlooked. Anyway, it's difficult (not impossible, though) to disallow it with JSON-Schema. Moreover, I think that we can't change the schema anymore.

@bghgary 
> To be consistent, perhaps, the note should go in the schema description?

:+1:

lexaknyazev:
Moved note to schema description.
@pjcozzi Please merge if it's fine.

pjcozzi:
I'm still a bit lost here - was there a discussion about this that we are able to reference here where we decided that undefined behavior was preferred compared to disallowing the divide by zero cases?

lexaknyazev:
> was there a discussion about this

I don't think so.

pjcozzi:
OK, why do we think undefined behavior is better than disallowing these cases?

lexaknyazev:
> OK, why do we think undefined behavior is better than disallowing these cases?

Do you think it's possible to make such change after spec release?

pjcozzi:
I think we could put this as a spec bug fix.  Let's confirm tomorrow.

lexaknyazev:
@pjcozzi 
Please check updated description and merge if it's fine.

pjcozzi:
Should this be non-zero or `> 0`?

pjcozzi:
Also, shouldn't the .md also be updated?

lexaknyazev:
Prop. reference will be updated on the next regen (it should happen soon).

lexaknyazev:
> Should this be non-zero or `> 0`?

Negative values could be used to flip final image. I'm not sure about disallowing such case.

pjcozzi:
Gotcha!

PR #1014, opened 2017-06-12
First attempt at normalizing schema titles with unique hierarchical values.

BeardedGnome:
Both the C# Loader and Wetzel expect the schema titles to be unique. Making them hierarchical is an easy way to achive that, but includes redundant data with the file name. Any solution that makes then unique and descriptive is acceptable. This is my first attempt.

@lexaknyazev 
@bghgary 
@pjcozzi

### DIFF
--- a/specification/2.0/schema/accessor.schema.json
+++ b/specification/2.0/schema/accessor.schema.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-04/schema",
-    "title": "accessor",
+    "title": "Accessor",
     "type": "object",
     "description": "A typed view into a bufferView.  A bufferView contains raw binary data.  An accessor provides a typed view into a bufferView or a subset of a bufferView similar to how WebGL's `vertexAttribPointer()` defines an attribute in a buffer.",
     "allOf": [ { "$ref": "glTFChildOfRootProperty.schema.json" } ],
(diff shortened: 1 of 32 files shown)


lexaknyazev:
Agree with human-oriented (capitalized with spaces) titles.

bghgary:
What happens with the links when running Wetzel?

BeardedGnome:
Secret gist of Wetzel output:
https://gist.github.com/BeardedGnome/55077c3f74b97137712f32b8ad0f9da1

bghgary:
> **Accessor Sparse Indices.componentType**

This looks strange to me. Perhaps we need to two properties for "title"?

BeardedGnome:
Looks like we're mixing and matching strings that represent types and strings that represent variables. I'm not sure of the solution off hand. I'll think on it.

BeardedGnome:
I've updated the secret gist. Does this look better?

> accessor.sparse.indices.componentType

BeardedGnome:
From @bghgary 
> is it possible to make the TOC not repeat the parent name?

Yes. See updated secret gist. Added to the separate Wetzel PR.

BeardedGnome:
Wetzel PR: https://github.com/AnalyticalGraphicsInc/wetzel/pull/22

lexaknyazev:
@pjcozzi 
OK to merge this?

pjcozzi:
Late to the party, but also +1 from me, thanks!!!

PR #1015, opened 2017-06-13
KHR_texture_transform

stevenvergenz:
As a prerequisite to an upcoming baked lighting extension, I needed the ability to atlas textures. This is a good start I think. Any comments?


lexaknyazev:
Couple notes:

- Are you sure that such extension must be a part of `sampler`, instead of `texture`? Will such extended samplers be usable for different textures?
- What are valid ranges for all new properties? E.g., could they be negative, could `tileS/T` be `0`?
- Are there any restrictions on used wrapping mode (e.g., it must be `REPEAT`)?
- Consider using `number` instead of `float`, to match JSON types.
- Consider providing an example JSON.

stevenvergenz:
@lexaknyazev 

1. You're right, offset/tile settings probably should depend on the source image. Updated to be part of textures.
2. See updated schema and README.
3. No restrictions on wrap mode, but I've added a caveat that mentions it in the README.
4. Done
5. Done

extensions/README.md
```diff
@@ -10,6 +10,9 @@ _Draft Khronos extensions are not ratified yet._
 * [KHR_materials_common](Khronos/KHR_materials_common/README.md)
 * [KHR_technique_webgl](Khronos/KHR_technique_webgl/README.md)
 
+#### Vendor extensions
+* [AVR_sampler_offset_tile](Vendor/AVR_sampler_offset_tile/README.md)
```
lexaknyazev: Please, update this line.

extensions/Vendor/AVR_texture_offset_tile/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+All of these values are restricted to the range [-1, 1]. While negative values are allowed, they will only produce sane results if the texture sampler's `wrap` mode is `REPEAT`, or if the result of the final UV transformation is within the range [0, 1] (i.e. negative tile settings and correspondingly positive offsets).
+
+### JSON Schema
+
+[AVR_texture_offset_tile.texture.schema.json](schema/AVR_texture_offset_tile.texture.schema.json)
+
+### Example JSON
+
+This example utilizes only the top left quadrant of the source image.
```
lexaknyazev: Given that `offsetS` is 0.5, it seems like top right quadrant is used. 
@stevenvergenz Could you check?

extensions/Vendor/AVR_texture_offset_tile/schema/AVR_texture_offset_tile.texture.schema.json
```diff
@@ -0,0 +1,36 @@
...
+        "offsetT": {
+            "type": "number",
+            "description": "How much a single repetition of the texture is offset from the beginning, in the T direction.",
+            "minimum": -1.0,
+            "maximum": 1.0,
+            "default": 0.0
+        },
+        "tileS": {
+            "type": "number",
+            "description": "How many times the texture is repeated across the surface, in the S direction.",
```
lexaknyazev: Since `tile` is `[-1, 1]`, could description be rephrased to not use "How many times"?
Text from readme (size of region) seems to be more precise, imho.

stevenvergenz:
Fixed those oversights, and reworded the descriptions in the schema.

lexaknyazev:
Not a very strong opinion, but there _could_ be benefits (as well as drawbacks) of moving this extension to `textureInfo` object.

---

With current design, `texture` object needs to be duplicated for each atlas element. Implementations may need to go through all textures and group them by `source`.
```json
{
    "materials": [
        {
            "emmissiveTexture": {
                "index": 0
            }
        },
        {
            "emmissiveTexture": {
                "index": 1
            }
        }
    ],
    "textures": [
        {
            "source": 0,
            "extensions": {
                "AVR_texture_offset_tile": {
                    "tileS": 0.5,
                    "tileT": 0.5
                }
            }
        },
        {
            "source": 0,
            "extensions": {
                "AVR_texture_offset_tile": {
                    "tileS": 0.5,
                    "tileT": 0.5,
                    "offsetS": 0.5
                }
            }
        }
    ]
}
```

However, this could be beneficial for glTF-to-glTF processing (e.g., individual textures could be converted to atlas without touching materials). Also, this scheme allows storing only texture atlas in glTF asset without anything else.

---

Other approach could be to move atlas information to `textureInfo`:
```json
{
    "materials": [
        {
            "emmissiveTexture": {
                "index": 0,
                "texCoord": 0,
                "extensions": {
                    "AVR_texture_offset_tile": {
                        "tileS": 0.5,
                        "tileT": 0.5
                    }
                }
            }
        },
        {
            "emmissiveTexture": {
                "index": 0,
                "texCoord": 0,
                "extensions": {
                    "AVR_texture_offset_tile": {
                        "tileS": 0.5,
                        "tileT": 0.5,
                        "offsetS": 0.5,
                        "offsetT": 0.5
                    }
                }
            }
        }
    ],
    "textures": [
        {
            "source": 0
        }
    ]
}
```

Pros: more compact representation and possibly easier runtime processing. 
Cons: Inability to store only texture atlas in an asset.

stevenvergenz:
The way I see it, there are two primary use cases:

1. A base color or metallic map texture is pulled from an atlas. In this case, it makes sense to have it on the textureInfo, as each material will probably have different settings here.
2. A [lightmap texture](https://github.com/KhronosGroup/glTF/pull/1017) is pulled from an atlas. In this case, each individual node in the scene graph will probably have its own portion of the texture, so if the tile/offset settings are not on the node, *something* down the chain (texture, material) will need to be duplicated (for both transmission and consumption).

Since the lightmap implementation is at a disadvantage either way, I think I'll move it to textureInfo, and handle the odd case in that extension's semantics.

donmccurdy:
> Pros: more compact representation and possibly easier runtime processing.
Cons: Inability to store only texture atlas in an asset.

Reading through https://github.com/mrdoob/three.js/pull/11577, my feeling is that the `textureInfo` may actually make runtime processing harder — one entry in the `textures` array previously mapped to a single Texture object, whereas now data required for texture creation is split between two places (`flipY` in textures, `scale`, `offset` in materials). If the asset only has a single texture with tiling specified, we'll end up cloning that texture needlessly while parsing `materials` because it might be reused by a later material with different tiling.

Despite the possible need for grouping by source, I think it's simpler to implement this extension with all properties in `textures`. IMO the compactness difference is a less critical issue.

lexaknyazev:
> If the asset only has a single texture with tiling specified, we'll end up cloning that texture needlessly while parsing `materials` because it might be reused by a later material with different tiling.

Why not use uniforms for handling `tile` and `offset`?

donmccurdy:
Oh, no not copying the image data (we do use uniforms for tile and offset) but the THREE.Texture object that wraps it must be cloned to have different tile/offset values. So not particularly expensive, but complicates implementation a bit.

emackey:
Any updates here?  Seems like this got a bit of an uptick in interest.  Also I notice that one of your sample usages provided in the docs here is the equivalent of a FLIP_Y flag, for people who wish that glTF's vertical texture coordinates went the other way.

```
"AVR_texture_offset_tile": {
    "offsetT": 1,
    "tileT": -1
}
```

stevenvergenz:
Are there any desired changes? I considered this extension complete. Though looking back, it's probably worthwhile to remove the [-1,1] restriction on the fields. There are use cases for having tile values greater than one that I hadn't considered when initially authoring this.

Plus, now that AltspaceVR has been acquired by Microsoft, I probably need to change the vendor extension.

bghgary:
> Are there any desired changes?

The link to the UnityGLTF can probably be updated to the Khronos one.

> change the vendor extension

Does it make sense for this to be a EXT_?

sonygod:
Any updates here?I'm waiting  for this feature in our project .
@stevenvergenz

stevenvergenz:
Okay, I've dusted off this extension. Any complaints? Should we add a rotation value, to complete the transform trifecta?

Changes:
* Changed AVR vendor extension to EXT. Is this good, or should it be MSFT?
* Changed "tile" to "scale" throughout. I realized I had mentally equated those terms, but they are reciprocal. Scale is easier to understand I think.
* Removed [0,1] restriction on values. I don't know what I was thinking with those.
* Updated UnityGLTF link

stevenvergenz:
For that matter, should the pairs be converted to arrays? Maybe `"offset": [0, 0]` makes more sense than separate `offsetS` and `offsetT`. That would be more consistent with the way node transforms are specified.

sonygod:
you can consider next vesion add rotation
1:EXT for shorter is better.
2:scale is standard, from unity and ue4 .
3:Removed [0,1] restriction on values ,good,no necessary.
4: I don't think array value better than object pair. it's need more and more documents.
  @stevenvergenz

stevenvergenz:
Okay, I've done one more pass. Now it's consistent with the way coordinate transforms are specified in `node`. `offset` and `scale` are now arrays, a scalar `rotation` value can be specified, and all these can be superseded by a `matrix` value. What do you think, @lexaknyazev @sbtron ?

extensions/README.md
```diff
@@ -10,6 +10,9 @@ _Draft Khronos extensions are not ratified yet._
 * [KHR_materials_common](Khronos/KHR_materials_common/README.md)
 * [KHR_technique_webgl](Khronos/KHR_technique_webgl/README.md)
 
+#### Vendor extensions
+* [AVR_texture_offset_tile](Vendor/AVR_texture_offset_tile/README.md)
```
pjcozzi: Should this be `AVR_` or `EXT_texture_transform`?  For `EXT_`, it should have multiple implementations.

For more info, see https://github.com/KhronosGroup/glTF/tree/master/extensions#promoting-extensions

stevenvergenz: Good review, and thanks for the link. Changing prefix to MSFT, and updating the extensions readme.

stevenvergenz:
@sbtron and I had a discussion offline, and determined that we've gotta wrap this up. We're keeping it simple: offset and scale. To support a good fallback for non-implementers, a `texCoord` field was added that overrides the textureInfo's field if supplied. This way, you can bake a transform into one set of coordinates for legacy systems, and efficiently tile them for implementers.

We're going to keep the EXT prefix for now, because by the time this extension is merged, we will have support in Babylon.js and UnityGLTF. Thus, it will be multi-vendor.

stevenvergenz:
Extension implemented in UnityGLTF: https://github.com/KhronosGroup/UnityGLTF/tree/feature/texture-transform

stevenvergenz:
@donmccurdy any objections if I implement this extension in the Three.js loader? As of a couple months ago, the loader didn't have any generic hooks for extension implementation, has that changed?

donmccurdy:
@stevenvergenz please do! there are still no generic hooks, but this seems straightforward enough to put in GLTFLoader.js... we may wait to merge that three.js PR until this one has been accepted, though.

extensions/Vendor/EXT_texture_transform/README.md
```diff
@@ -0,0 +1,82 @@
...
+| `scale`    | `array[2]` | `[1.0, 1.0]` | The scale factor applied to the components of the UV coordinates.
+| `texCoord` | `integer`  | `0`          | Overrides the textureInfo texCoord value if this extension is supported.
+
+Though this extension's values are unbounded, they will only produce sane results if the texture sampler's `wrap` mode is `REPEAT`, or if the result of the final UV transformation is within the range [0, 1] (i.e. negative scale settings and correspondingly positive offsets).
+
+> **Implementation Note**: For maximum compatibility, it is recommended that exporters generate UV coordinate sets both with and without transforms applied, use the post-transform set in the texture `texCoord` field, then the pre-transform set with this extension. This way, if the extension is not supported by the consuming engine, the model still renders correctly. Including both will increase the size of the model, so if including the fallback UV set is too burdensome, either add this extension to `extensionsRequired` or use the same texCoord value in both places.
```
donmccurdy: When implementing an exporter, will developers realistically follow this note? If I am generating UV coordinate sets with the transforms applied, is there any advantage to also including the extension?

extensions/Vendor/EXT_texture_transform/schema/EXT_texture_transform.textureInfo.schema.json
```diff
@@ -0,0 +1,33 @@
...
+                "type": "number"
+            },
+            "minItems": 2,
+            "maxItems": 2,
+            "default": [ 1.0, 1.0 ]
+        },
+        "texCoord": {
+            "type": "integer",
+            "description": "Overrides the textureInfo texCoord value if this extension is supported.",
+            "default": 0
```
lexaknyazev: Please add `"minimum": 0`.

lexaknyazev:
@stevenvergenz I'm going to implement this extension in glTF-Validator. Is it stable enough?

extensions/Vendor/EXT_texture_transform/README.md
lexaknyazev: > if including the fallback UV set is too burdensome, ... or **use the same texCoord value in both places**

How will clients not supporting this extension render an asset correctly in such case?

extensions/Vendor/EXT_texture_transform/schema/EXT_texture_transform.textureInfo.schema.json
```diff
@@ -0,0 +1,33 @@
+{
+    "$schema" : "http://json-schema.org/draft-04/schema",
+    "title" : "EXT_texture_transform textureInfo extension",
+    "type" : "object",
```
lexaknyazev: Is it intentional that this extension object doesn't inherit from `gltfProperty`, i.e., lacks `extras` and `extensions`?

extensions/Vendor/EXT_texture_transform/README.md
stevenvergenz: @lexaknyazev in this case, it would not be rendered "correctly", but in some cases a misplaced texture is preferable to a model failing to load entirely. It is up to creators/exporters to make this distinction.

extensions/Vendor/EXT_texture_transform/schema/EXT_texture_transform.textureInfo.schema.json
stevenvergenz: Oh, I didn't realize that the glTF semantics supported extending extensions. Doesn't that make the difficult job of loading extended properties even harder? I can add that if desired.

lexaknyazev: The SpecGloss extension (the only ratified 2.0 extension) follows that.

extensions/Vendor/EXT_texture_transform/README.md
stevenvergenz: @donmccurdy to be honest, I don't really expect anyone to pre-multiply UV sets. But I thought it was worth noting how to work around lack of support explicitly.

extensions/Vendor/EXT_texture_transform/schema/EXT_texture_transform.textureInfo.schema.json
stevenvergenz: It's added as of 60dcf94

stevenvergenz:
@lexaknyazev regarding this extension's stability, it might be worthwhile to wait a week or so. I don't anticipate any changes on my end, but someone might suggest something sensible.

extensions/Vendor/EXT_texture_transform/schema/EXT_texture_transform.textureInfo.schema.json
lexaknyazev: For schema to be correct, `extensions` and `extras` need to be mentioned in `properties` (see other schema files for example).

We don't have any guideline on whether or not allow extending extensions. I just think that it's a bit better to have uniform approach to all glTF objects.

extensions/Vendor/EXT_texture_transform/schema/EXT_texture_transform.textureInfo.schema.json
```diff
@@ -26,7 +28,8 @@
         "texCoord": {
             "type": "integer",
             "description": "Overrides the textureInfo texCoord value if this extension is supported.",
-            "default": 0
+            "default": 0,
```
lexaknyazev: I suggest to remove default value here, because it should default to texture's texCoord which isn't necessarily zero.

emackey:
This comment may be out-of-scope for this extension, but would it be desirable to drive the tile/offset values from animation somehow?  Lots of cool effects can be had from moving, scaling, and rotating texture coordinates.

extensions/Vendor/EXT_texture_transform/schema/EXT_texture_transform.textureInfo.schema.json
stevenvergenz: Added as of 0b22fc5

extensions/Vendor/EXT_texture_transform/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## glTF Schema Updates
+
+The `EXT_texture_transform` extension may be defined on `textureInfo` structures. It may contain the following properties:
+
+| Name       | Type       | Default      | Description
+|------------|------------|--------------|---------------------------------
+| `offset`   | `array[2]` | `[0.0, 0.0]` | The offset of the UV coordinate origin as a percentage of the texture dimensions.
+| `scale`    | `array[2]` | `[1.0, 1.0]` | The scale factor applied to the components of the UV coordinates.
+| `texCoord` | `integer`  | `0`          | Overrides the textureInfo texCoord value if this extension is supported.
```
lexaknyazev: Please, update texCoord's default here.

stevenvergenz: Fixed in d5ecfdc

emilian0:
Hello @stevenvergenz, can you please clarify why `rotation` was removed from the specs? 
The reason why I am asking is that we might be interested in using this extension, but we would need rotations. Thanks.

stevenvergenz:
@emilian0 Basically, because UV rotation is not commonly supported. Everyone already has offset/scale support, but the inclusion of rotation might discourage implementation for relatively little benefit. So we're keeping this extension light and easy.

donmccurdy:
just FYI, three.js (very recently) added support for texture rotation. That would probably extend to A-Frame VR and React VR as well down the road.

pjcozzi:
Hi all - just wanted to bump this discussion to see if folks are OK with including texture rotation.

@sbtron and all?  Any feedback?

mjmvisser:
Hi all, I can bring some clarity about Unity and texture rotation. It's on the roadmap and will be supported at some point, but it's not a priority right now. We're in the process of switching to a pluggable render pipeline (our Scripted Render Pipeline) and a shader graph, and right now that is the main focus of the graphics team. Texture rotation will eventually be supported in the new pipeline (along with scale and offset) via standard shader graph nodes, and SetTextureScale and Offset will be deprecated. In the new pipeline, texture scale/offset/rotation will be ordinary shader uniforms set via naming convention, e.g. SetFloat("_MainTexRotation", 90.0) and SetVector("_MainTexScaleOffset", new Vector4(1, 2, 0.5, 0.5). SRP is new in 2018.1, due around GDC.

extensions/Vendor/EXT_texture_transform/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+This is equivalent to Unity's `Material#SetTextureOffset` and `Material#SetTextureScale`, or Three.js's `Texture#offset` and `Texture#repeat`.
+
+## glTF Schema Updates
+
+The `EXT_texture_transform` extension may be defined on `textureInfo` structures. It may contain the following properties:
+
+| Name       | Type       | Default      | Description
+|------------|------------|--------------|---------------------------------
+| `offset`   | `array[2]` | `[0.0, 0.0]` | The offset of the UV coordinate origin as a percentage of the texture dimensions.
```
emilian0: Just a minor note: I believe offset is not a percentage, correct? It is encoded as a factor relative to the texture dimensions.

emilian0:
@stevenvergenz @sbtron . 
Last Wednesday @tparisi @pjcozzi and I discussed about wrapping up this extension including rotations for completeness. The main rationale is that, although less common, rotations can be useful (we have at least one use case). Including rotations will make this extension more future proof. 
What is your take?
Do we know of specific implementers opposed to including rotations?
Any other concern with this extension?
We can also discuss this live during the next 3D format group call. Ideally we would love to wrap this up in the next weeks. Thanks.

mjmvisser:
Hi all - on the Unity side, we're planning on supporting texture rotation as part of a standard node in our new shader graph and new renderer. I doubt the legacy renderer will ever support texture rotation, as it would require modifying every single shader to support it.

sbtron:
Thanks @mjmvisser. From the working group discussion we are going to go ahead with adding rotation back in with the assumption that various engines will support it in the future. 
@emilian0  please go ahead and add the rotation property back in.

//cc @stevenvergenz

donmccurdy:
If there's a sample model demonstrating this extension in a visually verifiable way, I'd be interested in testing that out. 🙂

extensions/Vendor/EXT_texture_transform/README.md
stevenvergenz: Oh, good catch. I'm used to percentages being [0,1)

stevenvergenz:
@donmccurdy We've been using this one for testing the UnityGLTF support: https://github.com/KhronosGroup/UnityGLTF/blob/master/UnityGLTF/www/EXT_texture_transform/pyramid.gltf

donmccurdy:
Thanks! Here is a [WIP implementation in three.js](https://github.com/mrdoob/three.js/compare/dev...donmccurdy:feat-gltfloader-texturetransform), based on @stevenvergenz's earlier implementation. I haven't added rotation yet, but don't expect any issues there.

| without extension | with EXT_texture_transform |
|---|---|
| screen shot 2018-02-22 at 11 21 28 am | screen shot 2018-02-22 at 11 20 42 am |

stevenvergenz:
Can someone verify the rotation direction? The descriptions say it rotates the UVs counter-clockwise (or the image clockwise) because that's what it does in Blender (the easiest way to test), but that might not be the case in the implementing engines.

donmccurdy:
three.js does the reverse; rotating the texture counter-clockwise for positive rotation values. But our API is conceptually dealing with textures not UVs; I don't have a strong opinion either way.

http://localhost:8080/examples/webgl_materials_texture_rotation.html

bghgary:
@PatrickRyanMS tried this in Maya and Maya seems to rotate the texture clockwise for positive rotation values.

I don't think this is going to be an issue for any implementation. It's just a matter of negating the rotation value on load if it's not the same convention, right?

emilian0:
@stevenvergenz @pjcozzi I went ahead and made the change
EXT_texture_transform -> KHR_texture_transform as discussed during the last 3dformat call. 
Please verify the 2 commits above. 
Thank you.

extensions/Khronos/KHR_texture_transform/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Example JSON
+
+This example utilizes only the lower left quadrant of the source image, rotated clockwise 90&deg;.
+
+```json
+{
+  "materials": [{
+    "emissionTexture": {
+      "source": 0,
```
MiiBond: This should be "index", I believe.

extensions/Khronos/KHR_texture_transform/README.md
```diff
@@ -0,0 +1,102 @@
...
+}
+```
+
+This example inverts the T axis, effectively defining a bottom-left origin.
+
+```json
+{
+  "materials": [{
+    "emissionTexture": {
+      "source": 0,
```
MiiBond: Ditto.

extensions/Khronos/KHR_texture_transform/README.md
```diff
@@ -0,0 +1,102 @@
...
+      "extensions": {
+        "KHR_texture_transform": {
+          "offset": [0, 1],
+          "scale": [1, -1]
```
MiiBond: Are we at all concerned that the `normalTextureInfo` structure already contains a `scale` value? I suppose there could be confusion that the scale value in the texture_transform is an override...
Maybe not but I thought I'd mention it.

MiiBond:
@lexaknyazev This extension also needs to move to match the new folder structure. I'm having trouble figuring out how to add my own commit to this pull request.

extensions/Khronos/KHR_texture_transform/README.md
bghgary: I think it might be confusing, but I don't have a better suggestion. For reference, here is what it might look like:

```json
      "normalTexture": {
        "index": 2,
        "scale": 1.5,
        "extensions": {
          "KHR_texture_transform": {
            "offset": [0, 1],
            "rotation": 1.57079632679,
            "scale": [0.5, 0.5]
          }
        }
      },
```

donmccurdy: It feels OK nested under the `*_texture_transform` namespace to me. Conceptually it then becomes `transformScale` or something like that in code.

donmccurdy:
Thanks @bghgary — the sample looks good in my WIP three.js implementation, except that the rotations are reversed. From the spec:

> Rotate the UVs by this many radians counter-clockwise around the origin. This is equivalent to a similar rotation of the image clockwise.

three.js does the opposite (the image rotates counter-clockwise for positive rotation values) so I was negating rotation values to compensate. Based on [this demo](https://threejs.org/examples/?q=rotation#webgl_materials_texture_rotation) I think three.js behaves as documented, and if so I think the screenshot above may be incorrect. Here's what I'm seeing:

screen shot 2018-04-24 at 7 09 33 pm

donmccurdy:
I'm wondering if we should include an implementation note that by rotation `around the origin` we mean UV coordinate `0,0` and not the center of the texture. This becomes obvious if you're implementing the transform in a shader, but might not be when dealing with existing code.

pjcozzi:
@bghgary the test model looks great, please add to the sample models repo when ready!

emackey:
@bghgary The sample model does look great.  One suggestion I have is that you might use rotation on an image of an arrow, and then provide a another arrow with pre-baked rotation, so users can check if the arrows line up.  The [OrientationTest](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/OrientationTest) model does this with geometry, but your model could do it with the rotation extension.

Otherwise it takes a sharp eye to spot the rotation going the "wrong" way as above.

bghgary:
_I'll address my test model issues a bit later._

First, I have a few observations that I want to call out with this extension and how tools/engines deal with texture rotation. Like everything else, they all use different conventions.

* Max/Maya rotates around the center of the texture.
* Blender rotates around the bottom-left of the texture.

This extension expects the rotation to rotate around the origin (0, 0) which is the [top-left for glTF textures](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#textures).

I'm ignoring offset and scale for simplicity.

Converting these to the conventions of this extension should be relatively straight forward, but there is some concern about animating the property of the texture transform. When the conventions are different between the DCC tool and this extension, it will likely require resampling the curves of the animation to correctly account for the convention differences.

For example, say someone is trying to animate a button with offset, scale, and rotation and is using spline curves for the animation. The only way I can see to export this animation curve is to compute each frame of the curve into the conventions of this extension and refit the curve.

I think this kind of export algorithm is unavoidable as no matter what we pick some tool will have a different convention.

Thoughts?

bghgary:
@donmccurdy 
> the rotations are reversed. From the spec:
>
>> Rotate the UVs by this many radians counter-clockwise around the origin. This is equivalent to a similar rotation of the image clockwise.

I think I am interpreting the spec differently?

The original image:
logo

My render:
image
For me, this is a clockwise rotation of the image.

Your render:
image
For me, this is a counter-clockwise rotation of the image.

My screenshot matches how I'm interpreting the spec, since it says "rotation of the image clockwise".

Perhaps we need to define what clockwise / counter-clockwise means.

donmccurdy:
I believe this was my mistake — three.js rotates images CCW for positive rotation values, the opposite of this extension, but we also use a different UV convention such that 0,0 is lower-left. If I understand right those two things cancel out, and our rotation becomes CW when applied to UVs following glTF's upper-left convention.

Visually I think CW / CCW is clear enough.

bghgary:
@emackey 
> One suggestion I have is that you might use rotation on an image of an arrow

[TextureTransformTest.zip](https://github.com/KhronosGroup/glTF/files/1953431/TextureTransformTest.zip)

image

How's this? I will write up a readme for it with pictures.

extensions/2.0/Khronos/KHR_texture_transform/README.md
```diff
@@ -0,0 +1,102 @@
...
+| `scale`    | `array[2]` | `[1.0, 1.0]` | The scale factor applied to the components of the UV coordinates.
+| `texCoord` | `integer`  |              | Overrides the textureInfo texCoord value if supplied, and if this extension is supported.
+
+Though this extension's values are unbounded, they will only produce sane results if the texture sampler's `wrap` mode is `REPEAT`, or if the result of the final UV transformation is within the range [0, 1] (i.e. negative scale settings and correspondingly positive offsets).
+
+> **Implementation Note**: For maximum compatibility, it is recommended that exporters generate UV coordinate sets both with and without transforms applied, use the post-transform set in the texture `texCoord` field, then the pre-transform set with this extension. This way, if the extension is not supported by the consuming engine, the model still renders correctly. Including both will increase the size of the model, so if including the fallback UV set is too burdensome, either add this extension to `extensionsRequired` or use the same texCoord value in both places.
+
```
donmccurdy: Let's add another note:

> **Implementation Note**: From the glTF core specification, the origin of the UV coordinates (0, 0) corresponds to the upper left corner of a texture image.

donmccurdy:
> For maximum compatibility, it is recommended that exporters generate UV coordinate sets both with and without transforms applied...

I can't quite justify this to myself. For maximum compatibility, the extension should not be used at all. The stated purpose of the extension is to support texture atlases for efficient runtime behavior, and including redundant UVs is contrary to that goal.

For those who might use this extension on the exporter/generator side (/cc @MiiBond, @Aurl), could you comment on whether and how you might implement this?

Considerations:

- We do want glTF assets to be widely supported, and for extensions to not break existing tools.
- On the other hand, this extension is fairly straightforward and likely to be well supported.
- In any case, creating a step in glTF-Pipeline or glTF-Toolkit that can strip the `KHR_texture_transform` extension by baking transforms to UVs may a good idea.

emackey:
For example:

```
    "materials" : [
        {
            "emissiveFactor" : [
                1.0,
                1.0,
                1.0
            ],
            "emissiveTexture" : {
                "index" : 2,
                "extensions": {
                    "KHR_texture_transform": {
                        "offset": [0, 0.1]
                    }
                }
            },
            "name" : "Material_MR",
            "normalTexture" : {
                "index" : 4,
                "extensions": {
                    "KHR_texture_transform": {
                        "offset": [0, 0.2]
                    }
                }
            },
            "occlusionTexture" : {
                "index" : 3,
                "extensions": {
                    "KHR_texture_transform": {
                        "offset": [0, 0.3]
                    }
                }
            },
            "pbrMetallicRoughness" : {
                "baseColorTexture" : {
                    "index" : 0,
                    "extensions": {
                        "KHR_texture_transform": {
                            "offset": [0, 0.4]
                        }
                    }
                },
                "metallicRoughnessTexture" : {
                    "index" : 1,
                    "extensions": {
                        "KHR_texture_transform": {
                            "offset": [0, 0.5]
                        }
                    }
                }
            }
        }
    ],
```

emackey:
So, using this extension could be a fast way to get your model a lot of extra UV sets, without the weight of all those extra vertex attributes, if the client implementations are smart about passing in the transform info per texture instead of baking it into new attributes.  Seems like it could quickly add a number of uniforms to the shader though.

emackey:
What I'm getting at is:  It seems like for a client to understand how many "truly different" UV maps are in use, the client must compare different transform extensions within a single material, and check for equality.  Equality would indicate that a single set of transformed UVs could be applied to multiple images in a material, whereas inequality (or presence-vs-absence difference) would indicate the need for different final sets of UVs.

donmccurdy:
> By "one" texture transform, I think that means the same transform must be equally applied to base, occlusion, normal, and emissive maps, with a different (or non-existent) transform able to be applied only to the occlusion map, correct?

I'm not sure what happens with an occlusion map actually, but otherwise that's correct.

>  If the user was diligent and took the time to correctly copy-and-paste the same transform to each of the textures in a material, would ThreeJS "diff" the offsets and understand that the same transform was applied?

three.js will take the first transform it finds, if any, and apply it to all textures (unsure about occlusion maps).

> In theory, given current limitations, the idealized correct thing to do is for ThreeJS to throw a warning/error if there is any difference between requested transforms for base, MR, Normal, Emit, or indeed if a transform is specified on any of those channels but not exactly repeated identically on each of those other channels.... right?

Agreed.

> It seems like for a client to understand how many "truly different" UV maps are in use, the client must compare different transform extensions within a single material, and check for equality. Equality would indicate that a single set of transformed UVs could be applied to multiple images in a material, whereas inequality (or presence-vs-absence difference) would indicate the need for different final sets of UVs.

Not sure I follow... I think you're describing how a client that doesn't support texture transforms but does support multiple UVs could bake the transform to UVs on the fly, if needed?

donmccurdy:
^I'm not sure what change(s), if any, would follow from this. Certainly there is no need to constrain this extension to three.js's current limits. I'm mostly trying to understand our fallback recommendation, and whether the `texCoord` index override makes sense and will be used.

Generally my sentiment is that extensions meant for optimization should never be included by model catalogs like Sketchfab or Turbosquid, but added by later tools, in which case making all extensions optional is less important. I think the Draco extension is a clear example of that. This one I'm not sure about... the "texture atlas" use case we describe is an optimization, but [texture animation](https://developer.playcanvas.com/en/tutorials/animated-textures/) is not. I'm hoping this extension will be widely implemented enough that it's a moot point.

emackey:
I'm not opposed to anything here, just trying to think this one through.  It strikes me as a bit odd that the seemingly common case of applying the same transform to multiple images can only be done by repeating the same transform information per image, and can only be detected by comparing such information.  But I don't have a better idea how to structure the extension without limiting its flexibility.

donmccurdy:
Yes, I suppose this also means that (in some future version where texture transforms may be animated) the animation must target each texture with a separate channel? It's verbose but does seem like the more flexible choice.

bghgary:
> in which case making all extensions optional is less important

If an optimization happens and stays within a closed ecosystem, then it's okay. If an asset ever leaves that closed ecosystem intentionally, the asset should not have any required extensions.

> Draco extension is a clear example of that

If someone serves an asset with Draco compression from a server intending for anyone to consume (i.e. not a closed ecosystem), then I would expect Draco to be optional. I would expect the same thing for this texture transform extension. These optimizations will offer better performance for clients that support the extension while keeping compatibility with clients that don't support the optimizations. If we don't do this, we risk fragmenting the ecosystem.

In the future, when we introduce a new version (probably 3.0 because of forward compatibility requirements for 2.x) where Draco and this extension is part of the spec, these kind of problems will go away.

donmccurdy:
> If someone serves an asset with Draco compression from a server intending for anyone to consume (i.e. not a closed ecosystem), then I would expect Draco to be optional. I would expect the same thing for this texture transform extension...

This makes sense, and I'm glad that we're ensuring optional extensions are possible. But, we're making a broad recommendation here that (as written) seems to imply all exporters should always include both. I don't expect exporters to do that, nor am I sure that they should. Things get complicated if an API wants to ensure an asset includes separate `.bin` files to prevent downloading unnecessary data:

- draco buffer
- uncompressed buffer
- shared buffer
- pre-transform UVs
- post-transform UVs
- pre/post transform UVs in draco/uncompressed variants?

Broadly I don't think there is any single recommendation we can make for all exporters. At some point I would like to write more detailed best practices for extensions in general, based on the functional role of the extension and the sort of ecosystem in which the model will appear.

I'm fine with moving forward with this extension as-is and potentially rephrasing this implementation note at a future date.

---

The points @emackey brings up are interesting; I imagine the first exporter implementation will run into some of that. So far it does still appear that this approach is the right one.

bghgary:
> ... Broadly I don't think there is any single recommendation we can make for all exporters. ...

I basically agree with this. +1 on rephrasing the implementation note.

> Things get complicated if an API wants to ensure an asset includes separate `.bin` files to prevent downloading unnecessary data

I think exporters will need to decide which optimization extensions to use. I don't think there will be a need for all possible permutations. For example, there can be one path for the optimized version (i.e. both Draco and texture transform together), all other cases can use core only. But I'm not sure this works out. Can we ensure that an asset is authored such that a client implementation will only load the optimized path if both Draco and texture transform are supported or else fall back to core?

extensions/2.0/Khronos/KHR_texture_transform/README.md
```diff
@@ -0,0 +1,102 @@
...
+mat3 matrix = translation * rotation * scale;
+vec2 uvTransformed = ( matrix * vec3(Uv.xy, 1) ).xy;
+```
+
+This is equivalent to Unity's `Material#SetTextureOffset` and `Material#SetTextureScale`, or Three.js's `Texture#offset` and `Texture#repeat`. UV rotation is not widely supported as of today, but is included here for forward compatibility.
+
+## glTF Schema Updates
+
+The `KHR_texture_transform` extension may be defined on `textureInfo` structures. It may contain the following properties:
```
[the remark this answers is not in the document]
bghgary: I would say yes since both `occlusionTextureInfo` and `normalTextureInfo` inherit from `textureInfo`.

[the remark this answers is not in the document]
lexaknyazev: OK. It would be nice to have sample models showing those cases (I've run into this internally).

donmccurdy:
> For example, there can be one path for the optimized version (i.e. both Draco and texture transform together), all other cases can use core only. But I'm not sure this works out. Can we ensure that an asset is authored such that a client implementation will only load the optimized path if both Draco and texture transform are supported or else fall back to core?

I don't think that's feasible, without resorting to requirements specific to these two extensions. Let's assume both the compressed and uncompressed mesh would need to include transformed and untransformed UVs, as the client makes a decision on each extension separately. An exporter might split that as follows:

- buffer1: compressed geometry, including UVs
- buffer2: uncompressed geometry, excluding UVs
- buffer3: uncompressed, transformed UVs
- buffer4: uncompressed, untransformed UVs
- buffer5: common data (e.g. textures, animation)

That's OK, and shouldn't be especially complex for the client to load efficiently. But whether that complexity is worthwhile probably depends on the situation:

- For APIs serving glTF assets for _immediate_ viewing in arbitrary engines, this approach seems appropriate. An alternative would be a dynamic serving mechanism like `foo.gltf?draco=true&textureTransform=false`, where the `.gltf` file is generated on the fly but buffers are stored statically.
- For a downloadable asset catalog like Sketchfab's website, prefer avoiding optimization extensions entirely and using `.glb`.
- For developer/artist workflows, where assets are being created in a modeling tool and exported for a particular experience, ... I'm not sure. I suspect the developer will want control in the application layer and not in their library's glTF loader, so they'd create different versions of assets targeted to desktop, mobile, etc. They'll know whether the engine supports Draco or texture transforms, and would enable/disable those optimizations themselves.

pailhead:
>For example, if a user applies a transform to the base color, but does not specify one on any other channel, ThreeJS seems likely to end up using that transform for all of the textures (sans occlusion) in that same material, right?

Three.js is somewhat confusing when it comes to this. Depending on what combination of textures is in your material, it will randomly apply the offset **to all the maps** but sampling from only one. 
https://github.com/mrdoob/three.js/issues/12788

pailhead:
>^I'm not sure what change(s), if any, would follow from this. Certainly there is no need to constrain this extension to three.js's current limits.

I agree with this. I had no idea that three.js is such a big authority on a format like glTF.

From which point should i try to observe this situation? 

Three.js on it's own may not be aware of a file format like `glTF`. The core of the library does not know how to handle this format at all. There is an example which to me reads as: 

"if you like this **rendering library**, you can write various loaders for it, including glTF."

So someone sat down, used the core of the library and wrote an example for loading the `glTF` format specifically. Now the issue is, the core itself may not be built to specifically handle this proposed extension. This shouldn't matter, since the core was never built to specifically handle `glTF` either.

Three.js users can use the library out of the box, but can also infinitely extend it. The complicated way would be to fork it and implement your features, but it does have various apis that allow one to extend the core features. 

I've made a quick demo to test this claim:

http://dusanbosnjak.com/test/webGL/three-material-includes/

I'm not sure if it tackled this specific issue as i've somewhat got lost in the comment but this is what it's supposed to do.

Outside the demo
1. Three's built in PBR material only supports roughness/metalness
2. Three's built in materials use texture.transform for transformations, according to that list posted above, impossible to use two different transforms for two different textures/channels

In the demo
1. Three's built in PBR can now use specular/gloss (while keeping everything else)
2. each `material.someMap` gets a bunch of transform properties `material.someMapOffset`, `material.someMapRepeat`. 

Would this be a valid way to approach this? Since `GLTFLoader` is not in the core of three.js, what requirements does it have for any of this stuff to also be in the core? 

This one uses the same extension applied to the GLTFLoader. Ie. this version of the loader has no limitations mentioned in this thread, but it depends on another 3rd part example, not the core:
http://dusanbosnjak.com/test/webGL/three-material-includes/webgl_loader_gltf_extensions.html

pailhead:
```
buffer1: compressed geometry, including UVs
buffer2: uncompressed geometry, excluding UVs
buffer3: uncompressed, transformed UVs
buffer4: uncompressed, untransformed UVs
buffer5: common data (e.g. textures, animation)
```

Could the loaders transform this without sending the duplicated uvs through the wire (or even storing them)? There could still be multiple vertex buffers for each of the duplicated uvs, but at least they would be transferred as a single one.

pailhead:
3dsmax

I've opened up an ancient version of 3ds max to compare this, and they seem to couple this with the texture object. I'm mostly curious about how this extension would affect three.js.

This screenshot corresponds to how three.js is already structured, the texture has what's called `Offset`, `Tiling` and `Angle`. It lacks the mapping i think, or it might actually know about "explicit map channel" and something else (like cube mapping) but it doesn't know about the uv channel. 

The source though is still the same bitmap, which could be a texture atlas. Animation though seems like it would be a completely different use case.

pailhead:
I have updated this demo, to include one of the gltf models from the three.js repo:

[bbsginst](http://dusanbosnjak.com/test/webGL/three-material-includes/)

Exposed with the dat.gui are some transform parameters for the specularMap, but how it's done is just one implementation. I think it would make sense for this to be obtained from some texture object, that could share some source bitmaps.

silvainSayduck:
Hi everyone,

This is not really directly a contribution to the discussion, but rather a general wondering of how using this kind of extension could be made optional? I mean, if we have a glTF using this extension:
```js
  ...
  "emissiveTexture" : {
    "index" : 2,
    "extensions": {
      "KHR_texture_transform": {
        "offset": [0, 0.1],
        "scale": [0.2,0.2]
      }
    }
  }
  ...
```

How can a client without support for it display anything remotely usable? I mean, if it's going to use a texture atlas without knowledge of the extra parameters in `extensions`, the rendering will be completely wrong, won't it?
What I mean is that, shouldn't the default `index` point to a non-atlas texture, and another `index` be added within the `extensions` object, pointing to the atlased texture? The default texture could be ignored (and not downloaded) by a client supporting the extension, while other clients still provide an acceptable result using the default `index`.
This reasoning could be applied to many extensions, so I'm wondering if I'm just missing something here. Or maybe it just means that we will always put this kind of extensions in  `extensionsRequired` (which will in turn prevent them to be widely used as they'd result in limited glTFs). 

Thanks a lot in advance for the clarification.

pailhead:
I can reason about this better if it weren't an extension. I fear that what you're describing can be mitigated by a network of fallback options, but i don't like those solutions.

I went to a recent meetup for glTF extensions, but I missed the introduction thus missing a lot.

najadojo:
@pailhead is the [Meetup: Get your glTF on with WebGL/WebVR at Microsoft!](https://www.youtube.com/watch?v=GuA1mYG0-xE) you are referring to?

najadojo:
@silvainSayduck There is a note about how fallback should apply with this extension using the texCoord field:

> Implementation Note: For maximum compatibility, it is recommended that exporters generate UV coordinate sets both with and without transforms applied, use the post-transform set in the texture texCoord field, then the pre-transform set with this extension. This way, if the extension is not supported by the consuming engine, the model still renders correctly. Including both will increase the size of the model, so if including the fallback UV set is too burdensome, either add this extension to extensionsRequired or use the same texCoord value in both places.

The texture would always be an atlas but one portion of it would be applicable for clients that don't implement the extension.

pailhead:
@najadojo 

Correct. The example that stuck in my head was the LOD presentation. I didn't see what benefits the extension brings over what's already available in the format. Ie. why can't this type of management be achieved with a naming convention `myMesh_master` + `myMesh_lod`. The argument over just storing multiple gltf files was i believe that it saves on the redundancy of duplicating the scene graph, but it left me wondering how much saving is achieved over for example adding an extra node called `variousLods` with just a flat list of corresponding LOD nodes.

I think the draco one discussed fallback options, so i have a bit of bias in this topic, when i try to understand something like this:

>The texture would always be an atlas but one portion of it would be applicable for clients that don't implement the extension.

I imagine an atlas that includes itself, and a fallback texture, ie. the client without the extension would just read (half?) the texture, which actually doesn't make sense since it would still require some kind of scale and offset. How does this sentence apply to a layout of a bitmap (that is to be used for this example)?

Why would two models, one using textures, another an atlas even be considered the same? This seems like something some server side API would be doing, based on client capabilities figure out what data the client needs. I understand this would now happen in the gltf header, it notifies the client of the presence of an additional system. I don't understand what happens next conceptually. 
Does this mean that a "gltf" loader needs to know how to handle the extensions, or that the client needs to know how to handle the extension with "gltf" loaders? 

I think i have a hard time understanding how extensions should fit the entire ecosystem, point me in the right direction as it seems to be a bit of a digression from this topic.

silvainSayduck:
> I think i have a hard time understanding how extensions should fit the entire ecosystem, point me in the right direction as it seems to be a bit of a digression from this topic.

I agree 100% with this... :) If someone knows a more appropriate place for the genera discussion around this, it would be awesome!

> The texture would always be an atlas but one portion of it would be applicable for clients that don't implement the extension.

We are actually planning to use this extension for tiling textures: basically, using a "low-resolution" texture that we tile across the whole mesh for high visual quality with low texture file size. And obviously in that case, unless I'm mistaken, including both a "non-tiled" texture and the tile as an atlas will defeat the original purpose. So I guess we'd have to go with the `extensionsRequired` way...

sbtron:
Some of the questions raised here are also discussed in https://github.com/KhronosGroup/glTF/issues/1256 
lets follow the discussion there for the general extension best practices so that this PR can be about KHR_texture_transform extension.

pjcozzi:
Is this ready to merge?

I think just update https://github.com/KhronosGroup/glTF/blob/master/extensions/README.md so that `KHR_texture_transform` is under the **Khronos extensions** section.

donmccurdy:
Merged via https://github.com/KhronosGroup/glTF/pull/1372 — opened a new PR to rebase.

PR #1025, opened 2017-06-21
Enforce extension lists dependency

lexaknyazev:
From [Specifying Extensions](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#specifying-extensions):
> All extensions used in a glTF asset must be listed in the top-level `extensionsUsed` array object...

> All glTF extensions required to load and/or render an asset must be listed in the top-level `extensionsRequired` array...

That implies, that all required extensions must also be listed in `extensionsUsed` array. However, schema allows `extensionsRequired` without `extensionsUsed`. This PR aligns schema with spec. 

@bghgary Please check whether such update is possible.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -1455,6 +1455,8 @@ All glTF extensions required to load and/or render an asset must be listed in th
 }
 ```
 
+`extensionsRequired` is a subset of `extensionsUsed`. All values in `extensionsRequired` must also exist in `extensionsUsed`.
+
 For more information on glTF extensions, consult the [extensions registry specification](../../extensions/README.md).
 
 # GLB File Format Specification


bghgary:
@lexaknyazev Is this change necessary? I don't see how it helps. I don't think this makes any difference to exporter/importers. Does it make a difference to the validator? Seems like the validator needs to validate that extensionsRequired is a subset of extensionsUsed regardless of the schema dependency.

Also, the wording seems to be confusing. Some of us have understood it to mean that is an extension is required, then it means automatically that the extension is used.

lexaknyazev:
> Does it make a difference to the validator?

No. Validation code could be adapted if needed.

Main point is: could an asset have only `extensionsRequired` array without `extensionsUsed`? Schema says yes, while spec implicitly says no.

> the wording seems to be confusing...

Could you propose updated wording?

bghgary:
Let's agree on the schema before changing the wording.

> could an asset have only `extensionsRequired` array without `extensionsUsed`?

If extensionsRequired is supposed to be a subset, then no, this would be invalid. But I don't think we need the schema to enforce this edge case. The schema can't enforce that extensionsRequired is a subset of extensionsUsed, so if an implementation wants to enforce, it has to do some work anyways. I don't think the schema enforcement adds much value.

lexaknyazev:
I don't know to what extent will applications rely on schema. Since it can't cover all possible issues, apps have to validate something on their own anyway. 
If we can't touch this anymore, let's focus on more clear wording.

bghgary:
I would suggest that we not change the schema unless we absolutely have to.

> let's focus on more clear wording
👍

bghgary:
> wording

How about adding a sentence like at the end:

> `extensionsRequired` is a subset of `extensionsUsed`. All values in `extensionsRequired` must also exist in `extensionsUsed`.

lexaknyazev:
PR updated.

Issue #1021, opened 2017-06-17
Clarification: Texture Image Orientation


mlimper:
Hi,

we are currently running into an issue with texture image orientation.

The [glTF 2.0 spec](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#images) says the following:
`
First image pixel (UV coordinates origin) corresponds to the upper left corner of the image.
Implementation Note: OpenGL-based implementations must flip Y axis to achieve correct texture sampling.
`

So, just to get this straight:
This means that exporters should store texture images in the typical standard orientation where the lower left corner of the image corresponds to the lower left corner of the UV space - correct? I am asking because this is what usually requires the flip, which the spec expects from the GL implementation.

Here's a simple (Collada) example loaded in MeshLab:
![image](https://user-images.githubusercontent.com/1853237/27256922-82d57900-5379-11e7-93de-96fe93a0be7f.png)

The image is loaded into the UV space "as-is". Meaning, the eye of the duck is still in the upper right corner, as it was in the original image. In this case, **the OpenGL-based implementation must flip**.

The issue we're observing right now is that *exporters* start to flip images, or UV coordinates, because *runtimes* seem to expect them in a non-standard (? at least uncommon) orientation. See, for example, this thread:
https://github.com/Kupoman/blendergltf/issues/59

This applies to BabylonJS (at least the version we are using) and Three.js, which seem to expect the images or UVs "upside down":
https://threejs.org/examples/webgl_loader_gltf2.html
https://threejs.org/examples/js/loaders/GLTF2Loader.js
(search for "_texture.flipY = false;")

So, given that those two engines already do it consistently, it could also be that I am getting this wrong?
Or are the runtimes indeed missing to flip the texture properly?

Thanks a lot in advance!

@bghgary Can I ask you again about your comments on this?

donmccurdy:
This was a recent correction to the spec, and I think our implementations and sample models haven't caught up yet. 

See also:
- https://github.com/KhronosGroup/glTF-WebGL-PBR/issues/16
- https://github.com/KhronosGroup/glTF/pull/1005

mlimper:
Oh I see, cool - thanks for the quick clarification.

javagl:
[...]

PR #1047, opened 2017-07-22
Rephrased UV origin specification and added example image

mlimper:
Based on https://github.com/KhronosGroup/glTF/issues/1021 and https://github.com/KhronosGroup/glTF-Sample-Models/issues/82:
* Removed irritating comment about OpenGL texture flipping
* Slightly rephrased / detailed UV origin specification
* Added example image, slightly modified from the original proposed by @javagl 

Current Version: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#images
Preview: https://github.com/mlimper/glTF/tree/uv-coords-origin-spec/specification/2.0#images


lexaknyazev:
Image shows gradient originating from lower left corner. I think that colors should be aligned with UV or removed.

mlimper:
Sorry, didn't think about that before, should now be correct.

One could probably add a comment to the spec about the semantics of the pixels in that image, but that may be too much (it would feel a bit overloaded to me, the current version with the coord annotations seems clear enough).

Issue #1052, opened 2017-08-03
Spec very ambiguous about multiple animation support.


reduz:
I'm very disappointed, I implemented 100% of the spec in Godot. Spent 4 days on it. At the time I'm done with it, I realize that, unlike what it proposes through the whole spec, glTF **does not** support multiple animations. 

Guys, It's 2017 and all DCCs and game engines support multiple animation clips, so it should be perfectly possible to export all character animations in a single file without relying on cheap tricks like segmenting a timeline or having to export one animation per file (with the whole scene cloned on it). Collada and FBX support this perfectly.

In the spec, the term **animations**, plural, is used through the whole thing. It totally makes sense, any animation works on a number of channels. By reading the spec it becomes obvious that multiple animations can be stored in the format.

Yet, when I go try the demos.. every demo I find uses an animation for a different node. For example on this demo:

https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BoxAnimated

To get the animation to look like in the gif, you have to play all animations together.
In this other demo:

https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/CesiumMan

Each bone has an animation of it's own. you must play all animations together in order to see the guy walking.

So, please change the spec to make it very clear to implementers that, in 2017, a new format does not support multiple animations for real, or that there is a strict requirement for them to be played at the same time for everything to look properly. Please also urgently add an official extension to support REAL multiple animations, so this format can be used for 21st century workflows.

I suggest: KHR_actual_multiple_animations or similar.

darksylinc:
+1 to this.

Juan is trying to convince me to support glTF in Ogre as well, and this is a deal breaker.

If popular exporters default to exporting animations as one node per track _which is further encouraged by the official samples_, the format is useless to us.

The specs should be very strict on that animations should be done by having multiple nodes being animated in one track; so that each animation is independent of each other.

donmccurdy:
Hi @reduz, this is an issue with our sample models, and (related) the Blender Exporter is missing support for exporting actions separately (https://github.com/KhronosGroup/glTF-Blender-Exporter/issues/39). However, the glTF spec itself very much does support what you're asking for here, there is no extension necessary.

We will certainly work on updating the sample models to be more in line with expectations here.

reduz:
[...]

PR #1069, opened 2017-08-10
Clarification for animation exporting best practices

reduz:


emackey:
This is per discussion in #1052.

pjcozzi:
Awesome, thanks for the update, @reduz!

@lexaknyazev and/or @bghgary do you want to review, make tweaks if any, and merge?

specification/2.0/README.md
```diff
@@ -1250,6 +1250,8 @@ glTF 2.0 also supports animation of instantiated Morph Targets in a similar fash
 
 > **Note:** glTF 2.0 defines only animation storage, so this specification doesn't define any particular runtime behavior, such as: order of playing, auto-start, loops, mapping of timelines, etc...
 
+> **Note:** glTF 2.0 does not specifically define how an animation will be used when imported but, as a best practice, it is recommended that each animation is self contained as an action. This clarification is necessary due to early glTF 2.0 exporters splitting the tracks of a single source animation into several glTF animations (behaviour that is not currently advised today).
```
donmccurdy: Couple things:

* [ ] For consistency with the rest of the document, I would title this paragraph: **Implementation Note:**
* [ ] IMO the second sentence, giving historical background about a (hopefully soon-to-be-fixed) bug in the sample models, can be omitted from the specification.
* [ ] Would suggest additional language explaining the concept of an action: *"For example, "Walk" and "Run" animations might each contain multiple channels targeting a model's various bones. The viewer or engine may choose when to play any of the available animations."*

reduz: Something like this?

*Implementation Note*: glTF 2.0 does not specifically define how an animation will be used when imported but, as a best practice, it is recommended that each animation is self contained as an action. For example, "Walk" and "Run" animations might each contain multiple channels targeting a model's various bones. The viewer or engine may choose when to play any of the available animations.

bghgary: A minor thing: we did a pass at some point to use the words "client implementation" instead of viewer or engine.

donmccurdy: @reduz that looks good, along with @bghgary's suggestion.

reduz: So, this could be final then? 

Implementation Note: glTF 2.0 does not specifically define how an animation will be used when imported but, as a best practice, it is recommended that each animation is self contained as an action. For example, "Walk" and "Run" animations might each contain multiple channels targeting a model's various bones. The client implementation may choose when to play any of the available animations.

donmccurdy: 👍

pjcozzi:
Thanks again @reduz!  I added you to the acknowledgments in 74c53694f4725a851849423d7447ebb60c514cf4.

PR #1071, opened 2017-08-12
Minor usage fixes

scurest:
Fixes some minor language errors in the spec: lots of missing articles, some agreement errors, a few mis-spellings, etc.

I also tried to be slightly more consistent in the placement of spaces and the use of punctuation before displays of code.

(On an unrelated note: in the example [here](https://github.com/KhronosGroup/glTF/blob/020c11c387ecdb005e88740e81055161815d807d/specification/2.0/README.md#transformations), the text says that the node "defines [a] non-default rotation" but the rotation appears to be the default one (`[0,0,0,1]`). Maybe the `1` should go in one of the imaginary components? If so, I can throw that in this PR too.)

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -727,7 +727,7 @@ Each attribute is defined as a property of the `attributes` object. The name of 
 
 Valid attribute semantic property names include `POSITION`, `NORMAL`, `TANGENT`, `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_0`, `JOINTS_0`, and `WEIGHTS_0`.  Application-specific semantics must start with an underscore, e.g., `_TEMPERATURE`.
 
-The valid accessor type and component type for each attribute semantic property is defined below.
+Valid accessor types and component types for each attribute semantic property are defined below.
 
 |Name|Accessor Type(s)|Component Type(s)|Description|
 |----|----------------|-----------------|-----------|
@@ -774,7 +774,7 @@ primitives[i].attributes.POSITION +
 ```
 Morph Targets are implemented via the `targets` property defined in the Mesh `primitives`. Each target in the `targets` array is a dictionary mapping a primitive attribute to an accessor containing Morph Target displacement data. Currently only three attributes (`POSITION`, `NORMAL`, and `TANGENT`) are supported. All primitives are required to list the same number of `targets` in the same order.
 
-The valid accessor type and component type for each attribute semantic property is defined below. Note that the *w* component for handedness is omitted when targeting `TANGENT` data since handedness cannot be displaced.
+Valid accessor types and component types for each attribute semantic property are defined below. Note that the *w* component for handedness is omitted when targeting `TANGENT` data since handedness cannot be displaced.
 
 |Name|Accessor Type(s)|Component Type(s)|Description|
 |----|----------------|-----------------|-----------|
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -4,7 +4,7 @@
 
 *Version 2.0*
 
-The GL Transmission Format (glTF) is an API-neutral runtime asset delivery format.  glTF bridges the gap between 3D content creation tools and modern 3D applications by providing an efficient, extensible, interoperable format for the transmission and loading of 3D content.
+The GL Transmission Format (glTF) is an API-neutral runtime asset delivery format. glTF bridges the gap between 3D content creation tools and modern 3D applications by providing an efficient, extensible, interoperable format for the transmission and loading of 3D content.
 
 Last Updated: June 9, 2017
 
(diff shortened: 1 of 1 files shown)


javagl:
Although I didn't "proofread" it in all depth, it seems like it fixes many of the minor grammar issues that I also noticed, but not being a native english speaker, didn't feel confident enough to fix. (An aside, there are languages where there is no such thing as "articles" - that's likely the reason of why they have been missing here...) 

For the person who may be responsible for accepting or declining this PR: 

It it valid to assume that these fixes are **only** targeting grammar, and do **not** change the *meaning* in any way? 

The latter may be subtle. For example, I noticed that at

https://github.com/KhronosGroup/glTF/pull/1071/commits/5f7192269e27209c307b4d37cf5e57daf9ade366#diff-a787a363bd89618cda440f087cd0e0e1L447

you changed 

> Buffer view could have `byteStride` property.

to

> A buffer view can have a `byteStride` property. 

While adding the articles is certainly correct, as far as I can tell, I'm not sure whether the change from **"could"** to **"can"** could (or can) change the meaning in a way that is relevant for a specification that has been ratified.

(I'm a big fan of https://www.ietf.org/rfc/rfc2119.txt for specification documents, by the way...)

scurest:
Thanks @javagl, you're right.

I did change the meaning in one place:

> The `asset` object must contain **glTF version** which specifies the target glTF version of the asset.

which became

> The `asset` object must contain **a `version` property** which specifies the target glTF version of the asset.

because the former sentence is almost redundant. The new sentence is still true though (`asset.version` is required).

----


By the way, I'm also somewhat uncertain about this:

javagl:
Regarding "could" vs "can": Based on my understanding of the language, I would have used "can" here as well. Or, more adhering to RFC2119, "MAY": 

> A buffer view MAY have a `byteStride` property. 

However, in many cases, the statements involve conditionals - something like "It *can* have this property, but it *must* have the property if (some condition holds)". One must really be careful here. (Must. Not may.) 

Regarding "are" vs. "is": Since it refers to "accessor type and component type", I think "are" is correct. 

But hepefully, a native english speaker (and/or someone who has a deeper understanding of the implications of changing a specification after its ratification) will chime in here.

specification/2.0/README.md
```diff
@@ -725,7 +727,7 @@ Each attribute is defined as a property of the `attributes` object. The name of
 
 Valid attribute semantic property names include `POSITION`, `NORMAL`, `TANGENT`, `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_0`, `JOINTS_0`, and `WEIGHTS_0`.  Application-specific semantics must start with an underscore, e.g., `_TEMPERATURE`.
 
-Valid accessor type and component type for each attribute semantic property are defined below.
+The valid accessor type and component type for each attribute semantic property is defined below.
```
bghgary: Minor inconsistency with this also. *The* was added to *valid* but *The* was not added to the previous *Valid* in the previous paragraph. My opinion is that *The* is unnecessary here.

scurest: _The_ is definitely needed here. It's not needed in the previous paragraph because "attribute semantic property names" is plural. We could make this "accessor types and component types" though (even though there is only one for each), which would obviate the _are_/_is_ thing here too.

scurest: Also, I think I'd prefer "Application-specific semantic property names must..." in the previous paragraph.

bghgary: I guess that depends on whether "accessor type and component type" is plural or singular as is being discussed already. I would be okay with changing it to "accessor types and component types."

scurest: Okay, let's do that here and for the next one too.

scurest:
Any more feedback on this? Should I push a commit for the changes talked about so far?

pjcozzi:
> Should I push a commit for the changes talked about so far?

Sure, I think that will keep things moving.

Please also merge in master; looks like there are minor conflicts.

Anyone else have input here?

emackey:
Are there hand-edits to wetzel output here?  If so, some of these changes may need to be copied back to schema files.

scurest:
I stopped checking when I got to the property reference to avoid that, so there shouldn't be, no.

emackey:
Sorry, did we drop the ball here?  Seems like there's some great edits here, and we just got stuck waiting for additional feedback that never came.

pjcozzi:
At quick glance these changes are OK with me.

specification/2.0/README.md
```diff
@@ -142,7 +142,7 @@ glTF has been designed to meet the following goals:
...
 
-The design of glTF takes a pragmatic approach. The format is meant to mirror the GPU APIs as closely as possible, but if it did only that, there would be no cameras, animations, or other features typically found in both modeling tools and runtime systems, and much semantic information would be lost in the translation. By supporting these common constructs, glTF content can not only load and render, but it can be immediately usable in a wider range of applications and require less duplication of effort in the content pipeline.
+The design of glTF takes a pragmatic approach. The format is meant to mirror the GPU APIs as closely as possible, but if it did only that, there would be no cameras, animations, or other features typically found in both modeling tools and runtime systems, and much semantic information would be lost in the translation. By supporting these common constructs, glTF content can not only be loaded and rendered, but can be immediately usable in a wider range of applications and require less duplication of effort in the content pipeline.
```
snagy: I feel like this part about "meant to mirror the GPU APIs as closely as possible" is a bit misleading; glTF is designed to be rendered via modern GPU APIs with as little data processing as possible, but isn't really a mirror of the APIs.

scurest:
This PR was out of date anyway. I'll leave the branch up for a while on the off chance it's useful to anyone who wants to check for typos in current master though.

specification/2.0/README.md
scurest: The only change on this line was "glTF content can not only load and render, but it can be" -> "glTF content can not only be loaded and rendered, but can be".

Issue #857, opened 2017-02-28
glTF 2.0: texture packing for Metallic-Roughness?


donmccurdy:
From [this paragraph](https://github.com/sbtron/glTF/blob/30de0b365d1566b1bbd8b9c140f9e995d3203226/specification/2.0/README.md#pbrmetallicroughnessmetallicroughnesstexture), I see that the Metallic-Roughness workflow will make use of texture packing by assigning:

| Channel | Map |
|---------|------|
| R | Metallic |
| G | Roughness |
| B | — |
| A | — |

Were there plans for using the `B` channel later? If Occlusion is the obvious next choice, is it worth altering the order to match [UE4's Occlusion/Roughness/Metallic R/G/B](https://support.allegorithmic.com/documentation/display/SPDOC/Unreal+Engine+4#UnrealEngine4-UnrealMaterialSetup) ordering?

I'm not very familiar with texture creation workflows, so maybe this is a non-issue and easy for exporters/converters to deal with. But thought I'd ask, as THREE.MeshStandardMaterial is already configured for that OcclusionRoughnessMetallic ordering. Thanks!

pjcozzi:
@mlimper?

mlimper:
That actually sounds like a reasonable proposal - especially as it's already being used elsewhere like that.

The ordering should not be an issue IMHO, if we can make it fit existing conventions, that sounds like a plus.

Only thing we might want to make sure is that people can use the extension safely without having an occlusion map... which then probably means that somewhere inside the glTF material description we will want to specify whether the texture content has an occlusion channel or not.

@sbtron and colleagues, what do you think?

lexaknyazev:
I fully support merging them and defining static channel mapping at least for MR. 
Texture swizzling isn't available in WebGL, so engines would need to maintain multiple versions of shader code to support different mappings.

How would SpecGloss model fit with such change?

javagl:
Breaking down the current texture reference...

    "metallicRoughnessTexture": {
        "index": 2,
        "texCoord": 1
    }

to 

    "occlusionChannel": {
        "index": 2,
        "texCoord": 1,
        "component" : 0
    }
    "metallicChannel": {
        "index": 2,
        "texCoord": 1,
        "component" : 2
    },
    "roughnessChannel": {
        "index": 2,
        "texCoord": 1,
        "component" : 2
    },

[...]

PR #1075, opened 2017-08-16
[KMC-extension] reworked intro, equation, shininess note

andreasplesch:
This PR is also available as https://github.com/UX3D-nopper/glTF/pull/1 since it was based on the UX3D-nopper fork. However, the suggestion was to provide the PR against KhronosGroup/glTF .

Please feel free to edit freely or ignore all suggestions and just use as input for discussion on phrasing, language and content.


pjcozzi:
Fixes #945.

McNopper:
-1

This extension should become KHR_cmn_lights extension, as it does not fit well to PBR. For common materials, we can take it 1:1.

andreasplesch:
Would it work better to complete this extension in the https://github.com/UX3D-nopper/glTF/ fork to avoid confusion ?

McNopper:
Basically, the extension is fine for for the common materials. But for PBR materials, we need too many exceptions, that I will suggest a separate one. Sorry for the confusion.

extensions/Khronos/KHR_lights/README.md
```diff
@@ -0,0 +1,173 @@
...
+#### Point
+
+Point lights emit light in all directions over a distance. Point lights support the `color` property. `constantAttenuation`, `linearAttenuation` and `quadraticAttenuation` light properties embedded in the `positional` property describe the different attenuations.
+
+```javascript
+"extensions": {
+    "KHR_lights" : {
+        "lights": [
+            {
+                "positional": [
```
warkid: typo: I guess "positional" is supposed to be a dictionary? should be curly brackets then not square...

donmccurdy:
> This extension should become KHR_cmn_lights extension, as it does not fit well to PBR

In the interest of expediting a material model that runs more efficiently on mobile devices, could we consider splitting this PR in two, and finishing up KHR_cmnBlinnPhong before we tackle lighting? Anecdotally, most users are using glTF for individual models, so I see blinn-phong as the more important (and perhaps less complex?) extension here.

extensions/Khronos/KHR_materials_cmnBlinnPhong/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+The following table lists the allowed types and ranges for the specular-glossiness model:
+
+|                            |Type         |Description|Required|
+|----------------------------|-------------|-----------|--------|
+|**diffuseFactor**           | `number[4]` | The reflected diffuse RGBA factor of the material.|No, default: `[1.0,1.0,1.0,1.0]`|
+|**diffuseTexture**          | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | The diffuse texture.|No|
+|**specularFactor**          | `number[3]` | The specular RGB color of the material.      |No, default: `[1.0,1.0,1.0]`|
+|**shininessFactor**         | `number`    | The shininess of the material.|No, default: `1.0`          |
+|**specularShininessTexture**| [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)|The specular shininess texture.|No|
```
donmccurdy: I assume this means specular is `RGB` channels, and shininess is `A`. We've [discussed a bit in the past](https://github.com/KhronosGroup/glTF/issues/857#issuecomment-284258812), and I don't remember where we ended up on this question, but there was some concern that because using RGB+A requires PNG, transmission size might actually end up worse than if two JPGs had been used (e.g. RGB + R).

donmccurdy:
Generally the blinn-phong proposal looks good to me.

One note, three.js and A-Frame can currently support `shininessFactor` but not `shininessTexture`.  There is some [ongoing discussion](https://github.com/mrdoob/three.js/issues/9339) about whether to change that.

extensions/Khronos/KHR_materials_cmnBlinnPhong/README.md
```diff
@@ -0,0 +1,134 @@
...
+### Blinn - Phong
+
+The Blinn-Phong material model is defined by the following properties:
+* `diffuse` - Reflected diffuse color of the material
+* `specular` - Specular color of the material
+* `shininess` - Shininess factor of the specular color
+
+TODO: EXPLAIN FURTHER 
+
+The following table lists the allowed types and ranges for the specular-glossiness model:
```
donmccurdy: s/specular-glossiness/Blinn-Phong

extensions/Khronos/KHR_materials_cmnBlinnPhong/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+The material is intended for shading according to the Blinn-Phong lighting model or a close approximation.
+
+This equation is not complex and detailed via the ACM. Refer to “Models of Light
+Reflection for Computer Synthesized Pictures,” SIGGRAPH 77, pp 192-198 [http://portal.acm.org/citation.cfm?id=563893](http://portal.acm.org/citation.cfm?id=563893), and in particular the "Simple Highlight Model" section.
+
+The following code illustrates the basic computation:
+
+```
+color = <emission> + <ambient> * al + <diffuse> * max(N * L, 0) + <specular> * max(H * N, 0)^<shininess>
```
donmccurdy: where is `<ambient>` defined?

zellski:
If I'm reading correctly, the discussion in https://github.com/KhronosGroup/glTF/issues/947 ended with:
- The **absence** of a term is a signal to the engine to downgrade to a shader that doesn't understand that term; e.g. no specular term means (at best) diffusion-only; no diffusive term means ambient/emissive only.
- We've scrapped the ambient factor.
The logical conclusion seems to be that a constant material is flagged as such by a completely empty cmnBlinnPhong extension? In other words, just a somewhat wordy boolean opt-in.

Was the explicit type really problematic? I'm all for simplified APIs, but I thought the type added clarity rather than complexity.

Of course it could be argued that we should then have a cmnLambert and a cmnConstant, with correspondingly tighter schemas -- indeed, like what @donmccurdy just suggested in https://github.com/KhronosGroup/glTF/pull/1095.

donmccurdy:
Thinking about this more, I am also worried about the wording: 

> This extension defines a single material type, Blinn-Phong. Other or simpler materials types such as Lambert or Constant can be described by having zero factors.

This implies the correct way to specify a Constant material would be:

```json
{
  "name": "RedConstant",
  "emissiveFactor": [1, 0, 0],
  "extensions": {
    "KHR_materials_cmnBlinnPhong": {
      "diffuseFactor": [0, 0, 0, 0],
      "specularFactor": [0, 0, 0],
      "shininessFactor": 0
    }
  }
}
```

This doesn't feel very natural. The Constant material is identified by the presence of a blinn-phong extension, even though no factors from the blinn-phong extension are actually used.

Also, we may someday want to allow animation of materials (say, opacity or color). In that case, you could define an initial diffuse state of `[1,0,0,1]` and animate it to `[0,0,1,1]`. But an initial state of `[0,0,0,0]` would imply an entirely different shading model, and introduces an edge case for loaders.

I think we should probably be more explicit about the intended shading model, rather than having things implicitly in `cmnBlinnPhong`. By comparison, a `"KHR_materials_cmnConstant": {}` extension with _no parameters_ (simplify defining a new shading model with a subset of parameters already available in the core pbr material) would be clearer.

extensions/Khronos/KHR_materials_cmnBlinnPhong/README.md
stevepg: on one of the other threads i did suggest adding the following (it is submitted as a PR against what i suspect is now a dead fork)

diffuseTerm   = diffuseFactor   * diffuseTexture 
specularTerm  = specularFactor  * specularShininessTexture.rgb 
shineTerm     = shininessFactor * specularShininessTexture.a
emissiveTerm  = emissiveFactor  * emissiveTexture // core Materials spec
occlusionTerm = occlusionFactor * occlusionTexture // core Materials spec 

color = emissiveTerm + occlusionTerm * (diffuseTerm * max(N * L, 0) + specularTerm * max(H * N, 0)^shineTerm)

where
 `N` – Normal vector
 `L` – Light vector
 `H` – Half-angle vector,calculated as halfway between the unit Eye and Light vectors, using the equation H= normalize(I+L)

Textures are optional, and if provided, will modulate the base Factor value. Note that `specularShininessTexture` contains the specular modulation as the rgb components and shininess modulation as the alpha component.

javagl:
Some of the questions here seem to be related to what I asked about in https://github.com/KhronosGroup/glTF/issues/947#issuecomment-308722750 (and I hope that what I wrote there was not considered as a "proposal", considering the disclaimer: I don't know which performance implications this might have).

But the point mentioned above, about the animations, is IMHO an important *conceptual* one: **If** the official statement will be that zero factors *imply* a different shading model, then loaders/viewers might be tempted to use (pseudo) code like

    if (material.everything == 0) loadConstantShader();
    else if (material.something == 0) loadLambertShader();
    else loadBlinnPhongShader();

(The goal of this would be (performance) optimizations, of course. I still wonder whether there will really be noticable performance differences between these imlementations, considering that even the most complex shader here is comparatively simple. But regardless of that: ) 

If some material value *changed* afterwards (be it though animations or something like a "material editor" UI component), then this would cause problems. 

I'm not so deeply involved here to make any proposals, but think that it *might* be possible to cover both directions with implementation notes:

- For different material types, one could mention: "For simplicity, viewers may choose to implement all these types with a single blinn/phong shader, where they only zero-out the unused properties to emulate the lambert/constant case"
- For a single material type, one could mention: ~"Even though some values might be 0 at the beginning, the viewer/shader has to anticipate that these values might change to non-zero values at runtime"

But this is just a vague idea.

steveghee:
i much prefer the idea of each extension being specific to its function e.g. material_blinnphong does only that.  If you want a _constant material, then define another material type.

donmccurdy:
> I'm not so deeply involved here to make any proposals, but think that it might be possible to cover both directions with implementation notes:
> * For different material types, one could mention: "For simplicity, viewers may choose to implement all these types with a single blinn/phong shader, where they only zero-out the unused properties to emulate the lambert/constant case"
> * For a single material type, one could mention: ~"Even though some values might be 0 at the beginning, the viewer/shader has to anticipate that these values might change to non-zero values at runtime"

Yes, these notes would clear up any ambiguity. The downside of the second is that (unless we add an explicit indication of shading model) there is then no option to switch to something cheaper.

> I still wonder whether there will really be noticable performance differences between these imlementations, considering that even the most complex shader here is comparatively simple.

I'm asking around, and while all of these shaders are comparatively simple, there are implementation differences that make a bigger difference (at least in three.js).

1. In the three.js Phong material, lighting is computed per-fragment unless `material.lighting` is turned off. But for the same reason as we can't switch to a Constant material based on zero parameters, it's probably not safe for the loader to turn lighting off without some explicit indication in the asset.
2. In the three.js Lambert material, lighting is computed per-vertex. This is an implementation detail, but means that devs may want to target Lambert for cheaper lighting. As a point of reference, Daydream provides a custom Unity renderer which (among other things) uses per-vertex lighting instead of per-fragment.
3. In the three.js Constant material, lighting is never computed.

For the reasons above, devs will have reasons to target a specific material type. I think it's fine for implementations to implement Blinn-Phong and use that in all cases (let's include this implementation note), but I don't think we should omit any indication of the shading model from the format, for implementations that need to distinguish between shading models.

zellski:
I will just add a note here, redundant perhaps, but for emphasis: that the relationship between performance and shader choice is far from predictable, as it passes through the domain of aesthetic consideration and art style. I think this extension should carefully avoid accidentally encoding any assumptions about **how** one would choose to downgrade shader complexity; it should simply retain as much expressive power as possible.

To take our case as an example: we decided early on that the user base we're hoping to attract early on is going to be largely on mid-grade android devices. In VR, reduced frame rate is not a reasonable path for graceful adaptation -- only scene complexity is. So we embraced in all our art a low-poly, unlit, flat-shaded style. This seems like exactly the kind of deliberate decision that you would want to encode **explicitly** in the asset file, not leave it as some kind of optional optimization path to the loader/viewer.

andreasplesch:
Although I fear that I am one of the least suitable persons to keep editing this PR, I went ahead and removed the lights extension language as it was included here by accident and adopted most (all?) of the specific suggestions made. Apologies to those I overlooked.
This PR should be seen solely as a way to continue to evolve this extension. I would love if a more responsive editor would take this draft over, in a new PR or fork, since I will only be able to contribute very sporadically and slowly.

extensions/Khronos/KHR_materials_cmnBlinnPhong/README.md
```diff
@@ -17,9 +17,9 @@ Written against the glTF 2.0 spec.
 
 This extension defines a Blinn-Phong model based material for use with glTF 2.0 in the same way as the core specification defines Physically Based Rendering (PBR) based materials. Other extension may define additional commonly used materials. For example, if a need for a Phong model arises, an extension named e.g. `cmnPhong` could be specified. 
 
-Given the long history of the Blinn-Phong model and the vast amount of collective experience with applying this model, the demand for convenient use of such materials is still there.
+Given the long history of the Blinn-Phong model and the vast amount of collective experience with applying this model, the demand for convenient use of such materials is still exists.
 
-This extension defines a single material type, Blinn-Phong. Other or simpler materials types such as Lambert or Constant can be described by having zero factors.
+This extension defines a single material type, Blinn-Phong. Other or simpler materials types such as Lambert or Constant may be emulated by applying zero factors. Please note, strict Lambert or Constant materials may be specified in dedicated extensions.
```
donmccurdy: I think I would just omit the second two sentences, so that this paragraph becomes:

> Given the long history of the Blinn-Phong model and the vast amount of collective experience with applying this model, the demand for convenient use of such materials is still exists. This extension defines a single material type, Blinn-Phong.

(Mentioning that simpler materials can be emulated via Blinn-Phong can be mentioned in the extensions for those materials)

donmccurdy:
@andreasplesch thanks for helping to move this forward! 🙂  Your changes look good to me; I just added one additional suggestion.

There's already a `KHR_materials_common` draft in the repo, so I would vote to go ahead and merge this PR, then in a new PR we can iterate on whichever we decide to use. I'll update #1095 to discuss the unlit/constant material.

andreasplesch:
I could also resubmit this PR against another branch if that would be preferable.

donmccurdy:
> I could also resubmit this PR against another branch if that would be preferable.

@andreasplesch Thanks, that might be best — could you resubmit this against a `gltf2-common-materials` branch?

andreasplesch:
Closing this PR after resubmitting to as https://github.com/KhronosGroup/glTF/pull/1111 against the gltf2-common-materials branch. Happy to reopen here for any reason.

PR #1095, opened 2017-09-10
KHR_materials_cmnConstant

donmccurdy:
Based on the syntax of #1075. For mobile devices trying to maintain 90+ FPS, I think it may be worth keeping an unlit shading model available.

### DIFF
--- a/extensions/Khronos/KHR_materials_cmnConstant/README.md
+++ b/extensions/Khronos/KHR_materials_cmnConstant/README.md
@@ -14,36 +14,23 @@ Written against the glTF 2.0 spec.
 
 ## Overview
 
-This extension defines an unlit ("constant") model based material for use with glTF 2.0 in the same way as the core specification defines Physically Based Rendering (PBR) based materials. Other extension may define additional commonly used materials. 
-
-In mobile devices with limited resources, unlit materials remain a performant alternative to higher-quality shading models.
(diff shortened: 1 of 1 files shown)


donmccurdy:
Hm as @zellski pointed out off thread, `emissiveFactor` and `emissiveTexture` are already in the default PBR material spec, so there's nothing new here. We can set an emissive and zero out the other factors.

If an exporter specifies that something should be unlit by zeroing-out the other factors, we expect the importer to check each factor and assign the simpler shading model? I thought we'd intended for exporters to be able to provide both metal/rough and spec/gloss and let the runtime choose which to use, but this leaves us no way of specifying whether unlit shading is explicitly a fallback or not. Maybe that's OK? I'm not sure.

donmccurdy:
On second thought, please ignore the content of this PR for now. I will follow up on #1075.

zellski:
Perhaps open this back up, in empty/boolean form, given recent developments in https://github.com/KhronosGroup/glTF/pull/1075?

donmccurdy:
Sounds good. I've updated this as an "empty" extension, and it should be ready for review or feedback.

donmccurdy:
Closing this; going to work toward including this along with blinnPhong as a single extension. See https://github.com/KhronosGroup/glTF/pull/1111.

PR #1097, opened 2017-09-13
Update blendergltf status to glTF 2.0

Kupoman:
With the 1.1 release of blendergltf, glTF 2.0 output is now supported.


Kupoman:
This project also supports glTF 1.0 output. Should it have a badge for all supported versions, or just the most recent supported version?

donmccurdy:
Nice work! I think just the most recent version supported should be mentioned on this readme.

PR #1102, opened 2017-09-14
test_glTF

dzbkqsw:
试一下


pjcozzi:
@dzbkqsw is this a translation of the glTF spec to Chinese?

If so, it could be a great contribution, but we would want this to be a separate document, and not have both the English and Chinese version in the same .md file.

realvictorprm:
Hm this only affects the first version of glTF. Would be awesome to see the second version translated too

pasu:
update to the latest

emackey:
Translations are most welcome, but need to be placed in separate documents.  I'm going to close this one since it's fairly old, but please do open a new PR if these changes are made.  Thanks!

PR #1104, opened 2017-09-19
Add BLENDER_physics extension from https://github.com/Kupoman/blendergltf

Moguri:
We've been working on this extension for a while, and we believe it is ready to share with a larger audience to continue to gather feedback and improve the extension. There has also been some demand for a physics extension, and we hope this PR will further discussion on such an extension.

### DIFF
--- a/extensions/Vendor/BLENDER_physics/README.md
+++ b/extensions/Vendor/BLENDER_physics/README.md
@@ -4,7 +4,7 @@
 
 * Mitchell Stokes, Blender, <mailto:mogurijin@gmail.com>
 * Daniel Stokes, Blender, <mailto:kupomail@gmail.com>
-* Steven Vergenz, AltspaceVR, <mailto:steven@altvr.com>
+* Steven Vergenz, Microsoft, <mailto:steven.vergenz@microsoft.com>
 
 ## Status
 
--- a/extensions/Vendor/BLENDER_physics/README.md
+++ b/extensions/Vendor/BLENDER_physics/README.md
@@ -32,7 +32,7 @@ The properties available are listed in the table below.
 |**mass**|`number`|The 'weight', irrespective of gravity, of the node|No, default: `1.0`|
 |**static**|`boolean`|Whether or not the Node should be moved by physics simulations|No, default: `false`|
 |**boundingBox**|`array`|The dimensions of the local (i.e., does not include the node's transform) bounding box of the collision shape centered on the origin|Yes|
-|**primaryAxis**|`number`|The axis to use for the height of the collision shape|No, default: `1`|
+|**primaryAxis**|`string`|The axis to use for the height of the collision shape (only applies to `CAPSULE`, `CYLINDER`, and `CONE`)|No, default: `Y`|
 |**mesh**|`glTF id`|The ID of the mesh to use for `CONVEX_HULL` and `MESH` collision shapes|No, default: `node's mesh` if it exists, otherwise use `BOX` shape|
 |**offsetMatrix**|`array`|A 4x4 transform matrix applied to the collision shape in addition to the node's transform|No, default: `[ 1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0 ]`|
 |**offsetRotation**|`array`|A rotation offset (as a quaternion) applied to the physics shape in addition to the node's rotation|No, default: `[0.0, 0.0, 0.0, 1.0]`|
--- a/extensions/Vendor/BLENDER_physics/README.md
+++ b/extensions/Vendor/BLENDER_physics/README.md
@@ -24,13 +24,22 @@ This extension extends nodes to include information necessary to construct rigid
 A `BLENDER_physics` object is added to the extensions list of any `node` that should participate in the physics simulation.
 The properties available are listed in the table below.
 
-**Properties**
+**BLENDER_physics Properties**
 
 |   |Type|Description|Required|
 |---|----|-----------|--------|
-|**collisionShape**|`string`|The shape a physics simulation should use to represent the node|No, default: `BOX`|
+|**collisionShapes**|`array`|Shapes a simulation should use to represent the node|Yes|
 |**mass**|`number`|The 'weight', irrespective of gravity, of the node|No, default: `1.0`|
 |**static**|`boolean`|Whether or not the Node should be moved by physics simulations|No, default: `false`|
+|**collisionGroups**|`integer`|A 32-bit bit field representing the node's group membership|No, default: `1`|
+|**collisionMasks**|`integer`|A 32-bit bit field representing what groups the node can collide with|No, default: `1`|
+
+
+
+**shape Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
 |**boundingBox**|`array`|The dimensions of the local (i.e., does not include the node's transform) bounding box of the collision shape centered on the origin|Yes|
 |**primaryAxis**|`string`|The axis to use for the height of the collision shape (only applies to `CAPSULE`, `CYLINDER`, and `CONE`)|No, default: `Y`|
 |**mesh**|`glTF id`|The ID of the mesh to use for `CONVEX_HULL` and `MESH` collision shapes|No, default: `node's mesh` if it exists, otherwise use `BOX` shape|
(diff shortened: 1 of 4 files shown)


extensions/Vendor/BLENDER_physics/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**collisionShape**|`string`|The shape a physics simulation should use to represent the node|No, default: `BOX`|
+|**mass**|`number`|The 'weight', irrespective of gravity, of the node|No, default: `1.0`|
+|**static**|`boolean`|Whether or not the Node should be moved by physics simulations|No, default: `false`|
+|**boundingBox**|`array`|The dimensions of the local (i.e., does not include the node's transform) bounding box of the collision shape centered on the origin|Yes|
+|**primaryAxis**|`number`|The axis to use for the height of the collision shape|No, default: `1`|
```
donmccurdy: I think you mean `Type=string`, `Required=No, default 'Y'` here, based on the schema below?

donmccurdy: And to verify — `primaryAxis` has no effect with shapes other than CAPSULE, CYLINDER, and CONE?

donmccurdy:
Did a quick read through — this looks awesome. Do you have an example scene demonstrating use of the extension, by any chance?

Moguri:
@donmccurdy Thanks for the input. [Here](https://github.com/KhronosGroup/glTF/files/1316402/physics_test.gltf.txt) is an example scene (with a txt extension to make GitHub happy). It has a static ground plane that uses a mesh shape, a cube with a box shape, and a cylinder with a cylinder shape.

stevenvergenz:
@Moguri Could you update my contributor's contact information? My email is now "steven.vergenz@microsoft.com".

stevenvergenz:
Also, it would be great if this spec supported compound colliders, i.e. multiple simple colliders per node. Perhaps make the root of the extension an array instead of an object? `"extensions": [ {...}, {...} ]`

Moguri:
@stevenvergenz I have updated your contact info. Also, I agree that compound colliders would be good to add. Some of the properties (e.g., mass, static) do not make sense to have per shape, so I think we need an array of shapes. I believe we can break the properties down as such:

Per shape:
  * collisionShape (might want to rename this, shapeType?)
  * boundingBox
  * primaryAxis
  * mesh
  * offsetMatrix
  * offsetRotation
  * offsetScale
  * offsetTranslation

Per node:
  * mass
  * static
  * collisionGroups
  * collisionMasks
  * collisionShapes/shapes


Does that sound good to you?

stevenvergenz:
Can engines do compound colliders of meshes? Seems like it'd be expensive. If not, I'd say the mesh field could be node-level.

Moguri:
@stevenvergenz I believe Bullet supports compound shapes with mesh shapes (e.g., triangle mesh, convex hull). I do not know what the performance is like.

donmccurdy:
CannonJS has compound shapes as well. It’s a significant performance improvement over using a triangle mesh. There was some work a while back to add compound physics shape support in Blender, but I don’t think that ever landed?

Moguri:
@donmccurdy Blender's game engine can do compound shapes, but Blender itself does not as far as I am aware. 

@donmccurdy and @stevenvergenz I have updated the spec and blendergltf for compound shapes.

@pjcozzi What are the next steps for moving this along? Waiting for more feedback?

donmccurdy:
I think what I had seen was this (experimental) implementation of compound shapes in Blender: https://wiki.blender.org/index.php/User:Sergof/GSoC2013/Documentation

pjcozzi:
> @pjcozzi What are the next steps for moving this along? Waiting for more feedback?

Did you decide to collaborate with the physics effort in #1135 or did you plan to keep this separate?

pjcozzi:
@Moguri did you see this update by @jratcliff63367:

https://github.com/KhronosGroup/glTF/issues/1135#issuecomment-346946312

What is your preference for moving this forward?  Does separate extensions make sense or is it possible to collaborate on one modular extension?

Moguri:
@pjcozzi 
Unfortunately, I think it makes more sense to continue as separate extensions at this point. For more details, you can take a look at my comment: https://github.com/KhronosGroup/glTF/issues/1135#issuecomment-348274045. If you have thoughts on the matter, I would love to hear them. Also, has there been much discussion on versioning glTF extensions for expanding feature sets? Do glTF extensions just follow the OpenGL convention of creating a new extension with an increment number to update the spec (e.g., ARB_transform_feedback2).

@Kupoman @donmccurdy @stevenvergenz 
I believe this extension should be renamed to make clear the more specific intentions of this extension. Something like BLENDER_physics_collisions or BLENDER_physics_rigid_bodies. We may also want to add a few more rigid body physics controls such as:
* Linear damping
* Angular damping
* Linear factor (this is what Bullet uses to lock linear velocity to particular axes)
* Angular factor (this is what Bullet uses to lock angular velocity to particular axes)
* Enable/disable gravity

We may also want controls such as friction and restitution, but maybe these are better suited for a physics materials extension?

pjcozzi:
> Also, has there been much discussion on versioning glTF extensions for expanding feature sets? Do glTF extensions just follow the OpenGL convention of creating a new extension with an increment number to update the spec (e.g., ARB_transform_feedback2)

The current thinking is to follow the OpenGL convention.

pjcozzi:
@Moguri is this extension in use?  Do you want this reviewed and merged - or has anything changed in the past year?

Moguri:
@pjcozzi I am still using the extension, but I do not know of others using it. However, I believe it really should be split into two extensions (collisions versus rigid bodies) as I mentioned in a previous comment. We could probably close this PR as I get the new extensions written up. Then I can file a couple of new PRs.

pjcozzi:
Thanks @Moguri, I'll close this PR to keep things tidy.  Please open new ones as your time allows.

extensions/Vendor/BLENDER_physics/README.md
JoshKlint: If there is an offsetMatrix, why bother having a primaryAxis value?

extensions/Vendor/BLENDER_physics/README.md
```diff
@@ -0,0 +1,182 @@
...
+|**mesh**|`glTF id`|The ID of the mesh to use for `CONVEX_HULL` and `MESH` collision shapes|No, default: `node's mesh` if it exists, otherwise use `BOX` shape|
+|**offsetMatrix**|`array`|A 4x4 transform matrix applied to the collision shape in addition to the node's transform|No, default: `[ 1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0 ]`|
+|**offsetRotation**|`array`|A rotation offset (as a quaternion) applied to the physics shape in addition to the node's rotation|No, default: `[0.0, 0.0, 0.0, 1.0]`|
+|**offsetScale**|`array`|A non-uniform scale offset applied to the collision shape in addition to the node's scale|No, default: `[1.0, 1.0, 1.0]`|
```
JoshKlint: What is the purpose of the offsetScale, if the shape's dimensions are already defined? The translation and rotation offsets, yes, you need those, but is a scale value actually needed?

extensions/Vendor/BLENDER_physics/README.md
```diff
@@ -0,0 +1,182 @@
...
+The properties available are listed in the table below.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**collisionShape**|`string`|The shape a physics simulation should use to represent the node|No, default: `BOX`|
+|**mass**|`number`|The 'weight', irrespective of gravity, of the node|No, default: `1.0`|
+|**static**|`boolean`|Whether or not the Node should be moved by physics simulations|No, default: `false`|
+|**boundingBox**|`array`|The dimensions of the local (i.e., does not include the node's transform) bounding box of the collision shape centered on the origin|Yes|
```
JoshKlint: I take it this gives the dimensions on the XYZ axes? Perhaps "size" or "scale" would be a better name for this?

extensions/Vendor/BLENDER_physics/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+**Collision Shapes**
+
+Below are the allowed values for `collisionShape`.
+When consuming this extension, an application should construct the `collisionShape` with parameters that best fill the `boundingBox`.
+The shape offset should be applied to the shape prior to the node's transform.
+
+* `BOX`
+* `SPHERE`
+* `CAPSULE`
```
JoshKlint: Perhaps note that the capsule's height must be equal or greater to the diameter.

extensions/Vendor/BLENDER_physics/README.md
```diff
@@ -0,0 +1,182 @@
...
+* `CYLINDER`
+* `CONE`
+* `CONVEX_HULL`
```
JoshKlint: What are the required components for the mesh? Is a point cloud sufficient?

PR #1105, opened 2017-09-21
Update minimal-gltf-loader status

shrekshao:
Finished most features of gltf 2 for the loader. Take off that 'geometry-only' hat.


pjcozzi:
Very nice @shrekshao!

PR #1115, opened 2017-10-06
Added unity collada/gltf exporter

RemiArnaud:


pjcozzi:
Thanks @RemiArnaud!

The links both redirect to https://www.assetstore.unity3d.com/en/.  Do I need to be logged in or do the links need to be updated?

emackey:
These links redirect for me even after logging into the asset store.

RemiArnaud:
looks like the asset store is broken at the moment..

pjcozzi:
@RemiArnaud can you provide new links?  I was able to get to it when I searched and it came up with a different URL.

RemiArnaud:
@pjcozzi @emackey - Asset store when back up, and I created two new links

Please give this a try.

Thank you

pjcozzi:
Looks good, thanks!

PR #1118, opened 2017-10-11
missing INT

sakrist:
Missing enum for INT type in scheme specification.

### DIFF
--- a/extensions/2.0/Vendor/MSFT_lod/README.md
+++ b/extensions/2.0/Vendor/MSFT_lod/README.md
@@ -1,0 +1,128 @@
+# MSFT\_lod 
+
+## Contributors
+
+* Saurabh Bhatia, Microsoft
+* Gary Hsu, Microsoft
+* Adam Gritt, Microsoft
+* John Copic, Microsoft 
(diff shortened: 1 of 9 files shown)


bghgary:
I don't believe anything in the core spec uses it, which is likely the reason why it is not included. Are you thinking it should be there for completeness? Perhaps you have an extension in mind that will use it?

sakrist:
> Are you thinking it should be there for completeness?
Yes :)

lexaknyazev:
To give you some historical perspective:
We've decided that rendering of glTF 2.0 core should be implementable on OpenGL ES 2.0 level hardware with couple extensions (like uint indices and texture shader lod). Because of this, 32-bit attributes aren't supported, i.e., even UINT type could be used only for indices. It's likely that future glTF revisions (2.x) will include 32-bit attribute types from OpenGL ES 3.0 (uint, int, and packed).

sakrist:
@lexaknyazev I understand that you guys  was developing glTF mostly for OpenGL use. 
I think this format is very good and should be more flexible for possible use with other graphics API.

PR #1126, opened 2017-10-20
Hint that S means U and T means V in wrapping modes

### LINKED ISSUE #1067: spec documentation nitpick
I am implementing a glTF loader library, but have little-to-no experience with the openGL API (hi~). while my target does use opengl under the hood, I am using the higher-level API provided, so I've got to do a little mental decompiling to get my higher-level primitives to match the lower-level ones glTF seems to be built around.

So I was a little thrown off by the fact that the spec uses the terms wrapS and wrapT in the samplers without ever defining what S and T are. it'd be nice to have a little note that explains that S means U means the horizontal axis and vice versa.

lexaknyazev:
Fixes #1067.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -1025,7 +1025,7 @@ Any colorspace information (such as ICC profiles, intents, etc) from PNG or JPEG
 
 ### Samplers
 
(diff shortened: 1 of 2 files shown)


pjcozzi:
Thanks @lexaknyazev!

PR #1127, opened 2017-10-20
Describe BIN chunk and buffer.byteLength relations

### LINKED ISSUE #1026: buffer.byteLength of GLB-stored buffer
Must the length of GLB-stored `BIN` chunk be equal to `buffer.byteLength`? Spec doesn't define their relations. Since GLB chunk must be padded to 4-byte boundary, converting ASCII glTF to GLB may require adjusting `buffer.byteLength`. 

Otherwise we can say that `buffer.byteLength` must be less then or equal to the length of `BIN` chunk.

CC @pjcozzi @sbtron @bghgary

lexaknyazev:
Fixes #1026.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -450,7 +450,9 @@ Buffers and buffer views do not contain type information. They simply define the
 
 #### GLB-stored Buffer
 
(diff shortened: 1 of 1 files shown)


bghgary:
+1 on the implementation note

lexaknyazev:
@bghgary 
Please review added impl. note.

bghgary:
LGTM

PR #1128, opened 2017-10-20
Mismatched component type for UNSIGNED_BYTE

my2iu:
In some parts of the documentation, it lists 5120 as being the component type for UNSIGNED_BYTE whereas elsewhere it lists 5121 as the component type.


bghgary:
That was my mistake. Thanks for the fix!

pjcozzi:
@my2iu good catch and thanks again for this fix!

PR #1131, opened 2017-10-24
Add more restrictions on indexed attribute semantics

### LINKED ISSUE #1023: Tightening indexed attribute semantics
From [meshes](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes):
> Client implementations must support at least two UV texture coordinate sets, one vertex color, and one joints/weights set. 

Atm, spec doesn't require indexed semantic to start with `0`. This leads to couple tricky cases:
```json
{
    "meshes": [
        {
            "primitives": [
                {
                    "attributes": {
                        "POSITION": 0,
                        "TEXCOORD_1": 1,
                        "JOINTS_1": 2
                    }
                }
            ]
        }
    ]
}
```

1. Could UV set `TEXCOORD_1` be declared without `TEXCOORD_0`?
2. Same question about `JOINTS_1`.
3. Could there be two UV sets `TEXCOORD_1` and `TEXCOORD_2` (let's say that material's texture info uses matching indices)?

Spec has a table with valid semantic names that start with `0`, but it's not as strict as it should be, imho.
Also, we should add that `JOINTS` and `WEIGHTS` semantics must be both present or both absent.

CC @pjcozzi @bghgary @UX3D-nopper @javagl

lexaknyazev:
Fixes #1023.


pjcozzi:
Looks good!  @sbtron please merge when ready.

Issue #882, opened 2017-03-19
Animation sampler rotation format in 2.0


stevenvergenz:
> Each of the animation's samplers defines the input/output pair: a set of floating point scalar values representing time; and a set of ... four-component floating-point vectors representing rotation.

What format should these rotation vectors take? Are they Euler angles? Quaternions? What order should the components come in? I'm guessing these were intended to be WXYZ quaternions, but they could also be XYZW quaternions. Some clarity here would be helpful.

McNopper:
They are for sure quaternions and in the order XYZW.

pjcozzi:
@McNopper is correct, but I'm sure if this is explicit enough in the spec.  There is this for `node.rotation`: https://github.com/KhronosGroup/glTF/tree/2.0/specification/1.0#noderotation

@lexaknyazev perhaps we should make the spec more clear, both 1.0 and 2.0?  Probably the same is true for translation and scale in keyframes.

pjcozzi:
I meant "I'm **not** sure..."

stevenvergenz:
It might be sufficient to add something to the Units section, declaring that all rotations are expressed as quaternions in XYZW order.

pjcozzi:
Will update 2.0 [Units section](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#coordinate-system-and-units) to include info on rotation, translation, and scale.

Will update 2.0 schema/reference doc for animations to include info like the table in animations (https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#animations):

![image](https://user-images.githubusercontent.com/782098/31783010-c77113b4-b4ca-11e7-916b-dfef6ee232f1.png)

[wetzel](https://github.com/AnalyticalGraphicsInc/wetzel) is used to generate the reference doc from the schema - or a manual update is probably OK here. (#1013)

Thanks @javagl for making the update.

pjcozzi:
Updated in #1136.

PR #1136, opened 2017-10-28
Added info about vector components

javagl:
As discussed for https://github.com/KhronosGroup/glTF/issues/882

The reference documentation is updated manually, but should
be the same as it will be generated by wetzel.


specification/2.0/README.md
```diff
@@ -249,6 +249,16 @@ All angles are in radians.
 
 Positive rotation is counterclockwise.
 
+The [node transformations](#transformations) and [animation channel paths](#animations) are 3D vectors or quaternions with the following data types and semantics:
+
+|Property|Type|Component Type(s)|Description|
+|----|----------------|-----------------|-----------|
+|translation|`"VEC3"`|`5126`&nbsp;(FLOAT)|XYZ translation vector|
+|rotation|`"VEC4"`|`5126`&nbsp;(FLOAT)<br>`5120`&nbsp;(BYTE)&nbsp;normalized<br>`5121`&nbsp;(UNSIGNED_BYTE)&nbsp;normalized<br>`5122`&nbsp;(SHORT)&nbsp;normalized<br>`5123`&nbsp;(UNSIGNED_SHORT)&nbsp;normalized|XYZW rotation quaternion|
```
bghgary: I think only FLOAT is allowed?

From https://github.com/javagl/glTF/blob/e637a434badd7558dd70bc54ed0e863f58d59a5d/specification/2.0/README.md#transformations
> rotation is a FLOAT_VEC4 unit quaternion value, (x, y, z, w), in the local coordinate system.

bghgary: Oh, I see where you got that from. Looks like we have a discrepancy here. TRS on a node are stored in the JSON where as TRS stored for animations are in a buffer.

javagl: Indeed, the component types refer to the animation channel paths. Should I just omit this column, making it just  

    Property     Type?        Description
    translation  3D vector    Translation (xyz)...
    rotation     4D vector    Quaternion (xyzw)...
    scale        3D vector    Scaling (xyz)...
    
or maybe omit the `Type` altogether? (Then I'd likely convert it into a bulletpoint list...)

bghgary: I don't have a strong preference. Bulleted list will be okay with me without the types. People who want the specifics should look at the corresponding links.

@lexaknyazev On a side note (I'm looking at the supported list for animation quaternions and weights), is normalized UNSIGNED_BYTE/UNSIGNED_SHORT useful for quaternions and normalized BYTE/SHORT useful for weights? Only positive values for quaternion seem unlikely and I believe weights are always positive?

lexaknyazev: Afair, we've decided to not limit those since exporters could optimize data for some cases while implementations will likely be able to read all kinds of accessors anyway.

emackey:
Merging this, since the original request was approved some time ago, and now there are additional corrections to the overview graphic.  These got accidentally mixed in the same branch, but we still want to take them.

PR #1137, opened 2017-10-30
EXT_animation_map extension proposal

msfeldstein:
I'd like to propose an extension we use at facebook to mark up animations to have semantic meanings, so client applications can know what to do with all the animations packaged inside a model.  The simplest use case is to target an animation to be the 'enter' animation, and clients will know what to play when the model first appears (think a flower opening, or person waking up).  More complicated use cases could be specifying idle/walk/die/birth animations for game assets.

It would be wonderful if this could be shared across the ecosystem so models can be more dynamic without requiring scripting.


pjcozzi:
Awesome! We'll help spread the word to get you some feedback.

RemiArnaud:
Related collada extension already available (maya) https://www.khronos.org/collada/wiki/Animation_clip_OpenCOLLADA_extension

space2:
In your example you mapped two animations (0 and 3) to the "enter" semantic. In such case (multiple animations mapped to the same semantic) how the animations will be played:

* in sequence (i.e. play anim 0 and then anim 3), or
* randomly selecting one?

lexaknyazev:
- Extension semantics refer to an "object". What does it mean? Arbitrary glTF asset could contain multiple meshes/nodes/scenes.

- glTF doesn't define any "official" way of looping animations, i.e., it's not defined how to interpolate between the last frame and the first frame. Maybe, desired behavior should be stated in the extension spec.

- Is it OK to have several bindings with the same semantic?

---

@pjcozzi We need to register `FB` extension prefix.

msfeldstein:
@space2 I added some clarification.  I believe they should all be played at once, though we could add a type modifier to allow you to play them in sequence, or random if desired.  I think playing at once default is good because you could imagine a glTF scene with multiple meshes and a separate animation targetting each.

@lexaknyazev by "object" i mean a literal json object in the array, not a glTF object.  Is there a more clear term i could use here?

> glTF doesn't define any "official" way of looping animations, i.e., it's not defined how to interpolate between the last frame and the first frame. Maybe, desired behavior should be stated in the extension spec.

What interpolation would there be?  Since there is 0 time between the last frame and first frame i would assume no interpolation.

> Is it OK to have several bindings with the same semantic?

I think so.  Do you see any problems with that?  I could see this being useful if we add the 'sequential' play modes suggested by @space2, so that you could have multiple sequential animation lists.

msfeldstein:
also how does naming prefix work? This is an extension proposed by us at FB but i don't think there's anything FB specific about it.

msfeldstein:
@lexaknyazev @space2 perhaps we should default to sequential playback of the animations in each binding object animation array, and just let people add multiple entries for one semantic to play them in parallel.

lexaknyazev:
My question is about this "object":
> Triggered when an object is picked up, for example grabbed in vr, or dragged in WebGL.

Does it mean the whole scene? Or should implementation perform lookup from `animation.channel.target` to determine affected objects?

---

> What interpolation would there be? Since there is 0 time between the last frame and first frame i would assume no interpolation.

That means that for looping animations, the first and the last frames must have the same values, right?

msfeldstein:
Ah i see, yes that is unclear.  I was assuming the target of the animation channel but now that i think about it that could add undue complexity to the clients (for an animation that targets a bunch of bones, and we expect clients to union all those bounding boxes for those bones meshes).

Perhaps it would be better to supply a target Node for any animation with interactivity?

>  the first and the last frames must have the same values

Yes i think that's up to the animation author to create properly looping animations, i don't think it should be the job of the engine to try to fix that.  I could also a non-seamless animation, think a bubble that rises from the ground over and over again, and it looks like a new bubble each time.

lexaknyazev:
> Perhaps it would be better to supply a target Node for any animation with interactivity?

Yes.

> we expect clients to union all those bounding boxes for those bones meshes

There was an effort to add pre-computed bounding boxes to skinned meshes. Maybe we should revisit that.

msfeldstein:
I added an optional `node` list for interactive semantics: 

`The following semantics require a list of glTF Node objects to specify interaction targets.  If the node is omitted, it can be assumed that the entire glTF Scene is the target.`

Does this seem satisfactory?

Pre-computed bounding boxes seems very useful, we're already fighting some performance problems trying to compute things on the client (we will get around this by pre-computing it on the server but it would be nice to have it already available in most models).

lexaknyazev:
Please use plural for array (`node` -> `nodes`).

I assume that any node from that list should trigger the animation, since it could be difficult for user to, e.g., grab 3 objects at the same time. On the other hand, there could be use cases for simultaneous conditions (e.g., gaze at several objects at once).

msfeldstein:
updated.  that assumption is correct, i specify that any node can trigger the event.  The use case for simultaneous conditions can be added with another property (simultaneous: true or something) but i think we can leave that out for simplicity now.

pjcozzi:
> @pjcozzi We need to register FB extension prefix.

See #1139

> also how does naming prefix work? This is an extension proposed by us at FB but i don't think there's anything FB specific about it.

@msfeldstein a vendor prefixed extension like `FB_` is for an extension that is created by a vendor that is only implemented by the vendor.  An `EXT_` prefix is for extension with multiple implementations.  The `KHR_` prefix is for Khronos ratified extensions that go through Khronos' IP process.

For this extension, it is very likely that engines like three.js, Babylon, and Cesium would be interested in implementing it so we might end up wanting to use the `EXT_` prefix to avoid namespace churn later - if there are multiple implementations before this is merged.

extensions/Vendor/FB_animation_map/schema/FB_animation_map.schema.json
```diff
@@ -0,0 +1,38 @@
...
+        "semanticBinding": {
+            "type": "object",
+            "properties": {
+                "semantic": {
+                    "type": "string"
+                },
+                "animations": {
+                    "type": "array",
+                    "items": {
+                        "type": "number",
```
lexaknyazev: `number` -> `integer`

extensions/Vendor/FB_animation_map/schema/FB_animation_map.schema.json
```diff
@@ -0,0 +1,38 @@
...
+                        "minimum": 0
+                    }
+                },
+                "nodes": {
+                    "type": "array",
+                    "items": {
+                        "type": "number",
```
lexaknyazev: `number` -> `integer`

lexaknyazev:
To follow existing glTF schema usages, please consider adding these properties to `animations` and `nodes` definitions:
- `"uniqueItems": true`
- `"minItems": 1`

sbtron:
I think we might be interested in implementing this too. Would it make sense to do a multivendor extension like EXT?

Are there any thoughts on allowing compound semantics like
- semantic1 && semantic2 
- semantic1 || semantic2
- semantic1 && !semantic2

/cc @thmignon @najadojo

msfeldstein:
I would be delighted to make this EXT_.

I have not thought of the compound semantics.  The OR case is already doable with the current spec, just have 2 separate entries.  My initial thought on things like AND NOT is that we're getting into logic rather than entity description.  What are some good concrete examples of where we would want to do this?

donmccurdy:
As semantics can be any string, it seems like clients could define their own combinations like `"PROXIMITY_AND_GAZE"` and interpret as they choose. But I'm having a hard time thinking of use cases from the example semantics, without getting into a much larger scope like Unity's state transitions.

najadojo:
Having a [Mixed Reality Model](https://developer.microsoft.com/en-us/windows/mixed-reality/creating_3d_models_for_use_in_the_windows_mixed_reality_home) where an author can express an animation to play when the customer is near and a separate animation that plays when near and gazing at the model.  The idea here is proximity can be used to gather a customer's attention and then switch a to different attraction when attention has been gained. This transition may not make sense if the model is across the room from the viewer.

msfeldstein:
Yeah i think for complicated cases like this it might make sense to just create a merged semantic.  I think its important to define the simple ones ahead of time, but many of them might come naturally (just make sure to post here or wherever when you implement them so we can keep things somewhat aligned if possible).

msfeldstein:
But i do see the PROXIMITY_AND_GAZE as kind of a PAY_ATTENTION or INSPECT or LOOK_CLOSELY semantic which i think is well deserving of it's own semantic vs a merging of 2

najadojo:
I does seem to me that we'll quickly get into a potential for a large number of combinations making more challenging for an author to create the scene desired.  PROXIMITY_AND_GAZE is one example but then you have PROXIMITY_ENTER_AND_GAZE, PROXIMITY_LEAVE_AND_GAZE, and so on.  Perhaps having a "loop" : Boolean field could solve this particular problem.

najadojo:
Also the proposal is silent on behaviors when two semantics' match which animate the same properties. If just PROXIMITY means animate thing X to value Y and PROXIMITY_AND_GAZE means animate thing X to value Z we have conflicting animations and one must win.  An author could define what happens here by saying PROXIMITY && !GAZE for one and PROXIMITY && GAZE for the other.  I believe this can extend to the other proposal semantics as well HELD && GAZE with HELD && !GAZE.  

Yes, a state machine specification would be welcome.

msfeldstein:
Good point on multiple animations with one target, the options are to add them, to average them, or to stomp them right?

donmccurdy:
> Yes, a state machine specification would be welcome.

I guess what I meant was that state machines are at a much higher level of abstraction than glTF. I do not think that glTF should include any concept of a state machine; that is better done as a wrapping format that references glTF assets, similar to [Cesium's 3D Tiles](https://github.com/AnalyticalGraphicsInc/3d-tiles) wrapping glTF to add geospatial metadata. 

Logical expressions for semantics feel like they're somewhere in between on that scale of abstraction... they could be in an extension, or in some wrapping format.

> options are to add them, to average them, or to stomp them right?

I suppose you could also do weighted averages, or crossfade from `animation_a` to `animation_b` over `duration_d`, but this quickly feels like a more complex extension. 🙃

stevenvergenz:
Why should these mappings go at the top level, instead of extending the animations themselves? Seems more intuitive to say:

```json
"animations": [{
    "channels": [],
    "samplers": [],
    "extensions": {
        "EXT_animation_map": {
            "semantic": "TRIGGER",
            "nodes": [0]
        }
    }
}]
```

msfeldstein:
I suppose that would work too.  I think having semantics be a top level element is more extensible though, you would lose the ability to do things like sequential vs parallel animations for one semantic.  Do you find other benefits to doing it the way you propose other than human readability?

stevenvergenz:
Oh good point. I suppose if it's a many-to-many mapping, it has to be in its own structure. Or else extend both animations and nodes.

thmignon:
@donmccurdy feedback taken on the level of abstraction concerns. I think we're okay with leaving that out of this initial rev and still adopting the baseline proposal with the tweaks discussed here. This keeps it simple and avoids unresolved edge cases or baking too many assumptions in. 

We'd love to still go with the multi vendor EXT_ on this one @msfeldstein . The only remaining proposal we have is adding a pointer concept to the initial proposed list of triggers. This would let you trigger animations based on controller pointing ray or hand pointing gestures. 
+* **POINTER_ENTER**: Triggered when pointer intersects an object
+* **POINTING**: Looped while a user is pointing at an object
+* **POINTER_LEAVE**: Triggered when a pointer gets exits an object 

CC/ @sbtron @najadojo

msfeldstein:
@thmignon updated to use EXT_ and added your pointer semantics

thmignon:
@msfeldstein thanks! I noticed that the example in the readme also needs to be updated to ext_.

msfeldstein:
Lets get it merged!  We haven't shipped an implementation yet but are still planning to, we're happy with where its at

msfeldstein:
do you know what else needs to be done to merge?

thmignon:
@sbtron will be reviewing with the rest of the working group when they begin meeting again next week

msfeldstein:
Excellent let me know if you need anything from me

lexaknyazev:
The extension spec says
> Animations should be played all at once

Does it mean that in the case like the following one all **four** animations must be playing at the same time? If so, the description should be more sound.
```json
{
    "extensionsUsed": {
        "EXT_animation_map"
    },
    "extensions" : {
        "EXT_animation_map" : {
              "bindings": [
                  {
                      "semantic": "ENTER",
                      "animations": [0, 3]
                  },
                  {
                      "semantic": "ENTER",
                      "animations": [1, 2]
                  }
              ]
        }
    }
}
```

msfeldstein:
yes, it looks like you're applying 4 animations to the ENTER semantic.  I'm not sure how to change the description for clarity do you have a suggestion?

punkoffice:
Just wondering, has this been added to Facebook? If so, is there any documentation on how to apply animations (armature / vertex) and incorporate it into a GLB for Facebook?

donmccurdy:
@punkoffice No, Facebook does not yet support animation from the core glTF spec. This extension (not required for animation in glTF) in particular should not be implemented unless and until it is out of draft status.

najadojo:
I created a test asset [semantics.zip](https://github.com/KhronosGroup/glTF/files/2060503/semantics.zip) to exercise the implementation of this extension in Windows Mixed Reality.

msfeldstein:
It's implemented in Windows MR home?

najadojo:
Yes we added it in our most recent update. Capabilities documented at https://docs.microsoft.com/en-us/windows/mixed-reality/creating-3d-models-for-use-in-the-windows-mixed-reality-home#animation-guidelines

msfeldstein:
Now that there is a shipped implementation, what can we do to finalize and solidify this extension?

extensions/Vendor/EXT_animation_map/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+It maps a semantic, which can be any string a client wishes to bind, to an array of animation indices, pointing to a list of gltf Animation.
+
+##### Example Semantics
+
+* **ENTER**: Trigger a single shot animation when the object first enters the scene or viewport
+* **LEAVE**: Trigger a single shot animation as the object is leaving the scene or viewport
+* **ALWAYS**: Constantly loop an animation
+
+The following semantics require a list of glTF Node objects to specify interaction targets.  The event is triggered by acting on any of the nodes listed.  If the nodes are omitted, it can be assumed that the entire glTF Scene is the target.
```
donmccurdy: I find this confusing — the section is labeled "example semantics", which leads me to believe it is not normative, but the language here claims these semantics "require" a list of nodes and presumably must be used in a particular way. 

I'm guessing that none of this section's content it is meant to be normative, but either way I think this should be spelled out somewhat.

extensions/Vendor/EXT_animation_map/schema/EXT_animation_map.schema.json
```diff
@@ -0,0 +1,42 @@
+{
+    "$schema" : "http://json-schema.org/draft-04/schema",
+    "title" : "EXT_animation_map extension",
+    "type" : "object",
+    "definitions": {
+        "semanticBinding": {
+            "type": "object",
```
donmccurdy: I'm not sure how to read these schema definitions so I'll just ask... can a bindings object have `extras`?

donmccurdy (commented):
Because the scope here is narrow and multiple vendors are interested in implementing, I think my concerns above can be considered resolved.

This extension is unusual compared to many others, in the sense that most of its implementation details are left undefined. That's OK with me, as I assume it will be most useful for application-specific situations, and we're guaranteed that models using the extension will load fine in engines that don't support it. But still, we should try to track and codify standard semantics and their use, if that seems to be turning into a pseudo-standard of its own.

extensions/Vendor/EXT_animation_map/README.md
msfeldstein: Ah i see the confusion here.  The semantics above can operate without any target nodes since they act on the entire object, whereas the ones below need some more info (which nodes to act on).  Perhaps better wording would be "Semantics can have a list of nodes indicating which nodes the semantic should act upon if needed".  Lemme try to rewrite this

extensions/Vendor/EXT_animation_map/schema/EXT_animation_map.schema.json
msfeldstein: i believe anything in a gltf schema can have extras right?

msfeldstein:
@donmccurdy thoughts on these changes? open to other wordings.  It might also make sense to formalize a subset, things like ENTER/LEAVE/ALWAYS are very simple, and useful in almost any context i can think of, while things like GRAB are only for vr-with-hands and probably not ready to be formalized.

msfeldstein:
actually after talking to @sbtron it sounds like it might be better to formalize the semantics listed, and let any other usages be considered non-normative, updated the spec to be stronger in that wording

najadojo:
In Windows MR we now have the ability to specify an animation reset state on the binding by appending `_RESET_ON_START,` `_RESET_ON_STOP,` or `_DONT_RESET` to any of the semantic names provided above. We could transition this to an extras object easily but I think we should have an easy way to add a sematic without editing this spec.

msfeldstein:
ah that might be worth specifying in spec (your addition makes things like enter animations not-compliant).  Can you elaborate on what those suffixes mean?  I assume DONT_RESET means to hold the final frame of the animation when it finishes, which would be good to add a how does RESET_ON_START work?

CSS calls this animation-fill-mode:forward, THREE calls it clampWhenFinished:true

najadojo:
The reset refers to the key frame time. When the suffix is `RESET_ON_START` the animation starts at time 0 and the initial pose is applied, the default behavior. When the suffix is `DONT_RESET` the animation starts at the last time it was running at and that pose is applied, if an animation was stopped before it completes it will pickup where it left off. If `DONT_RESET` is applied to a completed animation starting it again will do nothing as its already at its final keyframe. When the suffix is `RESET_ON_STOP` and the animation plays to completion the time will be set at time 0 and the initial pose is applied again.

extensions/Vendor/EXT_animation_map/schema/EXT_animation_map.schema.json
lexaknyazev: In that case, `semanticBinding` definition needs to contain
```json
"allOf": [ { "$ref": "glTFProperty.schema.json" } ],
```

msfeldstein:
Honestly i don't have enough info about exactly how this should work, i figured things like an ENTER animation will have its final frame match the rest position, so i didn't consider something like this.  Perhaps we can move that to a separate property or extra and try to finalize it later?

extensions/Vendor/EXT_animation_map/schema/EXT_animation_map.schema.json
msfeldstein: done, thanks

najadojo:
An extra is fine for this data but it highlights "semantic" is only meaningful in the context in which its applied and custom values should be allowed. That is unless we more completely define the semantics and how they apply to different client conditions. For example I assume ENTER, POINTER_ENTER, and GAZE_ENTER all mean the same thing to the FB feed viewing of a glTF. What does PROXIMITY mean in that context or any 3D in 2D context? I'd also like to have a way that we can extend semantics easily. say if we want to use a joystick or other gesture to start playback.

msfeldstein:
ENTER means the model has entered the scene, POINTER_ENTER is a user
interaction.  Would it be clearer if i renamed ENTER to APPEAR?

pjcozzi:
@donmccurdy are we close to being able to merge this as a draft?

donmccurdy:
I'm not opposed to merging this as a draft, but not sure whether the thread above is resolved. Will leave that to @msfeldstein and @najadojo, fine to merge when they're ready. Some of the suffixes, e.g. `_DONT_RESET`, sound like they could be properties of a binding rather than suffixes of a semantic, IMO.

Also wondering if it might be worth following vertex attribute semantics in requiring application-specific semantics to be prefixed with a `_`, thoughts?

msfeldstein:
It sounds like @najadojo is cool with moving the suffix stuff to an extra property which would unblock that.

The underscore suffix for non-standard semantics seems fine.

@najadojo you ask "For example I assume ENTER, POINTER_ENTER, and GAZE_ENTER all mean the same thing to the FB feed viewing of a glTF. What does PROXIMITY mean in that context or any 3D in 2D context?"

ENTER is different from POINTER_ENTER and GAZE_ENTER, pointer enter and gaze enter don't mean anything in the feed context since there is no pointer or gaze.  It should just be ignored.  Same with PROXIMITY, if you're not in ar/vr or have some other way to understand proximity then that animatino would be ignored.

pjcozzi:
@msfeldstein @najadojo is this extension in use at Facebook?  Is this PR ready to merge or does it need more discussion or changes?

msfeldstein:
It hasn’t shipped in any products at Facebook. I believe Microsoft is using
it however

najadojo:
That is correct Windows Mixed Reality is using this extension. Though we do have some current limitations and modifications that are not part of this extension description. Our documentation can be found at [Animation guidelines](https://docs.microsoft.com/en-us/windows/mixed-reality/creating-3d-models-for-use-in-the-windows-mixed-reality-home#animation-guidelines).

pjcozzi:
Thanks for the quick response, @msfeldstein @najadojo!

> Though we do have some current limitations and modifications that are not part of this extension description. Our documentation can be found at Animation guidelines.

@sbtron @najadojo do you think someone at Microsoft would be willing to update this so we can merge it?

msfeldstein:
Closing: no longer working on this. Focused on Grok Bot / Sand.

PR #1140, opened 2017-11-01
Add makeglb

sbtron:
Add makeglb to list of optimizers


pjcozzi:
Awesome, thanks @sbtron!

Issue #1043, opened 2017-07-19
Forward vector convention


bghgary:
The current text in the README.md says this:
> glTF uses a right-handed coordinate system, that is, the cross product of X and Y yields Z. glTF defines the y axis as up.

Which way is forward? This is apparently a lot of discrepancies between viewers, tools, and even the samples in the sample repo.

My understanding of the state of viewers/tools

> Maya/Max: right-handed, +Z is forward
> Toolbag: right-handed, +Z is forward
> Blender: right-handed, +Z is up, -Y is forward
> Unity: left-handed, +Z is forward
> Unreal: left-handed, +Z is up, +X is forward
> 

glTF is right-handed. Should -Z be forward or +Z be forward? Without this definition, it is hard for authors to decide which way to place the content.

Edit: Blender is -Y forward

emackey:
This is also mentioned in `README.md` for the [TextureCoordinateTest](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/TextureCoordinateTest) model.

> Looking at the POSITION attributes min/max, you can see that the left side is +X, the top is +Y, the back plane is +Z (facing away from the default camera), and the default camera sits on the -Z side looking toward the origin with +Y up. The "front" of the model faces -Z, and this is reflected in the NORMAL accessors of all except the back plane.

This model does not contain a camera.  It correctly faces the camera in BabylonJS, ThreeJS, and the [glTF-WebGL-PBR](https://github.com/KhronosGroup/glTF-WebGL-PBR) reference.

I suggest we embrace this behavior as correct, as many sample models face this direction.

lexaknyazev:
Just for convenience, from [Cameras](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#cameras):
> A camera defines the projection matrix that transforms from view to clip coordinates. The projection can be perspective or orthographic. Cameras are contained in nodes and thus can be transformed. Their world-space transformation matrix is used for calculating view-space transformation. The camera is defined such that the local +X axis is to the right, the lens looks towards the local -Z axis, and the top of the camera is aligned with the local +Y axis. If no transformation is specified, the location of the camera is at the origin.

emackey:
> the lens looks towards the local -Z axis

So, the spec says to look the opposite way from most of the implementations, unfortunately.

> the location of the camera is at the origin.

[...]

PR #1147, opened 2017-11-07
Update coordinate system and units section with forward convention.

bghgary:
See https://github.com/KhronosGroup/glTF/issues/1043#issuecomment-339132697


bghgary:
@pjcozzi @lexaknyazev @emackey @javagl Would like some feedback on the wording and/or if we need to explain this more.

bghgary:
I'll look into adding something, though the boom box model is a good example already as it has a clear front and back.

pjcozzi:
A screenshot with overlaid axes of any of these models in the spec would be good.

bghgary:
Ah, okay. I'll check with tech art.

bghgary:
@pjcozzi How's this? I can add a sample model for this as well.

pjcozzi:
Ah, good idea for the sample model.

The figure looks good but shouldn't we label the axes with X, Y, and Z?

bghgary:
We talked about this. We thought it was clear enough that red is X, green is Y, and blue is Z, but labels won't hurt.

cc @PatrickRyanMS

bghgary:
@pjcozzi I've updated it with axis labels.

pjcozzi:
Perfect!

PR #1151, opened 2017-11-14
Fill Appendix B with PBR math

snagy:
Filled out the PBR appendix with a baseline implementation that matches the implementation in the glTF-WebGL-PBR example.


emackey:
Thanks @snagy!

PR #1155, opened 2017-11-17
fix broken link

06wj:
Three.js  renamed `GLTF2Loader` to `GLTFLoader`. 
see [#11961](https://github.com/mrdoob/three.js/pull/11961)


donmccurdy:
Thanks!

PR #1162, opened 2017-11-24
Update A-Frame entry in Loaders and Viewers table

danmindru:
Hi there,
As per [A-Frame's documentation](https://aframe.io/docs/0.7.0/components/gltf-model.html), it supports glTF 2.0 (and in fact recommends it).


pjcozzi:
Nice!  Thanks @danmindru!

PR #1163, opened 2017-11-24
KHR_materials_unlit

donmccurdy:
Updated version of #1095 as alternative to #1150 (KHR_materials_common).

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_unlit/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_unlit/README.md
@@ -1,0 +1,144 @@
+# KHR\_materials\_unlit
+
+## Contributors
+
+* Don McCurdy, Google, [@donrmccurdy](https://twitter.com/donrmccurdy)
+* Norbert Nopper, UX3D, <mailto:nopper@ux3d.io>
+* Gary Hsu, Microsoft, [@bghgary](https://twitter.com/bghgary)
+* Scott Nagy, Microsoft, [@visageofscott](https://twitter.com/visageofscott)
+* Pär Winzell, Facebook, [@zellski](https://twitter.com/zellski)
+* Michael Feldstein, Facebook, [@msfeldstein](https://twitter.com/msfeldstein)
+* Robert Long, Mozilla, [@arobertlong](https://twitter.com/arobertlong)
+
+## Status
+
+Complete
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension defines an unlit shading model for use in glTF 2.0
+materials, as an alternative to the Physically Based Rendering (PBR) shading
+models provided by the core specification. Three motivating uses cases for
+unlit materials include:
+
+* Mobile devices with limited resources, where unlit materials offer a
+performant alternative to higher-quality shading models.
+* Photogrammetry, in which lighting information is already present and
+additional lighting should not be applied.
+* Stylized materials (such as "anime" or "hand-drawn" looks) in which lighting is
+undesirable for aesthetic reasons.
+
+These use cases are not mutually exclusive: artists may pick an unlit material
+for performance reasons, and make aesthetic decisions to complement that
+choice. As a result, client implementations capable of rendering PBR should not
+automatically "upgrade" to fully-shaded PBR. Any core PBR properties specified
+on an unlit material (except `baseColor`) are meant only as fallbacks for
+clients that do not support the `KHR_materials_unlit` extension. The extension,
+whether required or optional within the asset, indicates preference for an
+unlit visual style.
+
+## Extending Materials
+
+The common Unlit material is defined by adding the
+`KHR_materials_unlit` extension to any glTF material. When present, the
+extension indicates that a material should be unlit and use available
+`baseColor` values, alpha values, and vertex colors while ignoring all
+properties of the default PBR model related to lighting or color. Alpha
+coverage and doubleSided still apply to unlit materials.
+
+```json
+{
+    "materials": [
+        {
+            "name": "MyUnlitMaterial",
+            "pbrMetallicRoughness": {
+                "baseColorFactor": [ 0.5, 0.8, 0.0, 1.0 ]
+            },
+            "extensions": {
+                "KHR_materials_unlit": {}
+            }
+        }
+    ]
+}
+```
+
+### Definition
+
+The Unlit material model describes a constantly shaded surface that is
+independent of lighting. The material is defined only by properties already
+present in the [glTF 2.0 material specification](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#material).
+No new properties are added by this extension — it is effectively a boolean
+flag indicating use of an unlit shading model. Additional properties on the
+extension object are allowed, but may lead to undefined behaviour in conforming
+viewers.
+
+Color is calculated as:
+
+```
+color = <baseColorTerm>
+```
+
+`<baseColorTerm>` is defined by the [core glTF material specification](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#metallic-roughness-material).
+
+### Example
+
+The following example defines an unlit material with a constant red color.
+
+```json
+"materials": [
+    {
+        "name": "red_unlit_material",
+        "pbrMetallicRoughness": {
+            "baseColorFactor": [ 1.0, 0.0, 0.0, 1.0 ]
+        },
+        "extensions": {
+            "KHR_materials_unlit": {}
+        }
+    }
+]
+```
+
+### Extension compatibility and fallback materials
+
+When possible, authoring tools should provide a fallback material definition
+and mark the `KHR_materials_unlit` extension as optional. Models
+including the extension optionally will still render in all clients that
+support the core glTF 2.0 specification, falling back to a PBR metal/rough
+material. In clients that do not support the extension, direct lighting will
+be applied and performance may be degraded.
+
+Defining a fallback to the default PBR shading model is achieved by setting
+appropriate parameters on the default material.
+
+```json
+"materials": [
+    {
+        "name": "unlit_with_fallback",
+        "pbrMetallicRoughness": {
+            "baseColorFactor": [ 1.0, 1.0, 1.0, 1.0 ],
+            "baseColorTexture": { "index": 0 },
+            "roughnessFactor": 0.9,
+            "metallicFactor": 0.0
+        },
+        "extensions": {
+            "KHR_materials_unlit": {}
+        }
+    }
+]
+```
+
+> **Implementation Note:** For best fallback behavior in clients that do not
+> implement the `KHR_materials_unlit` extension, authoring tools may use:
+>
+> * `metallicFactor` is `0` and `emissiveFactor` is `[0, 0, 0]`.
+> * `roughnessFactor` is greater than `0.5`.
+> * Omit `metallicRoughnessTexture`, `occlusionTexture`, `emissiveTexture`,
+> and `normalTexture`.
+
+> **Implementation Note:** When `KHR_materials_unlit` is included with another
+> extension specifying a shading model on the same material, the result is
+> undefined.
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -4,11 +4,11 @@
 
 #### Khronos extensions
 * [KHR_materials_pbrSpecularGlossiness](2.0/Khronos/KHR_materials_pbrSpecularGlossiness/README.md)
+* [KHR_materials_unlit](2.0/Khronos/KHR_materials_unlit/README.md)
 * [KHR_draco_mesh_compression](2.0/Khronos/KHR_draco_mesh_compression/README.md)
 
 #### Draft Khronos extensions
 _Draft Khronos extensions are not ratified yet._
-* KHR_materials_unlit *(in progress)*
 * KHR_lights *(in progress)*
 * KHR_technique_webgl *(in progress)*
 * KHR_texture_transform *(in progress)*
@@ -34,7 +34,7 @@ _Draft Khronos extensions are not ratified yet._
 
 # About glTF Extensions
 
-glTF extensions extend the base glTF model format.  Extensions can introduce new properties (including properties that reference external data, and the extension can define the format of those data), new parameter semantics, reserved IDs, and new container formats.  Extensions are written against a specific version of glTF and may be promoted to core glTF in a later glTF version. 
+glTF extensions extend the base glTF model format.  Extensions can introduce new properties (including properties that reference external data, and the extension can define the format of those data), new parameter semantics, reserved IDs, and new container formats.  Extensions are written against a specific version of glTF and may be promoted to core glTF in a later glTF version.
 
 ## Extension Mechanics
 
@@ -78,7 +78,7 @@ This allows an engine to quickly determine if it supports the extensions needed 
 
 To create a new extension, use the [extension template](Template.md) and open a pull request into this repo.  Make sure to add the extension to the glTF Extension Registry (top of this file).
 
-If lack of extension support prevents proper geometry loading, extension specification _must_ state that (and such extension must be mentioned in `extensionsRequired` top-level glTF property).  
+If lack of extension support prevents proper geometry loading, extension specification _must_ state that (and such extension must be mentioned in `extensionsRequired` top-level glTF property).
 
 Extensions start as a vendor extension, then can become a multi-vendor extensions if there are multiple implementations, and can become a ratified Khronos extension (the multi-vendor extension is an optional step).
 
--- a/extensions/2.0/Khronos/KHR_materials_unlit/schema/glTF.KHR_materials_unlit.schema.json
+++ b/extensions/2.0/Khronos/KHR_materials_unlit/schema/glTF.KHR_materials_unlit.schema.json
@@ -1,0 +1,11 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_materials_unlit glTF extension",
+    "type": "object",
+    "description": "glTF extension that defines the unlit material model.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "extensions": { },
+        "extras": { }
+    }
+}


lexaknyazev:
@donmccurdy 
`baseColorFactor` and friends are defined within `pbrMetallicRoughness` object. Please update "compatibility" section to reflect that.

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+The Constant material model describes a constantly shaded surface that is independent of lighting. The material is defined only by properties already
+present in the [glTF 2.0 material specification](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#material): no new properties are added by this extension.
+
+Additional properties are allowed, but may lead to undefined behaviour in conforming viewers.
+
+The material is intended for unlit shading. Color is calculated as:
+
+```
+color = <emissiveTerm> + <occlusionTerm> * <aL>
```
snagy: Should an unlit mode include indirect light?  How indirect does the light need to be before it gets included?

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+Each term is defined as follows:
+
+* `emissiveTerm` – emissiveFactor * emissiveTexture
+    *  [*From core glTF material specification.*](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#additional-maps)
+* `occlusionTerm` – occlusionFactor * occlusionTexture
+    *  [*From core glTF material specification.*](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#additional-maps)
+* `aL` – Ambient light
```
snagy: The glTF core spec refers to 'indirect light' and not 'ambient light' when the occlusion term is mentioned.

It's pretty vague about what indirect light means, though.

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,111 @@
...
+            "KHR_materials_cmnConstant": {}
+        }
+    }
+]
+```
+
+### Extension compatibility and fallback materials
+
+When possible, authoring tools should provide a fallback material definition
+and mark the `KHR_materials_cmnBlinnPhong` extension as optional. By doing so,
```
snagy: This should reference KHR_materials_cmnConstant and not BlinnPhong.

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+```json
+"materials": [
+    {
+        "name": "unlit_with_fallback",
+        "emissiveFactor": [ 1.0, 1.0, 1.0 ],
+        "emissiveTexture": { "index": 0 },
+        "baseColorFactor": [ 0.0, 0.0, 0.0, 1.0 ],
+        "metallicFactor": 0.0,
+        "roughnessFactor": 0.0,
```
snagy: roughnessFactor of 0.0 gives a perfectly shiny surface, which is generally not a good default value.

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,111 @@
...
+]
+```
+
+### Extension compatibility and fallback materials
+
+When possible, authoring tools should provide a fallback material definition
+and mark the `KHR_materials_cmnBlinnPhong` extension as optional. By doing so,
+models created with the extension will still render correctly in all clients
+that support the core glTF 2.0 specification, although clients are less likely
+to optimize their shading model performance in this case.
```
snagy: This sentence should be rewritten for Constant instead of BlinnPhong.  The fallback material will probably have a different visual look for Constant, as well as worse performance.

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension defines an unlit ("constant") shading model for use in glTF 2.0 materials, as an alternative to the Physically Based Rendering (PBR) shading models provided by the core specification. In mobile devices with limited resources, unlit materials offer a performant alternative to higher-quality shading models.
+
+## Extending Materials
+
+The common Constant material is defined by adding the `KHR_materials_cmnConstant` extension to any glTF material. When present, the extension indicates that a material should use the Constant shading model. Unrelated default values in the default PBR shading model, such as `diffuseFactor`, should be ignored.
```
snagy: 'diffuseFactor' is called 'baseColorFactor'

snagy:
Pulling over a conversation from #1150 on whether the unlit values should be in base color or emissive.

1) Putting it in emissive will instantly work in almost all the current gltf renderers (with a little specular highlighting in engines that don't support the extension). However, it probably won't really work correctly with a renderer that actually emits light from the emissive channel, and it can't be defined in vertex data.
2) Putting it in base color will kinda work in renderers that don't support the extension; but, the visual look will diverge a lot more between renderers that are extension-aware and those that aren't.
3) Defining it in a separate property is clean and lets the exporting program decide how non-supporting renderers will work. My main concerns with this were intent (what should a renderer do if they encounter both? is unlit an optimization or the preferred look?), and whether it actually gets us a win in most situations. Renderers still need to be aware of the extension to handle things properly, and instead of coming to a general consensus about the preferred way to handle the drawbacks of the above two cases, we're leaving it to the exporters to figure out their own solution and thus potentially making it more difficult for future renderers to consistently interpret content from different sources.

I lean towards base color, with the assumption that most renderers will support this extension, and base gives us unlit vertex colors.

zellski:
The requirement that vertex colour functionality has a reasonable fallback seems to me to tower above the other considerations here -- so I am in favour of #2 as well. The low-poly vertex-colour aesthetic is one of the primary use cases for an unlit extension, aye?

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: ✅

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: Can't think of any reason to diverge from the glTF 1.0 extension here, https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_materials_common#constant

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: The distinction between ambient and direct lights is fairly well defined in most engines, yes? If there are notable exceptions to that, I would think we can leave this to interpretation.

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: ✅

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: ✅

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: ✅

donmccurdy:
Updated and applied feedback, including use of base color — thanks all.

I've added an explicit note on vertex colors as well.

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Extending Materials
+
+The common Constant material is defined by adding the
+`KHR_materials_cmnConstant` extension to any glTF material. When present, the
+extension indicates that a material should use the Constant shading model.
+Unrelated default values in the default PBR shading model, such as
+`roughnessFactor`, should be ignored.
+
+```
```
bghgary: put ```json here for syntax coloring

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,118 @@
...
+appropriate parameters on the default material.
+
+```json
+"materials": [
+    {
+        "name": "unlit_with_fallback",
+        "pbrMetallicRoughness": {
+            "baseColorFactor": [ 1.0, 1.0, 1.0, 1.0 ],
+            "baseColorTexture": { "index": 0 },
+            "roughnessFactor": 1.0
```
bghgary: `metallicFactor` defaults to `1.0` and having a fully rough and fully metal material is not typical. Might consider setting `metallicFactor` to `0.0`.

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,118 @@
...
+Term definitions:
+
+* **emissiveTerm** – `emissiveFactor` * `emissiveTexture`
+    *  [*From core glTF material specification.*](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#additional-maps)
+* **occlusionTerm** – `occlusionTexture.strength` * `occlusionTexture`
+    *  [*From core glTF material specification.*](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#additional-maps)
+* **baseColorTerm** – `baseColorFactor` * `baseColorTexture` * `vertexColor`
+    *  [*From core glTF material specification.*](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#metallic-roughness-material)
+* **aL** – Indirect light
+    * *Constant sum of all ambient light contributions from the scene.*
```
zellski: If we're saying that we're expecting baseFactor/baseTexture/vertexColor to define our unlit material (while tolerating emissive), then we shouldn't be multiplying it by **aL**, right? Unless the term's meaning here is meant to be entirely implicit.

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,118 @@
...
+models provided by the core specification. In mobile devices with limited
+resources, unlit materials offer a performant alternative to higher-quality
+shading models.
+
+## Extending Materials
+
+The common Constant material is defined by adding the
+`KHR_materials_cmnConstant` extension to any glTF material. When present, the
+extension indicates that a material should use the Constant shading model.
+Unrelated default values in the default PBR shading model, such as
```
bghgary: Can we be explicit about exactly which values should be ignored or is that something that is up to implementation?

extensions/Khronos/KHR_materials_cmnConstant/README.md
snagy: I was thinking about engines that use pseudo-directional lighting such as a hemispherical fill or environmental cube map instead of a flat ambient light.  They can just decide on their own interpretation.

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: Added explicit comment on `metallic` and `roughness`. 

I'm not sure what a client should do if it sees KHR_materials_cmnConstant _and_ KHR_materials_pbrSpecularGlossiness. Inclined to leave that up for application choice.

extensions/Khronos/KHR_materials_cmnConstant/README.md
snagy: I second this, I don't think occlusion or ambient light should be part of this spec.

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: ✅

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: ✅

extensions/Khronos/KHR_materials_cmnConstant/README.md
donmccurdy: Not sure I follow — are you asking whether an unlit material should ignore ambient light in addition to direct?

donmccurdy: Just saw @snagy's comment, the comment makes sense then —

On whether occlusion or ambient should be included, I'd point out that:

1. We did include them in the [constant material from glTF 1.0](https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_materials_common#constant)
2. [THREE.MeshBasicMaterial supports both](https://threejs.org/docs/#api/materials/MeshBasicMaterial)
3. Disallowing ambient light has no particular performance benefit.

Is there a benefit to omitting occlusion and ambient light?

zellski: It made sense to have an `aL` factor in the glTF 1.0 version because in that version we leaned on `emissive` to produce the unlit effect, and `emissive` is immune to scene light. That role is now to be played by baseColor & friends, instead. A material whose canonical visibility factor is multiplicatively attenuated by a light factor doesn't qualify as unlit, does it?

That is, unless we say that clients that do not have the concept of ambient light should assume a default value of 1.0... but that seems like a crazy can of worms...

extensions/Khronos/KHR_materials_cmnConstant/README.md
bghgary: What about `metallicRoughnessTexture`? What about vertex attributes like `NORMAL`, `TANGENT` and `normalTexture`? Should these be ignored?

bghgary: > KHR_materials_cmnConstant and KHR_materials_pbrSpecularGlossiness

Yeah, this one is a bit tricky. My gut feel is that this should not happen. If an asset has an unlit material, it probably wants to be fast, and I can't think of a good reason to have specular glossiness there also.

donmccurdy:
@snagy @zellski I'm moving this reply to a top-level comment for visibility —

I see what you mean; `ambient` in glTF 1.0 was a factor but did not allow a texture ID. In that case color should be:

```
color = <baseColorTerm>
```

And there's no point in differentiating direct vs indirect light, so that can be removed from the text as well. This seems [consistent with UE4](https://docs.unrealengine.com/latest/INT/Engine/Rendering/Materials/MaterialProperties/LightingModels/#unlit), also.

> A material whose canonical visibility factor is multiplicatively attenuated by a light factor doesn't qualify as unlit, does it? ~@zellski 

Yeah that had bothered me, but "constant" still made sense with ambient light... after doing some web searches on "constant shading" I didn't find much, and what I found isn't consistent:

screen shot 2017-12-11 at 3 08 19 pm

^[Essential computer animation fast](https://books.google.com/books?id=yt4LBwAAQBAJ&pg=PA44&lpg=PA44&dq=constant+shading+model&source=bl&ots=3aNi2aVkLz&sig=ot3cE8Ar6uKAKjBgMxbcu_I_sKg&hl=en&sa=X&ved=0ahUKEwjf7NnSiIPYAhUV72MKHTR8C-4Q6AEIXTAI#v=onepage&q=constant%20shading%20model&f=false)

[Here's a slide deck](https://www.slideshare.net/akbrightfuture/shading-10305107) also using "constant shading" interchangeably with flat shading.

[Another source](https://www.siggraph.org/education/materials/HyperGraph/scanline/shade_models/constant.htm) writes "In the Constant shading model we compute one shade or color for the entire object, e.g., there really is no shading being done," which would still be true with ambient lighting applied.

In that case, perhaps we should rename this to something like `KHR_materials_unlit`?

extensions/Khronos/KHR_materials_cmnConstant/README.md
```diff
@@ -0,0 +1,119 @@
...
+appropriate parameters on the default material.
+
+```json
+"materials": [
+    {
+        "name": "unlit_with_fallback",
+        "pbrMetallicRoughness": {
+            "baseColorFactor": [ 1.0, 1.0, 1.0, 1.0 ],
+            "baseColorTexture": { "index": 0 },
+            "metallicFactor": 0.0,
```
bghgary: Now that I think about this more. I think we should note that the fallback should not have `metallicRoughnessTexture` and `metallicFactor` should be 0 and `roughnessFactor` should be 1. Same for `normalTexture` depending on what we decide for that.

donmccurdy: Ok. Any thoughts on whether that should be normative, or simply an implementation note for recommended fallback behavior?

bghgary: I think since this is already in a fallback section it doesn't make sense to be an implementation note. Normative is my vote.

donmccurdy: Following this through, does this look right?

- metallicRoughnessTexture CANNOT be present
- metallicFactor MUST be 0
- roughnessFactor MUST be 1

- occlusionTexture CANNOT be present
- emissiveTexture CANNOT be present
- emissiveFactor MUST be 0
- normalTexture CANNOT be present
- NORMAL and TANGENT attributes ...
  - cannot be present?
  - may be present, but must be ignored during lighting calculations?

Inclined to say NORMAL attribute should not be disallowed, but ignored in shading. Arguably it could be used for things other than shading e.g. for aligning a reticle against the surface. Same is arguably true of `normalTexture` though...

bghgary: > Inclined to say NORMAL attribute should not be disallowed, but ignored in shading. Arguably it could be used for things other than shading e.g. for aligning a reticle against the surface. Same is arguably true of `normalTexture` though...

+1 for the NORMAL attribute. I don't think the same is true for the `normalTexture`. It would be weird if there is no visible changes for the `normalTexture` but something else (like a reticle or physics) did something with it.

snagy: Why should we dictate what the factors should be for the fallback case?  If I want a metallic or nonmetallic fallback, or shiny or rough, or whatever, I should be able to define that, right?

If you handle the extension you ignore everything that's not part of the lighting calculation above, and if you don't (or don't wish to) use the extension, you can use the fallback values.

snagy: Also, huge +1 to renaming this KHR_materials_unlit

I never even considered that constant would mean anything but unlit, so maybe we should be explicit in the name and remove any references to 'constant'.

bghgary: Perhaps this should be an implementation note then

snagy:
Does anyone have expected use cases for this extension that they'd like to share?  It might give us a little more clarity on exactly how we should suggest people handle things like fallback cases.

msfeldstein:
We have two use cases we envision.  The first is the most certain that we need.

1) Things like photogrammetry, where the lighting of the world is baked in since it's being captured via a camera.  These shouldn't have any extra lighting applied to it (unless we have some amazing method of removing world light).

2) Models which have super high quality ray-traced SSS caustic etc lighting baked in

zellski:
One of the projects I'm involved has the requirement that framerates stay high (in VR), even on weaker mobile devices. Rather than design for the high-end, and then be forced to implement progressively shittier fallbacks for old phones, we just rolled with the situation and opted into a low-poly, unlit vertex-colour art style from the beginning. And yet we've standardized completely around glTF 2.0, so I'm really eager for an extension that allows us to formally encode that art style decision in the model files, and have it be clear what it's intended to communicate: no dynamic light effects at all.

donmccurdy:
Updates:

* Renamed as `KHR_materials_unlit`
* Removed occlusion, emissive, and ambient light terms
* Added implementation note on recommended fallback property values

extensions/Khronos/KHR_materials_unlit/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+```json
+"materials": [
+    {
+        "name": "unlit_with_fallback",
+        "pbrMetallicRoughness": {
+            "baseColorFactor": [ 1.0, 1.0, 1.0, 1.0 ],
+            "baseColorTexture": { "index": 0 },
+            "metallicFactor": 0.0,
+            "roughnessFactor": 1.0
```
bghgary: `roughnessFactor` defaults to `1.0` and thus is not needed. I believe it will be flagged as a warning in the validator. Same goes for `baseColorFactor`, but I would recommend changing the values so that they are not the default like in the "red_unlit_material" above.

Nehon:
> Does anyone have expected use cases for this extension that they'd like to share?
Sky boxes,
Could be useful for particles... sometimes
Could be useful for GUI.

robertlong:
Other than @bghgary's comment, looks good to me. Thanks Don!

extensions/Khronos/KHR_materials_unlit/README.md
zellski: I agree it would be better to choose a fallback example here that doesn't use extreme values, like roughness 1.0. The fallback materials shouldn't necessarily try to look as close to individually "unlit" as possible, right? Rather, we know the fallback materials are only used in the situation where we're in a fully lit, PBR environment. Better to accept and embrace that fact and pick a fallback material that looks attractively natural in such a setting.

In that vein, while keeping it simple, perhaps drop roughness to 0.8, include a normal map?

amenzies:
I just want to second @msfeldstein's use case of Photogrammetry.  In this case we often have textures which have been painstakingly constructed to accurately represent the object as it was lit in the real world at the time of capture.  I suspect that there are a number of other cases in which an "unlit" material would be useful, such as achieving particular effects or for iconography.  

Another case I can think of is tilesets which have been pre-rendered.  For example, imagine you want to render the moon with accurate shadows.  You often have low incident light creating shadows from tall mountains that can be quite far away.  Real time shadows are impractical due to the distance and precision characteristics of the shadow caster.  In this case you might choose to bake in pre-rendered shading and shadow into your dataset.  I can also imagine tilesets that want to achieve a particular aesthetic, such as the 3D equivalent of street map tiles, that might want to have full control over their lighting.

It's such a dead simple material it would be a shame if it were impossible to achieve.

extensions/Khronos/KHR_materials_unlit/README.md
bghgary: I believe all the scenarios for an unlit material can be categorized into two buckets.

1. The unlit material is the primary material and the core PBR material is the fallback. In this case, if the client implementation can render unlit, it should render with unlit. The fallback core PBR material should be author to match as closely as possible to the simulate what an unlit material would do.

2. The core PBR material is the primary material and the unlit material is the fallback. In this case, if the client implementation can render core PBR metallic/roughness model, then it should render with PBR metallic/roughness model, even if there is an unlit extension. The fallback, likely due to runtime resources, is to render using unlit material as it is much cheaper to render. In this scenario, the core PBR material may have additional maps, various metallic/roughness values, etc. to support the higher fidelity material.

All of the scenarios that have been mentioned so far, including photogrammetry, seems to point to 1, but I think 2 is still possible and we shouldn't restrict someone from doing it.

Maybe we can incorporate this kind of language into this section? It would all be non-normative.

bghgary: > fallback core PBR material authored to match as closely as possible to simulate what an unlit material would do

To be clear, what I mean here is that the core PBR material should match what this PR currently already states:

> **Implementation Note:** For best fallback behavior in clients that do not
> implement the `KHR_materials_unlit` extension, authoring tools may use:
>
> * `metallicFactor` and `emissiveFactor` are `0`.
> * `roughnessFactor` is `1`.
> * Omit `metallicRoughnessTexture`, `occlusionTexture`, `emissiveTexture`,
> and `normalTexture`.

amenzies: This sounds good to me

bghgary: To add to this a bit more, in scenario 2, there is no reason to use the unlit extension at all. The client implementation can fallback to unlit if necessary on its own.

zellski: I am content to withdraw my hesitation above re: roughness 1.0 vs 0.8 and so forth. It's fine for this spec to recommend a somewhat extreme fallback material.

donmccurdy: I removed `roughnessFactor=1.0` from the example, as that's the default value anyway.

Perhaps we just want to loosen the recommendation in the implementation note?

From:

> `roughnessFactor` is `1`.

to something like:

> `roughnessFactor` is `> 0.5`

donmccurdy:
Thanks everyone for the feedback and use cases. 🙂

Current status of the PR is:

- [x] Schema and use of `baseColor` values and vertex colors seem solid at this point
- [ ] More discussion of use cases is welcome, as this could affect fallback recommendations
- [ ] Recommended fallback material should be clarified further
- [ ] One additional prototype implementation would be helpful to check things over (thanks @robertlong for the three.js and FBX2glTF examples!)

donmccurdy:
The PR currently recommends the following fallback behavior:

> **Implementation Note:** For best fallback behavior in clients that do not
> implement the `KHR_materials_unlit` extension, authoring tools may use:
>
> * `metallicFactor` and `emissiveFactor` are `0`.
> * `roughnessFactor` is `1`.
> * Omit `metallicRoughnessTexture`, `occlusionTexture`, `emissiveTexture`,
> and `normalTexture`.

Is there anything we want to change about this? Perhaps suggesting `roughnessFactor=1.0` is too specific per @zellski's comment; we could advise >= 0.5, or make no recommendation?

robertlong:
I've updated the THREE.GLTFLoader and FBX2glTF PRs to the newest spec.

https://github.com/mrdoob/three.js/pull/13136
https://github.com/facebookincubator/FBX2glTF/pull/61

donmccurdy:
Another question — we've specified that `emissive` must be ignored if present, but strictly speaking emissiveness is not incompatible with unlitness, e.g. if the user wants to do post-processing.

But IMO, post-processing and glow are not within our main use cases like high-performance mobile VR or photogrammetry,  and three.js doesn't support emissives on MeshBasicMaterial anyway, so I'd lean toward continuing to ignore emissive here sticking with `color = <baseColorTerm>`. 😅

donmccurdy:
On use cases and fallback behavior, here is one clarification per discussion with @bghgary  —

***

While a primary use case for `KHR_materials_unlit` is to enable performance benefits in mobile clients, client implementations capable of rendering PBR should not automatically "upgrade" to PBR shaders. Any core PBR properties specified on an unlit material (except `baseColor`) are meant only as fallbacks for clients that do not support the KHR_materials_unlit extension. The `KHR_materials_unlit` extension, whether required or optional within the asset, indicates preference for an unlit visual style.

***

^Should this be part of the specification, or just an implementation note? Other feedback or wording also welcome.

takahirox:
I want this note in Motivation chapter because

- engines/viewers can know they should support this extension in light (non-heavy) processing way
- we can remove this extension when (we think) mobile device's spec become high enough

msfeldstein:
> we can remove this extension when (we think) mobile device's spec become high enough

Please don't, our main motivation for this extension has nothing to do with performance.

takahirox:
Ah, ok. That's "one" primary use case, I see.

donmccurdy:
Great clarification, perhaps the overview section should also just make some mention of performance-agnostic motivations like photogrammetry.

takahirox:
Agreed. I want some more motivation mention to clarify.

donmccurdy:
I've extended the overview section as follows:

> This extension defines an unlit shading model for use in glTF 2.0 materials, as an alternative to the Physically Based Rendering (PBR) shading models provided by the core specification. Three motivating uses cases for unlit materials include:

> * Mobile devices with limited resources, where unlit materials offer a performant alternative to higher-quality shading models.
> * Photogrammetry, in which lighting information is already present and additional lighting should not be applied.
> * Stylized materials (such as "anime" or "hand-drawn" looks) in which lighting is undesirable for aesthetic reasons.

> These use cases are not mutually exclusive: artists may pick an unlit material for performance reasons, and make aesthetic decisions to complement that choice. As a result, client implementations capable of rendering PBR should not automatically "upgrade" to fully-shaded PBR. Any core PBR properties specified on an unlit material (except baseColor) are meant only as fallbacks for clients that do not support the KHR_materials_unlit extension. The extension, whether required or optional within the asset, indicates preference for an unlit visual style.

As well as loosening the language around recommended roughness fallback:

> * `roughnessFactor` is greater than `0.5`.

bghgary:
Is there a sample model with this extension that can be used to test implementations?

donmccurdy:
Here are two:

- https://khr-materials-unlit-ooddguuati.now.sh/#model=https://storage.googleapis.com/blocks-town/tmp/ftm_unlit/scene.gltf
- https://khr-materials-unlit-ooddguuati.now.sh/#model=https://cdn.rawgit.com/robertlong/three.js/dccc4242c43d4a42b90fff7183d9873945816374/examples/models/gltf/BotSkinned/glTF-MaterialsUnlit/Bot_Skinned.glb

bghgary:
Thanks!

bghgary:
The spec says:
> When present, the extension indicates that a material should be unlit and use available baseColor values and vertex colors, ignoring other properties of the default PBR model.

1. Vertex color still multiplies against baseColor correct?
2. Is alphaMode/alphaCutoff ignored and thus the material is always opaque?

donmccurdy:
> 1. Vertex color still multiplies against baseColor correct?

Yes, no change here.

> 2. Is alphaMode/alphaCutoff ignored and thus the material is always opaque?

I would expect alpha to to be used, too. Worth making that explicit, perhaps?

bghgary:
> I would expect alpha to be used, too. Worth making that explicit, perhaps?

Yes. As it is currently stated, it sounds like alpha should be ignored.

donmccurdy:
Added a note on alpha coverage. 👍

bghgary:
What about `doubleSided`? Is that allowed?

donmccurdy:
I'd think so, yes. And to generalize a bit, maybe we should be more explicit than the current language, `...ignoring other properties of the default PBR model.` How about:

```
... ignoring all properties properties of the default PBR model related to lighting or color.
```

bghgary:
That's sounds fine, but I prefer if we spelled out explicitly what properties are used to avoid misinterpretation.

donmccurdy:
That's just these two things, right? How about:

> ... ignoring all properties properties of the default PBR model related to lighting or color. Alpha coverage and `doubleSided` still apply to unlit materials. 

If not this, could you propose the wording?

donmccurdy:
Implementation in Blender exporter:

https://github.com/KhronosGroup/glTF-Blender-Exporter/pull/160

lexaknyazev:
@donmccurdy 
Could the extension spec have an implementation note about other material extensions (i.e., SpecGloss)? Should such combination be disallowed or should unlit extension override all other material extensions?

donmccurdy:
I'm not sure how to reason about all other possible material extensions. No material extension exists today that would sensibly be used alongside `KHR_materials_unlit`, and without such an extension I wouldn't feel comfortable prohibiting use of KHR_materials_unlit with another material extension, or claiming that this should always take precedence... For the example of spec/gloss, specifically, I can't imagine why an author would include both, or what behavior they would expect.

Does it seem reasonable to leave this explicitly undefined? E.g.

> When the `KHR_materials_unlit` is included with another extension specifying a shading model on the same material, the result is undefined.

lexaknyazev:
I'd say that explicitly undefined behavior is fine for now (/cc @pjcozzi).
My point is that JSON could contain anything and engines don't know anything about author's intentions, so they would have to make a decision which material model to use. Consider something like this:
```
if (material.specGloss)
    useSpecGloss()
else if (material.unlit)
    useUnlit()
else
    useMR()
```

donmccurdy:
Yes, makes sense. In the future I could imagine us doing something like specifying that order of precedence is based on order in `extensionsUsed` array, but it feels premature to decide now.

bghgary:
@donmccurdy 

> How about:
>> ... ignoring all properties properties of the default PBR model related to lighting or color. Alpha coverage and doubleSided still apply to unlit materials.
> If not this, could you propose the wording?

That's probably good for now. We can always update it if it is unclear.

donmccurdy:
Sounds good. Added language about alpha coverage, `doubleSided`, and undefined behavior when used alongside other material extensions.

extensions/2.0/Khronos/KHR_materials_unlit/README.md
```diff
@@ -0,0 +1,143 @@
...
+appropriate parameters on the default material.
+
+```json
+"materials": [
+    {
+        "name": "unlit_with_fallback",
+        "pbrMetallicRoughness": {
+            "baseColorFactor": [ 1.0, 1.0, 1.0, 1.0 ],
+            "baseColorTexture": { "index": 0 },
+            "metallicFactor": 0.0
```
fernandojsg: what about adding `"roughnessFactor": 0.5` as fallback value as recommended on the implementation note below?

extensions/2.0/Khronos/KHR_materials_unlit/README.md
```diff
@@ -0,0 +1,143 @@
...
+            "KHR_materials_unlit": {}
+        }
+    }
+]
+```
+
+> **Implementation Note:** For best fallback behavior in clients that do not
+> implement the `KHR_materials_unlit` extension, authoring tools may use:
+>
+> * `metallicFactor` and `emissiveFactor` are `0`.
```
fernandojsg: Maybe is clear enough already but maybe being even more explicit as `emissiveFactor` is a vector?:  `metallicFactor is 0 and emmisiveFactor is [0, 0, 0]`.

lexaknyazev:
> Additional properties on the extension object are allowed, but may lead to undefined behaviour in conforming viewers.

For formal completeness, we should provide a simple JSON-Schema like this:
```json
{
    "title": "KHR_materials_unlit glTF extension",
    "type": "object",
    "description": "glTF extension that defines the unlit material model.",
    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
    "properties": {
        "extensions": { },
        "extras": { }
    }
}
```

extensions/2.0/Khronos/KHR_materials_unlit/README.md
donmccurdy: Done 👍

extensions/2.0/Khronos/KHR_materials_unlit/README.md
donmccurdy: Done, using `0.9` here 👍

donmccurdy:
> For formal completeness, we should provide a simple JSON-Schema like this:

Done ✅

lexaknyazev:
Online validator has initial support for this extension. Please, drop some models into it.

@donmccurdy Is there any merit in warning on non-recommended fallback values?

donmccurdy:
Great thanks! Getting some errors in Firefox and Chrome, `Cannot read property 'style' of null`, though. I'd lean against warning about the fallback values, but curious what others think.

lexaknyazev:
> Getting some errors in Firefox and Chrome
Please force-reload web page.

donmccurdy:
Working now, thanks!

fernandojsg:
I'd say also to avoid warning about the fallback, as with the current traction of engines implementing gltf I believe most of them would at least use the fallback values in their implementation itself and hopefully in a near future most of them will support the extension, so we don't need to waste space on the files for these values.

extensions/README.md
```diff
@@ -8,7 +8,7 @@
 
 #### Draft Khronos extensions
 _Draft Khronos extensions are not ratified yet._
-* KHR_materials_common *(in progress)*
+* [KHR_materials_unlit](2.0/Khronos/KHR_materials_unlit/README.md)
```
pjcozzi: Please move this to the non-draft section above. 😄

PR #1166, opened 2017-11-28
Add gltf-import-export npm package

sbtron:
Also rearranging order so that 2.0 tools are on top.


pjcozzi:
Nice @sbtron @najadojo!

PR #1169, opened 2017-12-04
Tiny language tweak.

zellski:
I think these words in the sparse accessor section just got a little jumbled at some point.


emackey:
Thanks @zellski.

PR #1171, opened 2017-12-05
Remove Catmull-Rom spline from the specification

bghgary:
The working group last week agreed that Catmull-Rom should be removed from the spec. This is a PR to remove it. Let's discuss how to proceed.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -1413,7 +1413,7 @@ The following examples show expected animations usage.
 
 When `node` isn't defined, channel should be ignored. Valid path names are `"translation"`, `"rotation"`, `"scale"`, and `"weights"`.
 
-Each of the animation's *samplers* defines the `input`/`output` pair: a set of floating point scalar values representing linear time in seconds; and a set of vectors or scalars representing animated property. All values are stored in a buffer and accessed via accessors; refer to the table below for output accessor types. Interpolation between keys is performed using the interpolation method specified in the `interpolation` property. Supported `interpolation` values include `LINEAR`, `STEP`, `CATMULLROMSPLINE`, and `CUBICSPLINE`. See [Appendix C](#appendix-c-spline-interpolation) for additional information about spline interpolation.
+Each of the animation's *samplers* defines the `input`/`output` pair: a set of floating point scalar values representing linear time in seconds; and a set of vectors or scalars representing animated property. All values are stored in a buffer and accessed via accessors; refer to the table below for output accessor types. Interpolation between keys is performed using the interpolation method specified in the `interpolation` property. Supported `interpolation` values include `LINEAR`, `STEP`, and `CUBICSPLINE`. See [Appendix C](#appendix-c-spline-interpolation) for additional information about spline interpolation.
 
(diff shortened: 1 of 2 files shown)


pjcozzi:
Thanks @bghgary!

**Everyone** - did anyone implement Catmull-Rom splines in a glTF exporter or importer?  We are not aware of any in the wild so I think this PR may just be as simple as it currently is with perhaps a brief note about why Catmull-Rom splines were removed.  If there **are** implementations, perhaps we can provide some migration advice.

@javagl does this require any updates to the [glTF Tutorial Series](https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/README.md)?  I think not.

@lexaknyazev @emackey does this require any updates to [glTF Sample Models](https://github.com/KhronosGroup/glTF-Sample-Models)?  I think not.

Is there anything else we should consider before moving this forward?

emackey:
Our sample models don't use anything other than `"interpolation": "LINEAR"`.  😞

emackey:
The Khronos Blender exporter apparently does Cubic splines, not Catmull-Rom, but I haven't tried it yet.

emackey:
Looks like Three at least processes the `CATMULLROMSPLINE` enum, albeit with some other spline implementation, in https://github.com/mrdoob/three.js/pull/11825.  /cc @donmccurdy @zellski 

Looks like this could be fixed by removing one line of code.

javagl:
@pjcozzi  The https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_007_Animations.md sections only shows a `LINEAR` example as well. More details about STEP/CUBICSPLINE could/should be added. But I'm glad that CATMULLROM can be omitted.

pjcozzi:
@emackey ha, sounds good for three.js!  Do you happen to know what Cesium does?

emackey:
Cesium has a `CatmullRomSpline` math class that is never referenced by the glTF implementation.

For glTF, I only see one reference in the code to examining the contents of an `interpolation` field, and here it's only branching `LINEAR` off from all other (apparently spline-based) interpolations.

https://github.com/AnalyticalGraphicsInc/cesium/blob/cb237a32a1c9934d19824e0081828282995291b2/Source/Scene/ModelAnimationCache.js#L128

I don't fully understand the above code yet, but it's at least clear there's no implementation of Catmull-Rom for glTF in Cesium.  The enum name is missing from the source tree.

zellski:
My in-progress spline support for FBX2glTF only uses the full cubics. I was
not on the call but I assume the decision is based on the fact that C-R is
mostly a convenience? Although it could be seen as a compression format too
— right? For curves that satisfy the requirements, in models heavy with
animation data, it could be substantial savings...

bghgary:
@zellski 
> C-R is mostly a convenience?
We had a couple threads discussing this in the working group. The gist is that we didn't know anyone who implemented it and there were issues with overshooting if it's intended for authored animations.

> it could be seen as a compression format
True. The Draco compression team is already looking into this, so we decided it would be best to let them decide what compression method is best for compressing animation curves.

pjcozzi:
Thanks again @bghgary and everyone who looked into this!

Issue #1170, opened 2017-12-05
2.0/figures/dictionary-objects.png many vs single relationships


zellski:
This is definitely minor, but it looks to me like the dictionary-objects.png erroneously suggests:
- that a node can reference multiple meshes (removed between 1.0 and 2.0?)
- that a skin references multiple accessors (was this ever true?)

pjcozzi:
This could have been an oversight when updating the spec for 2.0.  Feel free to open a PR with both fixes.

zellski:
I was hoping whoever originally created the .png might still have some source document. But it not I can take a pass at photoshop copy-and-paste of pixel regions. :-)

lexaknyazev:
@zellski 
Source file is in the repo:
https://github.com/KhronosGroup/glTF/raw/master/specification/2.0/figures/pptx/figures.pptx

zellski:
Well today I learned something. Thank you!

pjcozzi:
I may also be the guilty party who made the figures...many years ago.

pjcozzi:
Fixed in #1175

PR #1175, opened 2017-12-08
Uupdated to 1 node per mesh, 1 accessor per skin.

zellski:
The replaced image suggests a 1-to-N relationship for nodes to meshes, and similarly from skin to accessors.

The font in this PPT export is a little heavier than in the .png being replaced; perhaps because I'm on a Mac. I'm not sure which I prefer; let me know if anybody prefers the older, lighter look.

See #1170.


pjcozzi:
+1.  Thanks!

Anyone else want to look at this?  @lexaknyazev?

bghgary:
The new image seems to be a screenshot of the entire slide which I don't think is what you want. Other than that, LGTM.

zellski:
What! I spent half an eternity cropping it.

I must’ve uploaded the wrong file. Will fix.

zellski:
Okay, let's try this again. Updated to properly cropped version.

pjcozzi:
Thanks again @zellski!

Issue #944, opened 2017-05-06
MIME Type & File Extension For Binary Buffers


lexaknyazev:
glTF 2.0 spec states that buffers with binary data must have `.bin` file extension and `application/octet-stream` MIME Type. 

The latter applies both to external files (e.g., when they're served over HTTP) and Data URI-embedded buffers (because Data URI syntax requires specifying MIME type when it isn't `text/plain`).

Despite that glTF binary buffer is useless without JSON description, usage of such generic MIME Type (`application/octet-stream`) and file extension (`.bin`) could be sub-optimal:

- lack of specific MIME Type complicates network data transfer, e.g., in cases when server wants to use transport-level compression (gzip-over-HTTP);
- lack of specific file extension complicates server's configuration for such cases;
- lack of specific file extension prevents desktop environments from recognizing such files (e.g., to show non-generic icon) because glTF buffers don't have any magic value at the beginning.

Proposal: register a new MIME Type for glTF binary buffers and use it instead of `application/octet-stream`. Also, optionally replace `.bin` file extension with a new one.

According to [RFC 6838, Section 4.2.5](https://tools.ietf.org/html/rfc6838#section-4.2.5), `application/` top-level type seems to be the right place:

> 4.2.5.  Application Media Types
>
>   The "application" top-level type is to be used for discrete data that
>   do not fit under any of the other type names, and particularly for
>   data to be processed by some type of application program.  This is
>   information that must be processed by an application before it is
>   viewable or usable by a user. 
>   ...
>  The subtype of "application" will often either be the name or include
>  part of the name of the application for which the data are intended.

I think that it seems quite consistent to propose `application/gltf-buffer`.

Thoughts?

CC @pjcozzi @sbtron @bghgary

pjcozzi:
CC #943

pjcozzi:
@sbtron OK to close this?

lexaknyazev:
As far as I can see, this isn't registered yet.

pjcozzi:
Ah, yes, of course, I was thinking of `gltf-binary` for .glb.

sbtron:
nope this is still pending just the gltf-binary is done and closed.

lexaknyazev:
@sbtron Any updates?

emackey:
https://www.iana.org/assignments/media-types/media-types.xhtml

https://www.iana.org/assignments/media-types/application/gltf-buffer

It looks like this is officially registered!

```
Additional information:

[...]

PR #1180, opened 2017-12-13
Updated mime type for binary buffer 

sbtron:
The registration request is still pending with IANA but getting the PR started for issue #944 

In addition to switching to a specific MIME type do we also want to change the recommended extension (.bin) to something else?
My current thinking is rather than creating a new extension we can keep the recommended extension to .bin but add an implementation note that the implementations could chose to have a different extension if they wanted. This would essentially mean any glTF loader should not explicitly check for a .bin extension and simply load whatever filename the buffer uri points to.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -164,7 +164,7 @@ Major version updates are not expected to be compatible with previous versions.
 ## File Extensions and MIME Types
 
 * `*.gltf` files use `model/gltf+json`
-* `*.bin` files use `application/octet-stream`
+* `*.bin` files use `application/gltf-buffer`
 * Texture files use the official `image/*` type based on the specific image format. For compatibility with modern web browsers, the following image formats are supported: `image/jpeg`, `image/png`.
 
 ## JSON encoding


emackey:
Consider that most web servers serve static files using a lookup table to map file extensions to HTTP MIME type headers.  One cannot reasonably expect web servers to be reconfigured to serve all `*.bin` files as the glTF bin MIME type.  Most web servers ship pre-configured to serve `.bin` as `application/octet-stream`.

For example, see the 6th entry in this table:  [MDN's Incomplete List of MIME Types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types)

Apache's list: https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types

```
application/octet-stream	bin dms lrf mar so dist distz pkg bpk dump elc deploy
```

We cannot save anything to a `*.bin` file on disk, put it on a web server, and expect the server to serve anything other than `application/octet-stream`.

emackey:
IIS 7 does this too:

```
            <mimeMap fileExtension=".bin" mimeType="application/octet-stream" />
```

donmccurdy:
Do we know if common servers like Apache or Nginx have rules for classic extensions like DAE or FBX?

emackey:
> Do we know if common servers like Apache or Nginx have rules for classic extensions like DAE or FBX?

They do not, and perhaps we should nudge them to add `.glb` and such.

But my objection here is that many servers, including at least Apache and IIS, already do have factory rules for `*.bin`, and it is not OK for us to override that.

donmccurdy:
> But my objection here is that many servers, including at least Apache and IIS, already do have factory rules for *.bin, and it is not OK for us to override that.

Makes sense. Just trying to gauge the likelihood of their handling `.glb` or a possible replacement for `.bin` appropriately.

emackey:
So, to serve glTF correctly, one must manually add the corresponding MIME types to the server configuration.  Moving away from `bin` and `octet-stream` just makes one more custom thing that needs registration.

Cesium registers all glTF-related files [here for IIS](https://github.com/AnalyticalGraphicsInc/cesium/blob/ec542b312d1a50b071813f134a2606804fe3ae44/web.config#L17-L22) and [here for Node Express](https://github.com/AnalyticalGraphicsInc/cesium/blob/ec542b312d1a50b071813f134a2606804fe3ae44/server.js#L47-L50).  Neither one needs to register `bin` to get this working, as `bin` is already registered.

specification/2.0/README.md
```diff
@@ -164,7 +164,7 @@ Major version updates are not expected to be compatible with previous versions.
 ## File Extensions and MIME Types
 
 * `*.gltf` files use `model/gltf+json`
-* `*.bin` files use `application/octet-stream`
+* `*.bin` files use `application/gltf-buffer`
```
emackey: This conflicts with most web servers' factory default settings.

sbtron:
Great points. Open suggestions for an alternative extension
.buf ??

donmccurdy:
couple more ideas in no order

- `.glbin`
- `.glbuf`

emackey:
No particular preference on my part.  I like `.glbin`, since it seems like we could more easily explain the rename from `bin`.   But any of these seems fine.

bghgary:
I don't have a strong preference either, but I prefer the extension to be easily distinguishable from `.glb` since GLB files can also reference external binary files. Both `.glbin` and `.glbuf` start with `.glb`. Is that confusing?

donmccurdy:
> I don't have a strong preference either, but I prefer the extension to be easily distinguishable from .glb

Since `.glb` is *Binary glTF* I could see `.glbin` causing confusion. `.glbuf` seems pretty distinct despite having the same first three letters IMO. Or we could consider more explicit alternatives like `.gltfbuffer`.

pjcozzi:
@sbtron? 😄😄😄

lexaknyazev:
Superseded by #1877.

Issue #1183, opened 2017-12-14
inconsistency around animation/sampler/output requirement


zellski:
So in the formal schema we're told only 'weights' may take an integer component type:

https://github.com/KhronosGroup/glTF/blob/9e6bdca9f7eeeb2eda7cd963c1b2e5e8899ea721/specification/2.0/schema/animation.sampler.schema.json#L35-L39

However, the table in the spec itself suggests the rotation quaternion can also accept normalized integers:

![image](https://user-images.githubusercontent.com/714167/34011777-1590bcda-e0c6-11e7-9914-6b7615bd1b8b.png)

Finally, glTF-Validator [does allow normalized integer rotation](https://github.com/KhronosGroup/glTF-Validator/blob/476e90d67b73bef38a3263d3add1e4b0cbb3c344/lib/src/base/members.dart#L154-L172).

lexaknyazev:
Spec's readme and the validator are correct, description in JSON-Schema should be updated.

zellski:
I'll see about a PR in a bit.

PR #1187, opened 2017-12-15
Allow normalized integers for 'rotation' path

zellski:
Change JSON-Schema to allow rotation animation values to be provided as normalized integers. This is already the law of the land according to the spec's README.md & glTF-Validator.

Fixes https://github.com/KhronosGroup/glTF/issues/1183.


pjcozzi:
Good catch, thanks @zellski!

Issue #1013, opened 2017-06-12
The textureInfo is not linked from pbrMetallicRoughness


javagl:
The [`pbrMetallicRoughness`](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#pbrmetallicroughness) currently refers to `baseColorTexture` and `metallicRoughnessTexture` as being of type `object`. In fact, they are of type `textureInfo`. 

Is this just a minor glitch in the generator? It would be nice if the type could be defined to be `textureInfo`, including the appropriate link. (There is a section for the [`textureInfo`](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#textureinfo), but it is not really used (i.e. linked to) right now...)

pjcozzi:
Discussed offline.  @javagl will see if he can fix [wetzel](https://github.com/AnalyticalGraphicsInc/wetzel) to make the link work.

pjcozzi:
@javagl bump for if/when you have bandwidth.

emackey:
Yesterday I was looking at the diffs between what wetzel puts out and what's embedded in the current 2.0 README.  The biggest differences appeared to be:

* A large number of lowercase titles became capitalized (for example, `bufferView` became `Buffer View`)
* The order has changed, perhaps unintentionally, making some diffs impossible to do in-place
* There do appear to be a number of hand-edits that need reconciliation
* The "child of glTF root" thing appeared, but I think I'm just missing the command-line switch to turn that one off.

This is the commandline I was using:

```
$ bin/wetzel.js -l2 --schemaPath=schema -a=cqo ../glTF/specification/2.0/schema/glTF.schema.json > ../wetzel_output/gltf.md
```

jcuda:
@emackey The capitalization of the titles was done in the *schema*. Unfortunately, this caused some ambiguities and quirks in wetzel: Before this change, the `title` of a schema was analogous to a "type name", and was used as such and referred to as such. This included several formatting routines, where the title was passed in "as-if" it was a type name, causing inconsistent links etc. 

However, I *tried* to handle this in this wetzel commit: https://github.com/javagl/wetzel/commit/0f7eb039c7a1424be65027cc7c293ce47237fbf6

The result of running this over the current spec is shown here:

https://github.com/javagl/glTF/tree/master/specification/2.0#objects

The links seem to work, and the formatting seems to be OK. 

I'm a bit hesitant to create the PRs:

[...]

PR #1193, opened 2017-12-29
Remove the max 1.0 restriction on emissiveFactor

### LINKED ISSUE #1083: HDR values for emissiveFactor?
Hi, 

Would it be possible to allow values greater than 1.0 for `emissiveFactor`?
I didn't find any thread or discussion about this, maybe it has just not been mentionned before.

[The specification](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/material.schema.json) enforces emissiveFactor values to be in [0,1], but Unity allows emissive HDR values in it's Standard materials, we also allow values in [0, 10] on Sketchfab, and it almost obvious that other softwares are allowing it too.

For now, exporter are clamping the values to be compliant with glTF specification, but all "glowing" models are not exported as they should be. 

Thanks!

emackey:
Fixes #1083 

I came across a model using 1.5 for the `emissiveFactor`, seems to work fine in the JS engines I normally test (Cesium, Babylon, ThreeJS).

Note however, Windows 10's apps (I tested Paint 3D, Mixed Reality Viewer) are very strict about this value, and will reject models that contain >1.0 emissiveFactor.  Is it OK for us to remove a restriction from the spec like this?

(The model I tested with was "blackvan" from AnalyticalGraphicsInc/gltf-vscode#80).

/cc @bghgary @sbtron

### DIFF
--- a/specification/2.0/schema/material.schema.json
+++ b/specification/2.0/schema/material.schema.json
@@ -31,8 +31,7 @@
             "type": "array",
             "items": {
                 "type": "number",
-                "minimum": 0.0,
-                "maximum": 1.0
+                "minimum": 0.0
             },
             "minItems": 3,
             "maxItems": 3,


pjcozzi:
Does the .md file also need to be updated?

emackey:
@pjcozzi No, the min/max values are not listed in the readme.

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#materialemissivefactor

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#material

emackey:
Although it looks like alphaCutoff has one, maybe this is already out of sync?

pjcozzi:
@emackey do you want to regenerate that part of the .md or update by hand?

emackey:
It would need a wetzel feature addition (min/max for array values).  I did some wetzel cleanup this morning but I don't think I'll have a chance to get to this one.

javagl:
"Off-topic": The task to familiarize myself with wetzel (in order to fix https://github.com/KhronosGroup/glTF/issues/1013 ) has been pending for far too long now (only justified by "not having highest priority"), but maybe I can have a look at this issue as well *when* I get the chance. I'll add it to the TODO list...

bghgary:
@emackey
> Is it OK for us to remove a restriction from the spec like this?

I don't think this is okay. We can't change the schema like this and still meet versioning requirements. For example, older versions of native apps that use the old schema to validate will fail to load these new assets.

The ways I can see this working is if we add another factor value that either replaces the old value if recognized or multiplies against it. The old `emissiveFactor` must continue to be populated with a reasonable fallback value for older clients.

emackey:
Closed this.  We should maybe clarify https://github.com/KhronosGroup/glTF/issues/1083#issuecomment-327263553 to make it clear this is not as simple as removing the 1.0 restriction.

PR #1194, opened 2018-01-03
Add info about normalizing normals

### LINKED ISSUE #993: Must supplied normal map vectors be normalized?
It took far too long to figure out what to expect from supplied normal maps in glTF. In the Properties Reference, it looks like this:

> material.normalTexture
> 
> A tangent space normal map. The texture contains RGB components in linear space. Each texel represents the XYZ components of a normal vector in tangent space. Red [0 to 255] maps to X [-1 to 1]. Green [0 to 255] maps to Y [-1 to 1]. Blue [128 to 255] maps to Z [1/255 to 1]. The normal vectors use OpenGL conventions where +X is right and +Y is up. +Z points toward the viewer.

But nowhere in that description does it say that the resulting vector must (or may not) be of length 1. The closest I could find was this formula in the description of normalTextureInfo.scale, where the computation explicitly calls `normalize`:

> scaledNormal = normalize((normalize(\<sampled normal texture value\>) * 2.0 - 1.0) * vec3(\<normal scale\>, \<normal scale\>, 1.0))

The sample models (the Corset and the FarmLand Diorama) are *not* normalized, which forces GPUs to spend time correcting it. I think it would be preferable to require supplied normal maps to be normalized, but in either case this should be discussed.

bghgary:
Fixes #993

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -743,7 +743,7 @@ Valid accessor type and component type for each attribute semantic property are 
 |Name|Accessor Type(s)|Component Type(s)|Description|
 |----|----------------|-----------------|-----------|
 |`POSITION`|`"VEC3"`|`5126`&nbsp;(FLOAT)|XYZ vertex positions|
-|`NORMAL`|`"VEC3"`|`5126`&nbsp;(FLOAT)|XYZ vertex normals|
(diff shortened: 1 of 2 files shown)


bghgary:
Anyone else want to review?

zellski:
Looks good to me. The words "normalize the normal" maybe don't exactly flow like poetry; a bit redundant -- but I'm not sure any variation would be an improvement.

PR #1196, opened 2018-01-03
Added glTF 2.0 Status for xeogl

sjfricke:
I was just playing around with xeogl the other day and when I was looking at the WebGL engine list I notice it was marked as 
status


bghgary:
I noticed the loader link is also dead. Maybe fix that also?

sjfricke:
Good catch, thanks.

bghgary:
There seems to be two links that are broken. You fixed the Tutorial one, but I actually meant the earlier one: https://github.com/xeolabs/xeogl/tree/master/src/importing/gltf

sjfricke:
missed pasted the link first time around, squashed and third time should be the charm

emackey:
Thanks @sjfricke!

pjcozzi:
Good stuff, thanks @sjfricke!

Issue #1058, opened 2017-08-05
Late-breaking glTF projects


pjcozzi:
This issue is for tracking late-breaking glTF ecosystem projects that haven't yet made it to the [master list](https://github.com/KhronosGroup/glTF#contents) in this repo's README.md.

To contribute, feel free to chime in here with projects you are aware of / working on, and open pull requests to update the master list.

_(The checkboxes are used to keep track of when something is added to the master list)._

_(Everything is glTF 2.0 unless otherwise noted)._

pjcozzi:
* [ ] Sketchup warehouse import/export extension. https://github.com/KhronosGroup/glTF/issues/1051#issuecomment-320078379

via @CentaurMare

pjcozzi:
* [ ] Unity HoloLens Tooklit. https://github.com/Microsoft/HoloToolkit-Unity/pull/793

via @keveleigh

pjcozzi:
* [ ] Qt 3D. https://twitter.com/sean_harmer/status/890247512971501569

pjcozzi:
* [ ] Metal renderer. https://twitter.com/warrenm/status/891558755657175040

via @warrenm

pjcozzi:
* [ ] PiXYZ export. https://twitter.com/PixyzSoft/status/892022576851169280

pjcozzi:
* [ ] Blender importer. https://github.com/ksons/gltf-blender-importer

via @ksons

pjcozzi:
* [ ] Godot. https://twitter.com/reduzio/status/892210751338930177, https://github.com/godotengine/godot/issues/7952

via @reduz

pjcozzi:
* [ ] React VR plans to support glTF. https://twitter.com/pjcozzi/status/837508186487676928 (via GDC meetup)

pjcozzi:
* [ ] assetkit. https://github.com/recp/assetkit

via @recp

pjcozzi:
* [ ] Flatbuffers. https://github.com/KageKirin/flatgltf_2_0

via @KageKirin

pjcozzi:
* [ ] gltf-bounding-box: Node module to compute asset's bounding box. https://www.npmjs.com/package/gltf-bounding-box

via @risq

pjcozzi:
* [ ] Chromium. https://groups.google.com/a/chromium.org/forum/#!topic/graphics-dev/NxG1JunmBO8 and https://codereview.chromium.org/2852533004

[...]

PR #1197, opened 2018-01-03
fixed broken jgltf links

sjfricke:
also added the jgltf-validator

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,13 +226,17 @@ To compare WebGL-based glTF loaders, see [gltf-test](https://github.com/cx20/glt
 #### Java
 
 * [JglTF](https://github.com/javagl/JglTF) - Java libraries for glTF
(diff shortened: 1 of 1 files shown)


javagl:
Thanks, I think it's OK to merge this. 

(The validator was not mentioned there originally, because it is not really intended for public use. The task to align this validator to yours (in terms of the output format etc.) is still on my TODO list - together with the task to update the readme and put the JglTF libraries into a table, to follow the pattern that is used for the other libraries). But these currently do not have highest priority)

pjcozzi:
Thanks again @sjfricke!  BTW what are you building with glTF?

sjfricke:
@pjcozzi Currently hacking around with WebGL2 and WASM and finally had some time to read up on glTF spec and tools since I learned about it at SIGGRAPH17 and realized it solves all the pipeline nightmares I have got trying to load assets via the web in past

pjcozzi:
@sjfricke very cool, please keep us posted on your progress!  More PRs are welcome and also feel free to chime in at #1058.

PR #1205, opened 2018-01-09
added gltf-enum.com to Utilities

sjfricke:


javagl:
Reminds me of the https://javagl.github.io/GLConstantsTranslator/GLConstantsTranslator.html that I hacked down a while ago (because I *needed* it when implementing JglTF, creating the tutorial etc...). Having the links to the glTF spec could really be handy in some cases 👍

emackey:
Nice.  Keeping track of these enums was one of the original reasons for the VSCode extension.

Thanks @sjfricke.

PR #1209, opened 2018-01-12
Updated Rust Loader link to active repo

sjfricke:
Current link is pointing to a fork that hasn't been updated/maintained like the original repo


pjcozzi:
Good catch, thanks for the fix as always, @sjfricke!

PR #1210, opened 2018-01-17
Fix typo in the specification 2.0, remove an extra "with"

takahirox:
This PR fixes typo in the specification 2.0, removing an extra "with".


pjcozzi:
Thanks for the fix @takahirox!

PR #1214, opened 2018-01-18
Organize extensions into version-specific directories.

### LINKED ISSUE #1208: Gltf Extensions should be split into 1.0 and 2.0 folders in the repo
I just spent 10 minutes trying to figure out why KHR_materials_common seemed so 'off', only to later realize its only for 1.0. This should be reflected in the folder structure.

Extensions/1.0/Khronos/KHR_materials_common

donmccurdy:
Fixes #1208.


pjcozzi:
Cool, thanks @donmccurdy!  Does this break any links in this repo, sample models, or elsewhere that we reasonably want to fix?

Anyone else want to look at this?

donmccurdy:
I've searched for links in this repo, and the [samples README](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/README.md) will also need an update. Beyond that I'm hoping others can fix broken links as they find them.

pjcozzi:
Thanks again @donmccurdy!

Issue #1177, opened 2017-12-09
glb has at most one BIN chunk ?


andreasplesch:
https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#glb-file-format-specification reads at first as if multiple chunks, including multiple BIN chunks are allowed. In addition,  https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#glb-stored-buffer seems to suggest that a second buffer object could refer to a second BIN chunk.
So it took looking at a few loaders to confirm that at least in praxis only a single BIN chunk is expected.
It is very possible that I overlooked more specific language. If not, it may be helpful to explicitly state that only the first BIN chunk is used for a (single) buffer. Subsequent chunks may exist for custom data, presumably.

javagl:
The second link says

> glTF Buffer referring to GLB-stored BIN chunk, must have buffer.uri property undefined, and **it must be the first element of buffers array;**

(e.b.m). Also, the "Table 1" from the first link says that there may be only "0 or 1" BIN chunk. So indeed, there is only one BIN chunk supported right now. Suggestions about how to make the wording clearer are certainly welcome.

Maybe helpful as a backtrace: The current structure is a generalization of the `KHR_binary_glTF` extension from glTF 1.0: https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_binary_glTF#binary-gltf-layout There, the structure was more rigid. The concept of "chunks" that was introduced in 2.0 indeed aims at supporting *additional* binary data - for example, for other extensions.

andreasplesch:
I missed the Occurrences column in Table 1. I think because the column is never referenced in the text.

The Binary buffer section at
https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#binary-buffer
says:

> This chunk contains the binary payload for geometry, animation key frames, skins, and images. See glTF specification for details on referencing this chunk from JSON.

What about

> This chunk contains the binary payload for geometry, animation key frames, skins, and images. _**There is at most one chunk of this type.**_ See the  GLB-stored Buffer section for details on referencing this chunk from JSON.

? How are other paragraphs or sections in the specification referred to best ?

The reference at
https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#glb-stored-buffer
> glTF Buffer referring to GLB-stored BIN chunk, must have buffer.uri property undefined,

just misses a definite article to help clarify:

[...]

PR #1231, opened 2018-02-03
Clarify that GLB BIN chunk must be second

bghgary:
See #1177 

To support existing GLB implementations, this change clarifies that the BIN chunk of a binary glTF asset must be the second chunk for GLB version 2. If a need arises that requires a custom chunk between the JSON and BIN chunk, we will increment the GLB version.


pjcozzi:
Thanks @bghgary!

PR #1241, opened 2018-02-09
Add Verge3D to the list of exporters and frameworks

alexkowel:
Verge3D uses glTF 2.0 as default  export/import format. Please consider adding it to the list. Thanks!


pjcozzi:
Thanks @alexkowel!

PR #1251, opened 2018-02-20
typo fix in Draco README

sakrist:
`extensionRequired` replaced with `extensionsRequired`
`extensionUsed ` replaced with `extensionsUsed `


pjcozzi:
Good catch, thanks @sakrist!

CC @FrankGalligan

PR #1261, opened 2018-02-22
Clean up image files format description

takahirox:
In [image.uri section](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#imageuri) we see

> The image format must be jpg or png.

then I think I should remove "etc." from

> Image files (.jpg, .png, etc.) for textures

in [glTF Basics section](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#gltf-basics) for the consistency.


pjcozzi:
Good catch, thank you!

Issue #1266, opened 2018-02-25
Sync spec with JSON schemas


lexaknyazev:
Properties Reference section is outdated and needs to be synced with schemas. Could we make this happen automatically on each schemas update? 

Also, "Last updated" date is incorrect.

/cc @pjcozzi @donmccurdy @javagl

javagl:
It's somewhat embarassing that I didn't manage to tackle https://github.com/KhronosGroup/glTF/issues/1013 yet. I'm not sure how the update could be done automatically, but I can try to run wetzel over the latest spec this week and provide the result as a PR (possibly, with manual tweaks for 1013)

pjcozzi:
@bghgary or one of his colleagues may be the best to advise on updating re-generating the schema.

It will be a bigger change to do it automatically for each update and may not be worth it until we start significant future spec work.

bghgary:
This is what I'm running locally from the wetzel\bin folder:
> wetzel\bin>node wetzel.js ../../glTF/specification/2.0/schema/gltf.schema.json -l 2 -p schema -i "['gltfid.schema.json', 'gltfchildofrootproperty.schema.json', 'gltfproperty.schema.json']"

If I remember correctly, we didn't add the auto link flag `-a` because there were cases where it incorrectly linked references that should not have been linked.

javagl:
The `-a` flag indeed seems to add references/links in texts very aggressively (basically, using a RegEx), and causes some not-so-pretty results. But I think it should be possible to at least add the respective links dedicatedly in the "Type" column of the Properties table. I'd give this a stab ASAP. 

(BTW: Maybe someone here want to share their thoughts on https://github.com/AnalyticalGraphicsInc/wetzel/issues/29 ?)

emackey:
Note the -a flag takes a parameter (from memory I think it's `-a=cqo`) that links code-quoted references only.  If you can get that working, it's far less aggressive than full auto-link.

javagl:
OK, I tried that. The resulting output when using `-a=cqo` is shown at https://github.com/javagl/glTF/blob/d0b6dee76f5c14ac2e13cb5f459ca3b2b406ab2f/specification/2.0/README.md#properties-reference 

**but** there are still some formatting issues:

- Capitalization of type names
- Camel-case types like `bufferView` are by default formatted as `buffer.view`, which is clearly not intended. 
[...]

PR #1268, opened 2018-02-26
Clarify the case where image.mimeType is required in Properties Reference section.

takahirox:
I think it's worth to mention the case where `image.mimeType` is required in Reference section

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#imagemimetype

as well as

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#images

----

Another thing (not included in this PR), I suppose adding "if defined" to

> Use this instead of the image's uri property.

in

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#imagebufferview

would make the sentence clearer.

> Use this instead of the image's uri property if defined.

What do you think?

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -2454,7 +2454,7 @@ The uri of the image.  Relative paths are relative to the .gltf file.  Instead o
 
 #### image.mimeType
 
-The image's MIME type.
+The image's MIME type. Required if `bufferView` is defined.
 
 * **Type**: `string`
 * **Required**: No


lexaknyazev:
This change should be done in JSON-Schema files since Property Reference section will be auto-generated (also #1266).

takahirox:
Thanks. I added it to image.schema.json.

It seems that auto-generation is not ready yet but should I revert Property Reference section?

lexaknyazev:
> should I revert Property Reference section

It doesn't really matter given that whole section will be replaced anyway.

> Use this instead of the image's uri property if defined.

Both current and proposed options are slightly misleading since `uri` and `bufferView` are mutually exclusive.

takahirox:
> Both current and proposed options are slightly misleading since `uri` and `bufferView` are mutually exclusive.

Having both `uri` and `bufferView` is invalid glTF? If so, I don't think it's clear in the spec. We should clarify in another PR.

lexaknyazev:
> I don't think it's clear in the spec.

The schema file formally disallows that (see `oneOf` block). The description should be updated though.

takahirox:
Right. We should update the description, it was actually misleading me.

PR #1269, opened 2018-02-28
Fix broken link

mkeblx:
Point to 2.0 spec file.


pjcozzi:
Good catch, thanks for the fix, @mkeblx.

PR #1276, opened 2018-03-07
Add MSFT vendor extensions 

sbtron:
Merging these Microsoft specific extensions back into master so that they are more discoverable. 

There are some ongoing discussions for others adopting some of these like the LOD extension. If there is interest in that, a separate multi-vendor extension can be added. Wanted to get these is in so that there is good documentation/snapshot of what we are currently doing.

### DIFF
--- a/extensions/2.0/Vendor/MSFT_lod/README.md
+++ b/extensions/2.0/Vendor/MSFT_lod/README.md
@@ -35,11 +35,18 @@ The following example shows how the `MSFT_lod` extension can be specified at the
         "mesh": 0,
         "extensions": {
             "MSFT_lod": {
-                "ids": [ 1, 2]
+                "ids": [
+                    1,
+                    2
+                ]
             }
         },
         "extras": {
-            "MSFT_screencoverage": [0.5, 0.2, 0.01]
+            "MSFT_screencoverage": [
+                0.5,
+                0.2,
+                0.01
+            ]
         }
     },
     {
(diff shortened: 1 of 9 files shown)


pjcozzi:
@sbtron are these already pretty well vetted and implemented?  Or do you want anyone to look at these before merging?  As a vendor extension, it is totally up to you.

sbtron:
These should be good to go with implementations in babylon, gltf-toolkit and windows mixed reality. Merging now.

pjcozzi:
Awesome, thanks!

takahirox:
Some of the extensions suggest that `image.mimeType` should be "image/vnd-ms.dds" for DDS textures. But it conflicts with glTF core spec. glTF core spec mentions `mimeType` should be "image/jpeg" or
"image/png".

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#imagemimetype

So we should move `"mimeType": "image/vnd-ms.dds"` in extensions?

    "images": [
        {
            "extensions": {
                "MSFT_texture_dds": {
                    "mimeType": "image/vnd-ms.dds"
                }
            },
            "bufferView": 1
        }
    ]

lexaknyazev:
All "enum" properties allow an arbitrary value of the same type, so using `"image/vnd-ms.dds"` doesn't conflict with the schema: 
https://github.com/KhronosGroup/glTF/blob/6c7eafcf9505c09f4fa7276327fe8edb22b7d1fa/specification/2.0/schema/image.schema.json#L15-L28

lexaknyazev:
> That's for extensions?

For extensions and for future minor version updates.

Issue #1279, opened 2018-03-12
Does the `extensions/2.0/Khronos/KHR_technique_webgl` references the `arrayValues.schema.json` in `specification/1.0`?


alexchicn:
In `extensions/2.0/Khronos/KHR_technique_webgl/schema/technique.parameters.schema.json`, the property - `properties.value` references the `arrayValues.schema.json` that just exists in `specification/1.0`.

Need copy the `arrayValues.schema.json` to `specification/2.0`?

donmccurdy:
Note that the `KHR_technique_webgl` extension is very much still a draft, and not implemented in tools yet — see https://github.com/KhronosGroup/glTF/issues/1217. But thanks we will make sure that is corrected as we move forward with the extension!

pjcozzi:
Should we consolidate this issue into #1217?

PR #1296, opened 2018-03-28
Updates to KHR_technique_webgl extension spec

### LINKED ISSUE #1217: Updates for KHR_technique_webgl
Let's get the draft [KHR_technique_webgl](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_technique_webgl) extension finished up. For background on why GLSL was removed from the core specification into an extension with glTF 2.0, see https://github.com/KhronosGroup/glTF/issues/733. The current draft is based almost exactly on glTF 1.0, so if we were to freeze it as-is there would be no significant GLSL-related changes for users upgrading from glTF 1.0. A few changes have been suggested that should also be considered:

* [ ] **(1)** Consider allowing shaders to be embedded in a binary bufferView (https://github.com/KhronosGroup/glTF/issues/868)
  * Change was made with #894, I think we just need to finish filling in details under the Shader object's properties in the spec?
* [ ] **(2)** Consider altering file extensions for external shaders (https://github.com/KhronosGroup/glTF/issues/738)
* [ ] **(3)** Consider changing listing of `technique.parameters` (https://github.com/KhronosGroup/glTF/issues/717)
* [x] ~~**(4)** Consider support for program pipelines (https://github.com/KhronosGroup/glTF/issues/840)~~ Out of scope.
* [ ] **(5)** Consider removing or simplifying the [render states](https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_technique_webgl#render-states) portion of the extension
  * (For example, keep blending and face culling, leave the rest to another extension or later versions.)

ggetz:
Fixes #1217 

Makes the suggestion listed in the above issue, with the most significant changes being:
* Inline `technique.parameters` in `technique.attributes` and `technique.uniforms` to reduce indirection, and allow the objects for each to be more specific to `attributes` and `uniforms` use
* Don't allow default value for attributes, only uniforms.
* Include list of WebGL 1.0 extensions in `program.glExtensions`
* Remove render states.
* `KHR_techniques_webgl` instead of `KHR_technique_webgl`
* Update `README`
    * Mentioning targeting WebGL 1.0
    * Address material extension specifically 

TODO:
 - [x] Add contributors to "Contributors" section
 - [x] Change "Status" to "Complete" once ready

@lilleyse let me know if you have any initial feedback.


pjcozzi:
@ggetz congrats on your first glTF extension work!

@donmccurdy @lexaknyazev would love your feedback here.

extensions/2.0/Khronos/KHR_technique_webgl/README.md
```diff
@@ -310,7 +236,8 @@ Shader source files are stored in the asset's `shaders` dictionary property, whi
     ]
 }
 ```    
-### Properties Reference
+
+## Properties Reference
```
donmccurdy: In this section, `.programs`, `.shaders`, and `.techniques` are now arrays with indices rather than an "ID in the global glTF namespace".

lilleyse:
@ggetz I started to review this afternoon and should have some feedback posted tomorrow.

extensions/2.0/Khronos/KHR_technique_webgl/README.md
```diff
@@ -14,122 +14,153 @@ Written against the glTF 2.0 spec.
...
+],
+```
+
+#### Technique
+
+The `technique` property points to the index of a technique listed in the asset's `KHR_techniques_webgl` `techniques` property.
+
+#### Values
+
+The `values` property defines the list of values, each of which is passed to the technique uniform of the same name, and when specified, overrides the corresponding uniform `value`.
```
donmccurdy: Since neither `.values` here nor `uniform.value` below are required, perhaps we need to say that at least one of the two must be present?

ggetz:
@donmccurdy 

> Could I talk you into writing a simple stub example? That would help me get a clearer picture of this.

When you say simple stub, I assume you mean the entire (but simple) glTF?

> l. Including blend states in KHR_techniques_webgl might be the quickest path to supporting that. But in the long run, I would prefer to see a GLSL-agnostic way to specify widely-supported blend operations, e.g. a KHR_blending extension... 

I agree the more agnostic approach is best in the long run. If tilt brush needs it now, and it doesn't rule out the longer-term approach, I'm fine with addressing it in this extension.

donmccurdy:
> When you say simple stub, I assume you mean the entire (but simple) glTF?

Yes, not necessarily including buffers or even meant to be rendered, but an example that shows all of the top-level properties cross-referencing one another.

donmccurdy:
Let's see if we can do blending in parallel, but keep it separate from this extension. I've started a thread for `EXT_blend` in https://github.com/KhronosGroup/glTF/pull/1302.

ggetz:
Thanks @donmccurdy and @lilleyse ! I've updated based on your comments, let me know if you have anymore feedback.

donmccurdy:
I think #1279 still applies here, do we want to fix that up in this PR?

ggetz:
@donmccurdy The one difference between `arrayValues` and this implementation is that we no longer accept strings as values, and instead use `textureInfo` to reference textures, see https://github.com/KhronosGroup/glTF/pull/1296#discussion_r178307596

lilleyse:
In https://github.com/KhronosGroup/glTF/tree/master/extensions rename `KHR_technique_webgl` to `KHR_techniques_webgl`.

lilleyse:
Rename `material.KHR_technique_webgl.schema.json` to `material.KHR_techniques_webgl.schema.json`

extensions/2.0/Khronos/KHR_techniques_webgl/README.md
```diff
@@ -0,0 +1,836 @@
...
+
+### uniform.value
+
+The value of the uniform. The length is determined by the values of the `type` and `count` (if present) properties.  A material uniform value with the same name, when specified, overrides this value.
+
+* **Type**: `any`
+* **Required**: No
+
+### uniform.name
+
```
lilleyse: Same comment: missing description.

extensions/2.0/Khronos/KHR_techniques_webgl/README.md
```diff
@@ -0,0 +1,836 @@
...
+|**technique**|`integer`|The index of the technique.| :white_check_mark: Yes|
+|**values**|`object`|Dictionary object of uniform values.|No, default: `{}`|
+|**extensions**|`object`|Dictionary object with extension-specific objects.|No|
+|**extras**|`any`|Application-specific data.|No|
+
+Additional properties are allowed.
+
+* **JSON schema**: [material.KHR_technique_webgl.schema.json](schema/material.KHR_technique_webgl.schema.json)
+
+### khr_techniques_webgl.material.extension.technique :white_check_mark:
```
lilleyse: Is it fine to drop the `khr_techniques_webgl.material.extension.` prefix for these? It looks like [KHR_materials_pbrSpecularGlossiness](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness) doesn't have the prefix.

extensions/2.0/Khronos/KHR_techniques_webgl/schema/material.KHR_technique_webgl.schema.json
```diff
@@ -0,0 +1,27 @@
...
+        },
+        "values" : {
+            "type" : "object",
+            "description" : "Dictionary object of uniform values.",
+            "properties" : {
+            },
+            "additionalProperties" : {
+                "$ref" : "uniform.value.schema.json"
+            },
+            "default" : {},
```
lilleyse: I forget if this came up before, but is to safe to just remove `"default" : {}` throughout?

extensions/2.0/Khronos/KHR_techniques_webgl/schema/technique.schema.json
```diff
@@ -0,0 +1,39 @@
...
+    "properties": {
+        "program" : {
+            "allOf" : [ { "$ref" : "glTFid.schema.json" } ],
+            "description" : "The index of the program."
+        },
+        "attributes" : {
+            "type" : "object",
+            "description" : "A dictionary object of `Attribute` objects.",
+            "properties" : {
+            },
```
lilleyse: Throughout, for empty property objects just condense to a single line: `"properties" : {},`

extensions/2.0/Khronos/KHR_techniques_webgl/examples/sample_techniques.gltf
```diff
@@ -0,0 +1,273 @@
...
+                        "u_projectionMatrix": {
+                            "semantic": "PROJECTION",
+                            "type": 35676
+                        }
+                    }
+                }
+            ]
+        }
+    }
+}
```
lilleyse: Add newline.

extensions/2.0/Khronos/KHR_techniques_webgl/schema/uniform.value.schema.json
```diff
@@ -0,0 +1,28 @@
...
+                "type": ["boolean"]
+                
+            },
+            "minItems": 1
+        },
+        { 
+            "$ref" : "textureInfo.schema.json" 
+        }
+    ]
+}
```
lilleyse: Add newline.

ggetz:
Thanks @lilleyse ! 

Addressed your comments, should we address these TODO items now then? Who all should be added to contributors? 

-  Add contributors to "Contributors" section
- Change "Status" to "Complete" once ready

extensions/2.0/Khronos/KHR_techniques_webgl/schema/technique.schema.json
```diff
@@ -0,0 +1,36 @@
...
+            "additionalProperties" : {
+                "$ref" : "technique.attribute.schema.json"
+            },
+            "gltf_detailedDescription" : "A dictionary object of `Attribute` objects.  Each object defines an attribute input with a type and a semantic."
+        },
+        "uniforms" : {
+            "type" : "object",
+            "description" : "A dictionary object of `Uniform` objects.",
+            "properties" : {
+            },
```
lilleyse: Change to `"properties" : {}`

extensions/2.0/Khronos/KHR_techniques_webgl/schema/material.KHR_techniques_webgl.schema.json
```diff
@@ -0,0 +1,27 @@
...
+    "properties" : {
+        "technique" : {
+            "allOf" : [ { "$ref" : "glTFid.schema.json" } ],
+            "description" : "The index of the technique."
+        },
+        "values" : {
+            "type" : "object",
+            "description" : "Dictionary object of uniform values.",
+            "properties" : {
+            },
```
lilleyse: Change to `"properties" : {}`

extensions/2.0/Khronos/KHR_techniques_webgl/README.md
```diff
@@ -0,0 +1,838 @@
...
+|**shaders**|Shader `[1-*]`|An array of [`Shader`](#reference-shader) objects.| :white_check_mark: Yes|
+|**techniques**|Technique `[1-*]`|An array of [`Technique`](#reference-technique) objects.| :white_check_mark: Yes|
+|**extensions**|`object`|Dictionary object with extension-specific objects.|No|
+|**extras**|`any`|Application-specific data.|No|
+
+Additional properties are allowed.
+
+* **JSON schema**: [glTF.KHR_techniques_webgl.schema.json](schema/glTF.KHR_techniques_webgl.schema.json)
+
+### khr_techniques_webgl.gltf.extension.programs :white_check_mark:
```
lilleyse: I forgot to mention, also remove the `khr_techniques_webgl.gltf.extension.` prefixes in this section. Related to https://github.com/KhronosGroup/glTF/pull/1296#discussion_r183505232

extensions/2.0/Khronos/KHR_techniques_webgl/README.md
```diff
@@ -0,0 +1,838 @@
...
+
+When a material instances a technique, the name of each supplied value in its `KHR_techniques_webgl.values` property must correspond to one of the uniforms defined in the technique.
+
+The above example illustrates several uniforms. The property `u_ambient` is defined as a `FLOAT_VEC4` type; and `u_light0Color` is defined as a `FLOAT_VEC3` with a default color value of white.
+
+##### Semantics
+
+Technique uniforms may also optionally define a *semantic*, an enumerated value describing how the runtime is to interpret the data to be passed to the shader.
+
+In the above example, the uniform `u_light0Transform` defines the `MODEL` semantic, which corresponds to the world space position of the node referenced in the property `node`, in this case the node `1`, which refers to the id of a node that contains a light source.
```
lilleyse: Bump - https://github.com/KhronosGroup/glTF/pull/1296#discussion_r183500300

lilleyse:
Is anyone else interested in reviewing this? @lexaknyazev, @pjcozzi, @stevesan?

lexaknyazev:
Will do by tomorrow's call.

pjcozzi:
I won't be able to review, but looks like this is getting enough attention.

For carry over credits from the 1.0 spec, probably include:

* Remi Arnaud
* Patrick Cozzi
* Tony Parisi
* Fabrice Robinet

ggetz:
Updated, completed TODO items.

extensions/2.0/Khronos/KHR_techniques_webgl/README.md
```diff
@@ -0,0 +1,842 @@
...
+        }
+    }
+}
+```
+
+#### Programs
+
+GLSL shader programs are stored in the asset's `KHR_techniques_webgl.programs` property. This property contains one or more objects, one for each program.
+
+Each shader program includes an `attributes` property, which specifies the vertex attributes that will be passed to the shader, and the properties `fragmentShader` and `vertexShader`, which reference the index of the shader in the `KHR_techniques_webgl.shaders` property for the fragment and vertex shader GLSL source code, respectively.
```
lexaknyazev: I've probably missed something, but why is `program.attributes` array needed? How should conflicts between it and `technique.attributes` be resolved?

extensions/2.0/Khronos/KHR_techniques_webgl/README.md
```diff
@@ -0,0 +1,842 @@
...
+            "fragmentShader": 0,
+            "vertexShader": 1
+        }
+    ]
+}
+```
+
+#### Shaders
+
+One or more shader source files are listed in the asset's `KHR_techniques_webgl.shaders` property. Each shader specifies a `type` (vertex or fragment, defined as GL enum types) and either a `uri` to the file or a reference to a `bufferView`. Shader URIs may be URIs to external files or data URIs, allowing the shader content to be embedded as base64-encoded data in the asset.
```
lexaknyazev: The spec should mention the preferred MIME type for Data URIs with shader sources.

ggetz:
Updated. 

* Removed `program.attributes` as it was not needed
* Removes `attribute.type`, as it is specified on the corresponding attribute's accessor
* Added "File Extension and MIME type" section for external shader files

stevesan:
Hey all, apologies for the delay!

I do have a request: For the example GLTF, could we add an example of an attribute with a custom semantic (_MY_ATTRIBUTE)? Tilt Brush has lots of these, and in glTF 1.0 they were implemented by overriding TEXCOORD's (and usually making them the 'wrong' type, like vec4). This was done because threejs's LegacyGLTFLoader didn't appear to handle custom semantics properly. So moving forward, it's probably worth explicitly calling this out.

lexaknyazev:
An optional nitpick (no strong opinion) - why do we need an intermediate object for binding attribute names with semantics?

```json
"attributes": {
  "a_normal":  {
    "semantic": "NORMAL"
  },
  "a_position": {
    "semantic": "POSITION"
  },
  "a_texcoord0": {
    "semantic": "TEXCOORD_0"
  }
}
```

vs

```json
"attributes": {
  "a_normal": "NORMAL",
  "a_position": "POSITION",
  "a_texcoord0": "TEXCOORD_0"
}
```

Do we expect app-specific use-cases, like this?
```json
"attributes": {
  "a_color":  {
    "semantic": "COLOR_0",
    "extras": {
       "disable": true
    }
  }
}
```

Also a few implementation details.
- Must all used semantics exist in the mesh data? Should an asset be flagged as invalid if some semantics are absent?
- Is it valid for technique to not use `POSITION`? If so, should `POSITION` be considered optional for assets that have `KHR_technique_webgl` as required extension?
- Sampler uniforms use `textureInfo` object. That object could be extended with, e.g., `KHR_texture_transform`. How should engines handle that (fail to load, patch shaders, resample texures, etc)?

ggetz:
> why do we need an intermediate object for binding attribute names with semantics?

I noticed this too. I don't have a concrete use case, but IMO we should keep it consistent with `uniform` which have their own object to add extras and extensions to.

> Must all used semantics exist in the mesh data? Should an asset be flagged as invalid if some semantics are absent?

Yes, updated.

> Is it valid for technique to not use POSITION? If so, should POSITION be considered optional for assets that have KHR_technique_webgl as required extension?

The glTF spec doesn't require the `POSITION` attribute explicitly, so I assume we can also not explicitly require `POSITION` here.

> Sampler uniforms use textureInfo object. That object could be extended with, e.g., KHR_texture_transform. How should engines handle that (fail to load, patch shaders, resample texures, etc)?

@lilleyse I think you have some thoughts on this?

lexaknyazev:
> The glTF spec doesn't require the `POSITION` attribute explicitly

Good catch! Opened https://github.com/KhronosGroup/glTF/issues/1332 to track that.

lilleyse:
> Sampler uniforms use textureInfo object. That object could be extended with, e.g., KHR_texture_transform. How should engines handle that (fail to load, patch shaders, resample texures, etc)?

Is there any precedence for extensions conflicting with each other? Maybe it is best to leave this decision up to implementations?

If my understanding of `KHR_texture_transform` is correct, `KHR_techniques_webgl` could achieve the same effect but through GLSL code and uniforms. So another answer might be to ignore `KHR_texture_transform` if present.

lilleyse:
> Disallowing it seems to be the most simple solution

Ok, sounds good to me.

So we just need a note somewhere that says textures referenced by the `KHR_techniques_webl` extension may not use the `KHR_texture_transform` extension, and a brief explanation why. This could all go in an `Extension compatibility` section.

lexaknyazev:
A couple follow-ups from the latest call:
- Should default GL state be assumed (depth test, alpha test, etc)?
- Should engines respect `doubleSided`, `alphaMode`, and `alphaCutoff` material properties?

lilleyse:
Yeah engines should respect `doubleSided`, `alphaMode`, and `alphaCutoff` (and later [EXT_blend](https://github.com/KhronosGroup/glTF/pull/1302)). That should be mentioned here.

Along with that, I think this [implementation note](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#alpha-coverage) applies well enough and default GL state should not be assumed, since depth test for example is off by default.

ggetz:
@lilleyse @lexaknyazev 

Updated and added a Conformance section. I believe all of the above comments have been addressed and this is ready to go.

ggetz:
Yes, updated wording to reflect.

extensions/2.0/Khronos/KHR_techniques_webgl/schema/program.schema.json
```diff
@@ -1,21 +1,10 @@
 {
     "$schema" : "http://json-schema.org/draft-04/schema",
-    "title" : "program",
+    "title" : "Program",
     "type" : "object",
-    "description" : "A shader program, including its vertex and fragment shader, and names of vertex shader attributes.",
+    "description" : "A shader program, including its vertex and fragment shaders, and names of vertex shader attributes.",
```
lilleyse: Remove `and names of vertex shader attributes`

ggetz:
@lilleyse Updated.

extensions/README.md
```diff
@@ -6,12 +6,12 @@
 * [KHR_materials_pbrSpecularGlossiness](2.0/Khronos/KHR_materials_pbrSpecularGlossiness/README.md)
 * [KHR_materials_unlit](2.0/Khronos/KHR_materials_unlit/README.md)
 * [KHR_draco_mesh_compression](2.0/Khronos/KHR_draco_mesh_compression/README.md)
+* [KHR_techniques_webgl](2.0/Khronos/KHR_techniques_webgl/README.md)
```
pjcozzi: Can we mark this as draft here and in the main extension spec and then merge this?

ggetz:
@pjcozzi Marked as draft. 

Ready to merge?

donmccurdy:
> ... modifying the supplied GLSL shader code at runtime if necessary

this one's a surprise to me — is that possible in the general case?

In either case I'm OK with merging this as an updated draft.

lexaknyazev:
@donmccurdy How else could an engine pass `alphaCutoff` value? For WebGL 1.0 shaders, it's probably not that hard given that `gl_FragColor` is a fixed name.

lexaknyazev:
> If the GLSL contains a line like this...

That line seems to enforce `0.5` as a hard-coded cutoff value. I don't think an engine should care about such case.

I was thinking about adding something like
```glsl
if ( gl_FragColor.a < cutOffUniformOrConstant ) discard;
```
to the very end of fragment shader's `main`.

donmccurdy:
> That line seems to enforce 0.5 as a hard-coded cutoff value. I don't think an engine should care about such case.

I don't follow this — if the asset contains `material.alphaCutoff: 0.5`, then the GLSL _should_ contain a `0.5` somewhere, either as a hard-coded value or a constant, unless there is a particular reason to need a uniform. And in any case, it seems like a decision that should be made by the exporter.

Even detecting the end of the fragment shader's `main` is not trivial — you need to count brackets and deal with multi-line comments. If our general philosophy is to most responsibility on the exporter, why shouldn't we require the exporter to include the alphaCutoff in the GLSL if it's writing it to the `material` entry? This gives the flexibility to use a uniform, a constant, or a hard-coded value as necessary.

One alternative might be to define a technique uniform semantic, `ALPHACUTOFF`, which is required to be present in this case.

lexaknyazev:
My understanding is that there could be one shader used by different materials that define different `alphaCutoff` values.

Requiring a pre-defined uniform name would definitely simplify engines.

pjcozzi:
@ggetz what is the latest here?  Is this close to being merged as a draft?

ggetz:
@pjcozzi We may want to discuss the implications of the material properties interacting with the shaders more, however, I think the consensus is this is good to merge as a draft.

ggetz:
Removed the requirement to have the engine patch the shader at runtime. Instead, require shaders to respect these values, and specified the `ALPHACUTOFF` uniform semantic for passing the value of `material.alphaCutoff' to the shader. Also added an implementation note for how to ensure the value of `material.doubleSided` is respected.

pjcozzi:
@ggetz please let me know when you want this merged.

ggetz:
Modified the Shader Requirement implementation note to instead include a TODO to reference once included in the main spec.

@pjcozzi I think this is good to merge.

pjcozzi:
Thanks @ggetz!

PR #1302, opened 2018-03-31
KHR_blend

donmccurdy:
Complement to #1296, in the hope that blending can be kept WebGL-agnostic. DirectX supports many blend operations that GL does not; currently those are omitted from this extension for compatibility. I also did not attempt to include `gl.blendColor`.

BabylonJS and Unreal appear to skip many of these blending equations. I wouldn't be opposed to cutting back to just the intersection, but would like to get more feedback on this first. @bghgary any preference, or rationale behind Babylon's choices here? Found [this thread](http://www.html5gamedevs.com/topic/15696-blend-modes/).

TODO:
* [ ] Clarify `alphaMode == BLEND` requirement. Is extension invalid when this is not the case, or simply ignored? Do we allow `alphaMode === MASK`, too?
* [ ] Clarify that MIN and MAX equations do not use `blendFactors`.
* [ ] Consider simplifying allowed blend states.
* [ ] Replace GL integer constants with string enums.

References:
- [Unity: Blending](https://docs.unity3d.com/Manual/SL-Blend.html)
- [Unreal: Blend modes](https://docs.unrealengine.com/en-us/Engine/Rendering/Materials/MaterialProperties/BlendModes)
- [OpenGL: Blending](https://www.khronos.org/opengl/wiki/Blending#Blend_Equations)
- [THREE: Blending equation constants](https://threejs.org/docs/#api/constants/CustomBlendingEquations)
- [BabylonJS: Blend modes](https://doc.babylonjs.com/how_to/how_to_use_blend_modes)
- [WebGL constants](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Constants)
- [Visual glBlendEquation tool](https://www.andersriggelsen.dk/glblendfunc.php)

Reference: https://user-images.githubusercontent.com/1848368/38159786-bca1e3ea-3464-11e8-9d3b-9d1a1e8cac74.jpg

### DIFF
--- a/extensions/2.0/Khronos/KHR_blend/README.md
+++ b/extensions/2.0/Khronos/KHR_blend/README.md
@@ -1,4 +1,4 @@
-# EXT_blend
+# KHR_blend
 
 ## Contributors
 
@@ -14,15 +14,15 @@ Written against the glTF 2.0 spec.
 
 ## Overview
 
-This extension defines blending operations on glTF materials. The `EXT_blend`
+This extension defines blending operations on glTF materials. The `KHR_blend`
 extension may be used alongside the default PBR materials, or materials defined
 in glTF extensions. When combined with other materials and techniques, blending
 may be used to achieve effects like fire, steam, or holograms.
 
-For any material with the `EXT_blend` extension, blending must use the
+For any material with the `KHR_blend` extension, blending must use the
 specified blend equation/operation and factors.
 
-For any materials without the `EXT_blend` extension, blending is assumed to be
+For any materials without the `KHR_blend` extension, blending is assumed to be
 off unless otherwise specified.
 
 ### Properties
@@ -78,7 +78,7 @@ Defines parameters to the blending equation.
         "name": "My_Additive_Material",
         "pbrMetallicRoughness": {},
         "extensions": {
-            "EXT_blend": {
+            "KHR_blend": {
                 "blendEquation": [32778, 32778],
                 "blendFactors": [1, 1, 1, 1]
             }
--- a/extensions/2.0/Vendor/EXT_blend/README.md
+++ b/extensions/2.0/Vendor/EXT_blend/README.md
@@ -1,0 +1,96 @@
+# EXT_blend
+
+## Contributors
+
+TODO
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension defines blending operations on glTF materials. The `EXT_blend`
+extension may be used alongside the default PBR materials, or materials defined
+in glTF extensions. When combined with other materials and techniques, blending
+may be used to achieve effects like fire, steam, or holograms.
+
+For any material with the `EXT_blend` extension, blending must use the
+specified blend equation/operation and factors.
+
+For any materials without the `EXT_blend` extension, blending is assumed to be
+off unless otherwise specified.
+
+### Properties
+
+**blendEquation** :white_check_mark:
+
+Determines how a new pixel is combined with a pixel already in the frame buffer.
+
+* **Type:** `number` `[2]`
+    - `0` *color* equation
+    - `1` *alpha* equation
+* **Required:** Yes
+* **Allowed values:**
+
+| name                     | blendEquation | description |
+|--------------------------|---------------|-------------|
+| FUNC_ADD                 | 0x8006        | Set an addition blend function. |
+| FUNC_SUBTRACT            | 0x800A        | Specify a subtraction blend function (source - destination). |
+| FUNC_REVERSE_SUBTRACT    | 0x800B        | Specify a reverse subtraction blend function (destination - source). |
+| MIN                      | 0x8007        | Produces the minimum color components of the source and destination colors. |
+| MAX                      | 0x8008        | Produces the maximum color components of the source and destination colors. |
+
+**blendFactors** :white_check_mark:
+
+Defines parameters to the blending equation.
+
+* **Type:** `number` `[4]`
+    - `0` : *source color* multiplier
+    - `1` : *destination color* multiplier
+    - `2` : *source alpha* multiplier
+    - `3` : *destination alpha* multiplier
+* **Required:** Yes
+* **Allowed values:**
+
+| name                | blendFactor | description |
+|---------------------|-------------|-------------|
+| ZERO                | 0           | Turn off a component. |
+| ONE                 | 1           | Turn on a component. |
+| SRC_COLOR           | 0x0300      | Multiply a component by the source's color. |
+| ONE_MINUS_SRC_COLOR | 0x0301      | Multiply a component by one minus the source's color. |
+| SRC_ALPHA           | 0x0302      | Multiply a component by the source's alpha. |
+| ONE_MINUS_SRC_ALPHA | 0x0303      | Multiply a component by one minus the source's alpha. |
+| DST_ALPHA           | 0x0304      | Multiply a component by the destination's alpha. |
+| ONE_MINUS_DST_ALPHA | 0x0305      | Multiply a component by one minus the destination's alpha. |
+| DST_COLOR           | 0x0306      | Multiply a component by the destination's color. |
+| ONE_MINUS_DST_COLOR | 0x0307      | Multiply a component by one minus the destination's color. |
+
+### Example
+
+```json
+"materials": [
+    {
+        "name": "My_Additive_Material",
+        "pbrMetallicRoughness": {},
+        "extensions": {
+            "EXT_blend": {
+                "blendEquation": [32778, 32778],
+                "blendFactors": [1, 1, 1, 1]
+            }
+        }
+    }
+]
+ ```
+
+## glTF Schema Updates
+
+**JSON schema**: TODO
+
+## Known Implementations
+
+TODO
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -3,6 +3,7 @@
 ## Extensions for glTF 2.0
 
 #### Khronos extensions
+* [KHR_blend](2.0/Khronos/KHR_blend/README.md)
 * [KHR_materials_pbrSpecularGlossiness](2.0/Khronos/KHR_materials_pbrSpecularGlossiness/README.md)
 * [KHR_materials_unlit](2.0/Khronos/KHR_materials_unlit/README.md)
 * [KHR_draco_mesh_compression](2.0/Khronos/KHR_draco_mesh_compression/README.md)


deltakosh:
For Babylon.js, we actually exposed only required and used ones. Not a big deal to add more but no user ask for them so far

stevesan:
Thanks, Don. Additive is definitely the big one, but multiplicative can be useful too for darkening things in an order-independent way. For example, for Blocks glass rendering (example: https://poly.google.com/view/atB26Z6BPd0), we use both a multiply pass and an additive pass.

recp:
For multiple lights, what will happen in this scenario if `EXT_blend` exists as different than **ADDITIVE**?

Multi-pass lighting (forward rendering):

```swift
for light in lights {
  /* additive blending for other lights */
  if light != firstLight {
      glDepthFunc(GL_EQUAL);
      glEnable(GL_BLEND);
      glBlendEq(GL_FUNC_ADD);
      glBlendFunc(GL_ONE, GL_ONE);
  }
  
  renderForLight(light);
}
```

if material needs multi-pass then the blend equations would make sense here:

```swift
for light in lights {
  /* Step 1: render material to FrameBuffer */
  for pass in material->passes {
     if pass != firstPass {
        /* set blend options specified in `EXT_blend`extension */
     }
  }

  /* Step 2: use/bind FrameBuffer texture as material (IF IT IS POSSIBLE FOR EVERY CASE) */

  /* Step 3: Additive Blending for other lights */
  if light != firstLight {
      glDepthFunc(GL_EQUAL);
      glEnable(GL_BLEND);
      glBlendEq(GL_FUNC_ADD);
      glBlendFunc(GL_ONE, GL_ONE);
  }

  renderForLight(light);
}
```

otherwise I don't know how to render `EXT_blend`-enabled glTF model :( **Thanks for implementation tips if it is possible in multi-pass lighting scenario.** 

Also some Order-Independed Transparency techniques e.g. Weighted-Blended OIT requires additional blending operations it may cause blending-conflicts too

MiiBond:
@recp I think you may need to separate transparent passes from your multi-pass lighting. If I remember correctly, UE3 (which did multi-pass lighting as you describe) didn't allow dynamic lighting on transparent objects for this reason.

Perhaps you could render transparent objects in a single pass with a clamped number of lights? Or, require baked lighting for transparent objects? I'm sure there are better options too.

These are implementation details and they really depend on how your runtime is handling its rendering. I'm not sure that the spec of KHR_blend can specify how to implement it in every case.

recp:
@MiiBond thanks for your advices,

Actually I render scene in three passes if transparency is active and it is working, it uses Weighted, Blended OIT (for now):
- Opaque Pass: Render opaque objects as above blending equations
- Transp Pass: Render transparent objects **using different blending equations** (multiple render targets)
- Composite Pass: Composite opaque and transparency passes **using different blending equations**

Full implementation: https://github.com/recp/gk/blob/master/src/transp/builtin/oit/weighted_blended.c#L91

I'll update lighting design after implementing morph/skin animations to support custom blending options in materials, then I can support this extension. 

Also will this extension be valid for opaque objects too (alphaMode != BLEND)? If opaque means that pixels/fragments behind opaque surfaces are completely invisible then why do we need to specify blending for that surfaces? Maybe it should only be valid if `alphaMode == BLEND`?

donmccurdy:
So far this extension uses the equation and factor primitives of OpenGL, which are combined for different effects. For example multiplicative blending would be:

```
"materials": [
    {
        "name": "MultiplicativeMaterial",
        "alphaMode": "BLEND",
        "extensions": {
            "EXT_blend": {
                "blendEquation": [32774, 32774], // ADD
                "blendFactors": [774, 772, 0, 0] // DST_COLOR DST_ALPHA ZERO ZERO
            }
        }
    }
]
```

Common blend types mentioned in Unity docs:

```
Blend SrcAlpha OneMinusSrcAlpha // Traditional transparency
Blend One OneMinusSrcAlpha // Premultiplied transparency
Blend One One // Additive
Blend OneMinusDstColor One // Soft Additive
Blend DstColor Zero // Multiplicative
Blend DstColor SrcColor // 2x Multiplicative
```

^All assume the blend equation is "add".

recp:
> Clarify alphaMode == BLEND requirement. Is extension invalid when this is not the case, or simply ignored?

Maybe if this extension is enabled we could have some options:
1. If **alphaMode** not exists AND **EXT_Blend** exists AND this extension is supported then alphaMode's default mode is assumed to BLEND instead of OPAQUE?
2. Both **alphaMode** and **EXT_Blend** exist AND `alphaMode == OPAQUE` then **EXT_Blend** could be ignored?
3. `alphaMode == BLEND` is a must for **EXT_Blend**

Having 1 and 2 could be an option, or just 3.

I couldn't see FUNC_MULTPLY in GL reference: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBlendEquation.xhtml

Is FUNC_MULTIPLY DirectX feature? I don't know about DirectX, but in OpenGL if know correctly multiplicative blending (if you meant this) is `glBlendFunc(GL_DST_COLOR, GL_ZERO)` as described here: http://www.learnopengles.com/tag/additive-blending/

MiiBond:
@donmccurdy Yeah, I need to have more coffee this morning. It's been a while since I've looked at blend equations. For some reason, I remembered there being an explicit multiply.

@recp I'm not sure that we'd want the presence of an extension to change the default on another property. I think we should just require the author to set the `alphaMode` value correctly.

donmccurdy:
To fill in details for any future readers, the core equation is:

```
output = (<sourceFactor> * sourceFragment) <operator> (<destinationFactor> * destinationFragment)
```

Where:

* `<operator>` is determined by `blendEquation`
* `<sourceFactor>` and `<destinationFactor>` are determined by `blendFactors`
* `sourceFragment` and `destinationFragment` are current fragment shader output color and existing color in the frame buffer respectively.

So in the multiplicative example above, we get this, which simply multiplies the two fragment colors:

```
output = DST_COLOR * sourceFragment + 0 * destinationFragment
```

[This article](http://www.learnopengles.com/tag/additive-blending/) was helpful for me, I'm having to read up on this as well. 😅

recp:
What about `glBlendFuncSeparate`? Could this extension also support it?

It seems WebGL also supports it: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/blendFuncSeparate

Since EXT_Blend allows custom blending equations it could also support `glBlendFuncSeparate` too for specifying different blending for alpha. I'm not sure if it would make the spec (or ext) more complex. On the implementation side, it is just another function with different parameters, after implemented **EXT_Blend** 🤔

donmccurdy:
@recp it is already covered — because `blendFactors` requires 4 values, it is implicitly _always_ using `glBlendFuncSeparate`. I'm wondering if there is a performance difference in that, or whether `blendFactors` and `blendEquation` should also allow 2 and 1 values respectively.

recp:
@donmccurdy thanks for clarification, I was thought that it is `glBlendFunc`, it should be obvious from params

On CPU side `glBlendFunc` may be faster (and less instructions) because RGB and A has same value, and it can be done with SIMD instructions e.g. SSE2. On GPU side, I don't know. 

But my render engine's state manager can replace `glBlendFuncSeparate` with `glBlendFunc` if both RGB and A blending options are same AND previous blend state is `glBlendFunc` not `glBlendFuncSeparate`. I like this idea and added to my TODOs. 🤗

ggetz:
Hi @donmccurdy and all, we have plans to support this extension in gltf-pipeline (PR https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/387#discussion_r198687435). Are there any more updates expected here?

ggetz:
Additionally, is `EXT_blend` the final name of this proposed extension, or will it be `KHR_` or a vendor prefix?

donmccurdy:
Renamed to `KHR_blend`.

The more likely updates to this extension are those mentioned in https://github.com/KhronosGroup/glTF/pull/1302#issuecomment-380823238. Most importantly, whether we should move away from the flexible but low-level approach here, toward just allowing a few common blend states. For example, only define `ADDITIVE` and `MULTIPLICATIVE` (or `MODULATE`). Those modes (and many others) can be composed from the low-level blend functions defined here.

Unreal only supports those two blend modes (in addition to the three already in glTF: opaque, masked, translucent), unless there is a mechanism to use more modes than [Unity docs describe here](https://docs.unrealengine.com/en-us/Engine/Rendering/Materials/MaterialProperties/BlendModes), so that's an argument for simplifying. On the other hand, Unity's docs claim the following modes are ["the most common blend types"](https://docs.unity3d.com/Manual/SL-Blend.html):

```
Blend SrcAlpha OneMinusSrcAlpha // Traditional transparency
Blend One OneMinusSrcAlpha // Premultiplied transparency
Blend One One // Additive
Blend OneMinusDstColor One // Soft Additive
Blend DstColor Zero // Multiplicative
Blend DstColor SrcColor // 2x Multiplicative
```

ggetz:
Thanks @donmccurdy!

We'd be fine with ether the abstracted blend states or the low level blend functions.

MiiBond:
From the perspective of Adobe Dimension, we will most likely only need the simple additive and multiplicative blending.

The Adobe Animate team are interested in this extension as well and have more complex uses.
https://helpx.adobe.com/animate/using/applying-blend-modes.html#blend_mode_examples

So, I think Adobe would prefer to keep this full specification rather than simplify it.

donmccurdy:
I think https://github.com/KhronosGroup/glTF/issues/822#issuecomment-274856972 makes an argument that `alphaMode=MASK` should also be allowed with this extension, not just `BLEND`. @erich666 or @AurL do you have use cases for alpha cutout/mask alongside blending?

donmccurdy:
Combining alpha blending with alpha cutoff seems practical, and probably worth including with this extension...  If so: core material `alphaMode` may be BLEND or MASK. If set to MASK, both `alphaCutoff` and blending specified by KHR_blend should be applied.

pjcozzi:
@donmccurdy is this ready to merge once it has a JSON schema?

@MiiBond have you implemented this?  If not, does it look like it will meet your needs and should we wait for you to implement it?

@ggetz did we implement this in Cesium?

donmccurdy:
There's an implementation of this extension in at least glTF-Pipeline, at least.

I'm still not really confident whether the low-level blend modes here are the right choice, or whether we want discrete named states like:

Add, Subtract, Multiply, Screen, Divide, Difference, Darken, Lighten, Overlay, Dodge, Burn, Hue, Saturation, Value, Color, Soft Light, Linear Light, ...

Before we merge the extension, additional feedback (@MiiBond, @AurL, others?) would still be welcome. It also seems important to know whether model sources like Sketchfab or authoring tools are likely to implement it, or whether it's just engines interested so far. I think Google Poly will likely use the extension, but need to re-confirm on that.

donmccurdy:
One other thought – I'm very tempted to pull the GL integer constants out in favor of string enums. In general I think new extensions should likely shift that way.

lexaknyazev:
> I'm very tempted to pull the GL integer constants out in favor of string enums.
> I'm still not really confident whether the low-level blend modes here are the right choice

I have the same concerns. Let's discuss on the next call.

---

For the reference, here's the current state of API-level options.

### Factors
```
ZERO
ONE
SRC_COLOR
ONE_MINUS_SRC_COLOR
DST_COLOR
ONE_MINUS_DST_COLOR
SRC_ALPHA
ONE_MINUS_SRC_ALPHA
DST_ALPHA
ONE_MINUS_DST_ALPHA
CONSTANT_COLOR
ONE_MINUS_CONSTANT_COLOR
CONSTANT_ALPHA
ONE_MINUS_CONSTANT_ALPHA
SRC_ALPHA_SATURATE
```
There are additional four factors for "Dual-Source Blending" that are supported by modern pipelines. They require specific fragment shader outputs, so we cannot expose them. 

### Operations
```
FUNC_ADD
FUNC_SUBTRACT
FUNC_REVERSE_SUBTRACT
MIN
MAX
```
`MIN` and `MAX` require OpenGL ES 3.0+ or `EXT_blend_minmax`.

#### Advanced Operations
There are 46 more blending operations available in [`GL_NV_blend_equation_advanced`](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_blend_equation_advanced.txt) or [`VK_EXT_blend_operation_advanced`](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VK_EXT_blend_operation_advanced) extensions. These blending equations do not use source and destination factors. Instead, each blend operation specifies a complete equation based on the source and destination colors. 

It's worth noting that Khronos-ratified version of the OpenGL extension ([`GL_KHR_blend_equation_advanced`](https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_blend_equation_advanced.txt)) exposes only 15 of them. The same 15 operations are available in HTML5 Canvas2D API and will probably be available in WebGL via proposed [`WEBGL_blend_equation_advanced_coherent`](https://www.khronos.org/registry/webgl/extensions/proposals/WEBGL_blend_equation_advanced_coherent/) extension.
```
MULTIPLY
SCREEN
OVERLAY
DARKEN
LIGHTEN
COLORDODGE
COLORBURN
HARDLIGHT
SOFTLIGHT
DIFFERENCE
EXCLUSION
HSL_HUE
HSL_SATURATION
HSL_COLOR
HSL_LUMINOSITY
```

ggetz:
This is implemented in Cesium as of https://github.com/KhronosGroup/glTF/pull/1302#issuecomment-425699516 in https://github.com/AnalyticalGraphicsInc/cesium/pull/6805.

https://github.com/KhronosGroup/glTF/pull/1302#issuecomment-425699516 would just need to be addressed.

donmccurdy:
Thanks @lexaknyazev, good points. Let's discuss more this week. In that list of 15 blend modes, I'm surprised not to see additive blending. Have I missed something, or are those 15 meant to be an addition to other blend modes?

Thanks for the update @MiiBond! That's a good point about render order, and perhaps depth test for that matter. I wonder if that should be in scope for this extension, or considered in in parallel.

I'm not opposed to the low-levelness so much as the possibility that many tools may only support certain discrete blend modes. If users write an arbitrary combination of factors and operations, the result won't load in most DCC tools but might be OK in most engines. And even if we stick with low-level modes, we shouldn't be too OpenGL-specific.

lexaknyazev:
>  are those 15 meant to be an addition to other blend modes?

Yes, since they are defined in an extension.

Original NVIDIA-proposed 46 modes include a full set of "high-level" blending operations including those that can be represented with low-level factors/operations.

> And even if we stick with low-level modes, we shouldn't be too OpenGL-specific.

I think there are two distinct question here:
- low-level (factors + operations) vs. high-level (like in Photoshop);
- string enums vs. OpenGL enums vs. our own integer enums.

Just for reference, all OpenGL ES 3.0 blending modes have 1:1 mappings in other graphical APIs.

donmccurdy:
Closing this PR for the time being. Currently we are more focused on the extensions listed here: https://github.com/KhronosGroup/glTF/tree/master/extensions#in-progress-khronos-and-multi-vendor-extensions

echadwick-wayfair:
Posting comments from a [recent conversation](https://khronosdevs.slack.com/archives/CDRQGHM3J/p1617839255011200) in the KhronosDevs Slack.

Would be nice to mimic the SparkARStudio BlendMode settings: 
https://sparkar.facebook.com/ar-studio/learn/reference/enums/MaterialsModule.BlendMode

I created this chart a few years back when I was developing for the Wii (not using glTF). I found it useful when making shaders in Unity as well. I'd love to see this available in glTF someday, so powerful.
image

mikeskydev:
Is there any interest in reviving an extension like this? `KHR_techniques_webgl` is now retired, and nearly 5 years later there's still no officially supported way for different types of blending, as far as I can see.

andybak:
Hi. I'd like to add my voice in favour of needing additive blending. I think reducing the scope of this to being purely about additive would maybe help move things forward. Other modes have *far* less utility and are used much less commonly.

My strongest argument would be that it is present as a core option in nearly every material editor I've ever used. 

It's essential to recreate many common effects and far beyond what has been referred to as "particle effects" (itself an incredibly broad area in terms of application)

Along with unlit - additive is used for many things where you want to model elements that aren't solid surfaces. It can be used for light beams, fog, a cheap glow/bloom, planet atmospheres, halos, fire and so on.

It's critical for us on Open Brush - most existing artworks make extensive use of additive brushes. But I suspect we're not alone in this respect.

PR #1303, opened 2018-04-01
Fix typo in normal vector calculation example

ale64bit:


pjcozzi:
Good eye, thanks @ale64bit!

PR #1304, opened 2018-04-02
Add `libgltf`, `libgltf_ue4` and `glTF for UE4` to C++ tool list

alexchicn:


donmccurdy:
I'm not sure I understand the difference between the 2nd and 3rd — what is "third-party" vs "plugin" and "load" vs "import"? Is one a dependency of the other?

alexchicn:
If you want to use a static/shared library in **Unreal Engine 4**, you need to make it as a **third-party library** by a `***.Build.cs`([ref: UE4's build tool](https://wiki.unrealengine.com/Linking_Static_Libraries_Using_The_Build_System)) file.
The **UE4's plugin** is dependent on the UE4's engine code, but the **UE4's third-party library** is not.
* `libgltf`: You can compile or link in your not-UE4 project.
  * This project is not only developed for UE4.
* `libgltf_ue4`: It links the `libgltf` to UE4 by `libgltf_ue.Build.cs`.
  * You can just use the `libgltf_ue4` without `glTF for UE4` in your UE4 project.
* `glTF for UE4`: It integrates the `libgltf` and UE4 by `libgltf_ue4` and convert the **glTF** to `UStaticMesh`, `UTexture2D`, `UMaterial`, etc.

:sweat_smile: :smiley:

pjcozzi:
@donmccurdy OK to merge this?

donmccurdy:
I'm still confused but I know nothing about UE4 😅 .. if this will make sense to a UE4 audience then looks good to me! If not we can just add a little more description to each.

emackey:
The description of the middle one in particular doesn't make sense (to me, also as a non-UE4 person)

> A third-party of UE4.10+ to load the glTF file

Googling around for UE4, I see some pre-existing confusion among the user base between UE4 plugins and UE4 modules.  One of these is listed as a plugin, the other is listed as "a thirdparty".  Is that a module?

alexchicn:
@emackey A third-party is not dependent the UE4 engine or another minor version of the UE4, but the engine depends on it by a `.Build.cs` file - `Engine/Source/ThirdParty/libcurl/libcurl.Build.cs`.

> **For example:**
> The UE4 use `curl` as a third-party library for the HTTP network. And you can compile the `curl` without the UE4 engine.

You can find a lot of third-party libraries in the UE4 engine. And most of them don't include the source code, just have some header files and some compiled `.so`, `.a`, `.lib`, `.dll` or another static/shared library files.

**About the module of the UE4:**

A third-party or plugin belongs to a kind of module in the UE4.

emackey:
Thanks @alexchicn.  Can we add the word `library` after `third-party`?  Also, it works for more than one glTF file, right?  So it should be plural, "glTF files" not "the glTF file".

Maybe update the descriptions like this:

>  A third-party library for UE4.10+ to load glTF files

>  A plugin for UE4.10+ to import glTF files

Would that be OK?

alexchicn:
Thank @emackey. Yes, you can and it is OK. 😃

emackey:
Wording update in fccf92eb1e

PR #1325, opened 2018-04-24
added cqparts as exporter

fragmuffin:
Added to readme exporter list.

**How to export**
Install `cqparts` with:

```bash
pip install cqparts
```

Then export `cube.step` -> `cube.gltf` & `cube.bin` (for example) with:

```python
from cqparts import Part
Part.importer('step')('cube.step').exporter('gltf')('cube.gltf')
```

**Docs**
gltf exporter class documentation: https://fragmuffin.github.io/cqparts/doc/api/cqparts.codec.html#cqparts.codec.gltf.GLTFExporter

**Example Output**
https://fragmuffin.github.io/cqparts/doc/tutorials/assembly.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Preview tools:
 | [jgltf-obj](https://github.com/javagl/JglTF/tree/master/jgltf-obj) | `OBJ` | ![status](https://img.shields.io/badge/glTF-1%2E0-yellow.svg?style=flat) | Java programmatic and command-line converters |
 | [Docker collada2gltf](https://hub.docker.com/r/winsent/collada2gltf/) | `COLLADA` | ![status](https://img.shields.io/badge/glTF-%3F-lightgrey.svg?style=flat) | Docker container for Python web service, built on COLLADA2GLTF |
 | [collada2gltf-web-service](https://github.com/AnalyticalGraphicsInc/collada2gltf-web-service) | `COLLADA` | ![status](https://img.shields.io/badge/glTF-%3F-lightgrey.svg?style=flat) | Node.js web service, built on COLLADA2GLTF |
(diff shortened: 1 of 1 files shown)


emackey:
Thanks @fragmuffin.  The links you provide in the modified README don't mention glTF, only the link you gave in the issue text above talks about glTF.  Presumably users looking at glTF's main README will want to link to the part of your project that deals with glTF, can you fix the link in this PR to do so?

@donmccurdy Any thoughts?

donmccurdy:
Very nice! The GLTFExporter-specific docs don't say as clearly how to actually do an export, I wouldn't get rid of any of the current links. Maybe:

> Python CAD for programmers ([import / export doc](https://fragmuffin.github.io/cqparts/doc/cqparts/import-export.html)) ([GLTFExporter](https://fragmuffin.github.io/cqparts/doc/api/cqparts.codec.html#cqparts.codec.gltf.GLTFExporter))

fragmuffin:
@emackey @donmccurdy thanks for pointing that out... I usually try to avoid documenting things twice (i this case, how to select and use an exporter), but given this context, it's definitely worth it.

I've updated the description text to that suggested by @donmccurdy 
Also, now the linked [`GLTFExporter`](https://fragmuffin.github.io/cqparts/doc/api/cqparts.codec.html#cqparts.codec.gltf.GLTFExporter) documentation to give some code examples.

emackey:
Thanks!

fragmuffin:
Thank _you_ @emackey & @donmccurdy  :smile:

dberardo-com:
old thread, but has anyone come up with a simple-to-replicate solution?

seems like many links from this thread return 404 by now. how to export step files into gltf ?

javagl:
@dberardo-com The project seems to have moved to https://github.com/cqparts/cqparts . Starting from there, one can find some documentation like https://cqparts.github.io/cqparts/doc/api/cqparts.codec.html?highlight=gltf#cqparts.codec.gltf.GLTFExporter

Beyond that, when looking for tools related to glTF, you can have a look at https://github.khronos.org/glTF-Project-Explorer/ . The STEP format does not seem to be soooo widely used as a basis for converting to glTF. But when searching the page (with the browser "CTRL+F" search), some of the projects seem to support STEP.

PR #1338, opened 2018-05-09
Cesium z-up extension

### LINKED ISSUE #1324: Up axis extension
Would there be any interest in an extension that allows redefining which axis is the up axis?

I come to this from a 3D geospatial angle where globe engines and 3D models often use the WGS84 coordinate system, whose origin is the Earth's center and up-axis is the north pole - essentially a z-up coordinate system.

In order to bring WGS84 (inherently z-up) models into Cesium we have to go through a few hoops including attaching a z-up to y-up transform to the root node of the glTF and then applying a reverse y-up to z-up transform at runtime. Related to this, we are building [3D Tiles](https://github.com/AnalyticalGraphicsInc/3d-tiles) with a z-up coordinate system and it is confusing that the glTF payload is y-up.

To summarize, it would be convenient if we could specify a different up axis than y.

The obvious downside to an extension such as this is now different glTF models are not following the same convention. Nonetheless would this extension be of interest to others?

ggetz:
Fixes #1324

@lilleyse Please let me know if you have any initial feedback.


lexaknyazev:
Couple optional comments.

- This seems to be the first extension without an extension object. Maybe, it's worth to mention in the readme that it doesn't exist just in case.
- > an implementation must treat the mesh geometry as _z_-up
  
  I assume that everything else (node and IBM matrices) should be treated as usual, right? Maybe, the readme could also emphasize that _z_-up applies to _all_ mesh geometry.

emackey:
I have to ask, is an extension really needed here?  Could one just place a parent node with a 90-deg rotation at the top of the node tree, rather than applying this extension?  Are there performance implications for adding such a parent node?

What's really the difference between a model that specifies this extension, vs a model that has a root node with a rotation added to it?

lilleyse:
Conceptually there isn't any difference, but the purpose of this PR is to avoid that extra node.

For someone who has WGS84 referenced data in their glTF it is a lot easier to tell them to include this extension than describe the changes they need to make to the node hierarchy.

It also allows glTF and 3D Tiles to share the same coordinate system. Although this may not be important broadly, it is really convenient in 3D Tiles pipelines.

ggetz:
Closing in favor of solution in https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/307

Issue #1342, opened 2018-05-17
KHR_draco_mesh_compression's bufferView decoded metadata can and does differ with the document's accessor values.


najadojo:
The Draco data stream contains vertex buffer metadata that mirrors that found in the accessor for the same buffer. KhronosGroup/glTF-Sample-Models#147 indicates that there are models where this metadata differs between the two sources data types. Additionally primitive vertex count can be different between the too. For example [BrainStem](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BrainStem/glTF-Draco) contains a primitive where the Draco vertex count is 806 and the accessor associated with that vertex attribute lists a count of 802. In this case there is no bufferView for the accessor, the accessor is simply giving incorrect information. 

The spec says that the bufferView be ignored when the draco extension is present for an attribute and the above presumes we also ignore the accessor to get expected result. Should we clarify that when no fallback is desired that the core attribute and its accessor not be present? This would eliminate the confusion above. It seems some clarifying language on the indices maybe necessary as well, the spec suggests to process the core but then read the data from draco. What does process mean here?

I'd also expect that when a fallback or mix and match attributes (ie some from draco some from core) is desired the data types and lengths match within a primitive. Ideally this is something that is validated by glTF-Validator.

lexaknyazev:
Accessors configuration must match Draco output, so that engines could build VAOs before decompression ends. If some assets have that wrong, they must be fixed.

najadojo:
Super! Can we assert this in the validator?

FrankGalligan:
So when compressing with Draco (most of the options), this can change the number of vertices. Draco will remove vertices if they are not being used. The problem is if we want uncompressed fallback option we cannot just change the count to match what Draco compressed. So I made a change to the spec here:https://github.com/KhronosGroup/glTF/pull/1343
This change says to ignore accessor.count for compressed primitives.

I also made a change to gltf-pipeline here:https://github.com/FrankGalligan/gltf-pipeline/tree/add_uncompressed_fallback
that adds an option to output compressed and uncompressed meshes.

lexaknyazev:
@FrankGalligan Can we keep `count` the same by replacing uncompressed data with decompressed (when fallback is present) as @najadojo says? That would be more aligned with general glTF extensions design.

PR #1343, opened 2018-05-20
Change Draco accessor properties to match decompressed data

FrankGalligan:
~~For compressed only primitives loacders must ignore `accessor.count`.~~

Exporters must use decompressed data to fill uncompressed accessors to satisfy the requirement that `accessor.count` be accurate for both compressed and uncompressed fallback data.

This fixes https://github.com/KhronosGroup/glTF/issues/1342


extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md
```diff
@@ -130,6 +133,8 @@ Below is the recommended process when a loader encounters a glTF asset with the
 
 To prevent transmission of redundant data, exporters should generally write compressed Draco data into a separate buffer from the uncompressed fallback, and shared data into a third buffer. Loaders may then optimize to request only the necessary buffers.
 
+To satisfy the requirement that `accessors` properties must match the decompressed data, exporters should write the uncompressed fallback primitives from the decompressed data.
```
donmccurdy: I think the "uncompressed" vs "decompressed" distinction was not clear enough in our thread earlier, and should be spelled out here:

> Draco compression may change the order and number of vertices in a mesh. To satisfy the requirement that `accessors` properties be correct for both compressed and uncompressed data, generators should create uncompressed attributes and indices using data that has been decompressed from the Draco buffer, rather than the original source data.

FrankGalligan: Thanks @donmccurdy, I updated with your text. I think it is much clearer.

lexaknyazev:
There's one more caveat to this issue. Since morph targets are stored as deltas, a special care is needed to not break them when compressing base geometry (pushing their data through compress-decompress pipeline should be enough).

emackey:
@lexaknyazev If you don't mind can you open a new issue to capture that?  Thanks.

PR #1348, opened 2018-05-26
Describe clamping before/after sampler input range

### LINKED ISSUE #1327: Animation start time behavior
If the first sampler input of an animation has a time value greater than 0 (say 1.5), what should the runtime behavior be?

1. Start at 1.5 seconds.
2. Start at 0 seconds using clamped behavior, wait for 1.5 seconds, then continue the rest of the animation.

three.js does 2
BabylonJS currently does 1
Sketchfab, as far as I can tell, does 1
Unity does 2

The original issue here:
https://github.com/BabylonJS/Babylon.js/issues/4224

donmccurdy:
Fixes #1327.


pjcozzi:
Thanks @donmccurdy!

PR #1350, opened 2018-05-29
MOZ_alt_materials Extension

robertlong:
In [Mozilla Hubs](https://hubs.mozilla.com), we support a wide variety of hardware with vastly different specs. Publishing glTF content that can take advantage of high end hardware while still running well on low end platforms is tough. Also, hardware requirements for VR are more demanding than 2D. Currently if you want to ship a glTF asset that uses real time lighting on high end hardware and unlit lighting on low end hardware, you need to distribute two separate glTF files and the loader needs to know about the availability of those files in order to determine which asset it should load. This extension aims to solve this problem by defining alternate materials in a glTF file that can then be loaded conditionally based on the client's hardware.

This extension is similar in some regards to the [MSFT_lod](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/MSFT_lod) extension which we considered for a while, but ultimately we decided that we wanted the loader to request a material with a explicitly defined rendering technique rather than a LOD level that maps to an arbitrary material. However, the two extensions could be used in tandem (ex. MOZ_alt_materials used to specify PBR and Unlit geometry LOD groups).

We are currently using this extension in Mozilla Hubs and our [gltf-bundle](https://github.com/MozillaReality/gltf-bundle) pipeline. This includes [a tool](https://github.com/MozillaReality/gltf-unlit-generator) that automatically generates an unlit map from PBR glTF model. We have a reference implementation in our fork of the [THREE.GLTFLoader](https://github.com/mozilla/hubs/blob/master/src/vendor/GLTFLoader.js#L889) which we would like to move into the ThreeJS repository once we receive some feedback from the glTF community.

### DIFF
--- a/extensions/2.0/Vendor/MOZ_alt_materials/README.md
+++ b/extensions/2.0/Vendor/MOZ_alt_materials/README.md
@@ -14,7 +14,9 @@ Written against the glTF 2.0 spec.
 
 ## Overview
 
-This extension adds the ability to define alternate rendering techniques for a given material. A typical use-case for this extension would be defining a physically based material for high end platforms and an unlit material on low end platforms. It could also be used for defining rendering techniques for multiple engines.
+This extension adds the ability to define alternate rendering techniques for a given material.
+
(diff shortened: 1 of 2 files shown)


pjcozzi:
Cool, do you also want to add `MOZ` to [Prefixes.md](https://github.com/KhronosGroup/glTF/blob/master/extensions/Prefixes.md) in a separate PR we could merge right away?

pjcozzi:
@robertlong is this still implemented in Mozilla Hubs?  Did anything change?  What is needed to merge this?  Just a review from interested folks?

pjcozzi:
@robertlong thanks for the prompt update!  Keep us posted on MobileStandardMaterial and if that requires any glTF extensions.

PR #1355, opened 2018-06-01
Add a note about no POSITION data

### LINKED ISSUE #1332: Meshes with no POSITION semantic
The 2.0 spec doesn't require POSITION attribute semantic to be present. We need to check with implementations if that's the case and update them or the spec (or add an impl. note).

- [x] **Validator** flags that as an error.
- [x] **three.js** - mixed (see below)
- [x] **Babylon** - no error
- [x] **Cesium** - bug (see below)

/cc @donmccurdy @bghgary @emackey

lexaknyazev:
Fixes #1332.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -761,6 +761,8 @@ All indices for indexed attribute semantics, must start with 0 and be continuous
 
 > **Implementation note:** Each primitive corresponds to one WebGL draw call (engines are, of course, free to batch draw calls). When a primitive's `indices` property is defined, it references the accessor to use for index data, and GL's `drawElements` function should be used. When the `indices` property is not defined, GL's `drawArrays` function should be used with a count equal to the count property of any of the accessors referenced by the `attributes` property (they are all equal for a given primitive).
 
+> **Implementation note:** When positions are not specified, client implementations should skip primitive's rendering unless its positions are provided by other means (e.g., by extension). This applies to both indexed and non-indexed geometry.
+
 > **Implementation note:** When normals are not specified, client implementations should calculate flat normals.
 
 > **Implementation note:** When tangents are not specified, client implementations should calculate tangents using default MikkTSpace algorithms.  For best results, the mesh triangles should also be processed using default MikkTSpace algorithms.


donmccurdy:
> ...client implementations should skip primitive's rendering...

What does this mean for a "loader" implementation, rather than an "engine" implementation? For three.js, the glTF loader does not do any rendering — it needs to pass generic three.js object data to a user callback. Should we omit primitives without position from the node hierarchy? Or mark them as invisible? Does it matter?

donmccurdy:
> There could be some app-specific use cases that would rely on extensions or extras to flag something to the engine.

We cannot assume that an extension the glTF loader doesn't recognize or support will still be passed through in a way that means anything to the engine. We can set `mesh.visible=false` in three.js I think. It will require intervention by application code to make that mesh visible, which I'm not entirely sure is better than requiring intervention by application code to prevent an error, but I'm OK with it.

donmccurdy:
In other words, shouldn't `KHR_techniques_webgl` really have been marked required if the mesh is going to be ignored without it? Or is there another use case here?

lexaknyazev:
> `KHR_techniques_webgl`

That's the one case we know about. 

I can think of another: an asset contains custom attributes data and the mesh has `extras` that says something about using engine's pre-defined app-specific custom shader. Can the glTF loader just pass-through everything instead of omitting?

> intervention by application code to prevent an error

Imo, the same asset loaded in off-the-shelf glTF viewer mustn't crash it.

lexaknyazev:
@donmccurdy Are you OK with this PR as it is or should we alter the text somehow?

donmccurdy:
The text is somewhat ambiguous. A loader could reasonably "skip primitive's rendering" either by (a) not adding the primitive to the scene at all, or (b) flagging the mesh as "invisible", if the engine has such a mechanism. 

> I can think of another: an asset contains custom attributes data and the mesh has extras that says something about using engine's pre-defined app-specific custom shader.

In this situation, if the loader does (a) the application code will not receive the primitive from the loader, and can do nothing with the extras. If the loader does (b) the application code just needs to remove  the 'invisible' flag from the mesh.

emackey:
@donmccurdy I don't think this spec needs to dictate how the engine handles this case internally.  I think it's enough to say that if there's no position data and no extension, nothing gets rendered.

donmccurdy:
Ok, works for me. 👍

PR #1356, opened 2018-06-01
Clarify accessor.count for morph targets

### LINKED ISSUE #1339: Morph Targets with different accessors count
What are the rules in place if I have a

```
"meshes": [
    {
      "primitives": [
        {
          ...
          "targets": [
            {
              "NORMAL": 3,
              "POSITION": 4,
              "TANGENT": 5
            }
        ...
```

And the `accessors` at 3, 4, and 5 have different `accessors.count`

First I am not sure if a model can have a different number of `NORMAL` vec3 then `POSITION` since its all a per-vertex change.

I guess not sure if this something that is implict and not stated in the spec or something to be aware of when parsing to not assume the same count, which if later then how does one handle that

lexaknyazev:
Fixes #1339.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -792,7 +792,7 @@ Valid accessor type and component type for each attribute semantic property are 
 
 `POSITION` accessor **must** have `min` and `max` properties defined.
 
-All Morph Target's accessors **must** have the same `count` which must match the `count` of accessors of the original mesh.
+All Morph Target's accessors **must** have the same `count` as the accessors of the original primitive.
 
 A Morph Target may also define an optional `mesh.weights` property that stores the default targets weights. In the absence of a `node.weights` property, the primitives attributes are resolved using these weights. When this property is missing, the default targets weights are assumed to be zero.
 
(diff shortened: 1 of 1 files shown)


specification/2.0/README.md
```diff
@@ -792,6 +792,8 @@ Valid accessor type and component type for each attribute semantic property are
 
 `POSITION` accessor **must** have `min` and `max` properties defined.
 
+All Morph Target's accessors **must** have the same `count` which must match the `count` of accessors of the original mesh.
```
emackey: You use `must` twice for the same directive.  How about something like this?

> The `count` of each Morph Target accessor **must** exactly match the `count` of the accessor for the original mesh.

lexaknyazev: > The `count` of each Morph Target accessor **must** exactly match the `count` of the accessor for the original mesh.

Is it clear what "the accessor" means?
We're trying to say that all vertex data accessors have the same `count` - no matter original mesh or morph target.

emackey: How about:

> All Morph Target's accessors **must** have the same `count` as the accessors of the original mesh.

or

> For a given mesh, all vertex data accessors, including morph target accessors, **must** have the same `count`.

emackey: or replace "mesh" with "primitive"

Issue #1321, opened 2018-04-22
Node tree in 2.0


dakom:
The docs state:

> For Version 2.0 conformance, the glTF node hierarchy is not a directed acyclic graph (DAG) or scene graph, but a strict tree. That is, no node may be a direct or indirect descendant of more than one node. This restriction is meant to simplify implementation and facilitate conformance. The restriction may be lifted later.

"The restriction may be lifted later" feels a bit ominous ;) - if a renderer adheres to the spec today, and then this restriction is lifted, that could break the renderer even if it doesn't break exporters (since a renderer would likely parse the gltf into its own runtime structure and make certain assumptions based on this restriction)

pjcozzi:
@dakom thanks for the feedback.  If the restriction was lifted, it would, of course, be in a future 3.x version where a breaking change is allowed.

@lexaknyazev @bghgary I suggest we just remove "The restriction may be lifted later."  I don't think it adds any value and clearly is creating some confusion.  What do you think?

bghgary:
> @lexaknyazev @bghgary I suggest we just remove "The restriction may be lifted later." I don't think it adds any value and clearly is creating some confusion. What do you think?

+1 from me. I have had discussions about this many times and it makes sense to me to remove it.

lexaknyazev:
+1, let's clarify these as well (question is virtually the same): #1086, #1096.

pjcozzi:
@lexaknyazev any chance you want to make the spec change? ☺️

pjcozzi:
Bump @lexaknyazev??? 😀

lexaknyazev:
Will open a PR to address those.

lexaknyazev:
Fixed with #1358.

Issue #1096, opened 2017-09-11
Node hierarchy clarification


recp:
Hi,

Recently I got some free time and started to implement glTF 2.0 in AssetKit (http://github.com/recp/assetkit).

I've confused about glTF's node hierarchy,  references are 
representing as array indices, this is clear but the part 
I don't understand is that why child nodes are representing as array indices?

wouldn't something like this be more clear?:

```JSON
"nodes": [
  {
      "mesh": 0,
      "children": [
        {
           "mesh": 1
        },
        {
           "mesh": 2
        },
      ]
  }
]
```

> "That is, no node may be a direct or indirect descendant of more than one node."

```JSON
{
    "nodes": [
        {
            "name": "Car",
            "children": [1, 2, 3, 4]
        },
        {
            "name": "wheel_1"
        },
        {
            "name": "wheel_2"
        },
        {
            "name": "wheel_3"
        },
        {
            "name": "wheel_4"
        }        
    ],
   "scenes": [
        {
            "nodes": [
                0 
            ]
        }
    ],
    "scene": 0
}
```

In this case; can `scene` pick `wheel_1` node as root node ("scene": 1)? 
Is it valid **(because wheel_1 is child node of nodes[0])**? 

I implemented like this; I moved all referenced child nodes inside node, 
since I'm confused I'm considering to set child nodes as instance node (just pointer) instead of moving,
in this way referenced nodes will still remain in same index as ROOT node

emackey:
This is similar to the first question in #1086.

javagl:
The second part of the question is certainly related to #1086 , and will hopefully clarified there.

Regarding the general approach, I can think of some points that may justify the current structure:

- Directly "linlining" the child nodes **in** their parent note might be easier to parse (for humans as well as for libraries), but I could imagine that this avantage vanishes when the node hierarchies become deeper. (It *might* then even become a disadvantage, in terms of performance, but this is just a guess)
- Having all nodes in a flat array allows modeling more than just trees. The section at https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#nodes-and-hierarchy says (emphasis by me):

  > For Version 2.0 conformance, the glTF node hierarchy is not a directed acyclic graph (DAG) or scene graph, but a strict tree. That is, no node may be a direct or indirect descendant of more than one node. This restriction is meant to simplify implementation and facilitate conformance. **The restriction may be lifted later**.

[...]

Issue #1086, opened 2017-09-02
Miscellaneous spec questions


scurest:
> For Version 2.0 conformance, the glTF node hierarchy is not a directed acyclic graph (DAG) or scene graph, but a strict tree. That is, no node may be a direct or indirect descendant of more than one node.

The nodes form a forest (a disjoint union of trees), not a single tree. For example, the sample model 'Cameras' has a node graph that is a forest but not a tree. For the same reason, the term "_the_ glTF node hierarchy" is somewhat unfortunate.

Also, saying a node cannot be the _indirect_ descendant of multiple nodes is somewhat misleading, since in eg. a tree, a node generally does have multiple ancestors. It's the "direct descendant" part that can't happen.

-----

Are the nodes referenced in `scene.nodes` required to be the roots of trees in the node forest? [This comment](https://github.com/KhronosGroup/glTF/issues/1035#issuecomment-313859346) suggests that they are but I'm not sure. The description for this property is

> nodes | integer [1-*] | The indices of each root node.

It's not clear whether "root node" is placing a requirement that the nodes be roots in the node graph, or whether they're being called root nodes just because they're roots for this particular scene and no requirement is being imposed.

If its the former, can this be clarified, and also, this being the only occurrence of "root node" in the spec, can "root node" be defined somewhere in the text?

-----

AFAICT, it is permitted for a primitive to define one but not both of the `JOINTS_0` and `WEIGHTS_0` attributes. Is this correct, and if so what are the semantics for skinning?

-----

> Implementations must use following equations to get corresponding floating-point value f from a normalized integer c and vise-versa: [table follows]

This appears in the "Animations" section. Am I correct in assuming that the spec means you must use these formulas _when reading an accessor for animation data_ (specifically, when reading an accessor for a sampler) and not in general?

For example, when drawing a mesh, it is acceptable to use whatever `glVertexAttribPointer` does when its `normalized` parameter is `true` (which I believe is f = (2c + 1)/(2^b - 1) for older versions of the GL) correct? But since animation data is decoded in the implementation (rather than the GL), it's being specified what formula to use?

If so, can this be clarified?

-----

According to the overview, the `byteStride` property is defined on an accessor, but it is actually defined in the buffer view (incidentally, the way the overview has it makes more sense to me. Why _isn't_ it defined on an accessor?)

![ex](https://user-images.githubusercontent.com/11024420/29992627-04b4ae32-8f66-11e7-9d7f-6013365497b0.png)

[...]

PR #1358, opened 2018-06-02
Clarify node hierarchy terms

lexaknyazev:
This PR incorporates feedback from #1086, #1096, and #1321.

Please confirm with implementations:
> All nodes listed in `scene.nodes` array must be the roots of trees in the node forest.

If it's true, no other action is needed (the validator has been checking for that since the beginning).

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -265,6 +265,8 @@ The [node transformations](#transformations) and [animation channel paths](#anim
 
 The glTF asset contains zero or more *scenes*, the set of visual objects to render. Scenes are defined in a `scenes` array. An additional property, `scene` (note singular), identifies which of the scenes in the array is to be displayed at load time.
 
+All nodes listed in `scene.nodes` array must be the roots of trees in the node forest (see the next section for details).
+
 When `scene` is undefined, runtime is not required to render anything at load time.
 
 > **Implementation Note:** This allows applications to use glTF assets as libraries of individual entities such as materials or meshes.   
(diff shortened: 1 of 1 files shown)


specification/2.0/README.md
```diff
@@ -265,6 +265,8 @@ The [node transformations](#transformations) and [animation channel paths](#anim
 
 The glTF asset contains zero or more *scenes*, the set of visual objects to render. Scenes are defined in a `scenes` array. An additional property, `scene` (note singular), identifies which of the scenes in the array is to be displayed at load time.
 
+All nodes listed in `scene.nodes` array must be the roots of trees in the node forest (see the next section for details).
```
emackey: I love the term `node forest` but in the next section you redeclare it as `node hierarchy`.  Perhaps pick one or the other?

lexaknyazev: It's not me. "node hierarchy" is used many times in the spec, while "forest" is quite stable term w.r.t. graph theory. Do you think we can introduce these notions better?

emackey: This just jumped out at me from reading this set of diffs.  If no one is actually bothered by it, I'm OK.

bghgary: I'm not bothered by "node forest" but I think its usage is unnecessary here. How about this?
> All nodes listed in `scene.nodes` array must be root nodes (see the next section for details).

donmccurdy:
Related, but not fully specified even after these clarifications – can two scenes reference the same 'root' nodes? For example:

```
{
  scenes: [
    {nodes: [0]},
    {nodes: [0]}
  ]
}
```

Issue #1252, opened 2018-02-20
Tangent-basis workflow for getting 100% correct normal-mapping


Ziriax:
Tools like Substance Painter have templates for targetting Unity 5 and Unreal 4, to make sure that the tangent-space normal-map baking uses exactly the same shaders as in these engines (so that the same tangent-basis is used).

What is the correct way to bake normal-maps for GLTF / PBR?

We tried some extreme models, but can't seem to get correct results.

Did anyone succeed in using e.g. Blender + Substance Painter for this?

Thanks,
Peter Verswyvelen

bghgary:
> What is the correct way to bake normal-maps for GLTF / PBR?

If the glTF includes tangents, then the normal maps should be in whatever tangent space is defined by the normal and tangent vectors + bitangent sign. If the glTF does not include tangents, the spec says this:

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#meshes

>> Implementation note: When tangents are not specified, client implementations should calculate tangents using default MikkTSpace algorithms. For best results, the mesh triangles should also be processed using default MikkTSpace algorithms.
>
>> Implementation note: When normals and tangents are specified, client implementations should compute the bitangent by taking the cross product of the normal and tangent xyz vectors and multiplying against the w component of the tangent: `bitangent = cross(normal, tangent.xyz) * tangent.w`

Tech art is telling me that Substance Painter currently does not write out tangents when exporting glTF files, so the normal maps needs to be in MikkTSpace. I'm not sure what Substance Painter is doing here. If you bring a model generated with MikkTSpace algorithms without tangents into Unity, for example, and set the import settings to generate tangents, then the normal maps should match.

donmccurdy:
As a warning to anyone testing on e.g. https://gltf-viewer.donmccurdy.com/, three.js currently ignores stored tangents and computes them in a shader, instead.

bghgary:
The same is true for BabylonJS [sandbox](http://sandbox.babylonjs.com/) though BabylonJS uses a slightly different tangent space calculation in the shader.

Ziriax:
Thanks for the info and fast reply.

We are writing a custom renderer (and the Maya2glTF exporter) and would
like to use glTF as our mesh format, so we need to understand all the
details.

So far using the MikkTspace source code from Blender does not result in
correct renderings. This code also accepts an `angularThreshold` to smooth
the tangents, reducing the number of vertex splits, so this argument must
[...]

PR #1364, opened 2018-06-06
Add a note about tangent.w values

lexaknyazev:
_Following up from https://github.com/KhronosGroup/glTF/issues/1252#issuecomment-395024604._

Should different `tangent.w` values be considered as a warning or an error?


bghgary:
> Should different `tangent.w` values be considered as a warning or an error?

I vote warning. If there are degenerate triangles with bad values, it won't display and the asset overall might still look fine.

pjcozzi:
@bghgary please merge when ready.

PR #1370, opened 2018-07-02
Add Maya2glTF to exporter list and improve getting started for artists section

robertlong:
I think Maya2glTF is mature enough to be featured on the readme page. It would also be helpful to show the 3DS Max and Maya exporters in the "Getting Started For Artists" section.

Thanks @j-conrad for pointing this out!


pjcozzi:
Great, thank you.

And huge thanks to @Ziriax!  :1st_place_medal:

PR #1377, opened 2018-07-05
EXT_lights_image_based extension

MiiBond:
Here's a super rough pass of the IBL extension that I've been working on with Microsoft (@bghgary).

It needs a lot more info added, of course, but we thought we'd get it out there early.


pjcozzi:
Should the name be `EXT_lights_image_based`?

Seems consistent with WebGL extensions: https://www.khronos.org/registry/webgl/extensions/

We basically followed this naming convention for glTF extensions, except for `KHR_materials_pbrSpecularGlossiness` (not sure if there was a reasonable, perhaps just an oversight).  https://github.com/KhronosGroup/glTF/blob/master/extensions/README.md

Also, if we expect quick and broad adoption, perhaps just start with `KHR_` prefix and it will be draft until ratified.

pjcozzi:
@ggetz @lilleyse @bagnell please take a look for Cesium.

McNopper:
Can we please put the definition of the HDR textures in another extension?
If one later wants to use other formats, this should not affect this extension.

McNopper:
Is there a reason, not to provide the whole mip map chain?

McNopper:
Finally, I would prefer to encode specularImages in such a way, that later e.g. one image can be referenced having all sides plus mip map levels.

gnagyusa:
Or, an even better way would be to allow IBL environment image changes per transform node:
```
"nodes":
[
  {
    "name":"MyNode1"
    "extensions" :
    {
      "EXT_lights_imageBased" : { "light" : 0 }
  }
}
]
```

bghgary:
@pjcozzi 
> We basically followed this naming convention for glTF extensions, except for KHR_materials_pbrSpecularGlossiness (not sure if there was a reasonable, perhaps just an oversight). https://github.com/KhronosGroup/glTF/blob/master/extensions/README.md

@MiiBond 
> I did talk to @bghgary about the naming and I believe we landed on the camelCase for imageBased because of the pbrSpecularGlossiness extension. I'm fine with changing it as I do think all lower case looks nicer.

I think when we added the KHR_materials_pbrSpecularGlossiness, we were (or at least I was) thinking the pattern was `<KHR|EXT|vendor>_<some category>_<something specific about that category>`. We also called the metal/rough object `pbrMetallicRoughness` in the core spec, and the casing for pbrSpecularGlossiness would be more consistent than pbr_specular_glossiness.

That said, +1 to using snake_case for image_based.

bghgary:
@McNopper 
> Can we please put the definition of the HDR textures in another extension?
> If one later wants to use other formats, this should not affect this extension.

We were thinking that an extension isn't necessary for HDR textures, just like we don't need a flag specifying the color space of a texture. It is implied by the object referencing the texture/image. The original proposal was to do this but I found that implementing it in Babylon was more difficult if we used an extension. Thoughts?

> Is there a reason, not to provide the whole mip map chain?

We went back a forth a few times on this. Most of the time, we don't want the whole mip map chain because the smallest mip maps don't have enough detail for the environment. It's the offset in some of the environment map creation tools (e.g. Lys). By not specifying all the levels, you implicitly also specify the offset and don't need to provide the unused images. The down-side is that some browsers don't support uploading partial chains and thus you have to fill the missing chains manually.

> Finally, I would prefer to encode specularImages in such a way, that later e.g. one image can be referenced having all sides plus mip map levels.

Can you elaborate on this? I'm not sure what you mean.

bghgary:
@gnagyusa 
> support per-mesh IBL maps, a.k.a. "light probes" / "reflection probes"
> allow IBL environment image changes per transform node

I'm not an expert in this, but I would think this should be a separate extension.

lexaknyazev:
> We were thinking that an extension isn't necessary for HDR textures, just like we don't need a flag specifying the color space of a texture. It is implied by the object referencing the texture/image.

I think we can start with a special PNG encoding that can be used without additional extensions (implying that encoding scheme and `image/png` are enforced by this extension). But I'd like us to have a way to use true HDR data in the future without deprecating this extension.

bghgary:
> I'd like us to have a way to use true HDR data in the future without deprecating this extension.

+1. Does KTX2 help at all with this?

lexaknyazev:
> +1. Does KTX2 help at all with this?

Sure, even KTX1 can hold BC6H cube map with mip levels.

bghgary:
I mean the texture transmission extension in conjunction with KTX2. Are there plans to support HDR?

lexaknyazev:
> Are there plans to support HDR?

Ultimately, yes. Can't say anything about timeline, though.

McNopper:
After this discussion, I recommend that `specularImages` links to one KTX1 file.
Having one KTX1 file solves the follwoing problems:
- Possibility to have true HDR now
- Provide whole mip-map chain
- `specularImageSize` can be removed
- `specularImages` gets simpler

McNopper:
Strongly disagree just having PNG for specular images, as the color channel resolution and finally the quality is too low.
Also, recommend to call this extension `EXT_lights_environment` or something similar, as using spherical harmonics is not image based.

McNopper:
Finally, providing the whole mip map chain and going down to 1x1 pixel should be sufficient. Please comapre with this tool:
https://github.com/derkreature/IBLBaker

MiiBond:
For the IBL-per-node idea, we did discuss that and I can't remember why we didn't allow it. Maybe lack of runtime support? I think Babylon and Three.js should both be able to handle this.

EXT_lights_imageBased -> EXT_lights_environment
I don't have a strong opinion here. The spherical harmonics are generated from an image but, yeah, they aren't an image themselves. I could go either way.

Leaving the door open for using other texture types would be nice. We just didn't want to support a long list of formats and layouts. e.g. equirectangular or cube layout, RGBE, RGBM, RGBD, logLuv-encoded PNG's, .hdr, .exr, etc.

As for the mip chain, I believe the common practice is to have an offset that specifies which mip represents a fully convolved (i.e. corresponding to roughness = 1.0) reflection. Then the shader interpolates down to that mip. Using the 1x1 mip for this isn't useful. So, we either specify an offset or we specify only a portion of the mip chain and let that implicitly define which mip corresponds to roughness = 1. Gary and I discussed both options but I'm curious what others think.

UX3D-nopper:
Regarding EXT_lights_imageBased -> EXT_lights_environment:

Unity: https://docs.unity3d.com/Manual/GlobalIllumination.html#Environment
Unreal: https://docs.unrealengine.com/en-US/engine/rendering/lightingandshadows/lighttypes/skylight
On quick look I found environment or sky light.

I personally would prefer environment, as it describes the ligth type and note the technology/algorithm used,

UX3D-nopper:
Images do become a texture:
```
"specularEnvironmentFactor" : 1.0,
"specularEnvironmentTexture" : {
   "index": 0
}
```

UX3D-nopper:
Finally, we need an extension for textures:
```
{
   "sampler" : 0, 
   "source" : 0
},
{
   "sampler" : 1, 
   "source" : [
      1, 2, 3, 4, 5, 6
   ]
},
{
   "sampler" : 2, 
   "source" : [
      [7, ..., n],
      [...],
      [...],
      [...],
      [...],
      [...]
   ]
}
```
First texture is for the existing use case or if the image contains all 6 sides plus all mip levels.
Second texture is for separate 6 sides plus all mip levels.
Third texture is for separate 6 sides plus separate mip levels.

This would allow to still use .png and .jpeg but also fiel formats like .hdr, .exr, .dds or .ktx

PatrickRyanMS:
> Finally, providing the whole mip map chain and going down to 1x1 pixel should be sufficient. Please comapre with this tool:
> https://github.com/derkreature/IBLBaker

To help explain what how we landed on removing the last two mip levels, I put together the image below for reference. What it represents is on the top two models the full mip chain used for calculating roughness reflections while the third is calculating roughness values ignoring the last two in the chain. The bottom model is our ground truth from a ray tracer so we had a target to aim for.

First, we are aiming for a perceptual roughness ramp instead of a linear roughness ramp. You can see in the IBL Baker version on the top model that we lose clarity in the reflections very quickly. The bottom two models (Lys to mip 0.8 and Arnold) have a more gradual loss of clarity in the reflection image. For this reason, we use Lys because we can control the offset on the blurring of the mip chain to emulate a perceptual roughness falloff where IBL Baker does not give us that control. 

Second, we took into account the quality of the blur we are getting and determined that using the last two levels of the mip chain were causing visible artifacts. First, we are seeing more seams in the reflections at roughness 0.5 and up. There is also much more noise in the reflection. This is mostly due to sampling from fewer pixels to interpolate the reflected colors from. Especially from 0.5 and rougher values, removing the last two mip levels gives us more colors in each cube map face to sample from. You can also see at the roughest levels with a 1x1 mip level that the transitions in the tones are harsher, magnifying the pixilation in the reflections. 

We know that there is more involved with loading a texture and ignoring the last two mips, but determined that the quality boost in the resulting reflections was worth the investment in load. 

ibl_lys_iblbaker_groundtruth

Here is the glb we used if you want to run comparisons as well.
 
[ENV_Spheres.zip](https://github.com/KhronosGroup/glTF/files/2221020/ENV_Spheres.zip)

UX3D-nopper:
Maybe u can post the math for skipping the last mip maps?

sebavan:
Hello,

We are currently using a percentage of the Lod Level we should fetch for the texture but it is not ideal. 

```
        float reflectionLOD = getLodFromAlphaG(textureSize, alphaG);
        // Apply environment convolution scale/offset filter tuning parameters to the mipmap LOD selection
        reflectionLOD = reflectionLOD * scale;
```

where the scale is computed on cpu to for instance map 9 to 6 if we wanted to skip the three last LOD.

Hope it makes sense.

OmarShehata:
We're working on implementing IBL in Cesium, and this has been a great resource as a reference! I'd love to contribute here as we get further into our implementation, so let me know if there's anything I can help with.

There seems to be a couple open questions here. This is what we were thinking for our implementation:

* Allow lights to be defined per node, so you can define an environment map for the whole scene or for individual models.
  * I don't think this adds too much complexity to the implementation and it makes the extension a lot more flexible.
* Use KTX as a container.
  * This is the one I'm least sure about, but it sounds like it would simplify `specularImages` quite a bit as McNopper said (https://github.com/KhronosGroup/glTF/pull/1377#issuecomment-405794096).
  * On the other hand, I appreciate that one of the goals of the extension is just to provide one specific, consistent, and widely supported way of loading HDR images. Would this contradict that goal? Maybe the fallback of using LDR JPEG images makes up for that?
* I think it makes sense to specify an offset for the mip level since that sounds like it would simplify loading if partial chains aren't supported. 
* Regarding `EXT_lights_imageBased -> EXT_lights_environment`, I think it makes sense to keep it as the former, since, if I'm reading it right, using spherical harmonics is just a way to concisely express the image, so it doesn't change the end goal.

MiiBond:
Should this extension be under the Khronos or Vendor folder? Or should there be a Multi-Vendor folder added for EXT extensions?

donmccurdy:
I think for now it should go in `Vendor/`, not `Khronos/`. I wouldn't be opposed to creating a third folder at some point in the future but it seems early now.

MiiBond:
Thanks. Since support for this extension is now live in Babylon and shipping on Monday with Dimension 2.0, should this PR be merged?

extensions/2.0/Vendor/EXT_lights_image_based/README.md
```diff
@@ -0,0 +1,127 @@
...
+}
+```
+
+## Prefiltered Specular Radiance Cubemaps
+
+The cubemap used for specular radiance is defined along with its prefiltered mipmaps with each face of the cube defined as separate images. The various mip levels evenly map to roughness values from 0 to 1 in the PBR material and should be generated with a principled multi-scatter GGX normal distribution. The data in the maps represents illuminance in lux (lm/m^2).
+
+The entire mip chain of images should not be provided. Instead, the lowest-resolution mip should have sufficient size to represent the maximally-blurred radiance map (say, 16x16) corresponding to roughness=1. The `specularImageSize` value defines the largest dimension of mip 0 and, taken together with the number of defined mips, should give the loading runtime the information it needs to generate the remainder of the mip chain and sample the appropriate mip level in the shader.
+
+*(say something about tooling for generating these mips)*
```
emackey: Remaining TODO?

extensions/2.0/Vendor/EXT_lights_image_based/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+Cube faces are defined in the following order and must be authored with the Y-axis inverted:
```
emackey: What does this mean?  Inverted with respect to what, glTF?

Should we specify face orientations, as well as axis assignments?  For example, for whichever face is above the model, how is the image on that face oriented?

extensions/2.0/Vendor/EXT_lights_image_based/README.md
```diff
@@ -0,0 +1,127 @@
...
+http://graphics.stanford.edu/papers/envmap/
+
+Generating SH Coefficients
+http://graphics.stanford.edu/papers/envmap/prefilter.c
+
+Using SH Coefficients
+http://graphics.stanford.edu/papers/envmap/envmaprender.in
+
+*(this section needs more about how the coefficients are used as well as a brief blurb about what SH are)*
+*(also, we should probably just put some shader code in here for using the coefficients)*
```
emackey: More TODO

extensions/2.0/Vendor/EXT_lights_image_based/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+RGBD has several advantages over similar packing schemes like RGBM and RGBE.
+
+1. Simple to pack and unpack. Cheaper than RGBE.
+2. Provides better range than RGBM.
+3. Interpolates better than RGBE. Important if the data is kept packed during rendering and filtering is used. Note that the interpolation isn't perfect so visual artifacts may still appear. This is why unpacking to float at load-time is recommended.
+4. Packed PNG doesn't appear as garbled as a one packed with RGBE.
+
+### Future HDR Image Support
+Because this extension just references regular textures, any future support added by glTF for HDR formats will automatically be supported by this extension. e.g. CTTF formats.
```
emackey: This statement contrasts oddly with the "exactly one way..." statement in the Overview above.

extensions/2.0/Vendor/EXT_lights_image_based/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Adding Light Instances to Scenes
+
+Esch scene can have a single IBL light attached to it by defining the `extensions.EXT_lights_image_based` property and, within that, an index into the `lights` array using the `light` property.
```
emackey: *Each

OmarShehata:
Also @MiiBond it would help to provide a sample glTF with this extension to test with for people implementing it. 

For explaining how to generate the spherical harmonics, @timvanscherpenzeel's tool (https://github.com/timvanscherpenzeel/cubegen/tree/octagen) is a wrapper around Filament's cmgen (https://github.com/google/filament#running-the-native-samples) which generates pre-filtered maps as well as the spherical harmonic coefficients.

MiiBond:
Updated the spec to address @emackey 's comments.
@OmarShehata , thanks for the Filament links. I'll take a look. Does the cmgen tool work on macOS?
For your other questions posted above, yeah, I think we could probably extend this to support IBL per node as well as CTTF (KTX) support when they become available.
This extension is already supported in Babylon.js and in Adobe Dimension's web-export feature but I don't think there would be big backwards compatibility issues. @bghgary, what do you think?

bghgary:
We haven't tried to update an existing extension before. I'm not sure what the implications are. I think it might be okay to do it if the update is both forward and backwards compatible, like glTF 2.0 core.

extensions/2.0/Vendor/EXT_lights_image_based/README.md
```diff
@@ -0,0 +1,136 @@
...
+https://metashapes.com/blog/realtime-image-based-lighting-using-spherical-harmonics/
+https://cseweb.ucsd.edu/~ravir/papers/envmap/envmap.pdf
+http://graphics.stanford.edu/papers/envmap/
+
+Generating SH Coefficients
+http://graphics.stanford.edu/papers/envmap/prefilter.c
+
+Using SH Coefficients
+http://graphics.stanford.edu/papers/envmap/envmaprender.in
+
```
lexaknyazev: GitHub strips one-time line breaks, so these links aren't formatted properly. Consider wrapping them with titles like:

- [Realtime Image Based Lighting using Spherical Harmonics](https://metashapes.com/blog/realtime-image-based-lighting-using-spherical-harmonics/)
- [An Efficient Representation for Irradiance Environment Maps (PDF)](https://cseweb.ucsd.edu/~ravir/papers/envmap/envmap.pdf)
...

extensions/2.0/Vendor/EXT_lights_image_based/README.md
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }            
+]
+```
+
+### Image-Based Light Properties
+
+| Property | Description | Required |
+|:-----------------------|:------------------------------------------| :--------------------------|
+| `name` | Name of the light. | No, Default: `""` |
```
lexaknyazev: There shouldn't be any default value for `name`.

extensions/2.0/Vendor/EXT_lights_image_based/README.md
```diff
@@ -0,0 +1,136 @@
...
+| `rotation` | Quaternion that represents the rotation of the IBL environment. | No, Default: `[0.0, 0.0, 0.0, 1.0]` |
+| `intensity` | Brightness multiplier for environment. | No, Default: `1.0` |
+| `irradianceCoefficients` | Declares spherical harmonic coefficients for irradiance up to l=2. This is a 9x3 array. | :white_check_mark: Yes |
+| `specularImages` | Declares an array of the first N mips of the prefiltered cubemap. Each mip is, in turn, defined with an array of 6 images, one for each cube face. i.e. this is an Nx6 array. | :white_check_mark: Yes |
+| `specularImageSize` | The maximum dimension (in pixels) of the first specular mip. This is needed to determine, pre-load, the total number of mips needed. | :white_check_mark: Yes |
```
lexaknyazev: Assuming that images must be squares (btw, the spec should explicitly say that), what does "maximum dimension" mean for the first mip? Also, are npot images supported?

extensions/2.0/Vendor/EXT_lights_image_based/schema/glTF.EXT_lights_image_based.schema.json
```diff
@@ -0,0 +1,16 @@
...
+    "title": "EXT_lights_image_based glTF extension",
+    "type": "object",
+    "properties": {
+        "lights": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "$ref": "light.schema.json"
+            },
+            "default": []
```
lexaknyazev: To align with the core spec and `KHR_lights_punctual`, `"default": []` should be changed to `"minItems": 1`.
Also, `lights` should be a required field to disallow empty extension object:
```json
{
  "asset": {
    "version": "2.0",
    "extesnionsUsed": ["EXT_lights_image_based"]
  },
  "extensions": {
    "EXT_lights_image_based": {}
  }
}

File: extensions/2.0/Vendor/EXT_lights_image_based/schema/light.schema.json [hunk 10]
```diff
@@ -0,0 +1,68 @@
...
+            "minItems": 9,
+            "maxItems": 9
+        },
+        "specularImages": {
+            "description": "Declares an array of the first N mips of the prefiltered cubemap. Each mip is, in turn, defined with an array of 6 images, one for each cube face. i.e. this is an Nx6 array.",
+            "type": "array",
+            "items": {
+                "type": "array",
+                "items": {
+                    "type": "number",
```
lexaknyazev: `"number"` -> `"integer"`

extensions/2.0/Vendor/EXT_lights_image_based/schema/scene.EXT_lights_image_based.schema.json
```diff
@@ -0,0 +1,16 @@
...
+    "properties": {
+        "light": {
+            "allOf": [
+                {
+                    "$ref": "glTFid.schema.json"
+                }
+            ],
+            "description": "The id of the light referenced by this scene."
+        }
+    },
```
lexaknyazev: `light` should be a required field to disallow empty extension object
```json
{
  "asset": {
    "version": "2.0",
    "extesnionsUsed": ["EXT_lights_image_based"]
  },
  "scenes": [
    {
      "extesnions": {
        "EXT_lights_image_based": {}
      }
    }
  ]
}
```

MiiBond:
Excellent feedback, @lexaknyazev. Thank you!

extensions/2.0/Vendor/EXT_lights_image_based/schema/scene.EXT_lights_image_based.schema.json
MiiBond: Done

extensions/2.0/Vendor/EXT_lights_image_based/schema/light.schema.json
MiiBond: Done

extensions/2.0/Vendor/EXT_lights_image_based/README.md
MiiBond: I added the stipulation that they must be square and power-of-two.

OmarShehata:
@MiiBond I haven't tried compiling cmgen on a mac but they do have prebuilt releases for all OS's (https://github.com/google/filament/releases). 

Also CesiumJS does have support for this almost ready which I think would help provide a second reference implementation. We've just been waiting to get our hands on some glTF models with this extension to test with. I know Adobe Dimensions can export it but do you know of any freely available models for testing?

MiiBond:
Thanks, @OmarShehata .

I'll see if I can export some examples from Dimension today.
@pjcozzi, would we be able to merge this extension in today?

emackey:
@MiiBond for this version of the extension, can it be applied only to `scene` objects in glTF?

emackey:
@MiiBond Actually it looks like the schema can be applied to the root or the `scene`, is that correct?

lexaknyazev:
The root gets an array of light definitions (like in punctual lights), the scene gets a light instance.

emackey:
@lexaknyazev Ah, of course.  Thanks!

TimvanScherpenzeel:
> @MiiBond I haven't tried compiling cmgen on a mac but they do have prebuilt releases for all OS's (https://github.com/google/filament/releases).

Just a heads up if you are going to use the spherical harmonics generated by `cmgen`: the authors have recently made a change that included 1 / PI in the spherical harmonics. This means the SH reconstruction in the shaders shouldn't include the Lambert BRDF.

extensions/2.0/Vendor/EXT_lights_image_based/schema/light.schema.json
```diff
@@ -0,0 +1,68 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "light",
+    "type": "object",
+    "description": "An image-based lighting environment.",
+    "allOf" : [ { "$ref" : "../../../../specification/2.0/schema/glTFChildOfRootProperty.schema.json" } ],
```
emackey: I know this just got merged, but I need to ask one more question.  None of our other extensions put this whole relative path on here, they all just say `glTFChildOfRootProperty.schema.json` without the path.  Would it be OK to remove the path for consistency?

emackey: Further discussion migrated to #1480.

MiiBond: Sure. I can make that change and do another pull request. Thanks.

emackey: @MiiBond I needed it for VSCode, so I made the change in #1480.  Can you take a quick look at that?  Thanks!

MiiBond:
@OmarShehata I added a sample asset here:
https://github.com/KhronosGroup/glTF-Sample-Models/pull/201

vorg:
@MiiBond Is there more info on "principled multi-scatter GGX normal distribution" or any reference code on how to generate assets compatible with this extension? Tried to have a look at BabylonJS repo but can't find anything.

MiiBond:
Actually, we don't really have any consistency in the exact lighting model or roughness curve used for generation or display of this data. However, we're currently in talks about defining a Khronos-ratified version of this extension which will spell-out, more concretely, the equations to use when generating the IBL maps and displaying them in the runtime.
I expect that we will have a rough spec written in the next few weeks. Stay tuned.

Issue #1387, opened 2018-07-13
Using the "node.extras" property to serialize game entity component data


robertlong:
At the [Silicon Valley WebGL/WebVR meetup](https://www.meetup.com/Silicon-Valley-HTML5-WebGL-Meetup/events/249872707/) I presented how we are using glTF as a scene serialization format in Mozilla Hubs.

My slides for the talk are available [here](https://docs.google.com/presentation/d/1suM_42_rigkV39zfAHmOEIzb0zVAV0fjXpsvE-pBMtI/edit#slide=id.g25275a8168_0_386)

After the talk I received some feedback about how we are using the `node.extras` property to serialize our entity component data. @sbtron suggested that we should be using vendor extensions instead of extras for this purpose. Our team has been thinking about this for a while now and there's still some confusion on when we should be using extras or vendor extensions that I'd like to get cleared up.

The entity component data that we are serializing is essentially a 1 to 1 mapping to AFrame components. After loading our glTF scene with the THREE.GLTFLoader, we pull that component data off of the generated object3D's `userData` property. We then use that data to attach AFrame components to the associated nodes.

### Example:

This glTF node:

```
 "nodes": [
   {
     "name": "fan",
     "translation": [1, 3, 0],
     "extras": {
       "MOZ_shadow": {
         "castShadow": true,
         "receiveShadow": true
       },
       "MOZ_loopAnimation": {
         "clip": "Fan01"
       }
     }
   }
 ]
}
```

Becomes this AFrame entity:

```
<a-scene>
 <a-entity
   class="fan"
   position="1 3 0"
   shadow="castShadow: true; receiveShadow: true;"
   loop-animation="clip: Fan01"
 ></a-entity>
</a-scene>
```

In Saurabh's [presentation](https://onedrive.live.com/view.aspx?resid=52B1DDAE0BF63BA3!106131&ithint=file%2cpptx&app=PowerPoint&authkey=!AI4LZBCb8rEgt1U) (on slide 16) he states that if your extension will only be used by your application then you should use a vendor extension:

In the spec, `extras` are defined as "application-specific data".

I'd like to know if our entity component data falls into the bucket of vendor extensions or application-specific data.

We don't intend for these components to be used outside of Mozilla Hubs in their current form. Many components are specific to our own application, are still changing, and definitely aren't applicable to the majority of the glTF community at this time. We do however want to ship our entity component data with the glTF scene.

[...]

PR #1388, opened 2018-07-14
XTRA Unity MetaData Extension

MackeyK24:


donmccurdy:
I'd recommend reserving the prefix as well — is "XTRA" a brand or application name, or a reference to the "extra metadata" purpose of this extension? See https://github.com/KhronosGroup/glTF/blob/master/extensions/Prefixes.md. Is Babylon Toolkit associated with BabylonJS? I might suggest "BABYLON" if so, and "BABYLONTOOLKIT" if not.

MackeyK24:
Well Babylon is just one implementation... I also have PlayCanvas and might throw ThreeJS as well

I like XTRA .... can I not use that anyways ???

MackeyK24:
Ok... I changed my prefix to CVTOOLS

robertlong:
I'm also looking into storing entity component data in a glTF for use in Mozilla Hubs (see #1387).

I noticed this extension stores the data on the `extras` property without a namespace. Using this extension essentially means: treat `node.extras.metadata` as Unity metadata.

I think if this is going to be an extension then it should store that metadata in an extension object on the node. However, I am still confused by what `extras` is to be used for if not for this type of application specific metadata.

Perhaps we can work together on figuring out a best practice for storing this type of data? I'm excited to see other people looking into this!

MackeyK24:
From GLTF readme

Extensions vs. Extras

In addition to extensions, the extras object can also be used to extend glTF. This is completely separate from extensions.

All glTF object properties allow adding new properties to an extras object sub-property, e.g.,

{ "shaders": { "a-vertex-shader-id": { "name": "user-defined-name", "uri" : "vertex-shader.glsl", "type": 35633, "extras" : { "precompiled_binary" : "path_to_precompiled_shader" } } } }
This enables glTF models to contain application-specific properties without creating a full glTF extension. This may be preferred for niche use cases where an extension would not be widely adopted.

MackeyK24:
Basically I want extension so I can ID the glTF has unity metadata ... but because of detail I encode as EXTRA metadata

MackeyK24:
OK... The CVTOOLS prefix has been registered :)

pjcozzi:
@MackeyK24 is this extension implemented in Canvas Tools or anywhere else?

@robertlong is this still of interest to your use case?

extensions/2.0/Vendor/CVTOOLS_unity_metadata/README.md
```diff
@@ -0,0 +1,225 @@
...
+    {
+      "mesh": 2,
+      "scale": [
+        1.25,
+        1.25,
+        1.25
+      ],
+      "name": "Sphere",
+      "extras": {
+        "metadata": {
```
aaronfranke, line 51: glTF extensions are not supposed to specify what goes in `"extras"`.

aaronfranke (commented):
Given that this extension is abandoned, the "Canvas Tools" project this was implemented in is abandoned, the extension is malformed in several ways, and the extension is inherently unstructured, I believe this extension should be rejected and the PR should be closed.

PR #1392, opened 2018-07-17
Add Ogre_glTF plugin to the list

Ybalrid:
As discussed in #1058, heading for the "Game Engines" section has been changed to "Game and Rendering Engines"

PR #1394, opened 2018-07-17
Update table of contents in main README

emackey:
TOC updated using [markdown-toc extension for VSCode](https://marketplace.visualstudio.com/items?itemName=AlanWalk.markdown-toc).

Reorganized a small section that contained the VSCode editor.

PR #1396, opened 2018-07-17
Fix label in Contents table

Ybalrid:
After PR #1392 the `Game Engines` heading was changed to `Game and Rendering Engines`. But I forgot to update the `Contents` table. This commit and PR fixes this


emackey:
This is duplicate with #1394.  Thanks!

Issue #753, opened 2016-10-20
I want an general extension for spatialized audio, but have no business writing one


amwatson:
Hey all,

As per some discussion at glTF next and W3C, I want to put forward a schema for adding spatialized audio support for glTF.  Given the great work and interest in both audio and glTF coming out of today's session, I'd love to have something linkable for tomorrow.  With that in mind, I think we could be more productive with something that isn't a vendor pull request.  How do we feel about this as a start? 
## Overview

This extension adds general support for spatialized audio in glTF scenes.  Audio sources are specified at the node level, and are colocated with the location of their node/joint in worldspace.       

| Property | Type | Description | Default Value |
| :-: | :-: | :-: | :-: |
| `source` | string | ID of the audio source container | None |
| `volume` | `float` | The factor by which the audio source's volume is scaled for playback. An Audio source with volume 0.0 is considered to be disabled, and may be ignored | 0.0 |
| `reflection` | string | Audio source attenuation mode | false |
| `attenuation` | `boolean` | Denotes whether reflection should be enabled (true) or disabled (false) | 'NONE' |
| `diameter` | `float` | Virtual diameter of the spherical audio source in meters. Larger sizes provide more envelopment for volumetric sounds | 0.0 |
| `maxSpeed` | `float` | Max speed at which the audio source can travel, in meters/second. Unanticipated jumps in motion may produce artifacts during positional interpolation | 0.0 |
| `startOffset` | `float` | The number of seconds between the time at which audio playback is requested, and the start time of the given audio source. If 'loop' is set to true, the audio source will restart playback 'startOffset' seconds after the end time of the previous iteration | 0.0 |
| `loop` | `boolean` | When true, the audio source is played continuously in a loop, restarting 'startOffset' seconds after the previous iteration has completed | false |
### glTF (gltf.XXX_audio.schema.json)

```
{
    "$schema" : "http://json-schema.org/draft-03/schema",
    "title" : "Spatialized Audio extension",
    "type" : "object",
    "properties" : {
        "audio" : {
            "type" : "object",
            "description" : "A dictionary object of audio sources.",
            "properties" : {
            },
            "additionalProperties" : {
                "$ref" : "audio.schema.json"
            },
            "default" : {}
        }

    },
    "additionalProperties" : false
}
```
### audio (audio.schema.json)

```
{
    "$schema" : "http://json-schema.org/draft-03/schema",
    "title" : "audio",
    "type" : "object",
    "description" : "Data used to create an audio source",
[...]

PR #1400, opened 2018-07-18
MSFT_audio_emitter extension

najadojo:
We at Microsoft have been developing this extension for use in various products and scenarios. We are sharing to document software we will be shipping and to get feedback on the extension design. Its possible that this extension maybe useful for other vendors as evidenced by #753.


lexaknyazev:
May I nitpick about audio data formats?
- Why it's only `audio/wav`?
- According to IANA [registry](https://www.iana.org/assignments/media-types/media-types.xhtml#audio),  `audio/wav` is not a valid registered MIME type (despite being widely used). Other seen options are `audio/vnd.wave` and `audio/x-wav`.
- Assuming that WAV means RIFF container, which of many possible [WAV encodings](https://tools.ietf.org/html/rfc2361) are supposed to be supported?

najadojo:
We picked wav because its commonly available and trivial to decode almost anywhere. Continuing with the runtime friendliness of glTF and the desire to use linear PCM decoded audio data at runtime it made reasonable sense. Our least capable client is able to decode wav with WAVE_FORMAT_PCM, and WAVE_FORMAT_IEEE_FLOAT. I'll update the mime type value.

najadojo:
I created a [demo video](https://najadojo.github.io/MSFT_audio_emitter/) that shows how this extension can be used in Babylon.js.

pjcozzi:
@amwatson and @tparisi had an interest in spatialized audio a couple of years ago, #753.  Perhaps they want to chime in here?

pjcozzi:
@najadojo is this spec/schema still up-to-date with the Babylon implementation?

@sbtron @lexaknyazev OK to merge from your perspective?  Perhaps the samples belong in the [glTF-Sample-Models repo](https://github.com/KhronosGroup/glTF-Sample-Models/)?

pjcozzi:
@najadojo very nice!

@sbtron @lexaknyazev please merge when you are ready.

faridskyman:
Video demo was great, having a 3d object with animation that also plays sounds is a terrific idea. However this thread is dead, Khronos people? what happen? u guys dont like this? then do u have a better idea? or no action?
can we have this or u guys comeout with a standard?
else its strange, we get user to upload audio and we try to sync with with the animation?
dont seem a good idea..
hope khoronos will reconsider.. its 2022, 4 years has passed...

extensions/2.0/Vendor/MSFT_audio_emitter/README.md
```diff
@@ -0,0 +1,356 @@
...
+The cone will be symmetrical and face the forward direction (Z+). The outerAngle value must be greater than or equal to the innerAngle value. If neither innerAngle or outerAngle are defined the emitter will behave as an omnidirectional source.
+
+* **Type**: `number`
+* **Required**: No, default: `3.141592653589793`
+* **Minimum**: `>= 0`
+* **Maximum**: `<= 3.141592653589793`
+
+#### emitter.outerAngle
+
+The cone will be symmetrical and face the forward direction (Z+). The outerAngle value must be greater than or equal to the innerAngle value. If neither innerAngle or outerAngle are defined the emitter will behave as an omnidirectional source.
```
aaronfranke, line 164: This is inconsistent with KHR_lights_punctual which defines spotlights as facing -Z. Is it expected that conical audio should be pointing the opposite direction as conical lights?

PR #1405, opened 2018-07-24
Add the AGI prefix.

emackey:
This is for an upcoming vendor extension for AGI's desktop software, STK (Systems ToolKit).

Also, clean up wording about applying labels to issues.  Only the maintainers can adjust issue labels.


pjcozzi:
Awesome, thanks!

PR #1407, opened 2018-07-27
Draco animation extension draft

FrankGalligan:
Adds an extension for compressing glTF animations using Draco technology.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -19,6 +19,7 @@ _Draft Khronos extensions are not ratified yet._
 * [MSFT_texture_dds](2.0/Vendor/MSFT_texture_dds/README.md)
 * [MSFT_packing_normalRoughnessMetallic](2.0/Vendor/MSFT_packing_normalRoughnessMetallic/README.md)
 * [MSFT_packing_occlusionRoughnessMetallic](2.0/Vendor/MSFT_packing_occlusionRoughnessMetallic/README.md)
+* [DRACO_animation_compression](2.0/Vendor/DRACO_animation_compression/README.md)
 
 
 ## Extensions for glTF 1.0


extensions/README.md
```diff
@@ -19,6 +19,7 @@ _Draft Khronos extensions are not ratified yet._
 * [MSFT_texture_dds](2.0/Vendor/MSFT_texture_dds/README.md)
 * [MSFT_packing_normalRoughnessMetallic](2.0/Vendor/MSFT_packing_normalRoughnessMetallic/README.md)
 * [MSFT_packing_occlusionRoughnessMetallic](2.0/Vendor/MSFT_packing_occlusionRoughnessMetallic/README.md)
+* [DRACO_animation_compression](2.0/Vendor/DRACO_animation_compression/README.md)
```
pjcozzi: Should this use the `GOOGLE` prefix?  Or do you want to use a `DRACO` prefix?  We have been allowing prefixes for both companies and projects.

extensions/2.0/Vendor/DRACO_animation_compression/README.md
```diff
@@ -0,0 +1,236 @@
...
+"extensionsRequired" : [
+    "Draco_animation_compression"
+]
+```
+
+If the root object of a glTF asset contains `extensions` property and the `extensions` property defines its `Draco_animation_compression` property, then the Draco animation compression must be used. If there are any compressed animations then the `Draco_animation_compression` must define the `compressedAnimations` property. The `compressedAnimations` property consists of an array of compressed animation objects. Each element in `compressedAnimations` must have the following properties:
+
+#### input
+The `input` property defines a single `input` of a compressed animation. The
+`input` corresponds to a `samplers` element's `input`.
```
donmccurdy: I'm not sure I understand the meaning of "corresponds to" here — what is this an index of? An accessor index, that is also referenced by a sampler's input? Or an index of the sampler whose input is overridden?

extensions/2.0/Vendor/DRACO_animation_compression/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Conformance
+
+Below is the recommended process when a loader encounters a glTF asset with the Draco animation extension set:
+
+* If `Draco_animation_compression` is in `extensionsRequired` and the loader does not support the Draco animation extension, then the loader must fail loading the asset.
+* If the loader does not support the Draco animation extension and `Draco_animation_compression` is not in `extensionsRequired`, then load the glTF asset ignoring `Draco_animation_compression` in `extensions`.
+* If the loader does support the Draco animation extension, but will not process `Draco_animation_compression`, then the loader must load the glTF asset ignoring `Draco_animation_compression` in `extensions`.
+* If the loader does support the Draco animation extension, and will process `Draco_animation_compression` then:
+    * The loader must process `Draco_animation_compression` first. The loader must get the data from `Draco_animation_compression`'s `bufferView` property and decompress the data using a Draco animation decoder to a Draco animation.
```
donmccurdy: > decompress the data using a Draco animation decoder to a Draco animation.

Is there a 1:1 relationship between a `Draco_animation_compression` bufferView and an animation? Or could two glTF animations share a single bufferView but reference a different subset of available attributeIds?

extensions/2.0/Vendor/DRACO_animation_compression/README.md
```diff
@@ -0,0 +1,236 @@
...
+            ]
+        },
+    ],
+    "accessors" : [
+        // Accessors of input/output of animations that use the compression extension will omit "bufferView" property.
+        {
+            "componentType" : 5123,
+            "count" : 100,
+            "type" : "SCALAR"
+        },
```
donmccurdy: Consider this a strawman idea, but could this extension also be structured:

```
{
    "extensionsRequired" : [
        "Draco_animation_compression"
    ],
    "extensionsUsed" : [
        "Draco_animation_compression"
    ],
    "extensions" : {
        "Draco_animation_compression" : {
            "compressedBufferViews" : [ 0, 1 ]
        }
    },
    "nodes" : [
      ...
    ],
    "animations" : [
      ... unmodified glTF animations ...
    ],
    "accessors" : [{
       "componentType" : 5123,
       "count" : 100,
       "type" : "SCALAR",
       "extensions" : {
          "Draco_animation_compression" : {
            "bufferView": 0,
            "attributeId": 1
           }
        }
     }],
    ...
}
```

extensions/2.0/Vendor/DRACO_animation_compression/README.md
```diff
@@ -0,0 +1,236 @@
...
+        {
+            "componentType" : 5123,
+            "count" : 100,
+            "type" : "VEC3"
+        },
+        {
+            "componentType" : 5123,
+            "count" : 100,
+            "type" : "VEC4"
```
donmccurdy: Do all properties of the original accessor (componentType, count, ...) remain valid for the compressed data?

extensions/2.0/Vendor/DRACO_animation_compression/README.md
FrankGalligan: So this is an accessor index, that is also referenced by the sampler's input.

I guess we don't really need to say anything about the sampler's input here.

How about:
The `input` property defines a single `input` of a compressed animation. The
`input` references the accessor to use for the input data.

extensions/2.0/Vendor/DRACO_animation_compression/README.md
FrankGalligan: Yes

extensions/2.0/Vendor/DRACO_animation_compression/README.md
FrankGalligan: The later (for "outputs"). Actually what we are currently doing is we parse the source gltf file and we gather all outputs that correspond to the same input. Then we compress the the one input with all of it's output in one animation compression. That is why we have an array of "attributeId" and an array of "output".

For each unique input we create one animation compression.

extensions/README.md
FrankGalligan: I'm fine with the DRACO prefix. But if people would rather us change to GOOGLE, that would be fine with me as well.

extensions/2.0/Vendor/DRACO_animation_compression/README.md
FrankGalligan: So I think this should work. The one issue is that using this format the gltf file will usually be bigger. I ran through some test files and I saw anywhere from a 1-27% increase in size (unzipped). CessiumMan and Monster saw increases of 9% in gltf size.  About a 2.5 and 4% total size increase with mesh and animation compression turned on.

donmccurdy: Oh yikes, that's 9% of total file size occupied by JSON in the `accessors` array with this layout? Difference might disappear when served with GZIP, but still not ideal. Or were you grouping accessors into compressed buffer views differently with this layout, e.g. one animation compression per accessor?

FrankGalligan: I wasn't grouping them differently, but I'm assumign every compressed animation accessor needs to have a bufferView and and attributeId (or I guess in case of input no attributeID, which should bring the size down a little)

Here is a link to what I think you are proposing:
https://drive.google.com/open?id=14_vRPj1Xd6X9TPHcPQkBCwbWHvzuKRah

 Ignore the actual values as I just put place holders in to get the size.

extensions/2.0/Vendor/DRACO_animation_compression/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+If the root object of a glTF asset contains `extensions` property and the `extensions` property defines its `Draco_animation_compression` property, then the Draco animation compression must be used. If there are any compressed animations then the `Draco_animation_compression` must define the `compressedAnimations` property. The `compressedAnimations` property consists of an array of compressed animation objects. Each element in `compressedAnimations` must have the following properties:
+
+#### input
+The `input` property defines a single `input` of a compressed animation. The
+`input` corresponds to a `samplers` element's `input`.
+
+### outputs
+The `outputs` property defines an array of `output` of sharing the same `input`.
+Each element of `outputs` corresponds to a `samplers` element's `output`.
```
donmccurdy: Are there any constraints on the types of the input/output accessors? Example code shows only Float32, but I assume the type would actually be read from the accessor or the bitstream?

extensions/2.0/Vendor/DRACO_animation_compression/README.md
donmccurdy: That seems clear enough then, yes. I think we should also specify that the accessor cannot be referenced by anything other than a sampler input/output.

Can a compressed `input` be used in a sampler alongside an uncompressed `output`, and vice versa? I would lean toward disallowing both.

extensions/2.0/Vendor/DRACO_animation_compression/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+### JSON Schema
+
+For full details on the `Draco_animation_compression` extension properties, see the schema:
+
+* [extension property (the entirety of this specification is normative and included in Scope)](schema/node.DRACO_animation_compression.schema.json)
+
+## Conformance
+
+Below is the recommended process when a loader encounters a glTF asset with the Draco animation extension set:
```
donmccurdy: The risk of calling this a "recommended process" is that it makes the actual requirements for exporters and clients implicit... can these be phrased as normative statements on the content and interpretation of the extension data, where possible?

extensions/2.0/Vendor/DRACO_animation_compression/README.md
donmccurdy: I think this is better described in terms of the data, rather than the order of processing. A loader might process the accessors to get componentType, type, count, etc? We should also mention that the accessor's `byteOffset` and `bufferView` may be undefined.

extensions/2.0/Vendor/DRACO_animation_compression/README.md
```diff
@@ -0,0 +1,236 @@
...
+* If the loader does support the Draco animation extension, but will not process `Draco_animation_compression`, then the loader must load the glTF asset ignoring `Draco_animation_compression` in `extensions`.
+* If the loader does support the Draco animation extension, and will process `Draco_animation_compression` then:
+    * The loader must process `Draco_animation_compression` first. The loader must get the data from `Draco_animation_compression`'s `bufferView` property and decompress the data using a Draco animation decoder to a Draco animation.
+    * Then the loader must process `samplers` properties of the `animations`. When loading each element of `samplers`, you must ignore the `bufferView` and go to the previously decoded Draco animation from `compressedAnimations` to get the data of `input` and `outputs`. A loader must use the decompressed animation data to fill the `accessors` (e.g. [ThreeJS (non-normative)](https://github.com/FrankGalligan/three.js/blob/add_draco_animation_compression/examples/js/loaders/DRACOLoader.js)).
```
donmccurdy: To clarify — when "when loading each element of `samplers`" should probably say each element of `samplers` that references an input/output pair with corresponding compressed data. Assuming that an asset could contain a mix of compressed and uncompressed animation — or should that be disallowed?

Presumably input and output must always come from the same compressed animation?

extensions/README.md
zellski: Given `KHR_draco_mesh_compression`, would `DRACO_animation_compression` ultimately then be `KHR_draco_animation_compression`? If so, it makes more sense to go `GOOGLE_draco_animation_extension` now, IMO...

Ziriax:
@pjcozzi I'm interested. People using our Maya2glTF exporter will need small file sizes (not a current requirement for our company yet). Since animation is sampled, it becomes very large. So animation compression is very welcome!

TimvanScherpenzeel:
I am definitely still interested and it would be very useful to me! 👍

I don't see anything that really stands out to me as wrong in the spec description, I mostly have some questions regarding the [implementation](https://github.com/google/draco/tree/master/javascript/npm/draco_animation) itself. Should I ask these over at the Draco repository?

pjcozzi:
@Ziriax @TimvanScherpenzeel thanks for the prompt response!

@TimvanScherpenzeel if you have questions specific to the Draco bitstream and how the compression works - not the glTF JSON extension and integration points - I think it would be most appropriate to ask on the Draco GitHub repo as you suggested.

extensions/2.0/Vendor/DRACO_animation_compression/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension defines a schema to use [Draco compression library (non-normative)](https://github.com/google/draco) to compress `animations` in glTF format. This allows glTF to support streaming compressed animation data instead of the raw data.
```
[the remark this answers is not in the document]
zellski: I've been curious about this, too: is the encoded bitstream streaming-compatible? If so, does that depend on the sequencing of the original uncompressed data? Pardon if this is a FAQ and/or obvious. I've found it a little tricky to locate authoritative documentation on the animation encoder.

zellski:
@pjcozzi this is one of those projects that will almost certainly be critically important at some point in my work, but it's impossible to say when a product will mature to where that happens... so for now, my interest is tentative, yet strong.

svensht2:
I'm following up here from feedback from @donmccurdy in #17878.
To the best of my knowledge, gltf doesn't provide a file size efficient way to represent morph target/per vertex animation.
I have interest in having morph targets draco compressed. I'm dealing with complex non-rigid motion where morph targets are a suitable solution. From my experiments, compressing morph target deltas using the edge breaker connectivity provides very efficient compression. 
My understanding is that this extension deals with compressing node animation, which doesn't correspond to a mesh topology. As a result, the underlying compression in draco is point cloud based compression.

extensions/2.0/Vendor/DRACO_animation_compression/README.md
```diff
@@ -0,0 +1,236 @@
+# Draco_animation_compression
```
aaronfranke, line 1: Extension prefixes should be all caps:

```suggestion
# DRACO_animation_compression
```

PR #1412, opened 2018-07-31
Appendix C wording

andreasplesch:


pjcozzi:
Thanks @andreasplesch!

@bghgary what do you think?

specification/2.0/README.md
```diff
@@ -3932,32 +3932,34 @@ Implementation of diffuse from [Lambert's Photometria](https://archive.org/detai
 
 Animations in glTF support spline interpolation with a cubic spline.
 
-The keyframes of a cubic spline in glTF have input and output values where each input value corresponds to three output values: in-tangent, spline vertex, and out-tangent.
+The keyframes of a cubic spline in glTF have input and output values where each input value corresponds to three output values of the same type: in-tangent, data tuple, and out-tangent.
```
bghgary: I'm not sure about using the term "data tuple" here. If the term spline vertex is confused with mesh vertex, then I would suggest spline point instead which looks like you use below.

specification/2.0/README.md
```diff
@@ -3932,32 +3932,34 @@ Implementation of diffuse from [Lambert's Photometria](https://archive.org/detai
...
 Given a set of keyframes
 
-&nbsp;&nbsp;&nbsp;&nbsp;Input *t*<sub>*k*</sub> and output in-tangent ***a***<sub>k</sub>, vertex ***v***<sub>*k*</sub>, and out-tangent ***b***<sub>k</sub> for *k* = 1,...,*n*
+&nbsp;&nbsp;&nbsp;&nbsp;Input *t*<sub>*k*</sub> with Output in-tangent ***a***<sub>k</sub>, tuple ***v***<sub>*k*</sub>, and out-tangent ***b***<sub>k</sub> for *k* = 1,...,*n*
 
-A spline segment between two keyframes is represented in a cubic Hermite spline form
+a spline segment between two keyframes is represented in a cubic Hermite spline form:
 
-&nbsp;&nbsp;&nbsp;&nbsp;***p***(*t*) = (2*t*<sup>3</sup> - 3*t*<sup>2</sup> + 1)***p***<sub>0</sub> + (*t<sup>3</sup>* - 2*t*<sup>2</sup> + *t*)***m***<sub>0</sub> + (-2*t*<sup>3</sup> + 3*t*<sup>2</sup>)***p***<sub>1</sub> + (*t*<sup>3</sup> - *t*<sup>2</sup>)***m***<sub>1</sub>
+&nbsp;&nbsp;&nbsp;&nbsp;***p***(*t*) = (2*t*<sup>3</sup> - 3*t*<sup>2</sup> + 1) ***p***<sub>0</sub> + (*t<sup>3</sup>* - 2*t*<sup>2</sup> + *t*) ***m***<sub>0</sub> + (-2*t*<sup>3</sup> + 3*t*<sup>2</sup>) ***p***<sub>1</sub> + (*t*<sup>3</sup> - *t*<sup>2</sup>) ***m***<sub>1</sub>
```
bghgary: Not that big a deal if you really want to add spaces, but see here https://en.wikipedia.org/wiki/Cubic_Hermite_spline#Interpolation_on_a_single_interval

specification/2.0/README.md
```diff
@@ -3932,32 +3932,34 @@ Implementation of diffuse from [Lambert's Photometria](https://archive.org/detai
...
 
 &nbsp;&nbsp;&nbsp;&nbsp;*t* = (*t*<sub>*current*</sub> - *t*<sub>*k*</sub>) / (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>)  
 &nbsp;&nbsp;&nbsp;&nbsp;***p***<sub>0</sub> = ***v***<sub>*k*</sub>  
-&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>0</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>)***b***<sub>k</sub>  
+&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>0</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>) ***b***<sub>k</sub>  
 &nbsp;&nbsp;&nbsp;&nbsp;***p***<sub>1</sub> = ***v***<sub>*k*+1</sub>  
-&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>1</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>)***a***<sub>k+1</sub>  
+&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>1</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>) ***a***<sub>k+1</sub>  
+
+The scalar-tuple multplications are per tuple component.
```
bghgary: I'm not sure what this statement means. Can you explain?

andreasplesch: Multiplication by a scalar is done per component. This is true for all data types without explicitly stating it here but I found it helpful to see that confirmed in implementation code. So I think it would be helpful to be explicit here as well.

specification/2.0/README.md
andreasplesch: I was looking for a generic term which applies scalars, 3d vectors, and quaternions, I only used point below to explain the p in the formula. Perhaps data point ? In the future other types may be animated as well.

specification/2.0/README.md
```diff
@@ -3932,32 +3932,34 @@ Implementation of diffuse from [Lambert's Photometria](https://archive.org/detai
...
+&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>1</sub> is the scaled ending tangent at *t* = 1  
+&nbsp;&nbsp;&nbsp;&nbsp;***p***(*t*) is the resulting tuple value  
 
-Where at input offset *t*<sub>*current*</sub> with keyframe index *k*
+and where at input offset *t*<sub>*current*</sub> with keyframe index *k*
 
 &nbsp;&nbsp;&nbsp;&nbsp;*t* = (*t*<sub>*current*</sub> - *t*<sub>*k*</sub>) / (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>)  
 &nbsp;&nbsp;&nbsp;&nbsp;***p***<sub>0</sub> = ***v***<sub>*k*</sub>  
-&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>0</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>)***b***<sub>k</sub>  
+&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>0</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>) ***b***<sub>k</sub>
```
bghgary: Probably should revert these also.

specification/2.0/README.md
```diff
@@ -3932,32 +3932,34 @@ Implementation of diffuse from [Lambert's Photometria](https://archive.org/detai
...
 
 &nbsp;&nbsp;&nbsp;&nbsp;*t* = (*t*<sub>*current*</sub> - *t*<sub>*k*</sub>) / (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>)  
 &nbsp;&nbsp;&nbsp;&nbsp;***p***<sub>0</sub> = ***v***<sub>*k*</sub>  
-&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>0</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>)***b***<sub>k</sub>  
+&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>0</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>) ***b***<sub>k</sub>  
 &nbsp;&nbsp;&nbsp;&nbsp;***p***<sub>1</sub> = ***v***<sub>*k*+1</sub>  
-&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>1</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>)***a***<sub>k+1</sub>  
+&nbsp;&nbsp;&nbsp;&nbsp;***m***<sub>1</sub> = (*t*<sub>*k*+1</sub> - *t*<sub>*k*</sub>) ***a***<sub>k+1</sub>  
+
+The scalar-point multplications are per point component.
```
bghgary: > multplications 

spelling

specification/2.0/README.md
bghgary: I personally would find this confusing since this is standard in math equations, but if you feel strongly about it, I'm okay with it.

andreasplesch: I actually would find it also helpful to state that bold letters represent vectors (or scalars) with that being a kind of standard as well. I just prefer being explicit but feel there is a line somewhere.

PR #1416, opened 2018-08-03
The AGI_articulations and AGI_stk_metadata extensions

emackey:
This extension adds several categories of extra metadata to the glTF model: Articulations, Stages, Pointing Vectors, Attach Points, and Solar Panel Groups.  These metadata are designed to be a 1:1 match with existing features of the product STK (Systems Tool Kit), produced by Analytical Graphics, Inc.

Generally this is an optional extension, and we do expect most other glTF implementations to ignore this.  However we certainly welcome the thought that other vendors might find some usefulness in building compatibility with some of this, wherever it makes sense to do so.

/cc @abwood @gbeatty  @DormantVision


emackey:
While this branch is open, here's the link to the branch's [main README.md](https://github.com/KhronosGroup/glTF/blob/agi-metadata/extensions/2.0/Vendor/AGI_metadata/README.md).

pjcozzi:
The name `AGI_metadata` may make this look like a vendor proposed general purpose metadata extension.  Perhaps `AGI_STK_metadata` or another more specific name is appropriate?

abwood:
Would it be wrong or unconventional to include your json schema files with this PR?

abwood:
It's a fair point @pjcozzi.  A wise man once told me that less is more.  I personally lean towards `AGI_metadata`, as it is bundling extra data we need on models for use in our core products.  With the inclusion of Solar Panel Groups, it's safe to say this extension is never intended to leave the Vendor category, so the AGI prefix may be enough.

emackey:
@pjcozzi @abwood I do think vendor-specific is strong enough.  I could imagine another vendor finding a use for pointing vectors, for example, and adding compatibility with that if they desired to.

pjcozzi:
Take a look at the vendor extensions for OpenGL - https://www.khronos.org/registry/OpenGL/index_gl.php

If we drop the vendor prefix, they still read pretty generic.

Do as you wish; I don't have the bandwidth to participate.  I am just trying to prevent the community from assuming this is a general metadata extension created by AGI, which I believe they will given the current name.

emackey:
After some discussion with @abwood and @gbeatty, the current plan of action is to break this extension into two separate extensions:

- `AGI_articulations` - will contain articulations and the associated pointing vectors, and will contain attach points.  All of these should be useful outside of STK, for example Blender has [rotation constraint limits](https://docs.blender.org/manual/en/dev/rigging/constraints/transform/limit_rotation.html) that are quite similar and could theoretically be exported to articulations.

- `AGI_stk_metadata` - will contain STK-specific things like Solar Panel Groups and possibly obscuration tool settings.

emackey:
There are now two extensions:

- [`AGI_articulations`](https://github.com/KhronosGroup/glTF/tree/agi-metadata/extensions/2.0/Vendor/AGI_articulations) - for others to consider

- [`AGI_stk_metadata`](https://github.com/KhronosGroup/glTF/tree/agi-metadata/extensions/2.0/Vendor/AGI_stk_metadata) - for STK use only

abwood:
Does `AGI_stk_metadata` also need properties to opt out of obscuration?  Otherwise, content looks really good to me.

emackey:
Added the noObscuration indicator.  I think this is ready.

pjcozzi:
@emackey perhaps wait for #1428 (CC @donmccurdy) then update that README.md to link to these new vendor extensions and merge?

emackey:
I am hoping to get this merged this week.  I intend to release an update to glTF-VSCode that includes schema validation using these files.  (KHR_* extension schema validation was added recently, as a precursor to this).

emackey:
This is ready.

PR #1423, opened 2018-08-13
Reorganising the Readme

ryanoshea-arm:
This pull request is to start a discussion about the best way of organising the README going forward.

The current style of referencing all known tools from the main README file is fast becoming too large to be useful, the current alternatives are either creating an ArtistsReadme and DevelopersReadme to contain all related tools (as is done in this PR) or to create a ToolsReadme that contains all the tools.

Which method do people prefer? Or does someone have an entirely different suggestion for organising it?


emackey:
This looks like a good breakup by user role.  Some tools are useful to both developers and artists, and so are listed in both files, which is great.  But the main README contains yet a third list of tools, separate from the developer tools and the artist tools, is that needed?

ryanoshea-arm:
I am thinking that it might be a good idea for the main README to have links to a few tools such as a viewer, converter, exporter, validator and perhaps some applications that use glTF? The main README is definitely still overburdened in its current state though.

donmccurdy:
Thanks for getting this going! I like the idea of having sections for artists and developers, although I'd been envisioning those as much shorter "best of" sections in the main readme, with the full list in a single place elsewhere. In this PR, certain long lists like *WebGL Engines* must be duplicated across both the artist and developer pages. 

For example:

- PAGE: README.md
  - SECTION: Artists (short "best of")
  - SECTION: Developers (short "best of")
- PAGE: TOOLS.md
  - ...Exhaustive list of tools...

Thoughts? In that case I think we could list even fewer tools on the main readme than we do here.

ryanoshea-arm:
I like the idea of a "best of" for both Artists and Developers in the main README as it gives an even more streamlined README.md which is a great thing although deciding what tools we would want to include in these "best of" sections could be difficult though.

We also need to make sure if we go that route that the TOOLS.md is very well formatted so that it isn't just an endless indecipherable list of tools.

pjcozzi:
@ryanoshea-arm is this ready to merge?

pjcozzi:
@ryanoshea-arm any update here? 😄

ryanoshea-arm:
Nothing yet. Sorry!

pjcozzi:
@ryanoshea-arm is this ready for another review?

ryanoshea-arm:
Yep, any feedback would be great!

donmccurdy:
Project list has since been consolidated into https://github.com/KhronosGroup/glTF-Project-Explorer – I'll close this PR but suggestions are welcome on the project explorer repo!

PR #1426, opened 2018-08-19
C sharp typo in contents

xfischer:
The sharp sign was not present.


emackey:
Thanks.  Seems likely this was from a bug in a TOC plugin I used recently.

PR #1429, opened 2018-08-23
NormalTexture value should not be normalized

06wj:
In the section normalTextureInfo.scale, the formula is
```
scaledNormal =  normalize((normalize(<sampled normal texture value>) * 2.0 - 1.0) * vec3(<normal scale>, <normal scale>, 1.0))
```

But it's not correct. 
e.g.: 
1. Assume that the original normal value is  (0, -1, 0)
1. In the texture it will be (0.5, 0, 0.5)
1. After the normalize it will be (0.7071, 0, 0.7071)
1. Finally it will be (0.4142, -1, 0.4142)，obviously, this is wrong.

Should the code be 
```
normalize((normalize(<sampled normal texture value> * 2.0 - 1.0)) * vec3(<normal scale>, <normal scale>, 1.0))
```
or just remove the normalize
``` 
normalize((<sampled normal texture value> * 2.0 - 1.0) * vec3(<normal scale>, <normal scale>, 1.0))
```

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -2937,7 +2937,7 @@ This integer value is used to construct a string in the format TEXCOORD_<set ind
 
 #### normalTextureInfo.scale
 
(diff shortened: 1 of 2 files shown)


pjcozzi:
Thanks @06wj!  @bghgary is this one you could review?

PR #1437, opened 2018-08-30
Fix broken anchor link

justinclift:
Just a trivial thing, but noticed it while looking through the README. :smile:


emackey:
Thanks!

PR #1451, opened 2018-09-21
Add implementation note about key/value extras.

### LINKED ISSUE #1120: Suggestion for "extras"
For `extension` glTF has `object` type in `additionalProperties`, which is representation of a dictionary. 

`extras` does not have any specifications. I suggest add something similar to code below:

```
"additionalProperties": {
    "anyOf": [
        {"type": "array"},
        {"type": "string"},
        {"type": "object"},
        {"type": "number"},
        {"type": "integer"}
    ]
}
```

Motivation: Usually not described properties are skipped.

donmccurdy:
Fixes #1120.

/cc @lexaknyazev

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -2292,7 +2292,9 @@ Additional properties are allowed.
 
 Application-specific data.
 
-
+> **Implementation Note:** Although extras may have any type, it is common for applications to
+store and access custom data as key/value pairs. As best practice, extras should be an Object
+rather than a primitive value for best portability.
 
 ---------------------------------------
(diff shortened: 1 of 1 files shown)


lexaknyazev:
> As best practice, extras should be an Object rather than a primitive value for best portability.

Should `null` be considered as "primitive value"? Like:
```json
{
  "asset": {
    "version": "2.0",
    "extras": null
  }
}

emackey:
Yeah, I think it's fine.  Since the note explains the intention of key/value pairs, and the schema can't be changed to disallow null, I think it follows that `null` can be disregarded as a lack of keys and values.

PR #1453, opened 2018-09-25
Add explicit type descriptor to integer enums

UX3D-hanselmann:
In [glTF-Blender-IO](https://github.com/KhronosGroup/glTF-Blender-IO) we are using [quicktype](https://github.com/quicktype/quicktype) to automatically generate python classes from the glTF JSON schema.

This commit adds explicit integer type specifiers to some enums, which prevents quicktype from wrongly deducing a type of float for these enum values ([refer to this warning in the JSON schema docs](https://json-schema.org/understanding-json-schema/reference/numeric.html#integer) for details).

### DIFF
--- a/specification/2.0/schema/accessor.schema.json
+++ b/specification/2.0/schema/accessor.schema.json
@@ -25,27 +25,33 @@
             "anyOf": [
                 {
                     "enum": [ 5120 ],
-                    "description": "BYTE"
+                    "description": "BYTE",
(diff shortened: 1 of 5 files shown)


pjcozzi:
Thanks @UX3D-hanselmann!  @lexaknyazev could you take a quick look?

pjcozzi:
Thanks @UX3D-hanselmann @lexaknyazev!

Issue #1226, opened 2018-01-27
'texture.source' property is not required, but no default value is specified?


tomolt:
From the spec:

| |Type | Description | Required |
| --| -- | -- | -- |
source | integer | The index of the image used by this texture. | No

I can't seem to find any information on how a complying GLTF2 loader should actually behave if texture.source is not defined.
Usually, the spec states a default value to be used instead, as in texture.sampler:

| |Type | Description | Required |
| --| -- | -- | -- |
sampler	| integer |	The index of the sampler used by this texture. When undefined, a sampler with repeat wrapping and auto filtering should be used. |	No

However, I can't find any information on a default value for texture.source, neither in the properties reference nor in the *Texture Data* section.

The only other options I might be missing are that texture.source is not needed when some
other properties are set (but I can't find anything on that either, and it would be kind of absurd to exist anyway) or that the texture is supposed to be filled with zeroes, much like a buffer without uri (or that's at least my understanding of how a buffer without uri behaves).

I believe the intended behavior to be simply using the first / only image in the GLTF file *per default* as the source, but I am not sure.

Can someone please clarify this?

Thank you in advance.

zellski:
I am curious, too. The use case I've assumed is that there are times when you author a model intending to provide a texture source at run-time (perhaps a photo frame), and you want artists to be able to control and export everything except the final actual pixel array. But that's just a guess.

As far as I can see, THREE.GLTFLoader simply NPEs if it's missing, so I'm going to suggest it's a bit of an edge case to leave it undefined.

takahirox:
What first came into my head is framebuffer.
They don't need an input image.

tomolt:
But why make framebuffers so implicit ?
Wouldn't it be much more in GLTF's spirit to explicitly define an array of framebuffers for the entire file
and then have the texture object either supply a 'image' property (instead of the current 'source' property) or a 'framebuffer' property pointing at the framebuffer you should use.
That way, it might be easier for loaders to bind their framebuffers to the textures.
Also, I'm sure the spec right now doesn't even mention framebuffers, so having them be an undocumented edge-case seems kind of wrong to me.

takahirox:
I just mentioned what texture without image can be for. I'd like to know the actual purpose of non-image-required texture in glTF spec and expected use case, too.

javagl:
[...]

PR #1458, opened 2018-09-28
Make source property required in texture

UX3D-becher:
Currently textures without an image source are valid glTF which leads to errors in some engines (e.g. Babylon.js, Cesium, Three.js). 

As it is unspecified how engines should display such 'empty' texures, the specification should probably be modified to prohibit them.

Example: 
[glTF.zip](https://github.com/KhronosGroup/glTF/files/2428149/glTF.zip)

### DIFF
--- a/specification/2.0/schema/texture.schema.json
+++ b/specification/2.0/schema/texture.schema.json
@@ -17,5 +17,6 @@
         "extensions": { },
         "extras": { }
     },
-    "gltf_webgl": "`createTexture()`, `deleteTexture()`, `bindTexture()`, `texImage2D()`, and `texParameterf()`"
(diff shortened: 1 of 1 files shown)


emackey:
I think this was intentional so that extensions could supply alternate texture sources...?

bghgary:
I think this thread #1226 has the most context.

pjcozzi:
Is it a breaking change in practice for existing software?  Or is this a spec bug fix?

If this is merged, the generated markdown should also be updated.

lexaknyazev:
`source` is optional because of extensions (like DDS or WebP).

emackey:
Right, thanks @lexaknyazev.  That means we can't merge this.

PR #1464, opened 2018-10-11
Clarify naming of application-specific semantics in morph targets.

### LINKED ISSUE #1363: Custom semantics in morph targets?
The spec says:
> Application-specific semantics must start with an underscore, e.g., `_TEMPERATURE`.

The spec should be more explicit whether this applies to morph targets (I'd say yes).

/cc @bghgary @donmccurdy @emackey

donmccurdy:
Fixes #1363.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -790,7 +790,7 @@ primitives[i].attributes.POSITION +
   weights[1] * primitives[i].targets[1].POSITION +
   weights[2] * primitives[i].targets[2].POSITION + ...
 ```
-Morph Targets are implemented via the `targets` property defined in the Mesh `primitives`. Each target in the `targets` array is a dictionary mapping a primitive attribute to an accessor containing Morph Target displacement data, currently only three attributes (`POSITION`, `NORMAL`, and `TANGENT`) are supported. If morph targets contain application-specific semantics, their names must be prefixed with an underscore (e.g. `_TEMPERATURE`) like the associated attribute semantic. All primitives are required to list the same number of `targets` in the same order.
+Morph Targets are implemented via the `targets` property defined in the Mesh `primitives`. Each target in the `targets` array is a dictionary mapping a primitive attribute to an accessor containing Morph Target displacement data. Currently only three attributes — `POSITION`, `NORMAL`, and `TANGENT` — are commonly supported. If morph targets contain application-specific semantics, their names must be prefixed with an underscore (e.g. `_TEMPERATURE`) like the associated attribute semantic. All primitives are required to list the same number of `targets` in the same order.
 
 Valid accessor type and component type for each attribute semantic property are defined below. Note that the *w* component for handedness is omitted when targeting `TANGENT` data since handedness cannot be displaced.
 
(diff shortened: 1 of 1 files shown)


specification/2.0/README.md
```diff
@@ -790,7 +790,7 @@ primitives[i].attributes.POSITION +
   weights[1] * primitives[i].targets[1].POSITION +
   weights[2] * primitives[i].targets[2].POSITION + ...
 ```
-Morph Targets are implemented via the `targets` property defined in the Mesh `primitives`. Each target in the `targets` array is a dictionary mapping a primitive attribute to an accessor containing Morph Target displacement data, currently only three attributes (`POSITION`, `NORMAL`, and `TANGENT`) are supported. All primitives are required to list the same number of `targets` in the same order.
+Morph Targets are implemented via the `targets` property defined in the Mesh `primitives`. Each target in the `targets` array is a dictionary mapping a primitive attribute to an accessor containing Morph Target displacement data, currently only three attributes (`POSITION`, `NORMAL`, and `TANGENT`) are supported. If morph targets contain application-specific semantics, their names must be prefixed with an underscore (e.g. `_TEMPERATURE`) like the associated attribute semantic. All primitives are required to list the same number of `targets` in the same order.
```
emackey: The "... currently only three" part is a run-on sentence, it should be the start of a new sentence.

Also, I'd like to change "supported" to either "widely supported" or "commonly supported" or similar, to indicate that we're talking about implementation support, not a spec requirement.

donmccurdy: Hm, true. I think I'm happy with "commonly supported."

emackey:
@bghgary Did you want to review?

specification/2.0/README.md
bghgary: Does this imply that there can be morph targets that displace say COLOR_0? If so, how does an implementation decide what to implement?

donmccurdy: That is the implication, yes. I think the latest language...

> ... only three attributes — `POSITION`, `NORMAL`, and `TANGENT` — are commonly supported.

...is probably enough to suggest a starting point for implementations. But I don't know of any implementations that support morphing COLOR_0 today.

bghgary: I still want to know what attributes implementations should support given this.

donmccurdy: Many implementations do not support animation at all, per @emackey's comment above, but we already have an implementation note saying:

> A conformant client implementation must support at least eight morphed attributes. This means that it has to support at least eight morph targets that contain a POSITION attribute, or four morph targets that contain a POSITION and a NORMAL attribute, or two morph targets that contain POSITION, NORMAL and TANGENT attributes.

I suppose that should mean conformant clients must implement POSITION, NORMAL, and TANGENT morph targets? I'll admit I don't see much value in implementing TANGENT morph targets, however.

lexaknyazev: There's nothing wrong with morph targets providing full tangent space. One of NVIDIA's GPU Gems examples showed that.

donmccurdy: Nothing wrong with it, agreed.

But for reference, three.js [removed support for tangents in favor of UV derivatives](https://github.com/mrdoob/three.js/issues/7094) in 2015, and to my knowledge no one has ever requested that we bring them back... requiring clients to implement tangent morph targets for glTF seems like a solution in need of a problem.

Perhaps we will need tangents to get anisotropic reflection in future PBR revisions? But I don't have first-hand experience with that.

PR #1465, opened 2018-10-11
Tighten requirements for semantic names.

### LINKED ISSUE #1360: Tighten format for attribute semantic names with indices
The spec says:
> `TEXCOORD`, `COLOR`, `JOINTS`, and `WEIGHTS` attribute semantic property names must be of the form `[semantic]_[set_index]`

> All indices for indexed attribute semantics, must start with 0 and be continuous: `TEXCOORD_0`, `TEXCOORD_1`, etc.

It's still unclear:
* Is using numbers with more than one digit allowed: `TEXCOORD_10`?
* Are leading zeros allowed: `TEXCOORD_02`?

As of now, the validator marks all semantics with more than one digit as invalid.

/cc @emackey @donmccurdy @bghgary

donmccurdy:
Fixes #1360.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -763,7 +763,7 @@ Valid accessor type and component type for each attribute semantic property are 
 
 `TEXCOORD`, `COLOR`, `JOINTS`, and `WEIGHTS` attribute semantic property names must be of the form `[semantic]_[set_index]`, e.g., `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_0`. Client implementations must support at least two UV texture coordinate sets, one vertex color, and one joints/weights set. Extensions can add additional property names, accessor types, and/or accessor component types.
 
-All indices for indexed attribute semantics must start with 0 and be continuous positive integers: `TEXCOORD_0`, `TEXCOORD_1`, etc. Indices should not use leading zeroes to pad the number of digits, and clients are not required to support more indexed semantics than described above.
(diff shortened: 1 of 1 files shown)


specification/2.0/README.md
```diff
@@ -763,7 +763,7 @@ Valid accessor type and component type for each attribute semantic property are
 
 `TEXCOORD`, `COLOR`, `JOINTS`, and `WEIGHTS` attribute semantic property names must be of the form `[semantic]_[set_index]`, e.g., `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_0`. Client implementations must support at least two UV texture coordinate sets, one vertex color, and one joints/weights set. Extensions can add additional property names, accessor types, and/or accessor component types.
 
-All indices for indexed attribute semantics, must start with 0 and be continuous: `TEXCOORD_0`, `TEXCOORD_1`, etc.
+All indices for indexed attribute semantics must start with 0 and be continuous positive integers: `TEXCOORD_0`, `TEXCOORD_1`, etc. Indices should not use leading zeroes to pad the number of digits, and clients are not required to support more indexed semantics than described above.
```
lexaknyazev: Maybe change "should not use leading zeroes" to "must not use leading zeroes" if we're going to enforce that?

donmccurdy: Done

emackey:
Looks like all comments were addressed.

PR #1470, opened 2018-10-21
Fix formating on WEIGHT_0's "type" in the primitive attributes

Ybalrid:
A '"' character was missing.

This is a minor edit.


Ybalrid:
I was checking the list of possible attributes in the spec, and that thing jumped to my eye ;-)

Issue #1142, opened 2017-11-03
Support for primitive restart


Themaister:
For triangle strip/line strip primitive types, it would be useful to support primitive restart (with fixed index 2^n - 1).

pjcozzi:
Thanks for the suggestion, @Themaister!  Will certainty be useful for graphics APIs beyond WebGL 1.

lexaknyazev:
Apparently, we need to discuss this within glTF 2.0 lifecycle.

WebGL 2.0 behaves as if primitive restart is always on with fixed index (255, 65535, or 4294967295). [The rationale from the spec](https://www.khronos.org/registry/webgl/specs/latest/2.0/#NO_PRIMITIVE_RESTART_FIXED_INDEX):
> This compatibility difference was introduced in order to avoid performance pitfalls in Direct3D based WebGL implementations. Applications and content creation tools can be adjusted to avoid using the maximum vertex index if the primitive restart behavior is not desired.

This means that some engines would need to manually rebuild index buffer in some cases which goes against glTF's goal.

Action items:
- [ ] Add a note about this to the spec.
- [ ] Add a validation warning on index buffers containing restart values.

Optionally, depending on ecosystem status:
- [ ] Disallow primitive restart values in index buffer.

/cc @donmccurdy @bghgary @emackey

lexaknyazev:
Same issue with Vulkan (although this feature can be disabled).
From `VkPipelineInputAssemblyStateCreateInfo`:
> `primitiveRestartEnable` controls whether a special vertex index value is treated as restarting the assembly of primitives. This enable only applies to indexed draws (`vkCmdDrawIndexed` and `vkCmdDrawIndexedIndirect`), and the special index value is either `0xFFFFFFFF` when the `indexType` parameter of `vkCmdBindIndexBuffer` is equal to `VK_INDEX_TYPE_UINT32`, or `0xFFFF` when `indexType` is equal to `VK_INDEX_TYPE_UINT16`. Primitive restart is not allowed for “list” topologies.

lexaknyazev:
Actual implementation details are even more cumbersome. D3D11 may have different behavior with different GPUs or driver versions. More details:
https://bugs.chromium.org/p/angleproject/issues/detail?id=708
https://bugs.chromium.org/p/angleproject/issues/detail?id=2035

Metal also seems to have primitive restart always on with max values.

[...]

PR #1476, opened 2018-11-01
Disallow primitive restart values.

donmccurdy:
Per #1142.

/cc @MiiBond @bghgary @lexaknyazev @emackey


bghgary:
@lexaknyazev This is expected since the 3rd value is a restart value (WebGL2 for BabylonJS and DirectX11 for 3D Viewer) and thus there is no triangle to show.

bghgary:
I haven't heard any issues with this, so I think this is good to go.

PR #1486, opened 2018-11-14
Add a note about URI normalization

### LINKED ISSUE #1459: URI normalization
From the other [discussion](https://github.com/KhronosGroup/glTF/pull/1100):
> Speaking of URIs, should we also disallow referencing parent files (e.g. using ../../..)? This was brought to my attention during a security pass.

> We should definitely add a note about security concerns (e.g., loading an asset with parent paths from untrusted source). I'm not sure about disallowing such URIs on spec level.

> Also related: should we encourage normalization on export (e.g., `dir1/./dir2/../tex.png` to `dir1/tex.png`)?

> In other words, canonicalized paths. I say yes, especially if caching is happening using this URI.

Possible action items:
- Require / endorse canonicalization
- Discourage / disallow parent dir references

/cc @donmccurdy @bghgary @emackey

lexaknyazev:
Fixes #1459.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -193,7 +193,7 @@ glTF uses URIs to reference buffers and image resources. Clients must support at
 
 Applications should consider applying syntax-based normalization to URIs as defined by [RFC&nbsp;3986, Section&nbsp;6.2.2.](https://tools.ietf.org/html/rfc3986#section-6.2.2), [RFC&nbsp;3987, Section&nbsp;5.3.2.](https://tools.ietf.org/html/rfc3987#section-5.3.2), and applicable schema rules (e.g., [RFC&nbsp;7230, Section&nbsp;2.7.3.](https://tools.ietf.org/html/rfc7230#section-2.7.3) for HTTP) on export and/or import.
 
-> **Implementation Note:** While the spec does not explicitly disallow non-normalized URIs, their usage may lead to unwanted side-effects such as security warnings or cache misses on some platforms.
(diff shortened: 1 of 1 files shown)


specification/2.0/README.md
```diff
@@ -191,6 +191,10 @@ glTF uses URIs to reference buffers and image resources. Clients must support at
...
 
+Applications should consider applying syntax-based normalization to URIs as defined by [RFC&nbsp;3986, Section&nbsp;6.2.2.](https://tools.ietf.org/html/rfc3986#section-6.2.2), [RFC&nbsp;3987, Section&nbsp;5.3.2.](https://tools.ietf.org/html/rfc3987#section-5.3.2), and applicable schema rules (e.g., [RFC&nbsp;7230, Section&nbsp;2.7.3.](https://tools.ietf.org/html/rfc7230#section-2.7.3) for HTTP) on export and/or import.
+
+> **Implementation Note:** While the spec does not explicitly disallow non-normalized URIs, their usage may lead to unwanted side-effects such as security warnings or cache misses on some platforms.
```
donmccurdy: Is it worth mentioning that some client implementations may reject non-normalized URIs for security reasons? E.g. I would expect Facebook to do so. The side effects could potentially be a little more severe than warnings or cache misses.

lexaknyazev: Valid point re per-app rejection. Could you please adjust the note?

donmccurdy: Done

PR #1490, opened 2018-11-15
Remove `name` from technique.uniform

lexaknyazev:
The schema for `technique.uniform.schema.json` inherits from `glTFProperty` so it shouldn't have `name` property. Uniform names come from `technique.uniforms` keys.


lexaknyazev:
OK to merge?

PR #1499, opened 2018-11-23
Clarify occlusion math.

### LINKED ISSUE #915: Clarify the occlusion color
The occlusion strength specification says:

> A scalar multiplier controlling the amount of occlusion applied. A value of 0.0 means no occlusion. A value of 1.0 means full occlusion. This value affects the resulting color using the formula: `occludedColor = lerp(color, color * <sampled occlusion texture value>, <occlusion strength>)`. This value is ignored if the corresponding texture is not specified. This value is linear.

What `color` is meant here?  Is it the `c_diff` color? Is it the final output color?  Is it something else?

donmccurdy:
Fixes #915.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -3012,7 +3012,7 @@ This integer value is used to construct a string in the format TEXCOORD_<set ind
 
 #### occlusionTextureInfo.strength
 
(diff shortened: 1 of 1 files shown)


peteroupc:
You specify multiply occlusion by the _diffuse_ color while the diagram you gave showed "the correct math" as multiplying occlusion by the _ambient_ color.  Can you explain this?

donmccurdy:
Will reply on the original thread, #915.

donmccurdy:
As written, my PR doesn't clarify the specification enough I think. Closing this until discussion in #915 resolves.

PR #1501, opened 2018-11-25
Reword buffer.byteLength

### LINKED ISSUE #1281: [bufferView.byteLength] how does stride affect this value?
Is the length the  buffer view "spans" the underlying raw buffer? Or is it the number of bytes in the view itself? e.g. if stride was 4, is it length *4 or just length?

I have not seen this specified in the spec. 

Chip

lexaknyazev:
Fixes #1281.


lexaknyazev:
@donmccurdy OK to merge?

PR #1502, opened 2018-11-25
Joint Hierarchy: small typos

andreasplesch:


andreasplesch:
Thanks, that was fast ;)

PR #1506, opened 2018-11-30
README upgraded from Blender Exporter to Blender IO

emackey:
I removed links to the old exporter and to julienduroure's old importer, and replaced them with a link to the new combined IO package.

/cc @donmccurdy


donmccurdy:
Thanks! Should we also add Blender to the "Applications" section? We list Substance and Modo there, since they include glTF support by default.

emackey:
@donmccurdy Done.  I pointed the link directly to their homepage, since the 2.80 beta has moved up to top billing there recently.

PR #1522, opened 2018-12-26
Use a more complete example

greggman:
I'm not sure if that's a good example but it seems important to point out that `accessor.count` is incomplete for the size of the buffer

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -696,7 +696,7 @@ Consider the following example:
             "byteOffset": 4608,
             "componentType": 5123,
             "count": 5232,
-            "type": "SCALAR"
+            "type": "VEC2"
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


javagl:
I agree, that could be an important hint. The [`accessor.count` spec](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/accessor.schema.json#L68) says 

> The number of attributes referenced by this accessor, not to be confused with the number of bytes or number of components.

but at the point of the example, people will be tempted to just copy+paste the given snippet. 

---

The `accessorTypeToNumComponentsMap` looks a bit elaborate (maybe only due to the formatting). I could imagine some pseudocode like 

    ... accessor.count * numOfComponentsFor(accessor.type));

Along that, one could also consider mentioning that `Uint16Array` is used due to `componentType` being 5123, i.e. `UNSIGNED_SHORT`, and maybe even that this last parameter will then be multiplied by  "BYTES_PER_ELEMENT" to obtain the actual number of *bytes*. But maybe that would go too far - I don't have a strong opinion on that one.

pjcozzi:
@greggman @lexaknyazev are any updates needed here or can this be merged as is?

pjcozzi:
Thanks again @greggman @lexaknyazev!

PR #1523, opened 2018-12-28
Added RapidCompact to list of converters / optimizers

mlimper:
Thanks @pjcozzi for the pointer
Preview here: https://github.com/mlimper/glTF


mlimper:
Alright... just wondering, is that OK for viewers / applications to show up multiple times? I saw Sketchfab is mentioned in various places, and obviously vendors have a natural interest to have their tools being mentioned up as often as possible. What's the policy here? Just wanna be fair with ourselves... ;-)

pjcozzi:
Generally I have only been merging PRs with things listed in one place.

mlimper:
OK, I'll keep that in mind for the future - many thanks ;)

PR #1532, opened 2019-01-10
Fix misplaced index maximum values.

donmccurdy:
Per https://github.com/KhronosGroup/glTF-Asset-Generator/pull/482.

/cc @bghgary @lexaknyazev


bghgary:
The property reference section is generated by [wetzel](https://github.com/AnalyticalGraphicsInc/wetzel) from the schema. Ideally, we should just re-run wetzel, but we still have an [issue](https://github.com/KhronosGroup/glTF/issues/1266). For now, we have been modifying both the property reference and the schema.

In this case, it'll be here: https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/mesh.primitive.schema.json#L19

pjcozzi:
@donmccurdy this just needs the schema update before merging, right?

lexaknyazev:
@donmccurdy Could you please update schema files as well?

donmccurdy:
Schema file updated. The second paragraph was missing from the schema file already, and I'm not sure how/if we can do linebreaks in the schema file (`\n`?), so it's just one paragraph there for now.

PR #1533, opened 2019-01-14
Fixes for Appendix B: BRDF Implementation.

neXyon:
- Clarify V and L vector directions and normalization.
- Correct Geometric Occlusion equations.


snagy:
Either direction works out to the same result from our lighting equations, but this PR lines up with the webgl-pbr sample.

Thanks for the clarification!

pjcozzi:
Thanks for the fix, @neXyon!

Does anyone else want to review this before merging?

Issue #1503, opened 2018-11-28
Extension for WebP for smaller images compared to jpeg


pjcozzi:
By Google.  From https://developers.google.com/speed/webp/:

> WebP is a modern image format that provides superior lossless and lossy compression for images on the web. Using WebP, webmasters and web developers can create smaller, richer images that make the web faster.
>
> WebP lossless images are 26% smaller in size compared to PNGs. WebP lossy images are 25-34% smaller than comparable JPEG images at equivalent SSIM quality index.

Note that this is not GPU compression; this is a replacement for JPG-like compression.

The Cesium team's initial experience with WebP is very promising.

**Ecosystem**
* Decode in Browsers
   * Chrome & Edge
   * Firefox support in progress
   * iOS & Safari - ?
* JS/wasm decoders
   * http://libwebpjs.appspot.com/
   * https://medium.com/@kennethrohde/on-the-fly-webp-decoding-using-wasm-and-a-service-worker-33e519d8c21e
* Decode in desktop / native apps
   * Reference implementation: https://developers.google.com/speed/webp/download
* Encode
   * Reference implementation: https://developers.google.com/speed/webp/download
   * https://sharp.dimens.io/en/stable/api-output/
* Probably others

**Questions**
* Should this be an EXT_ or KHR_ extension?
* How should glTF reference this?  Directly?  via KTX2?
* In the near- and long-term, how does this play with CTTF GPU compression?

lexaknyazev:
> iOS & Safari

Probably never, given that they support a different "JPEG-replacement" format.

> Should this be an EXT_ or KHR_ extension?

KHR would likely require bringing IP in question, so I'd say EXT.

> How should glTF reference this? Directly? via KTX2?

Directly, like JPEGs. KTX2 is for GPU-ready data.

> In the near- and long-term, how does this play with CTTF GPU compression?

Nothing in the near term. Looking forward, we may borrow some techniques from it for the future (past 1.0) CTTF versions.

MarkCallow:
Agree with @lexaknyazev, KTX2 is for GPU ready data.

If you can tolerate uploading uncompressed images to your GPU, i.e. you have vast amounts of memory, and transmission size is a problem then maybe webp is useful. However it is never going to be fully cross-platform. I think it not worth spending time on. We can already get JPEG-levels of compression with the under-development CTTF technology and there is likely room for further improvement and that is where time should be spent.

zellski:
[...]

PR #1534, opened 2019-01-15
EXT_texture_webp extension proposal

OmarShehata:
This is a proposed draft of an extension to support WebP textures following the discussion here: https://github.com/KhronosGroup/glTF/issues/1503.

[Convenient markdown link](https://github.com/OmarShehata/glTF/tree/master/extensions/2.0/Vendor/EXT_image_webp#ext_image_webp)

Since extensions cannot redefine existing properties, this adds a property on an `Image` object that mirrors the base image object, but allows the WebP format. This simplifies implementation (see PRs in [gltf-pipeline](https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/448) and [CesiumJS](https://github.com/AnalyticalGraphicsInc/cesium/pull/7486). The latter includes a sample model). 

Although I just noticed the [MSFT_texture_dds](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/MSFT_texture_dds) extension allows `uri` to point to a dds file, even though the base spec says it must be a PNG or JPEG. If this is allowed then this extension could potentially be simplified.

The only things I'm unsure of are:

* It's strange to define a `mimeType` inside the extension object, given that it will always be `image/webp`. But defining it there simplifies the implementation since clients can just treat it as a regular `image`.

### DIFF
--- a/extensions/2.0/Vendor/EXT_image_webp/README.md
+++ b/extensions/2.0/Vendor/EXT_image_webp/README.md
@@ -15,44 +15,60 @@ Written against the glTF 2.0 spec.
 
 ## Overview
 
-This extension allows glTF models to use WebP as a valid image format by extending the Image property. A client that does not implement this extension can ignore the provided WebP image and continue to rely on the PNG and JPG textures available in the base specification. Defining a fallback texture is optional. The [best practices](#best-practices) section describes the expected behavior when using this extension without a fallback texture.
+This extension allows glTF models to use WebP as a valid image format. A client that does not implement this extension can ignore the provided WebP image and continue to rely on the PNG and JPG textures available in the base specification. Defining a fallback texture is optional. The [best practices](#best-practices) section describes the intended use case of this extension and the expected behavior when using it without a fallback texture.
 
 WebP is an image format developed and maintained by Google that provides superior lossless and lossy compression rates for images on the web. It is typically 26% smaller in size compared to PNGs and 25-34% smaller than comparable JPEG images - [source](https://developers.google.com/speed/webp/).  
 
 ## glTF Schema Updates
 
-The `EXT_image_webp` extension may be defined on `image` structures. Its schema mirrors that of the the base Image property to simplify implementation.
+The `EXT_image_webp` extension is added to the `textures` node and specifies a `source` property that points to the index of the `images` node which in turn points to the WebP image.
 
 The following glTF will load `image.webp` in clients that support this extension, and otherwise fall back to `image.png`.
 
 ```
-"images": [
+"textures": [
     {
-        "uri": "image.png",
+        "source": 0,
         "extensions": {
             "EXT_image_webp": {
-                "uri": "image.webp"
+                "source": 1
             }
         }
+    }
+],
+"images": [
+    {
+        "uri": "image.png"
+    },
+    {
+        "uri": "image.webp"
     }
 ]
 ```
 
-It can also be used with a `bufferView` instead just like the Image property in the base specification:
+When used in the glTF Binary (.glb) format the `images` node that points to the WebP image uses the `mimeType` value of `image/webp`. 
 
 ```
+"textures": [
+    {
+        "source": 0,
+        "extensions": {
+            "EXT_image_webp": {
+                "source": 1
+            }
+        }
+    }
+],
 "images": [
     {
-      "extensions": {
-        "EXT_image_webp": {
-          "bufferView": 3,
-          "mimeType": "image/webp"
-        }
-      },
-      "bufferView": 4,
-      "mimeType": "image/png"
+        "mimeType": "image/png",
+        "bufferView": 1
+    },
+    {
+        "mimeType": "image/webp",
+        "bufferView": 2
     }
-  ]
+]
 ```
 
 ### JSON Schema
@@ -61,26 +77,27 @@ It can also be used with a `bufferView` instead just like the Image property in 
 
 ## Best Practices
 
-It is recommended to provide a fallback PNG or JPEG image for clients that do not support this extension. When a fallback image is defined, this extension should be defined in `extensionsUsed`. This will allow all clients to render the glTF correctly, and those that support this extension will request the optimized WebP version of the textures.
+Since WebP is not as widely supported as JPEG and PNG, it is recommended to use this extension only when transmission size is a significant bottleneck. For example, in geospatial applications the total texture payload can range from gigabytes to terabytes of data. In those situations, WebP textures can be used without a fallback to improve storage and transmission.
+
+When a fallback image is defined, this extension should be defined in `extensionsUsed`. This will allow all clients to render the glTF correctly, and those that support this extension can request the optimized WebP version of the textures.
 
 ### Using Without a Fallback
 
-To use WebP images without a fallback, define `EXT_image_webp` in `extensionsRequired`. The `uri` of the base image object can then be set to an empty string.
+To use WebP images without a fallback, define `EXT_image_webp` in `extensionsRequired`. The `texture` node will then have only an `extensions` property as shown below.
 
 ```
-"images": [
+"textures": [
     {
-        "uri": "",
         "extensions": {
-        	"EXT_image_webp": {
-        		"uri": "image.webp"
-        	}
+            "EXT_image_webp": {
+                "source": 1
+            }
         }
     }
 ]
 ```
 
-Not all browsers currently support WebP. If a glTF contains a WebP with no fallback texture and the browser does not support WebP, the client should either display an error or decode the WebP image at runtime.
+If a glTF contains a WebP with no fallback texture and the browser does not support WebP, the client should either display an error or decode the WebP image at runtime (see [resource](#resources) for decoding libraries).
 
 ## Known Implementations
 
--- a/extensions/2.0/Vendor/EXT_image_webp/README.md
+++ b/extensions/2.0/Vendor/EXT_image_webp/README.md
@@ -15,49 +15,57 @@ Written against the glTF 2.0 spec.
 
 ## Overview
 
-This extension allows glTF models to use WebP as a valid image format by extending the Image property. The [best practices](#best-practices) section defines what an implementation must do when encountering this extension and how this extension interacts with the images defined in the base specification.
+This extension allows glTF models to use WebP as a valid image format by extending the Image property. A client that does not implement this extension can ignore the provided WebP image and continue to rely on the PNG and JPG textures available in the base specification. The [best practices](#best-practices) section describes the expected behavior when using this extension without a fallback texture.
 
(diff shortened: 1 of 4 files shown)


emackey:
I had an offline conversation with @OmarShehata and @pjcozzi where I voiced some of Zell's fragmentation concerns in https://github.com/KhronosGroup/glTF/issues/1503#issuecomment-442897436.  The response, if I'm summarizing correctly, was that WebP has much faster decode & render than the core formats, so for systems like 3D Tiles where lots of images are being streamed down, the decode speed becomes as much of a win (or even more) as the network bandwidth reduction.

Is the intention here that WebP would never be used for "normal" glTFs in the wild, such as from the Blender exporter and other tools?  Would it be strictly contained to just 3D Tiles models?  Or would other engines see the benefit of WebP and want to adopt this as well?

I'm just wondering if the README needs to more clearly state the intended use, so that this doesn't needlessly become a source of fragmentation in the wider ecosystem of models.

donmccurdy:
> Is the intention here that WebP would never be used for "normal" glTFs in the wild, such as from the Blender exporter and other tools? Would it be strictly contained to just 3D Tiles models? Or would other engines see the benefit of WebP and want to adopt this as well?

And, would the answers to the questions above still be true if CTTF were complete today? 

It would clearly by ideal if CTTF were universally the best choice for texture data, but I don't know enough about the expected compression rates, transcoding times, and quality to make that claim.

lexaknyazev:
@donmccurdy 
It's safe to say that storage / transmission cost of WebP is better or the same as CTTF. It's also safe to say that WebP is definitely better than JPEG if you are willing to restrict runtime to WebGL and ignore Safari clients (Firefox will have WebP support later this year).

CTTFv1 will provide smaller GPU memory footprint for opaque natural images. Mileage may vary for images with alpha channel, text or vector shapes (think UI elements).

So, there's room for both, especially with regards to Cesium ecosystem.

emackey:
> ...DDS extension design is correct.
> 
> I would suggest moving the extension object to `texture`, while keeping `image` objects simple. Basically, replicating DDS extension.

Thanks @lexaknyazev.  Can we predict that the CTTF extension will be the same?

@OmarShehata I know you've got PRs open across a number of repos now.  Is it possible to change the extension to work like DDS?  It helps consistency of implementing multiple file types, and it helps GLB packing/unpacking since they don't need to understand the innards of each file type extension.

lexaknyazev:
> Can we predict that the CTTF extension will be the same?

Yes, I'm going to use `image` the same way for CTTF for the exact same reason - uri/bufferView packing.

OmarShehata:
Thanks for all the feedback so far! I've revised the draft and updated the implementation and sample model in the CesiumJS PR. 

> I'm just wondering if the README needs to more clearly state the intended use, so that this doesn't needlessly become a source of fragmentation in the wider ecosystem of models.

I do agree here, that using this for everyday models would destroy a glTF's cross-platform nature. I added a note in the _Best Practices_ section to warn about this (wording suggestions welcome). 

At this point, I think the reason to go for a `EXT_` as opposed to `CESIUM_` is that I believe this would be useful to multiple vendors given the use case described in https://github.com/KhronosGroup/glTF/issues/1503#issuecomment-442897436

> For situations like the work I do at Facebook, where we generate client-specific delivery versions of assets (or where folks are able and willing to host a fallback JPEG version online), having a formalised way to use WebP is inarguably powerful. Asset size is of overwhelming concern for us.

donmccurdy:
Thanks for the clarifications, sounds good. :)

extensions/2.0/Vendor/EXT_image_webp/README.md
```diff
@@ -0,0 +1,110 @@
+# EXT_image_webp
+
+## Contributors
+
+* Omar Shehata, Cesium, [@OmarShehata](https://github.com/OmarShehata)
+* TBA, Google
```
emackey: You have a TBA here

extensions/2.0/Vendor/EXT_image_webp/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Best Practices
+
+Since WebP is not as widely supported as JPEG and PNG, it is recommended to use this extension only when transmission size is a significant bottleneck. For example, in geospatial applications the total texture payload can range from gigabytes to terabytes of data. In those situations, WebP textures can be used without a fallback to improve storage and transmission.
+
+When a fallback image is defined, this extension should be defined in `extensionsUsed`. This will allow all clients to render the glTF correctly, and those that support this extension can request the optimized WebP version of the textures.
+
+### Using Without a Fallback
+
+To use WebP images without a fallback, define `EXT_image_webp` in `extensionsRequired`. The `texture` node will then have only an `extensions` property as shown below.
```
emackey: In this case, `EXT_image_webp` must be defined in both `extensionsUsed` and `extensionsRequired`.

OmarShehata:
@emackey thanks! I took care of your comments and added a link to webp_js as another decoder library that is potentially more complete/has better examples than libwebpjs.

emackey:
@lexaknyazev Did you want to take one last look before this gets merged?

donmccurdy:
Definitely a nit, but, why `EXT_image_webp` rather than `EXT_texture_webp`?

OmarShehata:
> Definitely a nit, but, why EXT_image_webp rather than EXT_texture_webp?

Oh, yeah. It was just that way because of the initial draft. I'll update this now.

extensions/2.0/Vendor/EXT_texture_webp/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+### JSON Schema
+
+[glTF.EXT_texture_webp.schema.json](schema/glTF.EXT_texture_webp.schema.json)
+
+## Best Practices
+
+Since WebP is not as widely supported as JPEG and PNG, it is recommended to use this extension only when transmission size is a significant bottleneck. For example, in geospatial applications the total texture payload can range from gigabytes to terabytes of data. In those situations, WebP textures can be used without a fallback to improve storage and transmission.
+
+When a fallback image is defined, this extension should be defined in `extensionsUsed`. This will allow all clients to render the glTF correctly, and those that support this extension can request the optimized WebP version of the textures.
```
lexaknyazev: This statement is a bit misleading. The extension must be present in `extensionsUsed` regardless of fallback presence.

extensions/2.0/Vendor/EXT_texture_webp/README.md
```diff
@@ -0,0 +1,110 @@
...
+        "bufferView": 2
+    }
+]
+```
+
+### JSON Schema
+
+[glTF.EXT_texture_webp.schema.json](schema/glTF.EXT_texture_webp.schema.json)
+
+## Best Practices
```
lexaknyazev: Can we provide suggestions regarding best-suitable WebP compression modes for each glTF texture type (ORM, normal map, etc)?

emackey: @lexaknyazev It would be handy to have such info, but I'm not sure that's in scope here.  Is there something you had in mind?

lexaknyazev: `cwebp` tool has a long list of command-line options. These could be of our interest:
```
  -q <float> ............. quality factor (0:small..100:big), default=75
  -alpha_q <int> ......... transparency-compression quality (0..100),
                           default=100
  -preset <string> ....... preset setting, one of:
                            default, photo, picture,
                            drawing, icon, text
     -preset must come first, as it overwrites other parameters
  -z <int> ............... activates lossless preset with given
                           level in [0:fast, ..., 9:slowest]
  -exact ................. preserve RGB values in transparent area, default=off
  -noalpha ............... discard any transparency information
  -lossless .............. encode image losslessly, default=off
  -near_lossless <int> ... use near-lossless image
                           preprocessing (0..100=off), default=100
```
JPEG-like artifacts are noticeably bad for normal maps. So having a hint (my guess is `-noalpha` and `-lossless`) in the spec regarding this use case should benefit users of this extension.

OmarShehata: At Cesium we've so far been using WebP for photogrammetry models which have all the lighting baked in so I don't know that I can speak with confidence about best practices for the general case. I did add a note about lossless compression for normal maps and I found this [WebP Compression Study](https://developers.google.com/speed/webp/docs/webp_study) a useful reading so I added it to the resources at the bottom as well.

lexaknyazev: I suggested stripping alpha just in case original file (most likely PNG) was saved in RGBA mode. It has nothing to do with artifacts. We can proceed with this extension as it is and update recommended compression settings later based on production experience.

OmarShehata: That sounds good with me. I removed that note but left the compression study in the resources.

OmarShehata:
@emackey @lexaknyazev this should be ready.

emackey:
@lexaknyazev Is this good to merge?

pjcozzi:
@OmarShehata can you please open a new PR to add this extension to the list: https://github.com/KhronosGroup/glTF/blob/master/extensions/README.md

chase-moskal:
@lexaknyazev 

> [...] if you are willing to restrict runtime to WebGL and ignore Safari clients (Firefox will have WebP support later this year).

i made a webp polyfill that allows webp to work in safari today. it runs the webp decoder clientside, in the browser. you exchange clientside computational effort for large bandwidth savings. much room for optimizations in the library, but it gets the job done today

 - [webp-hero](https://github.com/chase-moskal/webp-hero) — clientside browser polyfill for webp

eventually i'd be interested to see if i could integrate this for usage in web game engines like babylon, maybe even cache results to indexdb or something.. just wanted to drop a line in case somebody needs something like this. i started webp-hero initially to make 100MB of PNGs a more reasonable 5MB, which made an infeasible image-heavy website project suddenly feasible in the real world, webp really shines with it's ability to represent transparency in a lossy alpha channel

aaronfranke:
@OmarShehata Where can we find test glTF files that use WebP textures?

OmarShehata:
@aaronfranke this was the one originally used when drafting this extension: https://github.com/CesiumGS/cesium/blob/bd8be76b7707ddb9c858ae39538c18064c218f98/Specs/Data/Models/Box-Textured-Webp/CesiumBoxWebp.gltf

donmccurdy:
This should work to convert any glTF with PNG or JPEG textures:

```
npm install --global @gltf-transform/cli

gltf-transform webp input.glb output.glb 
```

It uses pretty stock compression settings at the moment. Using --slots, --quality, and --lossless flags you could experiment further: higher quality or lossless modes for normal maps may be worthwhile.

Issue #1028, opened 2017-06-28
Clarification of weights for animations


Fabrice3D:
so I got my parser and exporter working for TRS keyframes
also works nicely on Sketchfab (as far as nice this anim can be)
https://sketchfab.com/models/7f47f3fc84174a8bbe0e375f6a21e6a5

Next was to implement the new 2.0 morph. Unfortunatly, I do not get the mechanics of the weights from the provided examples.
For morphing/vertexAnimations, my program uses mechanics similar to md2 format. Where 2 sources vertices (or more are being used) and interpolated based on duration/ elapsed time . Looking at the stored weight 0-1 values, I can extrapolate its the interpolation value itself. But I'm lost at what and where the sources are.
Also I can't place why the provided example AnimatedMorphCube holds so many keyframes, like 30 per second. Looking at the provided example, 4 or 5 should be enough to reproduce the gif. So It's clearly intended to work in a way I do not grasp.

Aside this page explaining the storage, I wasn't able to find any info on this.
https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#animations

So if someone could explain me in big lines how its supposed to work.
That would be very helpful. thx by advance.

lexaknyazev:
@Fabrice3D 
Each element of `mesh.primitive.targets` array contains per-vertex differences from the base mesh data. 
Were you able to load targets w/o animations?

Fabrice3D:
![screen shot 2017-06-28 at 14 17 02](https://user-images.githubusercontent.com/550453/27636591-8f130c78-5c0c-11e7-807a-8e34c3a1a9a6.jpg)

I see the cube, I have all the data, keyframes data, I simply do not know yet what to do with it.

lexaknyazev:
Animation output accessor has following layout:
```
n - number of weights
c - number of keyframes

[W0, W1, ..., Wn] * c
```

To get morphed mesh, implementation should add weighted sum of all targets to the base mesh.

Fabrice3D:
aaaah! Starts makes sense now. Thank you Lexaknyazev.

javagl:
No formulas, but a snippet from the upcoming glTF Overview for glTF 2.0:

![gltfmorphtargets](https://user-images.githubusercontent.com/5597569/27641394-f84c0b90-5c1b-11e7-876e-32adea9a526f.png)

(Hoping that this appropriately conveys the basic, intuitive idea...(?))

Fabrice3D:
[...]

PR #1541, opened 2019-01-23
Clarify the structure of morph target weights

zellski:
Pretty much a copy-and-paste from https://github.com/KhronosGroup/glTF/issues/1028, to help explain that morph target scalar weights are all just concatenated together. If we wanted to be particularly fancy we could do some ASCII art.


pjcozzi:
@bghgary does this look good to you?

Issue #1270, opened 2018-03-01
[Clarification] Skeleton root and joints hierarchy


Selmar:
The [skins section](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#skins) of the specification mentions:
`The skeleton property points to node that is the root of a joints hierarchy.`

It is unclear to me whether this means that all joints must to part of the hierarchy below the skeleton root node or not, as it says `a joints hierarchy` and not `the joints hierarchy`.

I encountered an issue (or, thus, perhaps not?) with the Babylon exporter, which does not always export the skeleton root node as a parent of all the referenced joints.

najadojo:
The skeleton is rooted (with some additional transforms see #1195 ) with the given node. The joint list are those nodes that participate in the skinning, that is they have inverse bind values and are referred to in the JOINTS_0 mesh attribute by their index in this list. All nodes referenced in the tree described by the skeleton node are part of the skeleton. The nodes without joint index are attachments to the skinned mesh. In the asset you refer to the root is an attachment; [BrainStem](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BrainStem) is a sample that also presents this condition.

Selmar:
Alright, I understand all of that. In fact, the BrainStem model loads fine for me.

I will pose my problem as a question:
Is it allowed for a node in the `joints` array to NOT be a child (or grand-child, etc.) of the `skeleton` node?

bghgary:
> Is it allowed for a node in the joints array to NOT be a child (or grand-child, etc.) of the `skeleton` node?

If #1195 goes in, then the `skeleton` property should be ignored. We are considering removing the `skeleton` property if we can converge on the PR.

donmccurdy:
In addition to #1195, I think the other question here is whether the joints must actually be a strict tree. I.e., even if we remove the `skeleton` property, it is safe for clients to assume that the joint heirarchy (a) is a heirarchy, and (b) has a single root?

Would this be valid, where `Armature` is the node (currently) specified as `skeleton`?

```
<scene root>
├── Armature
│   ├── Bone1
│   ├── Bone2
│   └── Bone3
├── Bone4
└── SkinnedMesh
```

Selmar:
[...]

PR #1552, opened 2019-01-31
Add skinning clarifications

### LINKED ISSUE #1383: [Clarification] Should I animate nodes that are NOT part of the scene?
Hey,

If one of the nodes that is being animated is not directly part of the scene (for example, it may be a node from a skeleton hierarchy that is not contained within the scene itself), should I animate it or not?

For an example, see the attached gltf file. If I load it with ThreeJs, it does not animate, because the skeleton nodes are not part of the scene. But this is not specified in the specs anywhere as far as I can see. What should I do? Animate them or not?

[SimpleSkinNoChildren.txt](https://github.com/KhronosGroup/glTF/files/2175287/SimpleSkinNoChildren.txt)

lexaknyazev:
Fixes #1383 and #1270. 
Before merging, "Property Reference" section of the spec needs to be updated as well.

### DIFF
--- a/specification/2.0/schema/skin.schema.json
+++ b/specification/2.0/schema/skin.schema.json
@@ -11,7 +11,8 @@
         },
         "skeleton": {
             "allOf": [ { "$ref": "glTFid.schema.json" } ],
-            "description": "The index of the node used as a skeleton root. When undefined, joints transforms resolve to scene root."
+            "description": "The index of the node used as a skeleton root.",
+            "gltf_detailedDescription": "The index of the node used as a skeleton root. The node must be the closest common root of the joints hierarchy or a direct or indirect parent node of the closest common root."
         },
         "joints": {
             "type": "array",
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -857,7 +857,7 @@ After applying morph targets to vertex positions and normals, tangent space may 
 
 ### Skins
 
-All skins are stored in the `skins` array of the asset. Each skin is defined by the `inverseBindMatrices` property (which points to an accessor with IBM data), used to bring coordinates being skinned into the same space as each joint; and a `joints` array property that lists the nodes indices used as joints to animate the skin. The order of joints is defined in the `skin.joints` array and it must match the order of `inverseBindMatrices` data. The `skeleton` property points to the node that is the root of a joints hierarchy. 
+All skins are stored in the `skins` array of the asset. Each skin is defined by the `inverseBindMatrices` property (which points to an accessor with IBM data), used to bring coordinates being skinned into the same space as each joint; and a `joints` array property that lists the nodes indices used as joints to animate the skin. The order of joints is defined in the `skin.joints` array and it must match the order of `inverseBindMatrices` data. The `skeleton` property (if present) points to the node that is the common root of a joints hierarchy or to a direct or indirect parent node of the common root.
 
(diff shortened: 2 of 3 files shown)


specification/2.0/schema/skin.schema.json
```diff
@@ -11,7 +11,8 @@
         },
         "skeleton": {
             "allOf": [ { "$ref": "glTFid.schema.json" } ],
-            "description": "The index of the node used as a skeleton root. When undefined, joints transforms resolve to scene root."
+            "description": "The index of the node used as a skeleton root.",
+            "gltf_detailedDescription": "The index of the node used as a skeleton root. The node must be the closest common root of the joints hierarchy or a direct or indirect parent node of the closest common root."
```
donmccurdy: This no longer specifies what happens when the `skeleton` property is omitted... should we say that the joints must have a common root in a scene node? That may be a breaking change compared to `resolves to scene root` from before. Or should we say that implementations may try to find a common root, which may not exist, or may use the scene root?

lexaknyazev: The previous description has never defined what "scene root" was, so I don't see this as a breaking change. We've also agreed that `skeleton` doesn't affect node transforms. The updated readme says that the common root (node) must exist.

donmccurdy: It seems technically possible that assets may exist that do not have a common root node, although I don't know of any authoring tools that would create them that way, or could ingest them that way.

@bghgary assuming this seems OK to you as well i'm fine to make the spec more strict.

bghgary: Seems okay to me.

javagl: I'd have to review some of the discussion around the "skeleton" (particularly, the extensive information in 1270). But doesn't the statement

> "The node must be the closest common root of the joints hierarchy or a direct or indirect parent node of the closest common root."

boil down to something like "The node must be a common ancestor of all joints"? When it can be a direct or indirect *parent* of the closest common root, then the part of the statement that it must be the *closest common root* does not seem to make sense. 

(Sorry, maybe I overlooked or misunderstood something here...)

lexaknyazev: @donmccurdy @bghgary 
Any opinions on wording here wrt @javagl comment? I'm fine either way.

bghgary: I think both versions read okay to me. @lexaknyazev version calls out more explicitly that it doesn't have the be the _closest_ common root. I'm reading the statement as "The node be [A] or [B]" so they don't conflict with each other.

javagl: As mentioned in the comment below: I think that both wordings technically mean the same thing, so it may be just a matter of subjective preference.

pjcozzi:
@bghgary @donmccurdy OK to merge this?

lexaknyazev:
@donmccurdy 
"Property Reference" section should have been updated before merging.

Also, I think that we could streamline language a bit. The updated spec says:

> When a skin is referenced by a node within a scene, all joints used by the skin must belong to the same scene. Joint nodes must have a common root that may or may not be a joint node itself.

Since a common root is required, a node can have only one parent, and scenes can refer to only root-level nodes, we can rewrite this part like:

> Joint nodes must have a common root that may or may not be a joint node itself. When a skin is referenced by a node within a scene, the common root must belong to the same scene.

donmccurdy:
I'd rephrase the first sentence slightly:

> Each skin's joints must have a common root, which may or may not be a joint node itself. When a skin is referenced by a node within a scene, the common root must belong to the same scene.

lexaknyazev:
Opened #1573.

PR #1557, opened 2019-02-06
Add section to README.md for formats built on glTF

ggetz:
Adds a new section to `README.md` for formats built on glTF.

Following [the announced OGC and Khronos Liaison](https://www.khronos.org/news/press/ogc-and-khronos-form-a-liaison-to-improve-interoperability-within-the-geospatial-and-3d-graphics-communities), this seeks to promote glTF as standard foundation which can be referenced by other formats. To support that, should we include [Web3D's SRC format](http://www.web3d.org/src-and-gltf) and the [VRM format (still draft)](https://dwango.github.io/en/vrm/)?


donmccurdy:
Not sure about SRC, but I would omit VRM until we establish the extent to which it's "built on" glTF, as opposed to being one or more "extensions to" glTF.

ggetz:
Thanks @donmccurdy, sounds good. I've updated the table of contents.

PR #1579, opened 2019-03-12
Update the extension template to glTF 2.0?

zellski:
I don't know if people use this template, or if they just grab a ratified extension and modify that, but if there's to be a template, it should be up-to-date...

### DIFF
--- a/extensions/Template.md
+++ b/extensions/Template.md
@@ -10,7 +10,7 @@ TODO: Draft or Stable
 
 ## Dependencies
 
-Written against the glTF draft 2.0 spec.
+Written against the glTF 2.0 spec.
 
 ## Overview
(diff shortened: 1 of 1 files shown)


extensions/Template.md
```diff
@@ -10,7 +10,7 @@ TODO: Draft or Stable
 
 ## Dependencies
 
-Written against the glTF draft 1.0 spec.
+Written against the glTF draft 2.0 spec.
```
bghgary: Probably should remove the `draft` also. :smirk:

```suggestion
Written against the glTF 2.0 spec.
```

zellski: Haha, yes, that makes sense. :-)

PR #1580, opened 2019-03-12
FB_geometry_metadata extension

zellski:
This is a fairly rough write-up of an extension we use internally to provide some geometry-related summary information to parts of our system that can't or don't want to compute it themselves.

I'm not sure it has relevance for the rest of the world, but perhaps we can use this conversation thread to find out!

The schema files should be schematically valid JSON, but I'm not sure if they make semantic sense. In general, feedback is very welcome.

### DIFF
--- a/extensions/2.0/Vendor/FB_geometry_metadata/README.md
+++ b/extensions/2.0/Vendor/FB_geometry_metadata/README.md
@@ -46,6 +46,9 @@ The summary data is defined as an optional extension to a glTF `scene`, by addin
 }
 ```
 
+## Limitations
+This extension completely disregards animation: the static transforms, as supplied on each `node`, are taken at face value. No skeletal animations are applied, nor any morph targets.
+
 ## Computing the data
 
 Conceptually, computing and gathering this information is very straight-forward. On a hypothetical glTF `node` object:
--- a/extensions/2.0/Vendor/FB_geometry_metadata/README.md
+++ b/extensions/2.0/Vendor/FB_geometry_metadata/README.md
@@ -15,7 +15,7 @@ Written against the glTF 2.0 spec.
 
 ## Overview
 
-This extension annotates glTF `scene` objects with a summary of the cumulative geometric complexity and world-space extents of the scene's associated scene graph. Each recursively referenced mesh is enumerated for total vertex and primitive count, and each individual vertex is transformed into world-space for a perfectly computed bounding box.
+This extension annotates glTF `scene` objects with a summary of the cumulative geometric complexity and scene-space extents of the scene's associated scene graph. Each recursively referenced mesh is enumerated for total vertex and primitive count, and each individual vertex is transformed into scene-space for a perfectly computed bounding box.
 
 Some proven real-world applications for this information:
 - When a viewer needs to make a quick, high-level decision which `scene` to display. This is perhaps most meaningful e.g. when `scenes` are used to represent different variants (e.g. LoD levels) of a model.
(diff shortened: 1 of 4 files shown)


zellski:
One open question is whether or not this extension should note explicitly, in the glTF itself, that it applies to static geometry only –– to provide some future-proofing in case someone wants to add at least initial skeletal pose support later.

extensions/2.0/Vendor/FB_geometry_metadata/README.md
```diff
@@ -0,0 +1,108 @@
...
+                9.8512265772065,
+                2.6538096091834
+            ]
+        }
+    }
+}
+```
+
+## Limitations
+This extension completely disregards animation: the static transforms, as supplied on each `node`, are taken at face value. No skeletal animations are applied, nor any morph targets.
```
donmccurdy: I assume this means that no animation is applied when _reading_ the bounds? Or would you also like to impose that animation should not be applied when writing the bounds? Either way, may be good to specify here.

zellski: Hmm. I meant that these values were computed without applying any animation-related effects, and that a reader should interpret them accordingly. You're right, should be more specific.

extensions/2.0/Vendor/FB_geometry_metadata/schema/worldBounds.schema.json
```diff
@@ -0,0 +1,36 @@
...
+            "description": "The bounding box corner with the numerically lowest world-space coordinates",
+            "minItems": 3,
+            "maxItems": 3
+        },
+        "max": {
+            "type": "array",
+            "items": {
+                "type": "number"
+            },
+            "description": "The bounding box corner with the numerically highest world-space coordinates",
```
emackey: Here and above, would it be any clearer to say *scene-space* rather than *world-space*?  These numbers are expressed in terms of the glTF scene, not any particular node, correct?  I'm trying to accommodate the idea that the glTF itself may not be at the center of world space in whatever client has loaded it.

extensions/2.0/Vendor/FB_geometry_metadata/README.md
```diff
@@ -0,0 +1,108 @@
...
+                9.8512265772065,
+                2.6538096091834
+            ]
+        }
+    }
+}
+```
+
+## Limitations
+All these values should be computed with disregard for animation: an morph targets are inactive, all `node` transforms are in their static form, unaffected by any animation. Consequiently, any application that parses the model should interpret the values the same way.
```
emackey: `an morph targets` => `all morph targets`

`Consequiently` => `Consequently`

najadojo:
> I'm not sure it has relevance for the rest of the world, but perhaps we can use this conversation thread to find out!

Yes I was just thinking through a scenario where a geometry summary would be useful to one of our viewers.

extensions/2.0/Vendor/FB_geometry_metadata/README.md
zellski: Wow, I have gotten really reliant on spellcheckers.

extensions/2.0/Vendor/FB_geometry_metadata/schema/worldBounds.schema.json
zellski: That's a good point. I don't know if _scene space_ is an understood notion, but I suppose it's fairly self-explanatory!

emackey:
Left this open (a bit too long) for additional feedback.  Merging it now, any further feedback can be addressed in a new PR.

PR #1583, opened 2019-03-14
Add links to PDF and other forms of reference cards

emackey:
I knew these links existed somewhere, but I think it makes sense to gather them into the caption below the main overview card, for better discoverability.

via https://github.com/javagl/gltfOverview/issues/9

/cc @javagl


emackey:
The new links all came from the Khronos site, so I'm merging it.  Let me know any problems.

PR #1588, opened 2019-03-27
Changed Geometric Occlusion (G) formula.

UX3D-nopper:
As discussed during the 3D formats meeting.


bghgary:
@snagy Can you review?

specification/2.0/README.md
```diff
@@ -3935,9 +3935,9 @@ Simplified implementation of Fresnel from [An Inexpensive BRDF Model for Physica
 
 ### Geometric Occlusion (G)
 
-**Schlick**
+**Smith GGX**
 
-Implementation of microfacet occlusion from [An Inexpensive BRDF Model for Physically based Rendering](https://www.cs.virginia.edu/~jdl/bib/appearance/analytic%20models/schlick94b.pdf) by Christophe Schlick.
+Implementation of microfacet occlusion from [Geometrical shadowing of a random rough surface](https://ieeexplore.ieee.org/document/1138991) by Bruce G. Smith.
```
bghgary: I think we should be recommending the height-correlated Smith GGX as per http://jcgt.org/published/0003/02/03/paper.pdf?

bghgary: For at least two references:
https://google.github.io/filament/Filament.md.html#materialsystem/specularbrdf/geometricshadowing(specularg)
https://github.com/BabylonJS/Babylon.js/blob/master/src/Shaders/ShadersInclude/pbrBRDFFunctions.fx#L170

I haven't checked myself, but Unity and Unreal supposedly uses the same.

UX3D-nopper: Here is the one from Unreal Engine 4 (if u have access):
https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Shaders/Private/BRDF.ush
Yes, u are right, we should change to the one u suggested. But then we need to change the other images as well - or at least explain it e.g. like Filament does.

UX3D-nopper: But, UE4 also uses the one I use and has the `V_SmithGGXCorrelatedFast` as an alternative called `Vis_SmithJointApprox`. It does not have `V_SmithGGXCorrelated`. So, I suggest with the one we currently have.

UX3D-nopper: Also Disney BRDF still uses the "classic" one:
https://github.com/wdas/brdf/blob/master/src/brdfs/disney.brdf

UX3D-nopper: Also the same in Blender 2.80 Eevee ...

```
float G1_Smith_GGX(float NX, float a2)
{
	/* Using Brian Karis approach and refactoring by NX/NX
	 * this way the (2*NL)*(2*NV) in G = G1(V) * G1(L) gets canceled by the brdf denominator 4*NL*NV
	 * Rcp is done on the whole G later
	 * Note that this is not convenient for the transmission formula */
	return NX + sqrt(NX * (NX - NX * a2) + a2);
	/* return 2 / (1 + sqrt(1 + a2 * (1 - NX*NX) / (NX*NX) ) ); /* Reference function */
}

float bsdf_ggx(vec3 N, vec3 L, vec3 V, float roughness)
{
	float a = roughness;
	float a2 = a * a;

	vec3 H = normalize(L + V);
	float NH = max(dot(N, H), 1e-8);
	float NL = max(dot(N, L), 1e-8);
	float NV = max(dot(N, V), 1e-8);

	float G = G1_Smith_GGX(NV, a2) * G1_Smith_GGX(NL, a2); /* Doing RCP at the end */
	float D = D_ggx_opti(NH, a2);

	/* Denominator is canceled by G1_Smith */
	/* bsdf = D * G / (4.0 * NL * NV); /* Reference function */
	return NL * a2 / (D * G); /* NL to Fit cycles Equation : line. 345 in bsdf_microfacet.h */
}
```

UX3D-nopper: To compare with Unity:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.core/ShaderLibrary/BSDF.hlsl

`V_SmithJointGGXApprox` (Unity) ist the same as in `Vis_SmithJointApprox` Unreal Engine 4

UX3D-nopper: Let me check `V_SmithJointGGX` in Unity ...

UX3D-nopper: To summarize ... for G .. we should probably "allow" the following two formulas - to make all "major" Engines "compatible" and also to have not "too many" formulas:

Smith GGX
- Unreal Engine 4
- Blender eevee
- glTF sample viewer (I am fine to move to Smith Joint Approximation)

Smith Joint Approximation
- Unreal Engine 4
- Unity
- Babylon JS
- Filament

I am "painless", if we also "allow":

Smith Joint
- Unity
- Babylon JS
- Filament

Please also compare "Real-Time Rendering", Fourth Edition on page 334 to 336.
Smith Joint is mostly used and Smith GGX are using some practitioners.

bghgary: Another reference that has some good explanations and comparisons: https://twvideo01.ubm-us.net/o1/vault/gdc2017/Presentations/Hammon_Earl_PBR_Diffuse_Lighting.pdf (starting at around slide 57)

My understanding is:
- Uncorrelated Smith GGX is technically wrong, so I don't think we should recommend that in the spec.
- "Joint" means that it is height-correlated.
- All of the equations are GGX.

I think we should just put Smith Joint GGX in the spec and implementations can choose to approximate them for perf reasons. This is non-normative as indicated at the top of the section.

UX3D-nopper: Smith GGX is "Still, surprisingly good in practice" and is used in Unreal Engine 4 and eevee. Also, they both do not have the Smith Joint GGX in.
But to find an end on this, let's switch to Smith Joint GGX and it would be great if the folks from Unreal Engine 4 and eevee could add this formula as an alternative as well.
I will change this later.

UX3D-nopper:
New pull request here https://github.com/KhronosGroup/glTF/pull/1592

PR #1590, opened 2019-03-28
Update WebP best practices

lexaknyazev:
Minor update based on internal evaluation.

/cc @OmarShehata


OmarShehata:
I would suggest linking "extended file format" to https://developers.google.com/speed/webp/docs/riff_container#extended_file_format for context.

PR #1601, opened 2019-04-23
Update README.md

ibober:
New glTF Exporter for Revit plugin link added to Converters, importers and exporters table.


emackey:
Cool, thanks!

PR #1606, opened 2019-05-07
Reword colorspace descriptions

### LINKED ISSUE #1280: Where is texture format specified?
From looking at previous issues such as [this one](https://github.com/KhronosGroup/glTF/issues/835), it seems that there is a property `texture.format`. However, the spec doesn't list that as far as I see (neither [here](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-texture) nor [here](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#texture-data))

Additionally, it seems as though the format is determined by the specific material. For example:

> **material.emissiveTexture**
> This texture contains RGB components in sRGB color space. 

> **material.normalTexture**
> The texture contains RGB components in linear space.

> **pbrMetallicRoughness.baseColorTexture**
> This texture contains RGB(A) components in sRGB color space

Based on this- I'm a bit confused and would love some help understanding this better:

1. is "linear space" equivilent to gl.RGB/gl.RGBA? (e.g. for loading `normalTexture`)

2. The language for `pbrMetallicRoughness.metallicRoughnessTexture` could be a little clearer... it does say "these values are linear", but it might be helpful to use the same consistent language of "This texture contains components in x color space"

3. It feels natural to load a texture with all of the appropriate settings, and then reference it from a material, but the above setup requires crawling through the tree to find the material which references the texture, in order to determine the settings... I guess this isn't such a problem, but it does feel a little bit weird - wouldn't it feel a bit better to simply specify that each texture has a `format` field? (and then a sanity check should confirm that the texture pointed to by each material has the correct color space)

Long story short - all the above would be resolved by a `texture.format` field, but along with not seeing it in the spec, I'm also not seeing it in the demo gltf files... am I missing something? :)

lexaknyazev:
This PR makes color space descriptions more sound. No implementation changes.
Fixes #1280.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
@@ -65,7 +65,7 @@ The diffuse value represents the reflected diffuse color of the material. The sp
 
 The specular property from specular-glossiness material model is the same as the base color value from the metallic-roughness material model for metals. The glossiness property from specular-glossiness material model is related with the roughness property from the metallic-roughness material model and is defined as `glossiness = 1 - roughness`. See [appendix](#appendix) for more details on how you can convert between these two material models.
 
-The value for each property (`diffuse`, `specular`, `glossiness`) can be defined using factors or textures. The `specular` and `glossiness` properties are packed together in a single texture called `specularGlossinessTexture`.  If a texture is not given, all respective texture components within this material model are assumed to have a value of `1.0`. The factors (`diffuseFactor`, `specularFactor`, `glossinessFactor`) scale, in linear space, the components given in the respective textures (`diffuseTexture`, `specularGlossinessTexture`). Both textures are in sRGB space and must be converted to linear space before they are used for any computations.
+The value for each property (`diffuse`, `specular`, `glossiness`) can be defined using factors or textures. The `specular` and `glossiness` properties are packed together in a single texture called `specularGlossinessTexture`.  If a texture is not given, all respective texture components within this material model are assumed to have a value of `1.0`. The factors (`diffuseFactor`, `specularFactor`, `glossinessFactor`) scale, in linear space, the components given in the respective textures (`diffuseTexture`, `specularGlossinessTexture`). Both textures are encoded with the sRGB transfer function and must be converted to linear space before they are used for any computations.
 
 The following equations show how to calculate bidirectional reflectance distribution function (BRDF) inputs (*c<sub>diff</sub>*, *F<sub>0</sub>*, *&alpha;*) from the specular-glossiness material properties. In addition to the material properties, if a primitive specifies a vertex color using the attribute semantic property `COLOR_0`, then this value acts as an additional linear multiplier to `diffuse`.
 
@@ -101,7 +101,7 @@ The RGBA components of the reflected diffuse color of the material. Metals have 
 
 ### diffuseTexture
 
-The diffuse texture. This texture contains RGB(A) components of the reflected diffuse color of the material in sRGB color space. If the fourth component (A) is present, it represents the alpha coverage of the material. Otherwise, an alpha of 1.0 is assumed. The `alphaMode` property specifies how alpha is interpreted. The stored texels must not be premultiplied.
+The diffuse texture. This texture contains RGB components of the reflected diffuse color of the material encoded with the sRGB transfer function. If the fourth component (A) is present, it represents the linear alpha coverage of the material. Otherwise, an alpha of 1.0 is assumed. The `alphaMode` property specifies how alpha is interpreted. The stored texels must not be premultiplied.
 
 * **Type**: [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)  
 * **Required**: No
@@ -125,7 +125,7 @@ The glossiness or smoothness of the material. A value of 1.0 means the material 
 
 ### specularGlossinessTexture
 
-The specular-glossiness texture is a RGBA texture, containing the specular color (RGB) in sRGB space and the glossiness value (A) in linear space.
+The specular-glossiness texture is an RGBA texture, containing the specular color (RGB) encoded with the sRGB transfer function and the linear glossiness value (A).
 
 * **Type**: [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) 
 * **Required**: No
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -265,7 +265,7 @@ The [node transformations](#transformations) and [animation channel paths](#anim
 * rotation: A quaternion (x, y, z, w), where w is the scalar
 * scale: A 3D vector containing the scaling factors along the x, y and z axes
 
-
+RGB color values use sRGB (Rec. 709) color primaries.
 
 ## Scenes
 
@@ -1086,7 +1086,7 @@ This is illustrated in the following figure, where the respective UV coordinates
 <img src="figures/texcoords.jpg" /><br/>
 </p>
 
-Any colorspace information (such as ICC profiles, intents, etc) from PNG or JPEG containers must be ignored.
+Any colorspace information (such as ICC profiles, intents, etc) from PNG or JPEG containers must be ignored. Effective transfer function is defined by a glTF object that refers to the image.
 
 > **Implementation Note:** This increases portability of an asset, since not all image decoding libraries fully support custom color conversions. To achieve correct rendering, WebGL runtimes must disable such conversions by setting `UNPACK_COLORSPACE_CONVERSION_WEBGL` flag to `NONE`.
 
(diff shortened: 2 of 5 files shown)


extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
```diff
@@ -101,7 +101,7 @@ The RGBA components of the reflected diffuse color of the material. Metals have
 
 ### diffuseTexture
 
-The diffuse texture. This texture contains RGB(A) components of the reflected diffuse color of the material in sRGB color space. If the fourth component (A) is present, it represents the alpha coverage of the material. Otherwise, an alpha of 1.0 is assumed. The `alphaMode` property specifies how alpha is interpreted. The stored texels must not be premultiplied.
+The diffuse texture. This texture contains RGB components of the reflected diffuse color of the material encoded with sRGB transfer function. If the fourth component (A) is present, it represents the linear alpha coverage of the material. Otherwise, an alpha of 1.0 is assumed. The `alphaMode` property specifies how alpha is interpreted. The stored texels must not be premultiplied.
```
emackey: Each place we use "encoded with sRGB transfer function" should change to "encoded with the sRGB transfer function".

emackey:
One other topic:  This is marked as fixing #1280, but one of the central concerns there is that the colorspace of an image or texture can't be known without locating the reference to the corresponding texture in the materials list.  This PR doesn't fix that concern, but I don't think there's any fixing that within glTF 2.0.  Perhaps we should address that directly on #1280.

extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness/README.md
lexaknyazev: Done.

emackey:
Ah, missed that, thanks!

specification/2.0/README.md
```diff
@@ -1086,7 +1086,7 @@ This is illustrated in the following figure, where the respective UV coordinates
 <img src="figures/texcoords.jpg" /><br/>
 </p>
 
-Any colorspace information (such as ICC profiles, intents, etc) from PNG or JPEG containers must be ignored.
+Any colorspace information (such as ICC profiles, intents, etc) from PNG or JPEG containers must be ignored. Effective transfer function is defined by a glTF object that refers to the image.
```
emackey: I may be overthinking this, but... any glTF object?  Only a material, or material extension?  Will viewers ever get into decoding troubles if the same image is referenced with two different encodings (for example a BaseColor sharing an image with MetallicRoughness)?  Do viewers ever have to "search" the material list for this reference, and would it help them to halt that search as soon as the first reference to the image is found?  If so, maybe a clarification that any given image can be referenced multiple times but only as a single colorspace, so sRGB material components can't share an image/texture with linear material components?

lexaknyazev: So far, _images_ are referred by texture objects (from the core spec) and `EXT_lights_image_based.light.specularImages[]`.

> BaseColor sharing an image with MetallicRoughness

This sounds like a possible validation warning. There could be two material slots with different transfer functions pointing to the same texture object; or two texture objects (used by two material slots with different transfer functions) pointing to the same image.

specification/2.0/README.md
```diff
@@ -265,7 +265,7 @@ The [node transformations](#transformations) and [animation channel paths](#anim
 * rotation: A quaternion (x, y, z, w), where w is the scalar
 * scale: A 3D vector containing the scaling factors along the x, y and z axes
 
-
+RGB color values use sRGB (Rec. 709) color primaries.
```
bghgary: I don't know much about this, but from just some slight reading on it, I don't know if it's clear to me why Rec. 709 is here. Can we maybe add some links or maybe even the equation for converting to linear?

lexaknyazev: sRGB and Rec. 709 use the same primaries.

specification/2.0/README.md
emackey: Right, I'm asking if the spec itself should state that the model can't implicate multiple transfer functions (both linear and sRGB) onto the same image, regardless of the indirection through shared or non-shared texture objects.  Would that be beneficial to apps trying to determine the transfer function?  I don't know, but reading the linked issue makes it sound like this is desirable.

specification/2.0/README.md
bghgary: Okay, now I'm a bit confused what this statement means. Why mention sRGB or Rec. 709? Doesn't RGB already mean it's red, green, and blue and not some other primaries?

lexaknyazev: Color primaries define what "red", "green", and "blue" mean — that's why this statement is in Units section.

specification/2.0/README.md
lexaknyazev: I'm not sure about explicitly disallowing such edge case. Currently, it's possible to pack all textures into one and use multiple UVs or texture transforms to get proper texels. Multiple transfer functions can be handled by keeping two copies of the image on the GPU, or by applying correction in the fragment shader (less accurate). I doubt that engines will handle this case properly, though.

specification/2.0/README.md
emackey: Yep.  Monitor/TV/printer folks have standards for reproducing these colors, and I'm fine with glTF making a casual external reference to that here.

specification/2.0/README.md
emackey: Interesting.  I suppose it's easy enough to create a test model with this arrangement.  Doesn't sound like we should be encouraging it though.

specification/2.0/README.md
bghgary: Ok, I must be missing something. Is there some article I can read to understand this more? My searches haven't come up with anything useful.

emackey: I don't have any good links for this, but I'll try to offer some context here.  As Alexey points out, this particular sentence doesn't have anything to do with the transfer function.  If I understand correctly, it's talking about color reproductions, typically after the end of the graphics pipeline.  When an RGB image without a colorspace calls for color [1.0, 0.0, 0.0], we know it's asking for full red, but what does "full red" even mean?  Different monitors can have different capabilities as to the full gamut of colors they are able to express in hardware, and [1.0, 0.0, 0.0] may not map to exactly the same shade of red on every available monitor.  When sRGB calls for full red, they offer a precise definition of that red primary, and non-sRGB monitors should adapt the color as needed, to the best of the hardware capability.  Nailing down the exact primary colors like this means that we expect glTF to reproduce its colors the same way on a variety of different display technologies.

bghgary: Thanks. Alexey also gave me a few pointers which helps me understand what this statement means. I'm not sure whether people reading this spec will understand it. Perhaps it's just me.

Regardless though, what do glTF implementations do with this information?

lexaknyazev: > what do glTF implementations do with this information?

Exporters: convert texels to sRGB (or warn when it's impossible) when authoring was done in something else (e.g. if original textures were for some reason authored using AdobeRGB).

Importers: ensure proper look on non-sRGB displays (e.g., with wide color gamut support).

pjcozzi:
@bhouston do you have any feedback on this?  @donmccurdy suggested you might have insight here given all your colorspace knowledge with three.js.

donmccurdy:
^Oops @pjcozzi should have read this more closely first. The PR looks straightforward, and it's okay with me to merge it as-is. @bhouston if you have the bandwidth and interest there are some harder colorspace questions offline, to follow this pull request.

donmccurdy:
^typo, \*I\* should have read this more closely. 😅

specification/2.0/README.md
MiiBond: I agree with Gary that this statement is probably not going to be well understood by many people reading the spec. We could add a link to the Wikipedia page for Rec 709 and, ideally, add a few sentences to help demystify why we're talking about "primaries" and "transfer functions".
I don't think we should assume that the only people implementing exporters/importers are graphics engineers.

meshula: ILM published an introduction to color management here - https://www.openexr.com/documentation/OpenEXRColorManagement.pdf - it goes beyond what gltf needs, but it explains all the foundational concepts from photography to display and how they relate to reproducing pixel values.

bhouston:
It is pretty standard that the non-color textures should be in linear space thus roughness, metallic, opacity, displacement and normal are in linear color space and not sRGB.

This is also how it is in Arnold.  https://docs.substance3d.com/integrations/arnold-5-for-maya-157352171.html

This is best practice in UE4.  https://polycount.com/discussion/138671/matching-roughness-between-ue4-and-sd-sp

But has this been the practice in glTF so far?  It will confuse people if only a few non-color maps are linear and others are sRGB because then there is inconsistency.

donmccurdy:
> But has this been the practice in glTF so far?

This is the practice in glTF already. Just clarifying terminology in this PR. 👍

For completeness, vertex colors and material color factors are also linear.

bhouston:
@donmccurdy wrote:
> @bhouston if you have the bandwidth and interest there are some harder colorspace questions offline, to follow this pull request.

You can send me the requests.  I can make suggestions based on my understanding of industry best practices for sure.

lexaknyazev:
> This is the practice in glTF already. 

It's not just practice, it's the spec. 
glTF mandates a specific transfer function for each texture input.

specification/2.0/README.md
lexaknyazev: The formal standards in this area are not very beginner-friendly.
- [sRGB definitions](http://www.color.org/sRGB.pdf)
- [ITU-R BT.709-6](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.709-6-201506-I!!PDF-E.pdf)

I'd point to classic Poynton's FAQs, although they do not mention sRGB:
- https://poynton.ca/PDFs/GammaFAQ.pdf
- https://poynton.ca/PDFs/ColorFAQ.pdf

Another good guide for engine developers is NVIDIA's paper by Evan Hart: [UHD Color for Games](https://developer.nvidia.com/sites/default/files/akamai/gameworks/hdr/UHDColorForGames.pdf).

emackey: I'm not sure the glTF spec should be attempting to educate beginners on this complex topic.  I like the PR the way it is now.  It references the relevant specifications, such that experts understand that we're following the most common standard in that area, and if a beginner needs to dig into that, they can go research that themselves.

emackey:
That last commit seems like a great clarification, thanks!

PR #1614, opened 2019-05-17
Updated Readme: Adding Animate as exporter and a blog

Jaswant99:
Adding Animate in exporters list and a blog


emackey:
Thanks @Jaswant99

Issue #1036, opened 2017-07-10
Morph Target Names


feiss:
It's not unusual to name morph targets so artists and programmers can refer to them by name, not index. An example in Blender:

![image](https://user-images.githubusercontent.com/359872/28014373-ac9990f0-656c-11e7-9a09-0c0ae099c5e0.png)

But I don't see how 2.0 supports morph target names (https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#morph-targets). It seems it does not. Any thoughts on this?

lexaknyazev:
glTF 2.0 uses integer indices to refer to different entities, because they're stored in JSON arrays. If an application wants to add string identifiers, it could use `extras` property like this (not a part of spec, just an example):
```json
{
  "meshes": [
    {
      "primitives": [
        {
          "attributes": {
            "POSITION": 0
          },
          "targets": [
            {
              "POSITION": 1
            },
            {
              "POSITION": 2
            }
          ],
          "extras": {
            "targetNames": [
              "Happy",
              "Sad"
            ]
          }
        }
      ]
    }
  ]
}
```

feiss:
Yeah I guess that could work

donmccurdy:
Perhaps we should leave this open under `glTF Next` to consider named morph targets in a future version of the spec. It sounds like something that will be useful enough to implement via `extras` now, so bringing that into the spec eventually would be helpful.

lexaknyazev:
Having two ways of doing something (in this case, referring to a morph target) isn't a good thing for data format in general.

I would think of some sort of "Pose" extension which could cover this use case as well as named and predefined skeleton states. Sometimes, model "emotion" consists of both skeleton transforms and morph weights.

donmccurdy:
> Having two ways of doing something (in this case, referring to a morph target) isn't a good thing for data format in general.

Right, if anything this should be similar to the `name` properties used on nodes — available for engines that use them (three.js and BabylonJS both do) but never used as references within the asset.

takahirox:
I really want morph targets names as @feiss mentioned morph targets are often named for artists/programmers. I know Blender and other tools support morph targets names.

I think adding `name` property into `targets` elements is good (if morphTargets in glTF next is based on glTF 2.0). It's user-defined name, like node's, and isn't used as reference as @donmccurdy mentioned.

    "primitives": [
      {
        "attributes": {
          "POSITION": 0
        },
        "targets": [
          {
            "name": "happy",
[...]

PR #1631, opened 2019-06-18
Add non-normative note about the current usage of mesh.extras.targetNames

Ybalrid:
Morph target names are not required to perform morphing. A real time application that consume glTF assets doesn't need morph target names in the general case. 

Rightly so, glTF doesn't impose assets to have target names, but it also doesn't provide a way to optionally add them.

It seems that, since the release of glTF 2.0 in mid-2017, a number of implementation (three.js, the official Blender exporter, UnityGLTF, a and at least one Maya exporter (and probably some others software, I haven't made a list, these are the few I'm aware of)).

This has been discussed in issue #1036. 

The goal is to help implementations that wants to do so have a way to do so in an interoperably way. Currently this has been done by adding a `targetNames` extras properties to the object.

Maybe a more proper way would be to have an extension doing so, however, this practice is already quite widespread, so I would argue that at least documenting it here and integrate a better (and standardized) way of doing so in glTF 2.x/3.0 is the better thing to do.

I am not opposed to the idea of writing an extension to support morph target names, but that would add 2 *not really standard* way of doing it instead of the one already out in the wild.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -854,6 +854,46 @@ After applying morph targets to vertex positions and normals, tangent space may 
 
 > **Implementation note:** The number of morph targets is not limited in glTF. A conformant client implementation must support at least eight morphed attributes. This means that it has to support at least eight morph targets that contain a `POSITION` attribute, or four morph targets that contain a `POSITION` and a `NORMAL` attribute, or two morph targets that contain `POSITION`, `NORMAL` and `TANGENT` attributes. For assets that contain a higher number of morphed attributes, renderers may choose to either fully support them (for example, by performing the morph computations in software), or to only use the eight attributes of the morph targets with the highest weights. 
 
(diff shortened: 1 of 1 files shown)


donmccurdy:
Thanks! I'm in favor of adding the note, but would shorten it and omit the example:


> **Implementation note:** A significant number of authoring and client implementations associate
names with morph targets. While the glTF 2.0 specification does not provide a way to specify names, most tools use a list of strings, `mesh.extras.targetNames`, for this purpose. The `targetNames` list and all primitive targets must have the same length.

Ybalrid:
Hey @donmccurdy I agree with your proposition, it is way more concise and does tell basically the same information.

So, I have removed the example, and I have updated the note to what you proposed. I have replaced "list" by "array" (as these are JSON indexed "array").

Ybalrid:
I think it's good to be merged now?

emackey:
@lexaknyazev or @bghgary Any thoughts before this gets merged?

bghgary:
Ok to me.

donmccurdy:
Thanks for proposing the addition @Ybalrid!

Ybalrid:
@donmccurdy Thanks for accepting it :wink:

PR #1643, opened 2019-07-24
FRAUNHOFER_external_model extension.

tsturm:
State representation containing domain related payload and links to further resources are the key concept of data on the web. The abstract RFC2077 [1] definition for the “model/*” domain include 3D objects as payload and links inside the “model/*” domain. Some registered standards (e.g. X3D) include links to other 3D-Models as base concepts.

Those links allow a natural distribution of 3d data into more than one resource which support different change rates, reuse and logical packaging.

With our FRAUNHOFER_external_model extension we want to propose a simple way to support linkage of external models from within the glTF.


javagl:
This is basically one approach for tackling https://github.com/KhronosGroup/glTF/issues/37 , right? (The issue has been brought up in various forms later - some of them can be found via the links in 37 - but IIRC it has not been solved yet)

PR #1658, opened 2019-08-12
FBX_materials_lightmap

idorurez:
A draft pass at implementing a vendor extension for lightmap functionality


bghgary:
I haven't looked at either in great detail, but since they both have lightmap in the name, what are the differences between this extension and https://github.com/KhronosGroup/glTF/pull/1017?

bghgary:
Ah, sorry, should have read the other thread first. Never mind.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
```diff
@@ -0,0 +1,76 @@
...
+This extension adds an additional annotation to a material, providing a means to introduce a lightmap to be applied in the same scope.
+
+Some proven real-world applications for this information:
+1. Having multiple variations of lighting scenarios without having to bake lighting into a single, original base color/texture.
+2. Specify a separate UV channel (texCoord), to vary the new texture texel resolution with respect to the texture it's being applied to.
+3. Modulate the intensity of a lightmap, dimming or increasing it's effect.
+
+## Specifying Material extension
+
+The data is defined as an optional extension to a glTF `materials`, by adding an `extensions` property and defining a `FB_materials_lightmap` property with a `lightmapTexture` node inside it:
```
donmccurdy: I think we can confidently say that the `lightmapTexture` must be sRGB, right?

donmccurdy:
Thanks @idorurez!

Because this is a draft, I'll respond below on the assumption that FB is not yet using this extension, and is looking for either feedback on the approach or alignment with other vendors. If that isn't the case — e.g. the extension is already in use or locked in — and you'd simply prefer for this to be merged as-is, we can also make that happen.

***

> This does not define how the lightmap will be applied, just that one exists.

For a multi-vendor or Khronos extension, I think we would need to go farther than just identifying the texture as a lightmap. Ideally we would define how it will be applied, assuming some consistency in client applications is possible. If that isn't possible, we would at least want to understand the spectrum of implementations that exist, and provide some documentation and guidance for implementers. For occlusionTexture, as an example, [pre-existing implementations have quite a few differences](https://github.com/KhronosGroup/glTF/issues/915#issuecomment-460360473).
 
- **Unity**: See [Lightmaps: Technical Information](https://docs.unity3d.com/Manual/Lightmaps-TechnicalInformation.html). Offers several quality settings, with the "normal" setting being RGBM.
- **Unreal**: Haven't found much. One link suggested lightmaps are greyscale in Unreal?
- *threejs**: RGB channels are multiplied against lightmapIntensity, then added to indirect diffuse. Alpha channel is ignored.

^Not a lot of detail, hopefully others can help improve this! :)

extensions/2.0/Vendor/FB_materials_lightmap/README.md
donmccurdy: Or maybe not? See Unity docs and HDR / BC6H references..

extensions/2.0/Vendor/FB_materials_lightmap/README.md
```diff
@@ -0,0 +1,76 @@
...
+  ],
+```
+
+## Sample Implementation
+
+In it's simplemest form, Lightmap implementation predominantely involes a base, unlit texture or color in a material, and the supplied lightmap texture is sampled and multiplied against that unlit texture/color.
+
+```javascript
+    lowp vec4 lightmapColor = texture2D(LightmapTexture, oLightmapTexCoord);
+    color.rgb *= (LightmapFactor.rgb * lightmapColor.rgb);
```
donmccurdy: For threejs at least, this would be `+=` rather than `*=`... do we have other references here?

idorurez: Yes, this is a draft for a vendor extension and I'd like to merge it as-is, and it's already been implemented. **However**, I'm soliciting a broader range of practices in the event I'm completely missing an important component.  Furthermore, my limited research along with internal discussion hasn't revealed anything beyond a simple multiplicative application for a very basic lightmaps. I am aware of the additive application, but I am under the assumption that such an application implies that there still needs to be some level of

donmccurdy: > Furthermore, my limited research (I'll take a look at the Unity link above also) along with internal discussion hasn't revealed anything beyond a simple multiplicative application for a very basic lightmaps.

Ok thanks. I need to learn more about this myself, too. Everything I can find describes multiplicative lightmaps, so perhaps threejs is unusual here. We can dismiss the additive use if there are no other implementations.

I'm not sure I understand the advantage of a multiplicative lightmap over an ambient occlusion map – it's RGB, true, but can still only make the surface darker the base color?

> I am under the assumption that such an application implies that there still needs to be some level of lighting in the color/texture it is applied to...

If this extension is intended to replace other lighting in the scene, it could arguably require `KHR_materials_unlit` to also be present. But I would prefer to define both cases – both unlit and PBR application of the lightmap.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
donmccurdy: > My thought is that consider whatever platform this gets ingested on, it would be advantageous to leave color space to the adopter's discretion.

We have tried to be specific about the colorspace of other textures in the glTF spec; unless there's a clear reason not to I think we should do the same here. The colorspace could depend on the texture format if necessary.

donmccurdy:
Stepping back from lightmaps for a moment, there are several variations of lighting that either exist in glTF, or have been suggested at some point:

- [x] punctual lights
- [x] ambient occlusion maps
- [ ] ~~ambient lights~~ (not pursuing)
- [ ] area lights
- [ ] lightmaps
- [ ] spherical harmonics
- [ ] light probes / irradiance probes
- [ ] IBL

Looking at this list, I'm really not confident that _all_ of these things belong in the glTF format... Understandably users have different needs; some want efficient solutions for mobile devices today, others want higher quality or more future-proof features. But all the same, if we try to do everything the format will get complex very fast.

Of the not-yet-supported features on that list, where do we think glTF should be — say, 2-5 years out?

/cc @bhouston

robertlong:
I'd like to see all of those lighting features supported in glTF, although I think most of them should be implemented as extensions. I think that discussion should be moved into another issue.

As for lightmaps, I'd like to see a multi-vendor (EXT) supported lightmap extension, specifically one that was supported in Unity, Unreal, Babylon, and Three.js This one looks good as-is and I would imagine adding support for Three.js with a custom shader that used multiplicative lightmaps.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
idorurez: > Ok thanks. I need to learn more about this myself, too. Everything I can find describes multiplicative lightmaps, so perhaps threejs is unusual here. We can dismiss the additive use if there are no other implementations.

Without revealing too much about what we have under our hood, we also have an unpublished spec that uses an additive application. After some consultation, we realized we needed to define a more broader, simple approach with a multiplicative application.

> I'm not sure I understand the advantage of a multiplicative lightmap over an ambient occlusion map – it's RGB, true, but can still only darken the base color?

The multiplicative aspect, in general, is the same. The main difference is that the source itself is indirect (occlusion) vs direct. I could easily see a situation where we would want to brighten the environment with the lightmap.

> If this extension is intended to replace other lighting in the scene, it could arguably require `KHR_materials_unlit` to also be present. But I would prefer to define both cases – both unlit and PBR application of the lightmap.

Ah good point. Again, without revealing too much, I have been working in an environment that operates under the assumption that the (PBR) material will always be unlit. What would you suggest we do in this case?

idorurez:
Without getting too detailed (or do we need to?), should we define the limits on the encoded values based on the encoding method and color space?

I'm referring to the [Unity Technical Documentation on Lightmaps](https://docs.unity3d.com/Manual/Lightmaps-TechnicalInformation.html), where they mention various clamped values for RGBM and dLDR encoding.

Maybe it's worth keeping it to a single encoding method to keep things simple.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
bhouston: A lightmap technically can be unbounded in range (thus HDR) because a light map represents a measure of the incoming light.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
bhouston: > For threejs at least, this would be += rather than *=... do we have other references here?

Light maps, if you are doing them correctly, should be additive with all other light sources.

The reason is pretty simple.  Light in general is additive and a light map is not exclusive but rather just one more light source in a scene.  This way you can combine dynamic lights with precomputed light maps.

bhouston: > The multiplicative aspect, in general, is the same. The main difference is that the source itself is indirect (occlusion) vs direct. I could easily see a situation where we would want to brighten the environment with the lightmap.

>  source itself is indirect (occlusion) vs direct

A light source can not be occlusion and also a light source.  There are direct light sources and indirect light sources, and there are also occlusions, but all three of these things are separate.

I think you have an occlusion map and are calling it a light map.

> Yes, this is a draft for a vendor extension and I'd like to merge it as-is, and it's already been implemented.

If we basically want to say that this is a confused feature and that is okay because it is just a vendor extension, then sure, but then way standardize it at all?

bhouston:
In three.js parlance (which is based on Frostbite's), lightmaps should be considered indirect diffuse - search for "indirect diffuse" in this PDF:

https://seblagarde.files.wordpress.com/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf

The implementation that put them into indirect diffuse was this large PR here:

https://github.com/mrdoob/three.js/pull/7324/files#diff-29f010d669a36f5263d5351f4702e64fR62

Although we should upgrade the ```texture2D( lightMap, vUv2 ).xyz``` to be ```lightMapTexelToLinear( lightMap ).rgb``` so that encoded (e.g. HDR) textures can be supported with ease.

It was discussed how lightMap should interact with other lights (should it occlude other lights) here:

https://github.com/mrdoob/three.js/issues/7377#issuecomment-149366585

bhouston:
lightMaps to be physical should be specified in units at least once combined with the lightMapFactor - and it should be illuminance in lux.

The reason we put light maps into the indirect diffuse component is that it allows us to consider the indirect diffuse component correctly in the material's BRDF rather than just using a simplified material model of just (diffuse * lightMap) for lightmaps.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
idorurez: > I think you have an occlusion map and are calling it a light map.

That's a fair assessment.
 
> If we basically want to say that this is a confused feature and that is okay because it is just a vendor extension, then sure, but then why standardize it at all?

I'm not asking for a full EXT_, multiparty, approved extension. Does a vendor extension need to be standardized?

idorurez:
We're getting way into the weeds here, as Three.js has a much involved and proper model for a physically based implementation of lightmaps. 

Again, without giving a whole lot away, there is specificity on the unlit texture of what we're expecting to multiply against, so I was considering KHR_materials_unlit as a requirement.

However, @bhouston makes a good point: Don't call it lightmap. Name it something like FB_materials_modulationMap to avoid confusion on implementation, much like the thread he has linked. I think this is a much better nomenclature because it implies it's application.

Coincidentally, I called this FB_materials_multMap before refactoring it as FB_materials_lightmap.

What does everyone think?

donmccurdy:
On one hand this is "way into the weeds" for a vendor extension that is already sufficient for the vendor's needs. But it's definitely a discussion I would want to have in detail before an EXT_ or KHR_ extension. If threejs has a physically-based lightmap model now, I don't imagine we would implement the multiplicative version, although certainly threejs users can easily do this using custom shaders or THREE.NodeStandardMaterial.

From my perspective we can merge this (as a vendor extension) any time; it is just a public description of what Facebook needed and built. I agree that another name might be better, if that's ok.

Before an EXT_ or KHR_ extension that other tools would implement, I do think we should have a sense for how/if it fits our roadmap (https://github.com/KhronosGroup/glTF/pull/1658#issuecomment-523091780) and, if we decide to proceed, agree on a physical model.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
```diff
@@ -0,0 +1,80 @@
+# FB\_materials\_lightmap
```
donmccurdy: Consider renaming. Additional ideas, aside from those mentioned in the thread:

- FB_materials_base_color_multiplier
- FB_materials_unlit_lightmap

extensions/2.0/Vendor/FB_materials_lightmap/README.md
```diff
@@ -0,0 +1,80 @@
...
+```javascript
+"materials" : [
+  	{
+  		"name" : "material_test",
+  		"pbrMetallicRoughness" : {
+  			"baseColorTexture" : {
+  				"index" : 0,
+  				"texCoord" : 0
+  			}
+  		},
```
donmccurdy: Should include `KHR_materials_unlit` in the example? Or does FB render unlit even without this?

extensions/2.0/Vendor/FB_materials_lightmap/README.md
```diff
@@ -0,0 +1,80 @@
...
+  				]
+  			}
+  		}
+  	}
+  ],
+```
+
+## Sample Implementation
+
+In it's simplemest form, Lightmap implementation predominantely involes a base, unlit texture or color in a material, and the supplied lightmap texture is sampled and multiplied against that unlit texture/color.
```
donmccurdy: Couple typos: simplest, predominantly, involves.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
donmccurdy: > Does a vendor extension need to be standardized?

It does not. Very often contributors opening a "vendor" PR are also proposing standardization, or don't understand the difference, so we do tend to review liberally regardless of the prefix. :)

idorurez:
Thanks Don, your advice and help is appreciated. I'm going to rename this as appropriately after consulting with a few engineers here to be sure everyone's happy and update this pull request.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+```javascript
+    lowp vec4 lightmapColor = texture2D(LightmapTexture, oLightmapTexCoord);
+    color.rgb *= (LightmapFactor.rgb * lightmapColor.rgb);
+
+```
+
+## Limitations
+This does not define how the lightmap will be applied, just that one exists.
```
donmccurdy: As this is a vendor extension, and your implementation only applies it to unlit materials, I think this *Limitations section could be expanded:

> This extension depends on `KHR_materials_unlit`, and use with other materials is invalid.

With the requirement of unlit materials, the equation given is relatively specific in terms of how to apply the lightmap.

extensions/2.0/Vendor/FB_materials_lightmap/schema/material.extension.FB_materials_lightmap.schema.json
```diff
@@ -0,0 +1,35 @@
...
+        {
+            "$ref": "glTFProperty.schema.json"
+        }
+    ],
+    "properties": {
+        "lightmapTexture": {
+            "description": "The texture information for the lightmap.",
+            "allOf": [
+                {
+                    "$ref": "lightmap.schema.json"
```
donmccurdy: Should this be `textureInfo.schema.json` instead? Without that, other extensions like `KHR_texture_transform` would be considered invalid here.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
idorurez: Thank you for the suggestions!

extensions/2.0/Vendor/FB_materials_lightmap/README.md
idorurez: FB renders unlit even without this. As suggested earlier, it may avoid confusion if we specifically name the material based on application, so this may be moot.

extensions/2.0/Vendor/FB_materials_lightmap/README.md
idorurez: As noted earlier, this may be a moot point. However, do you think this is still necessary to include the unlit extension? Sorry, I should have asked for your opinion earlier.

extensions/2.0/Vendor/FB_materials_lightmap/schema/material.extension.FB_materials_lightmap.schema.json
idorurez: I'm still new to this, so I'm at your service. Can you point me to any documentation on how standardizing this part of the schema?

idorurez: Changed to `textureInfo.schema.json`

grimdeathr:
Where are we with this? Are lightmaps fully supported with clearcoat and metalness ? I'm having some issues with configuration

kitaedesigns:
Curious why this extension never made it. Is this the same as EPIC_Lightmap extension for gltf?

donmccurdy:
Note that this extension does not add lightmaps to glTF — it rather adds a broadly specified mathematical operation. It sounds like that structure was useful to Facebook in some (unspecified) project for lighting-related customization, but given the discussion under https://github.com/KhronosGroup/glTF/pull/1658#discussion_r316149184, this is not the right approach for a standardized lightmap extension.

Because this was a proposal for a vendor extension, I think it should be merged without further delay if the vendor is still interested in having the extension specification hosted here. If not, the PR can be closed. In either case, design for a lightmap extension probably needs a new PR, with volunteers to work on its definition.

extensions/2.0/Vendor/FB_materials_modmap/README.md
```diff
@@ -0,0 +1,81 @@
...
+],
+```
+
+## Sample Implementation
+
+In it's simplest form, Modmap implementation predominantly involes a base, texture or color in a material, and the supplied modmap texture is sampled and multiplied against that sampled value:
+
+```javascript
+    lowp vec4 modmapColor = texture2D(ModmapTexture, oModmapTexCoord);
+    color.rgb *= (ModmapFactor.rgb * modmapColor.rgb);
```
aaronfranke, line 67: This doesn't look like JavaScript. It looks like GLSL.

PR #1664, opened 2019-08-22
Updated to include Scapic

rosstopherkeen:
Added Scapic's Converter which converts any OBJ, STL, FBX, DAE or GLTF file into GLB (or USDZ) files.


rosstopherkeen:
If the page you're referring to is the [converter](scapic.com/create/usdz), then we do output the model as a glB, but not a glTF.

Sorry about adding into the middle!

donmccurdy:
Sorry for the delay here – we've reorganized the project list, so the best place to add this tool would probably now be https://github.com/KhronosGroup/glTF-Project-Explorer.

Issue #1665, opened 2019-08-23
Stricter Skinning Requirements


marstaik:
After messing around with the gltf spec and various engines, I feel there are quite a few cases with the current specification that make it extremely difficult for most importers to handle with their own internal rendering engines. Either this leads to a bunch of extraneous work and guessing on the importer side, or leads to duplication of data by the exporters.

Here are some ideas that I believe could tighten up the specification:

1) The skeleton root should be defined, otherwise the direct parent of the highest joint in the skin hierarchy must be used. The direct parent does *not* have to be a joint, to allow for multi rooted skeletons. Note that the direct parent may be the root, but it should *not* be the root as it was before when the skeleton property was undefined.

2) All joints in a skin *must* form a connected subtree with the skeleton root/direct parent (1)
This means there lie *no* non joints (or other skins joints) as connecting joints between skin joints. This however still allows you to have other nodes attached to joints, as long as no subsequent joints of the skin follow.
A lot of engines treat the skeleton/skin as a single entity, and allowing for non joints embedded in the tree of joints makes the engine have to create a bunch of shadow bones to get things to work properly.

3) A mesh can only bind to a single skin
I know this is pretty implicit, but I believe it should be defined.

4) As a per previous discussion here: https://github.com/KhronosGroup/glTF-Blender-IO/issues/566#issuecomment-523584953
All meshes skinned must be normalized to the local space of the skeleton.

Now I have another requirement that I am torn between two options: (5) and (6)

5) Each joint shall belong to only one skin
This is the ideal choice as it makes things extremely simple for importers as they do not need to resolve/union skinning trees to find the master skeleton/skin (see below)

6) Each skin shall either define a new tree of unused joints, or, explicitly be a subtree of a previously defined skin.
This allows for multiple skins per skeleton, but the subtree definition implies that all skins for a single skeleton must have a "master" skin that holds *all* of the joints for a skeleton. This makes it easy for an importer to map smaller skin definitions to a master skeleton containing all the joints.

vpenades:
Reusing nodes on multiple skins should not only be allowed, but it is also a great feature!

Consider a character model with 150 bones, made with a mesh and a skin; there's toolsets that might split the skin and the mesh to allow loading it in engines that have a limited number of bones per shader. For example MonoGame has a limit of 72 bones per skin.  And when you do the split, you really, really need that the extra skins keep pointing to the original nodes, so yes, you do need to share nodes between skins.

[...]

PR #1669, opened 2019-09-01
Stricter Skin Definitions

marstaik:
There are some common ambiguities with the current skin definition that leave a lot of implementation details up in the air, for reference see here: https://github.com/KhronosGroup/glTF/issues/1665

This is an initial PR to get some feedback as to how to potentially tighten up the Skinning specifications a bit and remove unnecessary ambiguity.


marstaik:
I have devised a way that should not break any of the current importers, and only benefit those engines trying to import with game engines, if you would still be interested in an updated PR.

donmccurdy:
@marstaik suggestions for a non-breaking alternative would be welcome, yes.

I would further clarify @emackey's comment, though, by saying that not breaking existing importers alone is not sufficient for a 2.x change – we also need to avoid invalidating existing glTF 2.0 models, unless they're egregiously unsupportable in some way. That would require a longer-term v3.0, or perhaps an extension (`EXT_skinning_strict`?) that imposes additional requirements on the core specification.

specification/2.0/README.md
```diff
@@ -862,7 +863,13 @@ After applying morph targets to vertex positions and normals, tangent space may
...
+All skins are stored in the `skins` array of the asset. Each skin is defined by the `inverseBindMatrices` property (which points to an accessor with IBM data), used to bring coordinates being skinned into the same space as each joint; and a `joints` array property that lists the nodes indices used as joints to animate the skin. The order of joints is defined in the `skin.joints` array and it must match the order of `inverseBindMatrices` data.
+
+All joints defined in a skin must form a connected subtree within the node hierarchy, and the highest order node is the `skin parent`. If a skin contains multiple distinct subtrees (multi-rooted, a joint is not defined which connects them) then the "roots" of the distict subtrees must be siblings in the node hierarchy, and their comment parent is the `skin parent` (note that in this case, the `skin parent` may not be a joint).
```
donmccurdy: Given the motivation for these restrictions, why allow multiple distinct subtrees at all? Unless I've missed something, couldn't any skin with multiple distinct subtrees and a common parent be made into a single connected subtree, by turning that common parent into a joint?

marstaik:
While I would welcome having them under a single parent joint, the multiple subtrees (that are siblings) don't pose a problem, as you can group them under the parent inside another entity (skeleton). I feel like whether or not you want to allow multiple joint subtrees is a smaller issue than the whole non-joints between skins issue.

In https://github.com/godotengine/godot/pull/32213 I successfully did some work interpolating overlapping/touching skins to a single skeleton in godot, but that code is very algorithmic in nature, and it seems there are some few edge cases still.

In that PR, you will see joint "stealing" and fake joint creation, expansion of skins to catch non-joints and create fake joints, as well as expanding skins vertically until they find a common parent, then merging skin subsets into skeletons. Before the scene based skin support there was also code to reroot the skin IBMs to the interpolated skeleton.

The point in trying to make is that while it's possible to handle this through code (everything is possible) it's still a huge challenge.

The exporter could make these decisions far for easily and also make sure it's making the correct decisions for the export due to every programs internal modeling differences.

But the PR would not have achieved as much without @reduz implementing Skins in godot.

marstaik:
Sorry that I took so long to respond, been busy.

No, there isn't a strong motivation for that exception. Some systems allow for multirooted trees, so I thought of also allowing it.

However, the strong counter argument is that if a system can handle a multirooted tree, it can definitely handle a single rooted tree.
But the reverse is not true.

So it may just be better to remove allowing multirooted trees.

But, after experiencing first hand the behavior of many importers and exporters, and analyzing gltf behavior, I will try to make a new PR thst captures the real essence of the problems, and tries to correct them in an intelligent way.

donmccurdy:
Closing this PR in favor of #1695 or #1747. We can't make breaking changes to the glTF 2.0 specification at this point, but an extension or changes in the next version of glTF (whenever that might be) are things we can consider.

Issue #1569, opened 2019-02-28
Early discussion around asset variants


zellski:
## Introduction

One of the use cases for 3D assets we often see referenced is the concept of compactly represented, easily selectable variants. This has obvious applications across the board, but has seen increased attention in the burgeoning realm of 3D-assisted commerce, whether on flat surfaces or in AR/VR/XR/MR. The AR use case is particularly interesting, because it comes with strict fidelity requirements in real-world measurements.

For a vivid example, consider the product page for a pair of sneakers: available in six colour combinations and a range of eight sizes, and additionally should be displayed atop a small chunk of ground plane that could be asphalt, track or trail. For extra credit, the sizes are not simple homogenous rescalings of the canonical asset, but also include subtle changes in geometry that need to be correct.

We could generate 6 * 8 * 3 assets, one per permutation. We could be a little more clever and load the ground plane asset separately, but we'd still end up with 48 assets, which seems prohibitive, and silly given the enormous redundancy across the assets.

## The Challenge

So we'd like to find a way to package all these variants up in a single asset in some way that maximises reuse of shared data. The efficiency with which we can do this is critically important: if it bloats the asset too much, then the initial time-to-interaction takes a blow, and the scheme becomes counter-productive.

A compliant loader would need the abililty to (very) quickly switch between assets variants on demand, at runtime.

To have real-world meaning, the representation scheme (which would surely take the shape of an glTF extensions, or perhaps a few interacting ones) must also be crafted with tooling in mind. It must not be prohibitively difficult or error-prone to generate and process these packaged-up assets, either for content applications tools, optimisation pipelines, or the actual humans tasked with creating the assets.

A final challenge is that interest in this realm is growing quickly, we can imagine that inside many tech companies efforts are already brewing to come up with ad-hoc solutions, to the ultimate detriment of retailers, who can't realistically be asked to produce a different asset for each platform. Balancing the need to for platforms to move quickly against the urge for future consensus will take work and discipline.

## Some obvious possibilities

While this post is intended to throw open discussion, some solutions are so obvious that to omit them seems disingenuous:
- Using simple tag strings to identify variants.
  - This lends itself well to composition, where variant-enabled sneakers rest atop a variant-enabled ground plane, and two tags are used to identify the combination.
- The ability to simply switch a mesh primitive from pointing to one **material** in the asset to another.
  - An interesting question is whether one might also wish to be able to switch between different sets of per-vertex colour.
[...]

PR #1681, opened 2019-10-07
The KHR_materials_variants extension

zellski:
**EDIT: Since this original write-up, the extension has entered the pipeline to become `KHR_materials_variants`**

This pull request introduces our vendor extension, `FB_material_variants`, and it is our hope that in this discussion thread, we'll see opinion converge on promoting it to EXT level.

I recommend anyone interested to catch up on the excellent & varied conversation in https://github.com/KhronosGroup/glTF/issues/1569, where much ground on the subject has already been covered.

In order to limit scope creep here, I also propose the likelihood of a future & more powerful `EXT_node_variants`, which would use the same type of simple tag-based mechanism, except to switch the children of a node, rather than the material of a mesh primitive. That's a far more ambitious project, though, and doesn't obviate the need for a simple extension like this that covers the common use cases.

Finally – there will shortly be a legally unencumbered tool on GitHub that produces files on this extension. It'll be a CLI tool at first, with a WebAssembly version later that runs entirely in-browser.


extensions/2.0/Vendor/FB_material_variants/README.md
```diff
@@ -0,0 +1,98 @@
...
+]
+```
+
+The tag-based approach allows highly selective material switching, e.g. changing the colour of shoe laces only, or wholesale changes across diverse geometry, by reusing the same tags in switches across many distinct mesh primitives.
+
+In the typical case, many tags will be active, e.g. `['sneaker_red', 'laces_gold']`.
+
+Composition also works well with tags: if a scene were built from several different variational models, each with their own set of tags, then it's easy to imagine an API call on the entire scene that takes a set of tags and passes them on to each constituent model, recursively.
+
+## Interaction with existing glTF functionality
```
donmccurdy: I think the intention is clear, but consider explicitly putting a non-normative note at the top of this section and the one below:

*This section is non-normative.*

extensions/2.0/Vendor/FB_material_variants/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+The primary purpose of this extension is to simply formalise the idea of configurability, so that members of the ecosystem can meaningfully communicate: those who produce asset and those who ingest them, and all the tooling inbetween.
+
+Beyond that, it's also a compression scheme: a way to compactly represent multiple assets that have a lot in common (e.g. most geometry, some textures). In this respect, it's most meaningful for the self-contained binary format GLB, where the asset file contains all its data up-front, both that which is shared between variants and that which differs.
+
+If we also allow external references, i.e. images and buffers to source from URLs,then we can additionally come up with sophisticated mechanisms where geometry and textures are loaded separately, dynamically. There's pragmatic pros and cons to going down this route, all beyond the scope of this document. Each application will make its own decisions there; meawhile, the extension proposed works well regardless.
```
donmccurdy: These two paragraphs (`Beyond that, ...` and `If we also...`) seem to contradict one another, and may be confusing as-is. I think you could effectively condense them to something like:

> As a secondary effect, material variants allow multiple assets — with shared geometry but different materials — to be stored more compactly. When using external URIs as references to textures, applications may (optionally) process geometry only once and lazily request texture assets only when needed for a particular variant.

***

On this section...

> If we also allow external references, i.e. images and buffers to source from URLs,then we can additionally come up with sophisticated mechanisms where geometry and textures are loaded separately, dynamically.

... I would consider this a certainty rather than an "if"; incrementally loading parts of a glTF asset is already possible and supported in engines like three.js. It's probably out of scope for this document, as you say, but I would strongly expect (though not require) that applications will take advantage of it.

extensions/2.0/Vendor/FB_material_variants/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+| Tags                               | Material                        |
+| ---------------------------------- | ------------------------------- |
+| `sneaker_yellow`, `sneaker_orange` | `shoelace_hole_material_brown`  |
+| `sneaker_red`                      | `shoelace_hole_material_purple` |
+| `sneaker_black`                    | `shoelace_hole_material_yellow` |
+
+(_the authors of this spec are not product designer, apologies for the dubious colour choices_)
+
+The currently active `material` for a `mesh primitive` is found by stepping through this array of mappings, and selecting the first one which contains any one of the currently active tag. If none match, fall back on vanilla glTF behaviour.
```
deltakosh: How does an app like Babylon.js Sandbox will know what is the active tag at load? Should we think about adding a "Default tag" information?

pushmatrix: I like the idea that the default tag is the gltf vanilla material. No need to specify it. That way authors are able to specify what the default should look like even on viewers that don't implement this extension.

extensions/2.0/Vendor/FB_material_variants/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Beyond that, it's also a compression scheme: a way to compactly represent multiple assets that have a lot in common (e.g. most geometry, some textures). In this respect, it's most meaningful for the self-contained binary format GLB, where the asset file contains all its data up-front, both that which is shared between variants and that which differs.
+
+If we also allow external references, i.e. images and buffers to source from URLs,then we can additionally come up with sophisticated mechanisms where geometry and textures are loaded separately, dynamically. There's pragmatic pros and cons to going down this route, all beyond the scope of this document. Each application will make its own decisions there; meawhile, the extension proposed works well regardless.
+
+## Implications for Applications and APIs
+
+How does an application communicate to a glTF engine what the initial variant state should be? How does it submit a runtime request for a different configuration? It's out of scope for this extension to constrain or mandate an engine's public API, but a useful implementation will require something of the sort.
```
edzis: Although we might not enforce users not making conflicting declarations I think we might need to cover the initial state by this spec in some way that helps avoid situations where the models by default look different on viewers that do not support this extension.
Maybe that could be solved by validation logic or a somewhat advanced json schema?

extensions/2.0/Vendor/FB_material_variants/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+Imagine a sneaker with shoelace holes that are made from materials that depend on the overall shoe colour in non-obvious ways:
+
+| Tags                               | Material                        |
+| ---------------------------------- | ------------------------------- |
+| `sneaker_yellow`, `sneaker_orange` | `shoelace_hole_material_brown`  |
+| `sneaker_red`                      | `shoelace_hole_material_purple` |
+| `sneaker_black`                    | `shoelace_hole_material_yellow` |
+
+(_the authors of this spec are not product designer, apologies for the dubious colour choices_)
```
edzis: Would you consider exposing in metadata all the choices that the users can make and all the values that they can specify? There might be more than just a single set of variants.
Those could drive the UI. And even if the UI is defined by the database this could be a way to ensure that the model is capable of representing all the same options exposed to the user.

If this ends up being applied, I would say that the tagnames would benefit from a well-defined format that combines both the property name together with the selected value. For example underscores might not be allowed for property  and value ids and be reserved only for concatenating those together.

sleroux: > Would you consider exposing in metadata all the choices that the users can make and all the values that they can specify?

Yea I think we would need to expose multiple tag 'sets' through metadata. When we talk about driving UI with content from the glTF extension I get a bit wary as we should try to keep the data stored in the extension as business-logic free as possible.

> If this ends up being applied, I would say that the tagnames would benefit from a well-defined format that combines both the property name together with the selected value.

I think having a clear and standardized way of tagging will make it easier for viewers to implement the spec so I can see this being useful regardless.

extensions/2.0/Vendor/FB_material_variants/README.md
```diff
@@ -0,0 +1,98 @@
...
+                "NORMAL": 1,
+                },
+                "indices": 2,
+                "material": 0,
+                "extensions": {
+                    "FB_material_variants" : {
+                        "mapping": [
+                            {
+                                "tags": [ "sneaker_yellow", "sneaker_orange" ],
+                                "material": 7,
```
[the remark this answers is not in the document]
sleroux: I'm not sure if a `visible` flag is needed as the mapping is only used as a look up table that lists all of the materials that can be applied to the mesh. In the case of volumetric logos that have associated geometry, I'd say that would fall out of scope of the material variants extension as that would start changing the nodes a level up from the material.

extensions/2.0/Vendor/FB_material_variants/README.md
pushmatrix: This has been brought up several times and consensus so far is to keep business-logic out of this extension. It can definitely lead to a rabbit hole of complexity as soon as there are thousands of combinations or rules start being needed (e.g The red laces can go with the yellow or red sole, but not the blue sole unless the purple heel is chosen).

It also puts a big burden on the 3D authoring tools. We can't expect 3D modelling tools to support defining the metadata of all possible configurations.

sleroux:
@donmccurdy @pushmatrix I feel like this extension has converged into a form that a few different companies are comfortable with using. With the exception of clarifying some of the description for this PR, do you see anything preventing renaming this to be an EXT extension? Most of the development on this is coming from the 3D commerce TSG instead of just Facebook so it would seem reasonable to transition it into an EXT. If there is other criteria let me know and we can move forward on those points.

extensions/2.0/Vendor/FB_material_variants/README.md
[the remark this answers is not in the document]
mikkoh: I've been thinking about `tags` more and my feeling is that `tags` don't represent the "first match" nature of this design well. In fact when I wrote a prototype application around this I actually implemented "best match" where the material with the most matching tags was applied.

I feel like they should be called `selectors` or something denoting that one the first matching material will be applied.

donmccurdy:
@sleroux Sorry for the delay — the use of an EXT prefix would be reasonable, yes. There is no additional formal requirement beyond the adoption of multiple vendors.

extensions/2.0/Khronos/KHR_variant_materials/schema/mesh.primitive.FB_material_variants.schema.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
```
pushmatrix: The filename for this should be KHR_variant_materials.schema.json no?

emackey:
I think `*_materials_variant` was more correct per the naming conventions:

https://github.com/KhronosGroup/glTF/tree/master/extensions#naming

pushmatrix:
@emackey The idea was to avoid confusing this extension with other PBR Next / material rendering extensions

emackey:
Oh, I see.  I haven't been following this one closely but I didn't get that impression from the name being different.  If it becomes a KHR extension, there may not be much need to separate it from PBR next, or there may be some better way to indicate that separation.  I'd rather see the ecosystem's names be consistent, I think it looks cleaner and more organized.  But that's just my $0.02.

donmccurdy:
+1 for _materials_variants. That it affects materials in a different way than existing material extensions doesn't change the fact that it is primarily scoped for materials, and reordering the words doesn't particularly clarify how it's different from existing extensions.

The same is true for `KHR_texture_basisu` and `KHR_texture_transform` — one adds a new *type* of texture, the other adds a new *feature* for all textures. The naming convention cannot capture that level of detail, which is fine, so we aim for consistency instead.

***

In https://github.com/KhronosGroup/glTF/pull/1681#issuecomment-578800783 it was suggested that this move from FB_ to EXT_, but it seems to have gone directly to KHR_. Was there other discussion or rationale on this?

MiiBond:
@emackey @donmccurdy This extension does not modify materials. Technically, it modifies primitives. Perhaps `KHR_primitive_material_variants` is more appropriate?  Maybe the current naming logic was meant to avoid this long of a name.

donmccurdy:
> This extension does not modify materials. Technically, it modifies primitives. 

I don't think the location in schema is the important thing here. Technically both the `KHR_lights_punctual` and `EXT_mesh_instancing` modify _nodes_, but it's conceptually clear that they are about lights and meshes in practice. All materials are attached to primitives, which are attached to meshes. The intention here is to allow "multiple material variants of an asset".

On that note, what was the reason that tags are attached to references to materials, rather than the materials themselves? I could imagine this structure instead:

```js
"meshes": [
    {
        "name": "shoelace_hole",
        "primitives": [
            {
                "attributes": {
                "POSITION": 0,
                "NORMAL": 1,
                },
                "indices": 2,
                "material": 0,
                "extensions": {
                    "KHR_variant_materials" : {
                        "materials": [0, 1, 2],
 // ...
],
"materials": [
{
  "name": "sneaker_yellow_lace_1",
  "extensions": {"KHR_variant_materials": {"tags": [ ... ]}},
  // ...
},
// ...
]
```

... which would reduce in less verbosity in an asset with many meshes. But I'm also fine with the current structure.

pushmatrix:
@donmccurdy 

> at note, what was the reason that tags are attached to references to materials, rather than the materials themselves? I could imagine this structure instead:

The reason being that a tag can represent a grouping of materials. In the reality of real products, it's not as common to have a shoe that is completely configurable (i.e select a lace color, sole color, inside color, logo color, etc...)

It's more that you have different "variants" of the shoe that are available. For example:
Screen Shot 2020-03-20 at 1 28 35 PM

It would be easier to pass in the active tag "street", and have all the materials that have the street tag on it be selected, versus having to pass in a whole bunch of individual material ids.

pushmatrix:
Regarding the EXT -> KHR naming. If this extension comes out of the Khronos group, then it would make sense for it to be a Khronos extension.

donmccurdy:
> It would be easier to pass in the active tag "street", and have all the materials that have the street tag on it be selected, versus having to pass in a whole bunch of individual material ids.

I'm not suggesting individual material IDs, no. But you're describing "materials that have the street tag", as if materials have tags, when the extension actually puts tags on each _reference_ to a material in the primitive.

Taking the example in the readme, of a "shoelace_hole" primitive. Suppose the shoe model has 100 total primitives, and 30 of them reference material `0`. Do we expect each of those 30 primitives to have the same tags for that material, or might they differ?

donmccurdy:
> Regarding the EXT -> KHR naming. If this extension comes out of the Khronos group, then it would make sense for it to be a Khronos extension.

Understood, and willing to discuss this more. But note that adoption of this extension is going to look very different from adoption of all previous extensions. Many ecosystem tools that have otherwise implemented all KHR extensions will likely not be able to implement this because it defines application behavior in a way that standalone _viewers_ can do, but engines like three.js and A-Frame cannot. Likewise, I'm not sure what DCC tools like Blender are going to do with it.

If the primary tools that support this are tools solely designed for retail, we will need to be very careful about how we position and set expectations for the extension.

pushmatrix:
> I don't think the location in schema is the important thing here. Technically both `KHR_lights_punctual` and `EXT_mesh_instancing` modify _nodes_, but it's conceptually clear that they are about lights and meshes in practice. All materials are attached to primitives, which are attached to meshes. The intention here is to allow "multiple material variants of an asset".

We also want to keep a consistent naming when we introduce geometry variants (toggling meshes based on tags). Would like those to share the same prefix.

> Taking the example in the readme, of a "shoelace_hole" primitive. Suppose the shoe model has 100 total primitives, and 30 of them reference material `0`. Do we expect each of those 30 primitives to have the same tags for that material, or might they differ?

Yeah we would expect that.

I'm trying to understand your approach a bit more. Why would you have the array `"materials": [0, 1, 2]` and not just material: 0, and then in material 0 have all the variantions of that material? I'm likely having a brain fart. It's been a long week, hah.



> If the primary tools that support this are much more focused on retail than the wider glTF ecosystem, we will need to be very careful about how we position and set expectations for the extension.

It's not only retail. That's the lens I'm looking at it under, but this would support complex configurators if needed. 

variants

Here you can see I have multiple active tags, and I can switch configurations on the fly. 

Ex:
[black_laces, pink_body, black_sole]
or
[black_laces, blue_body,white_sole]

For this approach though, I have tags representing specific parts of the model (_laces, _body, _sole)

Here's what parts of the file looks like:

Screen Shot 2020-02-25 at 6 43 26 PM

Screen Shot 2020-02-25 at 6 43 29 PM

donmccurdy:
> >     Do we expect each of those 30 primitives to have the same tags for that material, or might they differ?
> 
> Yeah we would expect that.

Yes to which, sorry? 

(A) Two primitives may reference the same material, with different tags.
(B) Two primitives may reference the same material, but must use the same tags.

The concept of materials "having" tags implies the latter. If that's really the case, the schema could be optimized for this. If that's not the case, that's totally fine.

> Why would you have the array "materials": [0, 1, 2] and not just material: 0, and then in material `0` have all the variations of that material?

I agree that the primitive should reference multiple materials, as you've set it up. I'm just wondering if the `tags` could be moved from the reference to the material itself. For a more 1:1  comparison:

**Current:**

```js
// primitives
[
{ extensions: KHR_materials_variants: { mapping: [
  { material: 0, tags: [ "A", "B" ] },
  { material: 1, tags: [ "C", "D" ] },
] } },
{ extensions: KHR_materials_variants: { mapping: [
  { material: 0, tags: [ "A", "B" ] },
  { material: 1, tags: [ "C", "D" ] },
] } },
{ extensions: KHR_materials_variants: { mapping: [
  { material: 0, tags: [ "A", "B" ] },
  { material: 1, tags: [ "C", "D" ] },
] } }
]

// materials
[
  { name: "Black" },
  { name: "Red" },
]
```

**Proposed:**

```js
// primitives
[
{ extensions: KHR_materials_variants: { mapping: [
  { material: 0 },
  { material: 1 },
] } },
{ extensions: KHR_materials_variants: { mapping: [
  { material: 0 },
  { material: 1 },
] } },
{ extensions: KHR_materials_variants: { mapping: [
  { material: 0 },
  { material: 1 },
] } }
]

// materials
[
  { name: "Black", extensions: KHR_materials_variants: { tags: [ "A", "B" ] } },
  { name: "Red", extensions: KHR_materials_variants: { tags: [ "C", "D" ] } },
]
```

Or, even more concise but equivalent...

**Proposed:**

```js
// primitives
[
{ extensions: KHR_materials_variants: { materials: [ 0, 1 ] } },
{ extensions: KHR_materials_variants: { materials: [ 0, 1 ] } },
{ extensions: KHR_materials_variants: { materials: [ 0, 1 ] } }
]

// materials
[
  { name: "Black", extensions: KHR_materials_variants: { tags: [ "A", "B" ] } },
  { name: "Red", extensions: KHR_materials_variants: { tags: [ "C", "D" ] } },
]
```

All three of these are equally flexible, if materials "have" tags. If primitives need to reference the same materials with different tags, of course, then my proposal is bad. :)

donmccurdy:
> We also want to keep a consistent naming when we introduce geometry variants (toggling meshes based on tags). Would like those to share the same prefix.

Consistency I can get behind, let's aim for that. 🙂 Fast-forward a year or two, and suppose we could end up in either of these states:

- **A.** KHR_mesh_variants + KHR_materials_variants
- **B.** KHR_variants_meshes + KHR_variants_materials

Personally I feel both (A) and (B) are internally consistent, whether "variants" is a suffix or a prefix. However, given that we have other extensions with `KHR_mesh_*` and `KHR_materials_*` in the pipeline, I feel (A) is much more consistent with the wider glTF ecosystem.

I would prefer to leave the term "primitive" out entirely, and consider it a minor implementation detail of a "mesh". Few tools have a "primitive" concept that resembles glTF's.

rsahlin:
Hi Don - good points!

I prefer A with `KHR_mesh_variants`

aurelienvaysset:
@donmccurdy 
The variants tags should not be defined within the materials because a same material could be applied to various primitives. The current implementation allows to assign materials to different primitives without the need to multiply the number of materials.

donmccurdy:
> The current implementation allows to assign materials to different primitives without the need to multiply the number of materials.

Can two primitives, associated with the same material, **use different tags for that association**?  

That assumption seems to be driving the schema, but it isn't mentioned and no use case is given for it. I'm not asking you to change it if there's a reason for it, but please give an example of why two primitives should share a material without using the same tags for that material.

aurelienvaysset:
> Can two primitives, associated with the same material, use different tags for that association?

Yes. I am sure you are not aware of all the history about it but this is actually the goal behind that extension from our perspective. 

> but please give an example of why two primitives should share a material without using the same tags for that material.

Lets say we have  a pair of socks with 2 primitives Left (L) and right (L)
We also have 2 materials black and white which can be applied to each of the 2 primitives
Variants we want to define are : 
variant_fullblack=L.black+R.black
variant_fullwhite=L.white+R.white
variant_blackandwhite=L.black+R.white

With the current implementation, the definition would look something like
```javascript
// primitives
	// L
	{ extensions: KHR_materials_variants: { mapping: [
	  { material: 0, tags: [ "variant_fullblack","variant_blackandwhite" ] },
	  { material: 1, tags: [ "variant_fullwhite" ] },
	] } }

	//R
	{ extensions: KHR_materials_variants: { mapping: [
	  { material: 0, tags: [ "variant_fullblack" ] },
	  { material: 1, tags: [ "variant_fullwhite","variant_blackandwhite" ] },
	] } }

// materials
[
  { name: "black" },
  { name: "white" },
]
```
In this case if we want to display variant_blackandwhite we can determine that L would use black and R would use white.

To take back your example where the tags are defined within the materials that would give something like

```javascript
// primitives
	//L
	{ extensions: KHR_materials_variants: { materials: [ black, white ] } }

	//R
	{ extensions: KHR_materials_variants: { materials: [ black, white ] } }

// materials
[
  { name: "black", extensions: KHR_materials_variants: { tags: [ "variant_fullblack", "variant_blackandwhite" ] } },
  { name: "white", extensions: KHR_materials_variants: { tags: [ "variant_fullwhite", "variant_blackandwhite" ] } },
]
```
In this case if we want to display variant_blackandwhite we could not determine which material to apply to which primitive.

To offer the same capabilities by defining the tags in the materials, we would have to multiply the number of materials which does not look the right option to me as that would create more complexity.

Another alternative would be a dedicated structure for the variants but I am not sure we want to go in that direction either.

donmccurdy:
Thanks – If the history you mention isn't in this thread I'm probably not aware of it.

That explanation makes sense to me. Knowing that it is an intentional case you mean to support, and not just a side effect of the schema, the current schema seems like a good choice for it.

extensions/2.0/Khronos/KHR_variant_materials/schema/mesh.primitive.FB_material_variants.schema.json
jercytryn: good catch, will fix with latest rename

Drigax:
Similarily to my concerns in [1790](https://github.com/KhronosGroup/glTF/pull/1790#issuecomment-624789030) is there a reason why we chose to implement this system of mesh/material variants over extending animation capabilities to allow for material swapping?

jercytryn:
@Drigax there is more detailed discussion on the motivation behind this extension in https://github.com/KhronosGroup/glTF/issues/1569 but primarily this extension aims to address material variant in a way that is lightweight, simple (deferring business logic to the application), compact, and above all explicit/uniform way.  The tag-based approach we hope is simple and formalizes the configurability aspect of material variants in their own right.  I don't think this precludes material swapping as animation, but the idea here is to give the application the language it needs to talk about tag-based variants and give applications control over swapping in a non-prescriptive way

extensions/2.0/Vendor/FB_material_variants/README.md
jercytryn: Thanks for the suggestions here @donmccurdy, I've gone ahead and condensed the language of the non-normative sections as per your proposal

emackey:
@Drigax Some context may help: The two "variants" extensions are of interest to a working group called 3D Commerce, where I think one of the goals is to enable a store product page showing a 3D model that can be configured in various ways.  The same product may be available in different colors or different configurations, causing certain materials or meshes to swap in and out as the user chooses between these variants.

Separately, I can see the desire to have mesh visibility as part of model animations.  But that can be in its own extension(s), I don't think it needs to intrude on the 3D Commerce ones.

MiiBond:
Have we discussed how this extension will work in conjunction with [EXT_mesh_gpu_instancing](#1691)? i.e. If a mesh has variants, are all GPU instances of the mesh expected to change at the same time when a variant is selected? Normally, I'd expect the logic of which mesh instances (nodes that reference the same mesh) are updated to be purely in the purview of the application. However, with GPU instancing, it complicates things somewhat for the engine to do this.
For example, if 10 GPU instances are defined in the glTF, and the application wants to change only one mesh to another variant, it'll have to convert the mesh so that it doesn't render as an instance.
Should we specify that this extension isn't to be used on a mesh that makes use of GPU instancing? Or should we leave it entirely up to the engine and application to handle this?

donmccurdy:
I believe that question is independent of `EXT_mesh_gpu_instancing`: This extension targets a mesh primitive, which could be referenced by any number of nodes throughout the scene even within the core spec. Semantically I would expect this extension to affect all instances, since it's defined at the mesh/primitive level. That's not a particularly _useful_ outcome, but it's much simpler than the alternative.

A note to this affect should probably be added to the extension; it would be normative.

bghgary:
We're running into some issues when implementing the UI for selecting variants.

The tagging system is (overly?) complex and makes it hard to create UI for it. I'll try to explain.

For the shoe example above:
Body
  Material 0 ["black_body", "variant_black"]
  Material 1 ["pink_body", "variant_pink"]
  Material 2 ["blue_body", "variant_blue"]
Sole
  Material 3 ["black_sole", "variant_black"]
  Material 4 ["white_sole", "variant_pink", "variant_blue"]
Laces
  Material 5 ["black_laces", "variant_black", "variant_pink"]
  Material 6 ["white_laces", "variant_blue"]

What should the UI show to the user to select the variants?

Initially, we thought maybe show a simple dropdown for the available tags, but a user might want to select multiple tags, say "pink_body" and "white_sole". Furthermore, it is hard to determine what the current selection should be for the dropdown.

Then, we thought maybe we should expose a list of checkboxes (or some kind of multiselect mechanism) for available tags, but this is even more tricky. Selecting tags is now fine, but what happens if a user deselects a tag. Furthermore, what if tags overlap in the material that is set? What happens if two tags are conflicting with each other? What does the UI do?

Perhaps the "variant_*" tags can be separated out into a separate list such that a variant (e.g. "variant_pink") stores a list of tags (e.g. ["pink_body", "white_sole", "black_laces"]) instead of using the tags to do it? The UI would use this list for the main selection and can still expose the tags individually if the user wants to customize.

donmccurdy:
I'd been assuming a general-purpose UI for this extension would be a tag multi-selection input:

Screen Shot 2020-06-10 at 5 25 53 PM

The extension does seem (to me) to be precise about how that would be interpreted:

> "The currently active material for a mesh primitive is found by stepping through this array of mappings in order, and selecting the first one which contains somewhere in its list of tags any one of the currently active tags. The corresponding material is assigned to the mesh. If none match, fall back on vanilla glTF behaviour."

Based on that...

> Selecting tags is now fine, but what happens if a user deselects a tag. 

If a user deselects a tag, the active material for all nodes is re-evaluated.

> Furthermore, what if tags overlap in the material that is set? What happens if two tags are conflicting with each other?

The first mapping matching any active tag will be used, regardless of any other tags.

***

To your question of "hard to determine what the current selection should be", I think that this section of the spec should be removed or clarified:

> How does an application communicate to a glTF engine what the initial variant state should be?  ... It's out of scope for this extension.

I think the default state, in the absence of some undefined context known to the application, should always be that no tags are selected, and the asset is rendered as though the extension weren't recognized.

***

> Perhaps the "variant_*" tags can be separated out into a separate list

I don't think I understand this suggestion... it seems to me like an enumerable list of variants would make this extension more complicated, not less?

bghgary:
> image

Maybe it's just me, but this seems to be an more advanced way for selecting variants for a UI? I think I would expect a simple dropdown of some kind and the more advanced customization is hidden to the typical user.

> The extension does seem (to me) to be precise about how that would be interpreted

I think I was missing the fact that the order of the active tags matters. The first one wins if I'm interpreting it correctly. A simple list of checkboxes, for example, would not be able to specify order.

> I think the default state, in the absence of some undefined context known to the application, should always be that no tags are selected, and the asset is rendered as though the extension weren't recognized.

+1 That is what we've done in our implementation.

> > Perhaps the "variant_*" tags can be separated out into a separate list
>
> I don't think I understand this suggestion... it seems to me like an enumerable list of variants would make this extension more complicated, not less?

I guess I'm not currently focused on how complicated the extension is yet. I think, for the shoe example at least, the tags are being used for two different purposes. The `variant_` tags are being used like a list of pre-defined choices and the rest are individual material options. Showing all of these together in one list seems like it would confuse a user trying to select a variant.

donmccurdy:
> I think I was missing the fact that the order of the active tags matters. The first one wins if I'm interpreting it correctly

The order in which tags are selected in the UI doesn't matter – it's the order of the `mappings` in the glTF file. Effectively that sets precedence of possible materials for the primitive, in case more than one material matches the active tags.

> Maybe it's just me, but this seems to be a more advanced way for selecting variants for a UI?

Hm, that's true... I think think there are some benefits to having tags and not (just) a flat list of variants, like the permutations of {laces, sole, inside color, outside color} in https://github.com/KhronosGroup/glTF/pull/1681#issuecomment-601823501. But to show a user-friendly UI there are "rules" that this extension doesn't contain, like the fact that each of those shoe parts (which might be multiple primitives) represents a single-selection choice.

I'm not sure whether something should be done about that. 😕

bghgary:
> The order in which tags are selected in the UI doesn't matter

I see. I need to read that text more carefully. Unless I'm still reading the text wrong, this logic is pretty strange? In the shoe example, if I select ["variant_blue", "pink_body"], then Material 1 (the pink body) is chosen. If I select ["variant_pink", "blue_body"], then Material 1 (the pink body) is still chosen. How would a user using these tags understand what is going on?

> some benefits to having tags and not (just) a flat list of variant

I agree tags can be used to avoid duplication as mentioned in the comment. If we add a list of variants (intended for UI) that point to an array of tags, that shouldn't duplicate the material ids. That said, I don't see a big issue with duplicating ids though and just have a single flat list. That would be much simpler.

donmccurdy:
> How would a user using these tags understand what is going on?

Yes, I think it's implicit that the app knows more (i.e. more than the extension contains) about how to display these options. The extension doesn't include human-readable names for tags, either. Maybe we need someone to weigh in on how this will be implemented in practice? Would a retailer typically have a sidecar file describing the valid combinations of options?

bghgary:
> The extension doesn't include human-readable names for tags

True. This is the same for animations. There are ways a client implementation can get away without display names though (e.g. use numbers, dots, etc.).

sleroux:
Thanks for all the feedback and questions around implementing UI with the extension Gary. We took this back and realized that the current tagging structure is too loose and ambiguous as it was trying to solve two problems at the same time.

Something we've been struggling with on this extension is providing support for both the 'configuration' and 'variant' use cases. Originally the extension was built around the idea of merging variants of a model into a single file and tagging the variant with a tag. For example, if I had two types of shoes that shared the same geometry but had a different colour scheme throughout, rather than having a different tag for each differing component (laces, sole, body), a single tag would represent the set of materials needed to match the new colour scheme. 

Over time, we modified the extension to support selecting multiple tags instead of just one which opened the door for the ability to implement a configurator. There's definitely value in having a scheme that can support describing the types of a materials each primitive can have but combining these two concepts into the same extension became unwieldy. I think the goal moving forward is to have this extension support the ability to select a variant instead of describing individual primitive material permutations for configurators.

How does this mental model work in the context of what you're implementing for Babylon?

jercytryn:
I've gone ahead and update the pull request with the changes @sleroux mentioned above around clarifying / simplifying multi-"tag" / multi-variant usage, and optimizing around the static configuration case based on discussion above and within the 3D Commerce working group. Feel free to leave any comments on questions here or inline in the latest

sebavan:
@jercytryn, It looks like the shoe.glb model (that we use in our babylon.js tests) does not have the 3 materials embedded but only 1  ?

PS: I was trying to update our support ;-)

jercytryn:
Ah you are right @sebavan, thanks for flagging this.  Looks like all the example models have this problem where they are not embedding the materials not used by default, possibly a bug with the [variant meld tool](https://github.com/facebookincubator/glTFVariantMeld) used to generate them  -- investigating and I will update PR with the fixed models

jercytryn:
@sebavan Models should be fixed now. Meld tool was fine, turns out I was inadvertently stripping them out when manually post-processing (`gltf-pipeline --keepUnusedElements` for the win)

sleroux:
Just curious, would adding a lookup table at the root of the glTF file make this a bit easier to implement into viewers? Right now, you'd have to loop through all the primitives and see if a variant exists to compile a list of available variants. With the change to a single variant representing a configuration for the entire node graph I wonder if we could lift this up:

```json
"extensions": {
    "KHR_materials_variants" : {
        "variants": ["sneaker_yellow", "sneaker_red", "sneaker_toupe"]
     }
}
```

This would be additive to the current proposal.

sleroux:
Just curious, would adding a lookup table at the root of the glTF file make this a bit easier to implement into viewers? Right now, you'd have to loop through all the primitives and see if a variant exists to compile a list of available variants. With the change to a single variant representing a configuration for the entire node graph I wonder if we could lift this up:

```json
"extensions": {
    "KHR_materials_variants" : {
        "variants": ["sneaker_yellow", "sneaker_red", "sneaker_taupe"]
     }
}
```

This would be additive to the current proposal.

emackey:
Can the example models be moved to a new PR on the [sample models repo](https://github.com/KhronosGroup/glTF-Sample-Models)?  They don't really belong in the spec repo (although a few have snuck in, I know).

extensions/2.0/Khronos/KHR_materials_variants/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+| Variants                           | Material                        |
+| ---------------------------------- | ------------------------------- |
+| `sneaker_yellow`, `sneaker_orange` | `shoelace_hole_material_brown`  |
+| `sneaker_red`                      | `shoelace_hole_material_purple` |
+| `sneaker_black`                    | `shoelace_hole_material_yellow` |
+
+(_the authors of this spec are not product designers, apologies for the dubious colour choices_)
+
+Application-specific logic defines, for a given instance of a glTF asset, a single active variant.  The currently active `material` for a `mesh primitive` is found by stepping through this array of mappings in order, and selecting the first one which contains somewhere in its list of variants the currently active variant. The corresponding material is assigned to the mesh.  If none match, fall back on vanilla glTF behaviour.
```
donmccurdy, line 54: If only a single variant may be active, then the order of the mappings shouldn't matter, right? If a variant appears twice in the mappings list, the later of the two can never be selected.

extensions/2.0/Khronos/KHR_materials_variants/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+In other words, **this is not a literal mapping** in its glTF form – exporters, take note.
+
+Implicitly, if the application has no active variant, it should also fall back on vanilla glTF behavior.  For most use cases, we recommend that one of the variants always corresponds to the default vanilla glTF material for each primitive, so that there exists some one-to-one with viewers which do not support the extension, but this is not a strict requirement. The decision of whether to include a separate option to switch to the no active variant case is left as a decision to compliant applications based on their respective use cases.
```
donmccurdy, line 58: What does "separate option to switch to the no active variant case" mean? A mapping with an empty array? Or creating a new variant as a stand-in for the "no variant" case?

extensions/2.0/Khronos/KHR_materials_variants/README.md
```diff
@@ -0,0 +1,131 @@
...
+| ---------------------------------- | ------------------------------- |
+| `sneaker_yellow`, `sneaker_orange` | `shoelace_hole_material_brown`  |
+| `sneaker_red`                      | `shoelace_hole_material_purple` |
+| `sneaker_black`                    | `shoelace_hole_material_yellow` |
+
+(_the authors of this spec are not product designers, apologies for the dubious colour choices_)
+
+Application-specific logic defines, for a given instance of a glTF asset, a single active variant.  The currently active `material` for a `mesh primitive` is found by stepping through this array of mappings in order, and selecting the first one which contains somewhere in its list of variants the currently active variant. The corresponding material is assigned to the mesh.  If none match, fall back on vanilla glTF behaviour.
+
+In other words, **this is not a literal mapping** in its glTF form – exporters, take note.
```
donmccurdy, line 56: I'm not sure I follow this line — does this refer to the dependency on order, mentioned above?

extensions/2.0/Khronos/KHR_materials_variants/README.md
jercytryn: Was actually referring to the UX of the viewer application.  For instance, if it chose to render all the different variants as a combobox or radio buttons, whether to include a separate item for "no variants".  I'll update the language to clarify

extensions/2.0/Khronos/KHR_materials_variants/README.md
jercytryn: This was left in from the initial version of the spec --- yeah, I think it was related to dependency on order. Now that there is a single active variant, we could probably just leave this line out to avoid confusion

extensions/2.0/Khronos/KHR_materials_variants/README.md
jercytryn: Yeah fair enough.  Technically the schema itself does allow for these never-selected mappings to exist, so the order piece just makes it unambiguous which mapping wins - the first should always be selected.  The alternative would be to just explicate/strictly require that for each primitive the variants mentioned in each mapping object must be locally unique -- and a validator for this extension should fail in these cases.. do you think that is preferable?

donmccurdy: If the spec will state normatively that only a single variant may be active, then I think it should also say normatively that variants cannot appear twice in the same primitive mappings list. And probably also that the 'variants' array cannot be empty for any mapping?

It sounds like a material index _could_ twice in the mapping, e.g. associated with two different names?

donmccurdy:
Sorry if this has already been discussed, but does a "mapping" have some meaning to the application, or is it primarily just a more compact representation when two variants use the same material for a specific primitive? It seems unintuitive to me that "mappings" can have human-readable names, extras, and extensions, but the central concept of the extension (variants) have only a tag that is explicitly not a human-readable name. If you wanted to attach metadata (XMP or otherwise) to a variant in the future, you'd need to attach it to a "mapping" instead, and decide some behavior if the mapping has multiple variants?

There would be some more flexibility in keeping a root-level list of variants:

```json
// root
{
  "asset": {"version": "2.0", "generator": "Fancy 3D Tool" },
  "extensions": {
      "KHR_materials_variants": {
        "variants": [
          {"name": "Standard" },
          {"name": "Gold Trim", "extras": { "cost": "$$$" } },
          {"name": "Red Trim" }
        ] 
      }
  }
}
```

... referenced by primitives....

```json
// primitive
{
  "name": "Lace_Left",
  "attributes": {  },
  "extensions": {
    "KHR_materials_variants": {
      "mappings": [
          { "variants": [0, 1], "material": 0 },
          { "variants": [2], "material": 1 },
          { "variants": [3, 4, 5], "material": 2 }
      ]
    }
  }
}
```

I guess that's somewhere between the current version of this PR and the previous iteration?

extensions/2.0/Khronos/KHR_materials_variants/README.md
jercytryn: Makes sense to me.  Will add this

Re: material index, yes though this would be semantically identical to including those different names in the same mapping item.  If we will explicitly prohibit duplicating variants and empty mappings, it probably makes sense to disallow this as well so that there is only one obvious way to do things

donmccurdy: The fact that a mapping is the only place you can attach names/extras/extensions is arguably some justification for allowing duplicate material indices, but see https://github.com/KhronosGroup/glTF/pull/1681#issuecomment-667424806 for more on that.

pushmatrix:
@donmccurdy I agree with the approach you suggest. It makes sense to bring the variants to the root and then reference them in the primitives. Seems to be the glTF way. We will make this amendment.

As for whether the mapping has meaning, it's not always the case but it definitely can. Here's an example:

Consider this shoe.
Screen Shot 2020-08-10 at 7 19 21 PM

The laces and the sole share the same material (white). However, the name might differ depending on whether it's on a lace or a sole. That would be the name of that colour used for marketing.

Imagine going to a retail store and seeing all the possible laces for that shoe. The "white" one is called snow white.

```
{
  "name": "Lace_Left",
  "attributes": {  },
  "extensions": {
    "KHR_materials_variants": {
      "mappings": [
          { "variants": [0, 1], "material": 0, name: "snow white" },
          ...
      ]
    }
  }
}
```

The white material on the sole is called "cloud"

```
{
  "name": "sole",
  "attributes": {  },
  "extensions": {
    "KHR_materials_variants": {
      "mappings": [
          { "variants": [0, 1], "material": 0, name: "cloud" },
          ...
      ]
    }
  }
}
```

So even though both sole & laces share the same material, they might be called differently at the business level. 
This is more common for configurators, and will not be the case for most products. By allowing for extensions, extras, name on the mapping, we can add this flexibility.

So my suggestion is to switch to your approach here:

```
// root
{
  "asset": {"version": "2.0", "generator": "Fancy 3D Tool" },
  "extensions": {
      "KHR_materials_variants": {
        "variants": [
          {"name": "Standard" },
          {"name": "Gold Trim", "extras": { "cost": "$$$" } },
          {"name": "Red Trim" }
        ] 
      }
  }
}```

But to also allow extra fields at the mapping level.

```
// primitive
{
  "name": "Lace_Left",
  "attributes": {  },
  "extensions": {
    "KHR_materials_variants": {
      "mappings": [
          { "variants": [0, 1], "material": 0, extras:...., name: ..... },
          { "variants": [2], "material": 1, extras:..., name: .... },
          { "variants": [3, 4, 5], "material": 2, extras:...., name:.... }
      ]
    }
  }
}
```

**Comment by donmccurdy** [comment]
> So my suggestion is to switch to your approach ... but to also allow extra fields at the mapping level.

Sounds good to me. 👍

**Comment by pushmatrix** [comment]
@bghgary To help bring the Babylon implementation up to date, here are the main changes.

Here's the old way:
*Old*
```
{
  "name": "Lace_Left",
  "attributes": {  },
  "extensions": {
    "KHR_materials_variants": {
      "mappings": [
          {
            "material": 0,
            "tags": ["street", "midnight"]
          },
         {
            "material": 1,
            "tags": ["beach"]
          }
      ]
}
```

In the new one, we renamed `tags` to be `variants`.
*New*
```
{
  "name": "Lace_Left",
  "attributes": {  },
  "extensions": {
    "KHR_materials_variants": {
      "mappings": [
          {
            "material": 0,
            "variants": [0, 1]
          },
         {
            "material": 1,
            "variants": [2]
          }
      ]
}
```
Also, instead of putting the variant name in the array, we put an index that points to the variant definition at the root of the file:
```

{
  "asset": {"version": "2.0", "generator": "Fancy 3D Tool" },
  "extensions": {
      "KHR_materials_variants": {
        "variants": [
          {"name": "street" },
          {"name": "midnight" },
          {"name": "beach" }
        ] 
      }
  }
}
```

Finally, in order to pick a variant, you no longer pass in an array of active tags. Instead you select a single active variant. This simplifies things and makes building a UI around this easier.

The current spec in this PR highlights most of this with the exception of the root level variants. @jercytryn will be updating the spec with that.

Let us know if you have any questions!

**Comment by bghgary** [comment]
@pushmatrix We are already updated to the point before Don's comments, so it should be a quick update. Has the sample been updated yet?

**Comment by pushmatrix** [comment]
@bghgary Not yet, but I'll have that updated tomorrow :)

**Comment by pushmatrix** [comment]
@bghgary I haven't made a PR yet with the updated models, but here's a link to them in the meantime: https://www.dropbox.com/s/ynity52sut4t6x2/khr_materials_variants_models.zip?dl=0.

@jercytryn I assume we can remove them from this PR now.

File: extensions/2.0/Khronos/KHR_materials_variants/schema/mapping.schema.json [hunk 11]
```diff
@@ -0,0 +1,41 @@
+{
```
bghgary, line 1: The schemas need to be updated to match the new changes.

donmccurdy:
@jercytryn and @pushmatrix — would it be possible to get the spec and schema in this PR updated in the next day or so? Just want to be sure the PR is aligned with the plan described in https://github.com/KhronosGroup/glTF/pull/1681#issuecomment-672995821. Thanks!

pushmatrix:
@donmccurdy Trying to!

extensions/2.0/Khronos/KHR_materials_variants/schema/mapping.schema.json
donmccurdy: Done.

donmccurdy:
Schema and spec updates are applied. ✅

extensions/2.0/Khronos/KHR_materials_variants/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Interaction with existing glTF functionality
+
+_This section is non-normative._
+
+The primary purpose of this extension is to simply formalise the idea of static configurability, so that members of the ecosystem can meaningfully communicate: those who produce assets and those who ingest them, and all the tooling in between.
+
+As a secondary effect, material variants allow multiple assets — with shared geometry but different materials — to be stored more compactly, particularly relevant for the self-contained binary format GLB. When using external URIs as references to textures, applications may (optionally) process geometry only once and lazily request texture assets only when needed for a particular variant.
+
+## Implications for Applications and APIs
```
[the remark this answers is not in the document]
sleroux: This was a carry over from an earlier version of the README and doesn't add any value to the spec. I vote for removing it.

extensions/2.0/Khronos/KHR_materials_variants/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+_This section is non-normative._
+
+How does an application communicate to a glTF engine what the initial variant state should be? How does it submit a runtime request for a different configuration? It's out of scope for this extension to constrain or mandate an engine's public API, but a useful implementation will require something of the sort.
+
+## Examples
+
+_This section is non-normative._
+
+For reference, we provide several simple examples of variants, both represented as distinct constituent GLB files without the extension and as multi-variant GLB files leveraging this extension.  We also include the respective net size of the individual variant GLB files and the size of the single combined multi-variant GLB, for comparison.
```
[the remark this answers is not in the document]
sleroux: Would it better to remove this section entirely? I can see an issue with the sample or examples being moved elsewhere since they are in another repo and the spec document falls out of date.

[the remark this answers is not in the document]
emackey: Yes, good point.  Many extensions include a small snippet of sample JSON for the extension itself, without including a complete working sample.  The samples do belong in a separate repo.

So, I think this section should be removed.

extensions/2.0/Khronos/KHR_materials_variants/schema/mesh.primitive.KHR_materials_variants.mappings.schema.json
```diff
@@ -0,0 +1,33 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_materials_variants mesh primitive extension",
+    "type": "object",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
```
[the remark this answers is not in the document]
sleroux: Got it 👍 Wasn't sure if the file name mattered for defining this as it included the 'mappings' in the name. I'll define `mappings` in the schema.

extensions/2.0/Khronos/KHR_materials_variants/README.md
```diff
@@ -0,0 +1,148 @@
...
+```javascript
+"meshes": [
+    {
+        "name": "laces",
+        "primitives": [
+            {
+                 ...,
+                "extensions": {
+                    "KHR_materials_variants" : {
+                        "mappings": [
```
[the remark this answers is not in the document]
sleroux: Should be `mappings`. I'll update the spec accordingly.

extensions/2.0/Khronos/KHR_materials_variants/schema/glTF.KHR_materials_variants.schema.json
```diff
@@ -0,0 +1,25 @@
...
+                "type": "object",
+                "description": "An object defining a valid material variant",
+                "gltf_detailedDescription": "An object defining a valid material variant",
+                "properties": {
+                    "name": {
+                        "type": "string",
+                        "description": "The name of the material variant",
+                        "gltf_detailedDescription": "The name of the material variant."
+                    },
+                    "extras": { }
```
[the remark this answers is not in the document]
sleroux: Yup `object` makes the most sense here.

extensions/2.0/Khronos/KHR_materials_variants/schema/mesh.primitive.KHR_materials_variants.mappings.schema.json
```diff
@@ -0,0 +1,33 @@
...
+        "variants": {
```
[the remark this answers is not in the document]
sleroux: Uniqueness of each mapping is important so I can add this in. Just wondering how JSON schema validators will work if it's an object type with keys instead of scalar values.

extensions/2.0/Khronos/KHR_materials_variants/schema/mesh.primitive.KHR_materials_variants.mappings.schema.json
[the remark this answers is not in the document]
sleroux: For the validation in your screenshot, are you using the glTF Validation tool with the spec? Just curious how you are able to run it.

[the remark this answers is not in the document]
emackey: Actually I just wrote some instructions on that in this comment: https://github.com/KhronosGroup/glTF-Blender-IO/pull/1094#issuecomment-676585367

But the comment says use branch `various-pbr-next`, and for this there's a different branch to use, `materials-variants`.

Once the schema is stable I'll probably merge that branch and publish a new release of gltf-vscode that includes these changes.  If engines support it, a UI chooser for variants could even be created, I think.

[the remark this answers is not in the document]
pushmatrix: Thanks @emackey. That validator worked like a charm. I updated the sample models.

extensions/2.0/Khronos/KHR_materials_variants/schema/glTF.KHR_materials_variants.schema.json
```diff
@@ -3,7 +3,7 @@
     "title": "KHR_materials_variants glTF extension",
     "type": "object",
     "description": "glTF extension that defines a material variations for mesh primivites",
-    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "allOf": [ { "$ref": "../../glTFProperty.schema.json" } ],
```
emackey, line 6: By convention we don't add this `../../` relative path to the official spec.  My gltf-vscode project has a script (`util/importAll.sh` with [docs](https://github.com/AnalyticalGraphicsInc/gltf-vscode/tree/master/util)) that adds these just to the gltf-vscode project internally.

jercytryn: Makes sense, will fix

extensions/2.0/Khronos/KHR_materials_variants/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension allows for a compact glTF representation of multiple material variants of an asset, structured to allow low-latency switching at runtime.
+
+A typical use case is digital commerce, where a user might be presented with e.g. a pair of sneakers and the ability to switch between different colours.
+
+![pair of sneakers](examples/shoes/photo.png)
```
emackey, line 32: Oops, this became a broken link.  Maybe add this image back alongside README.md (not in a folder)?

extensions/2.0/Khronos/KHR_materials_variants/schema/mesh.primitive.KHR_materials_variants.mappings.schema.json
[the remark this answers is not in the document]
emackey: Great!  /cc https://github.com/AnalyticalGraphicsInc/gltf-vscode/pull/189

jercytryn:
Made a last round of updates to clean up / simplify the spec's description of the schema and fixed broken image link (Good catch @emackey )

extensions/2.0/Khronos/KHR_materials_variants/schema/mesh.primitive.KHR_materials_variants.schema.json
```diff
@@ -0,0 +1,44 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_materials_variants mesh primitive extension",
+    "type": "object",
+    "allOf": [ { "$ref": "../../glTFProperty.schema.json" } ],
```
emackey, line 5: One last little `../../` to be removed here.  (No need to re-do the ratification review document just for this).

jercytryn: derp, sorry don't know how I missed this -- should be fixed now

PR #1683, opened 2019-10-09
Update extras.schema.json to support extras.

fire:
Work in progress for https://github.com/KhronosGroup/glTF-CSharp-Loader/issues/18#issuecomment-364659431

Please review and comment.

Edited:

The goal is for a load and then output to not lose the extras information.

### DIFF
--- a/specification/2.0/schema/extras.schema.json
+++ b/specification/2.0/schema/extras.schema.json
@@ -1,5 +1,11 @@
 {
     "$schema": "http://json-schema.org/draft-04/schema",
     "title": "Extras",
-    "description": "Application-specific data."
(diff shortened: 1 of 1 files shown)


emackey:
Hi, unfortunately we can't take this change directly.  See conversation in #1120, and spec note added in #1451.

donmccurdy:
Closing some old PRs out, as @emackey mentions we cannot make this change to the 2.0 schema file. Extras are expected to be objects, and probably that requirement should be formalized in the next version of glTF (whenever that might be). Feel free to open an issue to track the request, if this is still helpful to you.

Issue #1682, opened 2019-10-09
KHR_materials_basisu and independent data channels


donmccurdy:
Opening a new thread here, based on discussion around https://github.com/KhronosGroup/glTF/pull/1612.

Storing three independent data channels in compressed textures introduces significant artifacts with current approaches (see #1612). As a result, these two cases in the current glTF material specification require consideration:

- occlusion/rough/metal (RGB) texture
- three-channel normal (RGB) texture

These could be solved with:

- **(1)** occlusion (R) and rough/metal (RG) textures, and
- **(2)** two-channel normal (RG) texture

A general-purpose texture-swizzling extension would be an alternative to (1). Currently that is harder to support on the web, but it won't always be so, and might be worth considering today.

Assuming (1) and (2) are the correct solutions here, what are the right mechanisms to bring them into the format? A couple ideas:

- **(A)** Introduce new extensions, e.g. KHR_packing_normal, KHR_packing_roughness_metallic, analogous to [MSFT_packing_normalRoughnessMetallic](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/MSFT_packing_normalRoughnessMetallic/README.md).
- **(B)** Include them with the upcoming "PBR Next" material model's extensions; do not backport to existing metal/rough, spec/gloss, or unlit materials.
- **(C)** Include these changes, and perhaps other KTX2/Basis-related changes, in a glTF 2.1 release to simplify the perceived complexity, as compared to specifying multiple interdependent extensions.

MarkCallow:
How would general-purpose texture-swizzling solve the occlusion/rough/metal case? In current compressed textures you can only have 2 independent channels: the RGB channel & alpha.

lexaknyazev:
@donmccurdy 
Possible channel mappings from Basis to GPU formats
https://github.com/KhronosGroup/KTX-Specification/issues/100#issuecomment-538029332

donmccurdy:
@MarkCallow the format already allows occlusion to be separated from rough/metal, but even when they're separated rough/metal still points to the GB channels in the spec, the R channel isn't used. A texture swizzle extension would allow rough/metal to be read from different channels.

zeux:
General-purpose swizzling doesn’t solve normal map storage - shader code is necessary to reconstruct the third component.

[...]

PR #1688, opened 2019-10-17
KHR_materials_sheen

sebavan:
The idea is to start the discussion about the sheen extension, @bghgary and I will keep the thread updated with our implementation progress.


extensions/2.0/Khronos/KHR_materials_sheen/schema/glTF.KHR_materials_sheen.schema.json
```diff
@@ -0,0 +1,36 @@
...
+            },
+            "minItems": 3,
+            "maxItems": 3,
+            "description": "Color of the sheen layer.",
+            "default": [1, 1, 1]
+        },
+        "sheenTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "The sheen layer texture.",
+            "gltf_detailedDescription": "The sheen layer texture. Stored in channel RGB the sheen color in Gamma and in A the linear factor."
```
donmccurdy: Do we have any existing references on efficiently packing sheen color and sheen factor?

On one hand, a single texture is elegant. On the other hand, I'm not sure whether one PNG is always going to be more efficient than two JPGs. Or whether this packing scheme will remain useful with `KHR_texture_basisu`.

sebavan: Yup I totally agree, and the main intent for us is to reduce the overall number of samplers due to the small allowed number on some devices.

sebavan:
Live example in the playground: https://www.babylonjs-playground.com/frame.html#BNIZX6#1

The model is available here: https://github.com/BabylonJS/MeshesLibrary/tree/master/Sheen so you can see the setup.

extensions/2.0/Khronos/KHR_materials_sheen/schema/glTF.KHR_materials_sheen.schema.json
lexaknyazev: For consistency, please replace "Gamma" with "sRGB transfer function".

sebavan: thanks @lexaknyazev I have just updated it

emackey:
Looks great.  One comment, this uses the word "Factor" in a way that is not consistent with glTF's many other factors, all of which multiply with their respective textures.  Textures and factors always come in pairs multiplied together.  So, I would suggest a rename:

* `sheenFactor` -> rename to `sheenIntensity`
* `sheenColor` -> rename to `sheenColorFactor`
* `sheenTexture` -> rename to `sheenColorTexture` if this is only color information

On second reading it's not clear if `sheenTexture` actually contains intensity information as well as color information.  If so, a different set of names could be used:

* `sheenFactor` -> rename to `sheenIntensityFactor`
* `sheenColor` -> rename to `sheenColorFactor`
* `sheenTexture` -> keep the name, include intensity in alpha channel

Alternately the texture could be split up into `sheenColorTexture` and `sheenIntensityTexture`.

Thoughts?

proog128:
> The sheen roughness is identical to the material roughness to simplify the configuration. `sheenRoughness = materialRoughness`

I am wondering why you decided to couple sheen and microfacet roughness. In my experience, sharing parameters is difficult if their visual effect is different, it breaks layering/parameter blending (see Sec. 5.7 in [Disney paper](https://disney-animation.s3.amazonaws.com/library/s2012_pbs_disney_brdf_notes_v2.pdf)). Assuming there are two materials: material A has sheenIntensity=1 and roughness=0.2, material B has sheenIntensity=0 and roughness=0.5. If an artist now creates a mask that layers material A on top of material B, for example to have a 30% semi-transparent fabric on top of a metal, the roughness behaves weird. The blended material has sheen=0.3*1 + 0.7*0=0.3 (ok) and roughness=0.3*0.2 + 0.7*0.5=0.41, shared by the sheen BRDF and the microfacet BRDF (not ok). This doesn't behave similar to BRDF blending (which is the ground-truth), in which case you would end up with 30% sheen with roughness 0.2 and 70% microfacet with roughness 0.5.

> f_sheen = lightColor * sheenTerm * PI * ndotl

`lightColor` and `ndotl` shouldn't be part of the BRDF. Same for `PI`, I'd guess, not sure where it comes from.

> sheenTerm = sheenColor * sheenIntensity * sheenDistribution * sheenVisibility;

This is different from Imageworks ["Production Friendly Microfacet Sheen BRDF"](http://www.aconty.com/pdf/s2017_pbs_imageworks_sheen.pdf). Did you deviate from their solution to make it simpler to compute? What I like a lot about the Imageworks sheen is the fact that it fits nicely into the microfacet framework. They also provide a (not so cheap to evaluate) shadowing term that is properly normalized. In the end, this gives an energy conserving and reciprocal BRDF (except their terminator softening), whereas the sheen BRDF proposed here is not energy conserving and, therefore, not well-suited for GI rendering.

> specularity = mix(metallicF0, baseColor, metallic)
> [...]
> final = f_emissive + f_diffuse + f_specular + (1 - reflectance) * f_sheen

I couldn't find a description of `metallicF0`. Maybe I misunderstood the weighting of diffuse/specular and sheen, and the following is wrong: This combination is simple and energy conserving. On the other hand, it may be confusing to users because the strength of the sheen effect is connected to the color of the base material. The directional-dependent albedo-scaling technique described in the Imagework paper doesn't have this problem, but is a bit more expensive to evaluate.

Maybe we could tweak the equations in this extension a bit to make it more raytracing-friendly (like the microfacet-based Imageworks model) and define it as the reference. This would allow rasterizers to do whatever approximation is best for the intended target platform and performance constraints. We plan to soon release an adaptation of the Imageworks model as part of the [Enterprise PBR Material](https://github.com/DassaultSystemes-Technology/EnterprisePBRShadingModel).

sebavan:
You are totally right about layering, we were only thinking about reducing the number of parameters in our implementation to help with deferred renderers for instance. 

> lightColor and ndotl shouldn't be part of the BRDF. Same for PI, I'd guess, not sure where it comes from.

Yes, totally, this is just the way we are using it in our code but could benefit from more clarity. I ll update it on Monday.

> This is different from Imageworks "Production Friendly Microfacet Sheen BRDF". Did you deviate from their solution to make it simpler to compute?

Yes, the idea was too make it faster too compute and easier too integrate in real time rendering engines as detailed in the Filament documentation.

> I couldn't find a description of metallicF0

This is basically the default F0 being 4% in metallic workflow but customizable through the proposed specular extension.

> On the other hand, it may be confusing to users because the strength of the sheen effect is connected to the color of the base material. The directional-dependent albedo-scaling technique described in the Imagework paper doesn't have this problem, but is a bit more expensive to evaluate.

Yes, this is currently the main issue (aside of layering as you described above) and usually forces us to darken a lot the base color. 

> Maybe we could tweak the equations in this extension a bit to make it more raytracing-friendly (like the microfacet-based Imageworks model) and define it as the reference. This would allow rasterizers to do whatever approximation is best for the intended target platform and performance constraints. We plan to soon release an adaptation of the Imageworks model as part of the Enterprise PBR Material.

Would be nice if the chosen model was also really efficient for real time engines. Here, one of the nice trick by not relying on the fresnel for instance is the ability to store precomputed values in the BRDF lookup texture for the IBL part.

bghgary:
See https://github.com/KhronosGroup/glTF/pull/1677#issuecomment-545526435

We should do the same to this extension.

proog128:
There is now a first draft (work in progress) for integrating Imagework's sheen into [Enterprise PBR version 2021x](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2021x.md.html#components/sheen). It adds two parameters in addition to `sheen`: `sheenColor` and `sheenRoughness`. Layering uses albedo-scaling as proposed in the paper which is energy conserving and avoids darkening. There is no Fresnel term to keep it simpler.

Optimizations for real-time rendering are still to be done. Using the Ashikhmin visibility term might be a good approximation, the error needs to be checked. Layering is an open issue.

sebavan:
About the factor renaming, what should be the final naming so that I could update the PR ?

I have seen asked to rename :

sheenFactor -> sheenIntensityFactor
sheenColor -> sheenColorFactor

is that all the changes we want ?

About the texture, I would like to keep it only one if possible merging color and intensity, do you think it makes sense ?

At least this part would not change even if we adapt the formula. 

Also we should choose if the roughness needs to be a parameter as well or if it makes sense to reuse the material one  to limit the number of params and be more Deferred rendering friendly ?

bghgary:
> renaming

My opinion (assuming we use one texture):
- `intensityFactor`
- `colorFactor`
- `colorIntensityTexture`

No need to keep putting sheen in the name since it's already under the sheen extension.

```json
    "materials": [
        {
            "extensions": {
                "KHR_materials_sheen": {
                    "intensityFactor": 0.9,
                    "colorFactor": [
                        0.0235294117647059,
                        0.0352941176470588,
                        0.2745098039215686
                    ],
                    "colorIntensityTexture": {
                        "index": 0
                    }
                }
            }
        }
    ]
```

emackey:
+1 for @bghgary's suggestion.

donmccurdy:
> About the texture, I would like to keep it only one if possible merging color and intensity, do you think it makes sense? 

This will prevent the use of JPG textures with this extension. Is that OK? I would like to think that Basis Universal will just become best practice for all of this, but I don't have a good sense for the pros/cons of packing these textures together with Basis.

sebavan:
Ok so I ll first update to Gary's suggestion and depending on the outcome for the rest I ll go ahead an fix accordingly.

I guess the texture packing discussion is probably bigger than only this extension :-)

I will try to have it updated for Monday.

donmccurdy:
> I guess the texture packing discussion is probably bigger than only this extension.

Yes, good point. Happy to move that discussion elsewhere until we can come up with some best practice.

sebavan:
All factors have been renamed according to @bghgary 's proposal and Babylon.js has been updated to reflect the changes as well as the babylon sample model.

We can now focus on the need for roughness and equations to use :-)

sebavan:
@proog128, I ll try soon the albedo scaling technique as despite the extra lookup it looks quite nice.

How is the rendering quality of your analytical approach in the end (thinking of smaller devices fallbacks) ?

sebavan:
Also It wonder what do you think about not splitting the roughness parameter and reusing the base layer one ?

proog128:
> How is the rendering quality of your analytical approach if any (thinking of smaller devices fallbacks) ?

We plan to use the albedo-scaling with a precomputed texture in the WebGL renderer (probably with just one look-up: for VdotN). We do the same for mixing diffuse and specular on the base layer and performance was good enough. In the beginning we tried to use an analytic approximation instead of a texture for diffuse/specular, but there was no performance benefit.

> Also It wonder what do you think about not splitting the roughness parameter and reusing the base layer one ?

Feedback from our artists was more in favor of keeping the separation because of its flexibility. It was one of the reasons for us to extend the sheen model.

In the following example, we have a material with a small sheen roughness (step, grey gradient at grazing angles), but high reflection roughness (makes the surface soft).
example

For skin or plants it's also possible to texture the reflection roughness (base layer) due to irregularities at the surface, but you still want to have uniform brightening at the rim (coming from small, uniform but perpendicular fibers on the surface).

romainguy:
@proog128 An analytical approximation is useful not necessarily for performance but to avoid using more samplers, which are limited in numbers on mobile

I agree there definitely should be a separate sheen roughness parameter.

proog128:
Quick and dirty analytical approximation:

```
E = 0.65584461 * pow(1.0-cos_theta, 3) + 1.0 / (4.16526551 + exp(-7.97291361*roughness+6.33516894))
```

The error is around 3% (RMSE), which is quite high, but I didn't spend a lot of time on it.

romainguy:
@proog128 Thanks, I was going to try (again) to come up with one. I'm also tempted to just put that lookup table in an extra channel of the DFG LUT (we already use the blue channel for the cloth/sheen BRDF). It would be higher resolution than needed but at least it would save a sampler.

sebavan:
@romainguy @proog128 I ll go ahead and add the separate roughness parameter over the week end as well as update the tbd in the spec with @bghgary's advices.

@proog128, do you have a sample implementation code in GLSL in order to compute the required texture values, and about the albedo scaling (this might speed up a bit the implementation) ?

proog128:
@sebavan You mean sample code to compute the texture containing directional albedo E as given [here on slide 58](https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides_v2.pdf)? I may be able to share some Python/OpenCL code, we use it to generate [this texture](https://github.com/DassaultSystemes-Technology/EnterprisePBRShadingModel/blob/master/res/Sheen_E.exr). I may also be able to give you some GLSL code that shows how to use it for albedo scaling.

sebavan:
Yup this would be amazing :-)

sebavan:
It would actually help a great deal to avoid mistake in the implementation.

proog128:
Attached a Jupyter notebook that computes the texture (needs numpy and pyopencl) and a GLSL sample implementation that can be used in Disney BRDF Explorer. Besides, I noticed that a sqrt is missing in the fit above, fixed it.

[sheen.zip](https://github.com/KhronosGroup/glTF/files/4076548/sheen.zip)

sebavan:
Thanks a ton really appreciate it !!!

sebavan:
I did not forget about this one but having trouble to find time at the moment :-( sorry about the long wait I am trying to see if some babylon contributors want to tackle it. In the mean time, do not hesitate to PR over this proposal as I love the separate roughness and albedo correction ideas :-)

bhouston:
I spoke with Vladimir Koylazov of V-Ray about this specification.  He said the implementation notes were incorrect as it doesn't account for Fresnel.  As such this implementation would be non-physical.

Popov72:
> He said the implementation notes were incorrect as it doesn't account for Fresnel

I would say the implementation notes are correct as they are in line with what the people who created those formulas published:
* http://www.aconty.com/pdf/s2017_pbs_imageworks_sheen.pdf
* https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides_v2.pdf see slide 106

They explicitely states there's no Fresnel ("Sheen energy conservation: we store the albedo of the sheen lobe in a 16 × 16 2D table indexed by roughness and incident angle, since **we don’t take Fresnel into account**").

They also state "With our `min{(1 − E(µo)),(1 − E(µi))}` scaling we avoid energy gain and suffer only subtle energy loss.".

So, they indeed acknowledge there's is a small energy loss (you can see the white furnace test in slide 106) - maybe the enterprise PBR spec should indicate this somewhere (?) - but should that disqualify it for being the KHR sheen extension implementation?

Popov72:
> the implementation notes

As it was not obvious in my previous comment, I was speaking here about the enterprise PBR document, not the current specification notes, if that's what Vladimir Koylazov commented about.

Regarding the specification for this extension, I think a number of things remains to be cleared up. Will try to do it in following posts.

(note that I'm working with @sebavan about this)

Popov72:
Here's a sum up of what I think we still have to discuss / validate:
* separate `roughness` parameter. I think we have agreed that having a separate roughness parameter for sheen is the way to go
* remove `f_sheen = lightColor * sheenTerm * PI * ndotl` from the doc as this one is up to the implementer. Or at least don't relate the formula to `f_sheen` but simply show it as a possible implementation for calculating the final contribution of the sheen layer (I too don't understand where `PI` is coming from ?).
* for the visibility term:
  * add a note that using Ashkimin will make the implementation not energy conserving.
  * add the shadowing term as described in the original paper and indicate that using this term is heavier on computation but makes the implementation energy conserving
  * it's up to the implementer to use one or the other, depending on the constraints
* replace the last part about the layering by the albedo-scaling solution for layering the sheen brdf over the base brdf. Link to the pre-computed texture as provided by @proog128. @proog128, you said: "We plan to use the albedo-scaling with a precomputed texture in the WebGL renderer (probably with just one look-up: for VdotN)". Have you been able to perform some testings and evaluate the impact on the enery conservation side by not doing the second lookup with LdotN? Maybe this specification should indicate that the proper scale factor is `min(1-E(VdotN, roughness), 1-E(LdotN, roughness))` but can be replaced by `1-E(VdotN, roughness)` to save some performances (E(...) being retrieved through a lookup in the texture)?

Let me know if I missed something.

proog128:
Thanks for the summary and sorry for the late reply. Yes, we did some testing regarding simplification of the albedo scaling.

Visual output is very close, so it is not an issue to use the simplified version of it in WebGL and the more complex in ray tracing.

Simple:
fabric_estevezkulla_scaling_simple

Full:
fabric_estevezkulla_scaling_full

Both images are rendered with a ray tracer, so that all other differences are excluded in this comparison.

Regarding energy conservation: As expected, the simplified version does not work well in bidirectional path tracing, as it is not reciprocal. The white furnace test shows that (in our case) the simplified version is not energy conserving.

Moreover, we compared visual output for Ashikhmin and Estevez-Kulla G. As before, consistency is good, so it is possible to use the one in WebGL, the other in ray tracing.

Ashikhmin:

fabric_ashikhmin

Estevez-Kulla:

fabric_estevezkulla

Popov72:
@proog128 Thanks for your answers and for the pictures!

I think I will update the document with my proposed changes, that will be easier for people to comment.

I have two more questions for you if you don't mind:
* regarding the albedo scaling factor (`min(1-Ewi, 1-Ewo)`), am I right when I say it should be applied to all contributions calculated so far, meaning the diffuse / specular terms of both the direct and indirect lighting? So something like `(fd_direct+fd_indirect+fs_direct+fs_indirect)*albedo_scaling + f_sheen`
* for the IBL part of the sheen calculation, it seems the lookup table can be reused to calculate the environment reflectance, meaning `sheenEnvReflectance = sheenColor * albedo_lut(NdotV, roughness)`?

EliphasNUIT:
Hello Popov, 
I have been summoned by @proog128 as reinforcement. To answer your questions:
- Not exactly, in the model you are going to sum for each light, if you have n direct lights and one indirect light, your complete equation would look like 
CodeCogsEqn
- Could you clarify what do you mean by `albedo_lut`? Is it the directional albedo? If so sheenEnvReflectance should not depend on it and look something like, assuming split sum approximation, `sheenEnvReflectance = sheenColor * sheenIntensity * SampleSheenEnvMap(sheenRoughness) * SampleSheenGVis(sheenRoughness)`

Regarding using only the `VdotN `part for layering, it was only recently implemented on our side. We still need to properly quantify energy differences but we are confident that it is going to be consistent enough visually. 

I'm working on a PR on the DSPBR repo right now (https://github.com/DassaultSystemes-Technology/EnterprisePBRShadingModel/pull/19), I do intend to add everything @proog128 have showed image wise + energy stuff for the sheen. After that I intend to work on another one, assuming it is acceptable, where we alternatively only use one lookup for sheen layering but also diffuse/specular energy conservation/preservation formulas.

Popov72:
Hi @EliphasNUIT,

Thanks a lot for your answers!

Regarding albedo scaling, if we only use `1 - E(VdotN)` as the scaling factor, I think it is the same for both direct and indirect lights?

I was speaking about the directional albedo lut indeed. I wasn't including the sheen radiance in my formula (as the mult is done later in my code) and `sheenColor` is already including `sheenIntensity`, so I'm ok with your formula:
* `SampleSheenEnvMap` is the prefiltered env cube map lookup
* I think `SampleSheenGVis` is:
image
?
In 6.2.3 in the enterprisePBR doc, directional albedo is defined as: 
image
I think it is the same thing, as Mb is the sheen BRDF?

EliphasNUIT:
Indeed, if you limit yourself only to `VdotN` then the formula is greatly simplified, as the same value can be computed/sampled once and for all in your code and be used for any kind of lights, direct or indirect.

Regarding the 2nd point, I was going for a traditional split sum approximation where we have that pre-filtered env map lookup but also another lookup, traditionally called GF textures in GGX, that, in our case, contains pre-computed values for the `VCharlie` terms or `VAshikhmin` if you are going for a simpler G term in your model. If you decide to drop `LdotN` from your albedo scaling formula then the directional albedo computation does not intervene anywhere in your sheen pre-filtering operations as this is where you'd get access to `L` vectors.

The directional albedo is only used for layering purposes, so it shouldn't appear in the sheen reflectance computation formula: `finalReflectance = (diffuseReflectance + specularReflectance) * (1.0 - E(VdotN)) + sheenReflectance` (ignoring `LdotN`)

Popov72:
If not mistaken, the split sum approximation is:
image
The first integral is the lookup in the prefiltered cube env map and the second integral is 
image if we use Charlie for both D and V.

To evaluate this second integral, one can simply do a lookup in the directional albedo table with NdotV + roughness, as this table is built using 
image which is the same integral than above as Mb = Dcharlie * Vcharlie.
Or did I miss something?

As you say you store only Vcharlie (or Vashikhmin) in the table, then it can't be used to evaluate this second integral indeed.

> The directional albedo is only used for layering purposes, so it shouldn't appear in the sheen reflectance computation formula

I understand that albedo scaling has been introduced for layering purpose only, but as I see it it is the same integral we need to compute as part of the IBL lighting process?

Thanks for confirming one can use the same albedo scaling for both direct and indirect lighting if using only the `1 - E(VdotN)` part.

EliphasNUIT:
Oh sorry, I was misunderstanding the situation, I did not get that you were trying to optimize the computation. 
Yes, absolutely, you can do that, the formula is the same.

Sorry for the confusion, I generally proceed with importance sampling when possible so the D term is "simplified", with only the Visibility term and the half vector transformation Jacobien remaining so I got used to calling it GVis texture or GFVis texture when the fresnel is also present. I'll call it BRDF texture in our future exchanges, to avoid that confusion.

Popov72:
No problems, I understand that offline raytracers are not in the same league than realtime renderers!

I'm going to update the GLTF doc with everything discussed so far, hopefully we should be able to come to an agreement about the spec.

Popov72:
@proog128 Can I reuse your picture in the specification:
70046988-75f6b880-15c7-11ea-9147-263f47390654
?

proog128:
Yes, you can use it.

Popov72:
> Yes, you can use it.

Thanks!

sebavan:
Just updated the PR with @Popov72 new version to not lose track of the PR discussion. As discussed, renaming as been done, roughness is now splitted and albedo scaling technique has been described.

It looks overall pretty good and we ll have the implementation following in Babylon pretty soon.

What would now be missing to go one step further with the PR ?

Popov72:
Here's the implementation in Babylon:

https://popov72.github.io/BabylonDev/showSample.html?sample=sheen

The red material is best seen with the "Studio" environement because it is brighter.

We use the Ashikhmin visibility term instead of the more computational intensive one described in Estevez & Kulla. Also, for albedo scaling, we only use the `1-E(NdotV)` term.

aidinabedi:
Hello @Popov72,

Following formulas in the spec (so far) seem to indicate that `intensity` is just a scalar factor for `color`:
- `sheenTerm = sheenColor * sheenIntensity * sheenDistribution * sheenVisibility;`
- `albedoScaling = 1.0 - sheenIntensity * max3(sheenColor) * E(VdotN)`

If so, then why separate them?

This goes back to a question asked by @sebavan:

> About the texture, I would like to keep it only one if possible merging color and intensity, do you think it makes sense ?

aidinabedi:
Another probably dumb question.

Looking at sheen distribution in the spec, I notice that instead of using `roughness` directly, it is substituted by the term `alphaG` (which is equal to `roughness` squared):
```glsl
alphaG = sheenRoughness * sheenRoughness
invR = 1 / alphaG
cos2h = NdotH * NdotH
sin2h = 1 - cos2h
sheenDistribution = (2 + invR) * pow(sin2h, invR * 0.5) / (2 * PI);
``` 

However, looking at the source material, I can't seem to find any mention of the term or squared. Neither in the [Production Friendly Microfacet Sheen BRDF](https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_sheen.pdf) paper:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;image

Nor in the [Cloth Shading](https://knarkowicz.wordpress.com/2018/01/04/cloth-shading/) shader implementation:
```glsl
    float CharlieD(float roughness, float ndoth)
    {
        float rcpR  = 1. / roughness;
        float cos2h = ndoth * ndoth;
        float sin2h = 1. - cos2h;
        return (2. + rcpR) * pow(sin2h, rcpR * .5) / (2. * PI);
    }
```

In fact, the only place I can find a `roughness` squared, is in the older distribution function
[Ashikmin et al. 2000], which is not from ImageWorks and not otherwise suggested in the current specs:
```glsl
    float AshikhminD(float roughness, float ndoth)
    {
        float m2    = roughness * roughness;
        float cos2h = ndoth * ndoth;
        float sin2h = 1. - cos2h;
        float sin4h = sin2h * sin2h;
        return (sin4h + 4. * exp(-cos2h / (sin2h * m2))) / (PI * (1. + 4. * m2) * sin4h);
    }
```

Please adivse.

sebavan:
About the separate intensity, it is because it might help transitioning from sheen to no sheen without loosing the color set in the first place. I guess we have the same in clear coat if I copy pasted it correctly :-)

sebavan:
About the roughness, we are using a perceptual roughness value in the setup to be consistent with the rest of the material it still varies from 0 to 1 but in the same way the roughness would do in the main part of the material.

I guess it is a nice question for the thread ?

proog128:
My guess is that the distribution proposed by Imageworks already accounts for perceptually linear behavior, therefore this correction is not needed. Comparison:

Linear and squared roughness for specular:
specular_roughness_linear
specular_roughness_squared

Linear and squared roughness for sheen:
sheen_roughness_linear
sheen_roughness_squared

While the squared roughness looks more "linear" for specular, it's not so clear in case of sheen. I'd prefer to stick to the source and use the linear roughness.

The separate sheen intensity could be useful to reduce texture size. If a material has a constant sheen color, but spatially-varying intensity, only a 1-channel texture is needed. On the other hand, if we combine sheen intensity and sheen color into one 3-channel parameter, we could pack it together with a 1-channel sheen roughness texture into a 4-channel texture (RGB = sheen color/intensity, A = sheen roughness).

aidinabedi:
Regarding packing, I suggest keeping things simple and consistent. Merging intensity and color into a single 3-channel parameter simplifies implementations and keeps behavior consistency with similar parameters like `specularFactor` (in [`KHR_materials_pbrSpecularGlossiness`](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_pbrSpecularGlossiness#extending-materials)), which also represents both intensity and color.

proog128:
Another thought with regard to merging intensity and color: the base layer weight (`albedoScaling`) currently respects `sheenIntensity` and `sheenColor`.

`albedoScaling = 1.0 - sheenIntensity * max3(sheenColor) * E(VdotN)`

As a result, it is possible to merge sheenIntensity and sheenColor.

However, if we redefine `albedoScaling` in the following way:

`albedoScaling = 1.0 - sheenIntensity * E(VdotN)`

we would gain one more degree of freedom. It allows artists to make the sheen darker without adding more of the base layer. Here is an example:

This is pure diffuse without sheen:
sheen_1_diffuse

Let's now set sheenIntensity=1 and sheenColor=(1,0,0). A red sheen effect at the edges becomes visible, taking energy away from the base layer.
sheen_2_red_sheen

Now let's make the sheenColor darker in the current approach: sheenColor=(0.3,0,0). The sheen effect becomes less intense, the remaining energy is filled up with the base layer.
sheen_3_dark_red_sheen

Compare this to the proposal in this comment. As before, the sheen effect becomes less intense, but the base layer stays dark too (because `sheenIntensity` is still 1).
sheen_4_dark_red_sheen_intensity_only

No idea if this is useful, but it might be a reason to have sheenIntensity and sheenColor separate.

sebavan:
@proog128 I wonder if it is not going to far, and in order to keep things simpler I tend to agree with @aidinabedi on merging both.

Would that work for everybody ???

MiiBond:
Did we come to a conclusion about merging intensity and colour into the RGB channels of the texture? I'd vote for this if it lets us put roughness texture data into the alpha channel. Otherwise, I'd like to add a separate roughness texture.

emackey:
For what it's worth, I'd prefer a potential sheen roughness texture to explicitly use the G channel, not alpha.  This way the ORM map can be re-used in this slot in what I think might be the common case where sheen roughness matches base material roughness.  Does that sound reasonable?

Also +1 for combining color and intensity.

MiiBond:
@emackey I wonder if that would be possible too. I've put the question to a few people here and I'm trying to track down some example assets that would use sheen. I don't believe that Substance or any Adobe product supports it yet (though, it's coming).
Another possibility (that I think the Disney model might use?) is to make colour a uniform and then use a greyscale tint map to blend between that colour and white. I think that would cover the most common types of coloured sheen and it would allow tint to be stored in a single channel. Thoughts?

MiiBond:
Another thought... and this isn't specific for sheen but it would be nice to somehow allow separate `textureInfo` objects for each of these maps. I'm often running into an issue with having to share the `KHR_texture_transforms` between packed textures. In Adobe Dimension, you can apply different transforms to metallic and roughness, for example. Or baseColor and opacity. I like how `occlusion` is defined as its own `textureInfo` object but can point to the same sampler as metal-rough as an optimization. 
I know this potentially allows an asset to define lots of unique textures which is an issue for WebGL (and mobile, I assume). It would be nice to allow the separate transforms while enforcing that they reference the same sampler... not sure how that would work.

donmccurdy:
> It would be nice to allow the separate transforms while enforcing that they reference the same sampler...

This amounts to requiring the client to support the use of separate textures for the two roles, I think. At least if the client has a developer-facing API or UI (i.e. isn't solely dedicated to rendering glTF models), the transforms are associated with texture slots and not with a fraction of a texture slot.

If ability to use separate transforms is important, I think we should allow the textures to be separate.

proog128:
I don't think there are many cases in which the roughness of base and sheen are identical. Maybe you can compensate the differences via the constant scale factor sometimes, but I am skeptical... Similar to the specular extension, I would prefer to keep it as simple as possible: one RGBA texture with sheen color and sheen roughness in the alpha channel. There are so many cases in which materials could be optimized easily if there was a generic swizzling extension, so maybe that should be our goal?

sebavan:
Totally aligned with @proog128 on this one. I ll reshape the extension like this today. 

And lets worry about a more generic swizzling/packing solution for later :-)

MiiBond:
As discussed in the glTF call today, can we rename the properties back to having `sheen` as a prefix on each so that we can keep the parameter names for a possible glTF 3.0 (or 2.x) where this extension goes away and the params become part of the core spec.
On a similar note, we decided that the default value for the weight parameter for these extensions should be `0` like it is with `KHR_materials_clearcoat`. The logic being that the extension shouldn't automatically turn on functionality without parameters defined.

sebavan:
ok sounds good I ll do that today :-)

MiiBond:
I'd suggest:
`sheenFactor`
`sheenRoughnessFactor`
`sheenColorFactor`
`sheenTexture`

sebavan:
Ok spec updated:

Now one last question for everyone then. As we do not have the factor anymore, merged within color, I need to default to a black color to disable by default like clear coat.

So if there is a texture but no color defined, unlike other part of the spec the factor would not multiply by default with the texture (same goes for the default roughness value).

@proog128 , @emackey  @bghgary  and @donmccurdy  any preferences ?

bghgary:
I think this is okay. It is already the case for emissive whose factors default to [0, 0, 0].

MiiBond:
If we remove `sheenIntensity` completely, maybe `sheenColorFactor` should just be renamed `sheenFactor`? That would also match how emission is handled.

sebavan:
not sure cause we still have the sheenRoughnessFactor ?

MiiBond:
Good point. I don't have a strong feeling about it either way.

sebavan:
Perfect let s keep it like this and wait to see if there are more comments else we should try to get other implementations ?

MiiBond:
Since we're now defaulting to extensions being disabled when defined with no properties, is it fair to, instead set `sheenColorFactor` to be required?

Same goes for other extensions. e.g. `transmissionFactor` would be required in the transmission extension.

sebavan:
No Preferences on my side I can change it pretty easily :-) @bghgary any feedback ?

donmccurdy:
I think the motivation behind `extensions being disabled when defined with no properties` is really that extensions should be disabled with _default_ properties, where omission of a property is exactly the same as specifying default values for that property. None of these properties would be required if they were in the core spec, so none should be required in these extensions. Instead:

- transmissionFactor: defaults to 0 when not specifiied, which disables the transmission effect entirely.
- sheenColorFactor: defaults to [0,0,0] when not specified, which disables the sheen effect entirely.

sebavan:
Thanks @donmccurdy makes total sense in prevision of a next core revision.

sebavan:
So I am glad to think the extension is ready for stronger review and other implementations ? Any thoughts ?

extensions/2.0/Khronos/KHR_materials_sheen/README.md
```diff
@@ -0,0 +1,158 @@
+# KHR\_materials\_sheen
+
+## Khronos 3D Formats Working Group
```
emackey, line 3: Please change this line to `## Contributors` and add a contributors list similar to the one in the [XMP extension README](https://github.com/KhronosGroup/glTF/blob/8832254c547c642cddbd4e4b1e70aedd4eb1434e/extensions/2.0/Khronos/KHR_xmp/README.md).

Also, please copy the Khronos full copyright statement appendix from that extension.  Thanks!

extensions/2.0/Khronos/KHR_materials_sheen/schema/glTF.KHR_materials_sheen.schema.json
```diff
@@ -0,0 +1,36 @@
...
+                "type": "number",
+                "minimum": 0.0,
+                "maximum": 1.0
+            },
+            "minItems": 3,
+            "maxItems": 3,
+            "description": "Color of the sheen layer (in linear space).",
+            "default": [0, 0, 0]
+        },
+        "sheenTexture": {
```
emackey, line 20: Would it be OK to rename `sheenTexture` to `sheenColorTexture`?  That way we keep the factor+texture naming convention going.

extensions/2.0/Khronos/KHR_materials_sheen/schema/glTF.KHR_materials_sheen.schema.json
```diff
@@ -0,0 +1,36 @@
...
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "The sheen layer texture.",
+            "gltf_detailedDescription": "The sheen layer texture. Stored in channel RGB the sheen color in sRGB transfer function and in A the linear factor."
```
emackey, line 23: I think this description doesn't match the README anymore.

extensions/2.0/Khronos/KHR_materials_sheen/schema/glTF.KHR_materials_sheen.schema.json
emackey: Nevermind, roughness is now bundled here.

MiiBond:
Have we discussed how this extension layers with other extensions like clearcoat? It should definitely be specified in the definition of these extensions to ensure consistent implementations.
I think, typically, clearcoat is layered on top of sheen but I wonder if there might be use cases for the other way around (e.g. dust on top of a clearcoat).

sebavan:
Always below and for dust on top I guess clear coat roughness and normals could do ?

rsahlin:
> I think, typically, clearcoat is layered on top of sheen but I wonder if there might be use cases for the other way around (e.g. dust on top of a clearcoat).

I would see sheen as a surface parameter (much like metallic/roughness) and clearcoat as a layer.
This way sheen can be a parameter to the base layer and to the clearcoat layer.

extensions/2.0/Khronos/KHR_materials_sheen/README.md
```diff
@@ -0,0 +1,226 @@
...
+                }
+            }
+        }
+    ]
+}
+```
+
+### Sheen
+
+All implementations should use the same calculations for the BRDF inputs. Implementations of the BRDF itself can vary based on device performance and resource constraints. See [appendix](/specification/2.0/README.md#appendix-b-brdf-implementation) for more details on the BRDF calculations.
```
rsahlin, line 57: I think it would be beneficial to mark the sheen (BRDF) implementation as non-normative - this goes along with the glTF 2.0 BRDF implementation.
Implementers are free to choose some other formula for the BRDF but this is the one we suggest for realtime rasterizers and will be implemented in the Sample-Viewer.

sebavan: We are using the exact same phrasing than the other extensions which should be ok ??? or we might need to revisit the one on clear coat and so on as well. I do think it is actually what this means ???

bsdorra:
> > I think, typically, clearcoat is layered on top of sheen but I wonder if there might be use cases for the other way around (e.g. dust on top of a clearcoat).
> 
> I would see sheen as a surface parameter (much like metallic/roughness) and clearcoat as a layer.
> This way sheen can be a parameter to the base layer and to the clearcoat layer.

Is there a specific use-case where this is needed? Otherwise, I'd agree with @sebavan and would suggest to stick to the "sheen as base layer property" concept. I'm afraid, that making it a generic layer property would increase implementation complexity for no significant reason.

extensions/2.0/Khronos/KHR_materials_sheen/README.md
bsdorra: I understand the current phrasing as exactly what you're saying @rsahlin. The BRDF input computations based on the parameters need to comply to the spec, which is crucial for consistency. But the implementation itself is non-normative and may vary depending on the constraints. (this might also play a significant role in consistency, but is inevitable.)

emackey:
> making it a generic layer property would increase implementation complexity for no significant reason.

Yes.  At some future point we may have to address layered materials with a new layering extension of some kind.  Sheen is not the correct scope for that work.

extensions/2.0/Khronos/KHR_materials_sheen/README.md
rsahlin: @sebavan In the referenced glTF documentation (Appendix B) linked by clearcoat it states:
> This section is non-normative.

This goes for the whole BRDF implementation - I think that all parameters added by extensions must follow that. 
So yes - I think we must update clearcoat to match.

For glTF to incease adoption as a  file transmission format I believe that it is very important to keep the mindset that we are defining a set of parameters.
Exactly how these parameters are used is up to implementations.
Mandating how the values are used is up to certification, whose purpose is to enable consistensy and predictability in viewers.

emackey: So the parameter list is normative, and the suggested implementation with formulas are non-normative, I think that sounds correct.  This means the parameter list will need to appear in a different section from the non-normative implementation notes and formulas, which means a change to the wording in clearcoat.

proog128: When writing extensions (#1718, #1719, #1825, #1726) I separated the list of parameters and their behavior, described with text and a set of images, from the implementation part. While the implementation is non-normative, the behavior and intent of parameters is not. Behavior and intent includes how factor and texture are combined, how the effect blends with other effects (layering, mixing, Fresnel, etc.), a description of the physical effect we want to simulate, and so on.

In other words:
* Information you need to create an MDL or OSL shader should be part of the normative section.
* Information you need to create a renderer for this MDL shader (i.e., implement the MDL/OSL nodes/functions in a renderer with functions like sample/eval/pdf/...) are part of the non-normative section.

This ensures that the material renders consistently everywhere (same parameters with same behavior, for example base color changes color of diffuse BRDF and microfacet BTDF), but still gives enough room for custom, renderer-specific optimizations (for example the choice of BSDF for global illumination vs. real-time, like Estevez-Kulla vs. Ashikhmin visibility term).

In the future I hope we can adopt a common representation for the normative part which is used in all extensions. Describing the behavior of paramters with text may lead to misunderstandings. I drafted this in #1717 and #1738. Please have a look at especially the rewrite of the clearcoat extension in #1717. I am planning to update this for all extensions as soon as they are merged.

emackey: I opened #1836 to update clearcoat.

extensions/2.0/Khronos/KHR_materials_sheen/README.md
```diff
@@ -0,0 +1,226 @@
...
+Where this Specification includes normative references to external documents, only the
+specifically identified sections and functionality of those external documents are in
+Scope. Requirements defined by external documents not created by Khronos may contain
+contributions from non-members of Khronos not covered by the Khronos Intellectual
+Property Rights Policy.
+
+In particular, the referenced paragraphs of ISO 16684-1:2019 are normative and so are
+INCLUDED in the Scope of this Specification and may contain contributions from non-members
+of Khronos that are not covered by the Khronos Intellectual Property Rights Policy.
+The links to https://github.com/adobe/xmp-docs are purely informative and so are EXCLUDED
```
emackey, line 186: Totally my fault, I knew XMP had updated trademarks, but I didn't realize XMP had its own paragraph snuck in here.  Please remove this paragraph from Sheen, from lines 183 to 189 in this revision.  The result should match the new version in clearcoat in #1836.

extensions/2.0/Khronos/KHR_materials_sheen/README.md
sebavan: Done as proposed on clear cloat in #1836

emackey:
The spec document here uses shader code to define the behavior.  Can we add a more formal BSDF mathematical definition, similar to KHR_materials_transmission and KHR_materials_volume, etc?  It's fine to also keep the shader code by way of example.

sebavan:
Yup we asked offline @proog128 for more guidance on the formula to add as details here :-) or if you know the one you want to see, do not hesitate to point me to them and I ll add it.

emackey:
@sebavan There was a suggested patch from @proog128 on the mailing list, did you see it?  Let me know if it should be delivered by some other vehicle.

sebavan:
Sorry for the delay, I am currently in vacation and just noticed the message, I have just pushed the patch in.

Thanks a ton @proog128 !!!

Just a quick question roughness squared has been replaced by roughness here. @proog128 is it intentional ?

proog128:
@sebavan I am sorry, I forgot to mention this! It was intentional, yes, because it is not mentioned in the linked source article from Imageworks. We should either provide a good reason for squaring and explain it in the text, or keep it as it is. I just noticed that this was already discussed here in the comment thread in April including sample renderings. I also had a look at the [Filament cloth model](https://google.github.io/filament/Filament.md.html#materialsystem/clothmodel), [Standard Surface](https://github.com/Autodesk/standard-surface/blob/master/reference/standard_surface.osl#L91) and [Enterprise PBR](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2021x.md.html#components/sheen), all of these materials stick to the source and do not square the roughness (as far as I can tell).

romainguy:
Actually we do in Filament; roughness is the “perceptual roughness” (user input) squared.

proog128:
@romainguy OK, thanks for the clarification, it wasn't clear to me from the documentation. There is a term for squared roughness in D_Ashikhmin (Listing 17), but not in D_Charlie (Listing 18). So I guess the remapping from perceptualRoughness to roughness described in Section 4.8.3.3 does not only apply to GGX roughness, but also to sheen roughness; D_Ashikhmin and D_Charlie are called with the squared roughness, so D_Ashikhmin actually has a2=perceptualRoughness^4 and D_Charlie has invAlpha=1/(perceptualRoughness^2).

proog128:
@romainguy Hmm, but it still leaves the question open why you square it? In case of GGX it's clear, it's obvious that the squared roughness is perceptually linear. For sheen it's unfortunately not so clear, in the renderings [here](https://github.com/KhronosGroup/glTF/pull/1688#issuecomment-613900816) it's hard to say (for me) which one looks "more linear". Do you have a scene which shows it better?

romainguy:
@proog128 That's correct the "Charlie" NDF doesn't square ⍺. For squaring the roughness there is no reason to do it in the first place. It goes give a roughly perceptually linear result (although not exact) and other engines use widely different remappings. In my initial tests this remapping felt better with the cloth model than without the remapping. I agree it's not obvious but when in an editor it did feel like it gave me better control over low roughness values (your screenshots show that actually.

Note that I don't know whether the sources do a remapping either. The remapping is independent of the BRDF (Blender doesn't do remapping for its roughness parameter in the principled BSDF for instance).

sebavan:
Same for us, we actually preferred the sheen ramp with the square. Also it was more consistent with the rest of our roughness inputs as preventing some extra explanation to new joiners.

@proog128 do you want to give it a try and we chose the most pleasant one ???

proog128:
I tried the [Babylon.js playground](https://www.babylonjs-playground.com/#FEEK7G#33) and I think I understand now what you mean. It works well for low roughness values which are hard to handle without the remapping. I didn't notice it earlier, because we clamp the roughness to linear range [0.07, 1] (which is squared range [0.25, 1]). You can see that in the renderings in the comment from April, the first 3 items in the row (0.0, 0.1, 0.2) look identical. With clamping the linear roughness feels better to me, without clamping the squared roughness.

Now the interesting question is: why the clamp? We added it because we observed problems with energy conservation of the sheen BRDF if linear roughness < 0.07 in the white furnace test. See below, red pixels indicate values > 1.0, linear sheen roughness is 0.0225 (= squared roughness 0.15):

stellarium_image_718

And on a plane, here it's clearly visible at the horizon, no need to mark it with red pixels:

stellarium_image_719

I suspect that the fit for Λ given in the paper is not precise at low roughness value, but that's just a guess. Could also be a bug in our implementation :-). Can you confirm this problem?

romainguy:
Filament (and Babylon.js I think) does not use the shadowing term from "Production Friendly Microfacet Sheen BRDF". We use [a cheaper term](https://google.github.io/filament/Filament.md.html#mjx-eqn-clothSpecularBRDF).

Popov72:
Indeed Babylon uses Ashikhmin for visibility too.

There's a nice shader toy here that compares the sheen BRDFs: https://www.shadertoy.com/view/4tfBzn

sebavan:
Coming back to it, I do find the clamp weirder conceptually than the squared definition which brings consistency to the overall roughness definition ???

@proog128  could you try with the same kind of term ?

proog128:
@sebavan The Ashikhmin V term is also not energy conserving unfortunately.

I did a few tests today. First of all I tried a few visualizations to get a better sense for the perceptual linearity. The following images helped me a lot. It shows spheres with different sheen roughness values from left to right and top to bottom. The spheres at the crossings are sheen roughness inbetween the row and column, e.g., the sphere in the fifth column (0.8) and second row (0.2) has a roughness of 0.5.

Smith masking, linear roughness:
sheen_smith_interpolated_roughness_linear

Smith masking, squared roughness:
sheen_smith_interpolated_roughness_squared

I think that shows that squared roughness is the better choice, as you propose.

That leaves as with the precision issue at low roughness values. Clamping is not an option I think because values in the squared range [0,0.25] wouldn't have an effect. The Ashikhmin V term doesn't work either.

As an alternative, I tried the V-Cavities shadowing/masking term (used in the Cook-Torrance model).

V-Cavities masking, squared roughness:

sheen_vacv_interpolated_roughness_squared

What I like about it is the fact that it does not need a fit for Λ (there is no Λ) and it is much cheaper to evaluate. It's also physically-based, energy conserving etc. Downside: it's significantly brighter than Ashikhmin if roughness approaches 1. So I am not sure if we can use Ashikhmin in real-time and V-Cavities in offline rendering.

Ashikhmin masking, squared roughness:

sheen_ash_interpolated_roughness_squared

Any chance the V-Cavities masking is fast enough for rasterizer? I implemented `G2 = min(G1(wi,wm), G1(wo,wm))` from eq. 3 in [Heitz, D’Eon 2014: Importance Sampling Microfacet-Based BSDFs using the Distribution of Visible Normals](https://hal.inria.fr/hal-00996995v1/document).

EliphasNUIT:
Before talking about performances, I'm concerned by how brighter V-Cavities is, even compared to the original model. The highlights are noticeably stronger, even around 0.4 roughness. Ashikmin masking feels much closer to the original paper.

If we want to ditch that look and push for the V-Cavities route, I believe we should get an artistic feedback as well, Estevez-Kulla + Smith was favored for its smooth look. V-Cavities looks definitively sharper.

As far as performances are concerned, looking at the formula, it is indeed costlier than Ashikmin masking but nowhere to a point that it'd make the proposal a no-go in my opinion. It is still cheap to evaluate.

romainguy:
The benefit of Ashikmin and the main reason I picked it in Filament is that
it is visually fairly close to the term used by Estevez-Kulla. It is darker
at high roughness but it overall behaves fairly well (see the shadertoy
posted earlier which lets you switch between both terms easily).

sebavan:
We had a similar conclusion in Babylon than @romainguy exposed here and followed this path to already bring a bit of consistency in the ecosystem.

proog128:
OK, understood. I guess we can add the squared sheen roughness to the normative section without describing the "perfect" implementation in the non-normative section for now. As in the base material, implementations may choose any approximation they want, and the exact masking-shadowing term used in an implementation is certainly something that lends itself for approximations (so Ashikhmin, Smith, V-Cavities are all fine, but have different trade-offs regarding consistency, performance, physical accuracy, ...). I am somewhat optimistic that we can improve the fit given in the Estevez-Kulla paper so that it works well for low roughness values. I cannot tell at the moment when we will revisit this topic, but at some point it is probably going to happen. Once we have a better solution we can adapt the non-normative section.

proog128:
Would be great if we could add a sentence to the introduction that mentions the squared roughness, like in the attached patch: 
[roughness.zip](https://github.com/KhronosGroup/glTF/files/5150200/roughness.zip)

sebavan:
You are right, done and thanks a lot

donmccurdy:
Earlier in this thread, we wrote:

> > About the texture, I would like to keep it only one if possible merging color and intensity, do you think it makes sense?
> 
> This will prevent the use of JPG textures with this extension. Is that OK?

We then decided to punt on the question, which affects other textures too, as a general question for `KHR_texture_basisu`. That extension is now all but complete, but there's still no magic solution for texture packing. In https://github.com/KhronosGroup/glTF/issues/1682 we concluded:

> We will also need to be careful with channel decisions on the upcoming PBR next (transmission, volume, sheen, ...) texture types.

If we do someday produce a generic swizzle extension, it's presumably going to allow something like:

```
material.sheenTexture = inputTexture.rrrg
```

So, by including only a single texture slot for color and roughness, we permanently require them to share a single source texture, and a swizzle extension can't lift that requirement. As we did with clearcoat, my preference might be to allow _but not require_ the packing arrangement.  This would mean introducing another property:

- sheenColorFactor
- sheenColorTexture (.rgb)
- sheenRoughnessFactor
- sheenRoughnessTexture (.a)

This retains the capability to use only a single sampler for both textures, but also:

1. Allows the extension to benefit from swizzle (if we create that extension later)
2. Allows sheenRoughnessTexture to be packed with other textures (ORM?), if sheen uses a solid color, actually _reducing_ required samplers further than the current design.

More generally — I believe our upcoming PBR Next extensions should avoid combining separate pieces of information into a single `propTexture` slot. By doing so, we allow the the possibility that an unoptimized material might use one more sampler, but we also allow more flexibility for optimizing the material. To me, that seems worthwhile.

donmccurdy:
Another advantage of splitting `sheenTexture` into `sheenColorTexture` and `sheenRoughnessTexture` — it's probably not uncommon for sheen to have the same hue as baseColor, or perhaps a lighter version of it. For example, the only kind of sheen color that [Blender's Principled BSDF](https://docs.blender.org/manual/en/latest/render/shader_nodes/shader/principled.html) supports is a "sheen tint" — which just controls the mix of base color vs. white in the sheen reflection.

In that scenario you'd probably want sheenColorTexture and baseColorTexture to be the same texture (with different factors?) but that's not possible if the `.a` channel is defined to be alpha for one and sheen roughness for the other.

sebavan:
If everybody agrees, I am happy to make the change but still afraid by the number of samplers we can already not support on some devices. It is probably edge case but I wonder how we could enforce and if we should enforce a max number of samplers ?

rsahlin:
> This retains the capability to use only a single sampler for both textures, but also:
> 
> 1. Allows the extension to benefit from swizzle (if we create that extension later)
> 2. Allows sheenRoughnessTexture to be packed with other textures (ORM?), if sheen uses a solid color, actually _reducing_ required samplers further than the current design.

I agree with this assessment, in my view it it also an engine implementation thing how the different texture channels may be packed into fewer samplers.
The above solution seems to follow the current pattern and allows flexibility from a client perspective.
Thumbs up from me :-)

rsahlin:
> devices. It is probably edge case but I wonder how we could enforce and if we should enforce a max number of samplers ?

I am against this type of enforcement - it sort of implies that we are tied to one type of technology and I would rather that we are agnostic in this area.
At the same time we should take care to avoid duplication of channels and do everything we can to make it easy to pack channels together (in order to reduce total number of samplers used)

donmccurdy:
> It is probably edge case but I wonder how we could enforce and if we should enforce a max number of samplers?

Not an edge case — using a reasonable number of samplers is certainly important! — but I'm worried about downstream effects of enforcing a maximum. A safer option would be similar to the skin weights language, saying that clients are required to support at least __ samplers per material, so that exporters are more conscious of the constraints. But we can't do that in this extension; it may have to be a future 2.x or 3.0 thing...

In the meantime I do think flexible packing will likely serve us better than strict pairings.

donmccurdy (approved):

bghgary:
SheenCloth model is now updated: https://github.com/KhronosGroup/glTF-Sample-Models/pull/266

extensions/2.0/Khronos/KHR_materials_sheen/README.md
```diff
@@ -0,0 +1,237 @@
...
+* Gary Hsu, Microsoft, [@bghgary](https://twitter.com/bghgary)
+* Eliphas NUIT, Dassault System, [@EliphasNUIT](https://github.com/EliphasNUIT)
+* Aidin Abedi, Epic Games, [@aidinabedi](https://github.com/aidinabedi)
+
+Copyright (C) 2018-2020 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Experimental
```
bghgary, line 21:
```suggestion
Draft
```

extensions/2.0/Khronos/KHR_materials_sheen/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+* Sebastien Vandenberghe, Microsoft, [@sebavan](https://github.com/sebavan)
+* Alexis Vaginay, Microsoft, [@Popov72](https://github.com/Popov72)
+* Tobias Häußler, Dassault System, [@proog128](https://github.com/proog128)
+* Ed Mackey, Analytical Graphics, Inc.
+* Romain Guy, Google, [@romainguy](https://github.com/romainguy)
+* Mike Bond, Adobe, [@MiiBond](https://github.com/MiiBond)
+* Don McCurdy, Google, [@donmccurdy](https://twitter.com/donrmccurdy)
+* Gary Hsu, Microsoft, [@bghgary](https://twitter.com/bghgary)
+* Eliphas NUIT, Dassault System, [@EliphasNUIT](https://github.com/EliphasNUIT)
```
EliphasNUIT, line 13: I would prefer if my real name, Jekfer Bichon, were used here instead of my alias.

sebavan: All good and changed, sorry about that. I did not find it when filling the section :-) Thanks for checking.

extensions/2.0/Khronos/KHR_materials_sheen/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+|                                  | Type                                                                            | Description                            | Required                       |
+|----------------------------------|---------------------------------------------------------------------------------|----------------------------------------|--------------------------------|
+|**sheenColorFactor**                   | `array`                                                                         | The sheen color in linear space        | No, default: `[0.0, 0.0, 0.0]` |
+|**sheenColorTexture**         | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)             | The sheen color (RGB).<br> The sheen color is in sRGB transfer function | No               |
+|**sheenRoughnessFactor**               | `number`                                                                        | The sheen roughness.                   | No, default: `0.0`             |
+|**sheenRoughnessTexture**         | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)             | The sheen roughness (Alpha) texture. | No               |
+
```
rsahlin, line 65: I guess to stay consistent with the core spec - for instance pbrmetallicroughness:
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#reference-pbrmetallicroughness

Then we should declare usage of each of the properties, for instance what texture component(s) the value is sampled from_

### sheen.sheenRoughnessTexture 
The sheen roughess value, roughness is sampled from the A channel.

etc etc

sebavan: I tried to stay consistent with the other extensions: https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md

Should they be reviewed as well then ?

bghgary: These are mentioned in the schema (e.g. https://github.com/KhronosGroup/glTF/pull/1688/files#diff-30c33ab0604a6cc121306aa2e409155dR23) which is kind of the equivalent of the core spec since the properties reference is generated from the schema.

rsahlin: Ok thanks @bghgary  could you elaborate what that means in this particular case?
Do we need to change anything in this extension to be consistent with the core spec in the future?

Issue #1675, opened 2019-09-29
Animating morph targets individually


donmccurdy:
The animation spec currently allows a `channel.path` value of `"weights"`, for a sampler that must provide keyframe values for all available morph targets on the mesh. It would be useful to provide a mechanism of animating the weights of a single target, or perhaps an arbitrary subset, for two reasons:

1. **Filesize**: If only one target needs to be animated for an effect, providing 0s for other targets is wasted space.
2. **Composability**: Animation like facial expressions may be composed from multiple targets. For example, consider an animation that moves the character's mouth, and another that animates the eyebrows. Ideally those animations could be played together or separately, but playing them together may have unwanted effects if the eye animation includes weight=0 for the mouth targets.

I mentioned this in https://github.com/KhronosGroup/glTF/pull/1301, as I think that extension (if it moves forward) would be a good place to address this feature. Filing a new bug, though, just so that it's tracked for reference.

Bug-Reaper:
I think the above post summarizes things quite nicely. As someone working on animation editing software for morph target lip sync animations, I can confirm the above statements to be real pain-points.

I'd also like to add:

3. **Practicality**: If you intend to do any additional modifications to an animation, it is ideal to have separated tracks for each individual morph target. Editing an all in one track is significantly more complex/difficult.
4. **Understandability**:  A track structure where morph targets are targeted individually would allow you to understand the purpose of that animation by just looking at the track data.  Conversely, the merged track structure makes the purpose of that track very hard to understand by just looking at the data.


In the interest of making this format more accessible and easier for developers to work with, I believe this would be an extremely welcome addition.

donmccurdy:
> Editing an all in one track is significantly more complex/difficult.

I think this one might be specific to how threejs imports the animation – while the glTF file stores all the morph target keyframes in one "sampler", glTF has no opinion about what defines an editor's "track", and when  THREE.GLTFLoader creates a THREE.AnimationClip with a single THREE.KeyframeTrack for all targets, that could also be split into a clip with one THREE.KeyframeTrack for each morph target. Might be a little less efficient to play back, but YMMV.

> A track structure where morph targets are tracked individually would allow you to understand the purpose of that animation by just looking at the track data.

👍

zeux:
[...]

Issue #1660, opened 2019-08-14
Mesh Instancing + LOD extension proposal


vpenades:
This is an extension proposal I've been pondering about after my work with [SharpGLTF](https://github.com/vpenades/SharpGLTF), that aims to solve a number of issues in one shot:

- Simplify scene rendering evaluation: makes glTF more _JPEG of 3D_
- Clean LOD support, no weird childs of childs, and much more.
- Multiple meshes per node.

The extension would depend primarily in two objects:

**MeshInstance** would be included in the **Scene** Extensions as a collection.

It essentially defines a Mesh Draw Call:

```c#
MeshInstance
{
    int? VisibilityRule; // should we render it?
    int Mesh;           // what to render.
    int? Node;         // where to render it (If no Skin).
    int? Skin;        // how to deform it. (If Skin)
}
```

So, a client that wants to render a scene, instead of traversing the node
tree graph, it would simply loop over the **MeshInstances** of the scene.

Now the interesting part is that a **MeshInstance** references a _VisibilityRule_
object, which are stored in the glTF model extensions.

**VisibilityRule** must be evaluated at runtime to a _Visible | Invisible_ state. If
a **MeshInstance** references a **VisibilityRule** that evaluates to _Invisible_, then that
mesh should not be rendered.

Initially I thought that the rule could simply be the LOD distance or something like
that, but then I noticed that I could be extended to incorporate many other visibility
and occlusion mechanims, other than LOD, so I came with something like that:

```c#
VisibilityRule
{
    LODRule    LOD
    PortalRule Portal    
    ... Any other visibility rule we can imagine; Planes? BSP trees?
}

```

Now I am not sure if this is the right way of structuring a VisibilityRule, my
guess is that it is certainly possible to have multiple concurrent rules, for
example, a given rule would only evaluate to visible if it's close enough by a
LOD level, and its portal is visible.

But I believe this mechanism would help into bringing rendering large scenarios
into glTF in a way that is simple to understand and implement.

#### How a client would render the scene?

```c#
for each Node in SCENE.Nodes
{
        Node.EvaluateWorldTransform();
}

for each VRule in SCENE.MeshInstances.VisibilityRules
{
	VRule.EvaluateVisibility(Projection, View, World);
}

for each MInstance in SCENE.MeshInstances
{
        if MInstance.VisibilityRule is Visible
        {
              if MInstance has Skin RenderSkinnedMesh(MInstance, Skin);
[...]

Issue #1478, opened 2018-11-02
simple extension for geometry metadata


zellski:
Hey folks,

Throwing together a simple internal extension to allow glTF to summarise some computed geometry data –– I thought I might as well post here in case anyone else is interested.

The information we are interested in is **vertex count**, **triangle count**, and **world-space bounds** for each mesh. Since many nodes can instantiate a mesh, nodes are the meaningful place to insert the extension data.

Every node includes data on the cumulative counts and world bounds for the subtree they parent. If the node references a mesh, we also include the world bounds of just that node.

Limitations:
- This is just for the default static pose. We make no attempt to accomodate animations.
- We traverse the scene graph from the default scene. Other scenes would typically result in different per-node world transforms.

Proposed output:
```
{
    "asset": { "version": "2.0" },
    "extensionsUsed": [ "FB_geometry_metadata" ],
    "scenes": [
        {"nodes": [ 0 ]},
        {"nodes": [ 2 ]}
    ],
    "nodes": [
        {
            "children": [ 1 ],
            "extensions": {
                "FB_geometry_metadata": {
                    "nodeWorldBounds": {
                        "min": [ -0.26, 0.24, -0.88 ],
                        "max": [ 2.24, 1.88, -0.22 ]
                    },
                    "vertexCount": 48,
                    "triangleCount": 24
                }
            }
        },
        {
            "mesh": 0,
            "children": [ 2 ],
            "extensions": {
                "FB_geometry_metadata": {
                    "meshWorldBounds": {
                        "min": [ -0.13, 0.12, -0.44 ],
                        "max": [ 1.12, 0.94, -0.11 ]
                    },
                    "nodeWorldBounds": {
                        "min": [ -0.26, 0.24, -0.88 ],
                        "max": [ 2.24, 1.88, -0.22 ]
                    },
                    "vertexCount": 48,
                    "triangleCount": 24
                }
            }
        },
        {
            "scale": [ 2.0, 2.0, 2.0 ],
            "mesh": 0,
            "extensions": {
                "FB_geometry_metadata": {
                    "meshWorldBounds": {
                        "min": [ -0.26, 0.24, -0.88 ],
                        "max": [ 2.24, 1.88, -0.22 ]
                    },
                    "nodeWorldBounds": {
                        "min": [ -0.26, 0.24, -0.88 ],
                        "max": [ 2.24, 1.88, -0.22 ]
                    }
                    "vertexCount": 24,
                    "triangleCount": 12
                }
[...]

PR #1691, opened 2019-10-23
EXT_mesh_gpu_instancing

ultrafishotoy:
start instancing discussion


ultrafishotoy:
I'm not sure where the new attributes (TRANSFORM4x3, and ID) should get defined.

donmccurdy:
Thanks @ultrafishotoy!

We should also state that this extension's goal is to enable _GPU_ instancing<sup>1</sup>. The definition given in the [Unity docs](https://docs.unity3d.com/Manual/GPUInstancing.html) is quite good, I think:

> Use GPU Instancing to draw (or render) multiple copies of the same Mesh at once, using a small number of draw calls. It is useful for drawing objects such as buildings, trees and grass, or other things that appear repeatedly in a Scene. GPU Instancing only renders identical Meshes with each draw call, but each instance can have different parameters (for example, color or scale) to add variation and reduce the appearance of repetition. GPU Instancing can reduce the number of draw calls used per Scene. This significantly improves the rendering performance of your project.

Of course the bit about "each instance can have different parameters" doesn't fully apply here; as we haven't yet defined — and may not define — ways to override arbitrary material parameters per-instance.

But otherwise I think those are the rough goals for this extension. Do others agree, or have I missed anything important? I suppose transmission size could also be mentioned.

***

<sup>1</sup> I just want to disambiguate "GPU instancing" from "instancing" because the latter could also mean simple reuse of transmitted data (accessor, texture, mesh, material, etc.) throughout a glTF asset. glTF already allows this, and the feature is widely used. It does not, however, directly enable efficient reduction of draw calls as this extension might.

donmccurdy:
And... would you be able to share some example use cases? We've discussed at various points, but just want to add that context in the thread too.

If by any chance it would be easy to generate a sample asset using the proposed extension that would also be very neat to see, but no need for anything yet if that's a nontrivial amount of work.

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
```diff
@@ -0,0 +1,64 @@
...
+        {
+            "mesh": 0,
+            "name": "teapot",
+            "extensions": {
+                "KHR_instancing": {
+                    "attributes": {
+                        "TRANSFORM4x3": 0,
+                        "ID": 1
+                    },
+                }
```
donmccurdy: I suspect this is the right design, but just to mention an alternative, the `KHR_instancing` object could also be left empty and the attributes put into the mesh's `primitive.attributes` instead. The per-instance would have different counts, compared to the native mesh attributes, and this would break implementations that don't recognize the extension, if they tried to use those attributes. Recommend keeping as-is.

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Instancing is defined by adding the `KHR_instancing` extension to any glTF mesh node.  The attributes section contains accessor ids for new TRANSFORM4x3 and ID attribute buffers. 
+For example, the following defines some instancing attributes to a mesh node in the graph. Instancing only applies to mesh nodes.  Applying to nodes rather than meshes allows the 
+same mesh to be attached as normal to nodes while also being instanced.  
+
+```json
+{
+    "nodes": [
+        {
+            "mesh": 0,
```
donmccurdy: One option we should consider is enabling the KHR_instancing extension to (optionally) override the `node.mesh` property. For example:

```js
name: "teapot",
mesh: 0,
extensions: {
  KHR_instancing {
    mesh: 1,
    attributes: { ... }
  }
}
```

This simply provides exporters with some flexibility to control how/if fallback happens for viewers that don't recognize the extension:

1. When the `mesh` override is omitted, clients that don't recognize the extension will render a single instance of the mesh with the node's transform.
2. When the `mesh` override is provided, it should reference a single instance of the mesh. The fallback `node.mesh` value could then point to anything the exporter chooses – a merged mesh of all instances, a single point, or a textured plane saying "404 Not Found".
3. When the `KHR_instancing` extension is marked as required, clients that don't recognize the extension are expected to fail fast without attempting to proceed.

ultrafishotoy:
I chose this design based on what Octane allows; any given mesh can be
referenced by multiple nodes some of which may be instanced.  While it
is possible to collapse all references to a single group of instance
transforms, I prefer to keep the octane node structure intact so that it
can be leveraged at run-time for things like hide/show, procedural
animation, culling, etc.

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Khronos 3D Formats Working Group
+
+* Alexey Knyazev
+* Don McCurdy
+* Ed Mackey
+* Norbert Nopper, UX3D [@UX3DGpuSoftware](https://twitter.com/UX3DGpuSoftware)
+* Romain Guy
+* Tobias Haeussler
+* John Cooke
```
donmccurdy: Fine to just leave this list as a "John Cooke + TODO" for now, we usually fill it in at the end.

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
donmccurdy: > I chose this design based on what Octane allows; any given mesh can be
referenced by multiple nodes some of which may be instanced...

Agreed, that's another good reason. I'll resolve this suggestion.

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
ultrafishotoy: oops.

ultrafishotoy:
I think the optional mesh is a good idea.  Another property I think that would be helpful is a bounding box for the entire transform group.   It could be useful for culling and picking and calculating it at run-time would be very time consuming.

ultrafishotoy:
I can definitely provide some sample assets.  I've got some teapot scenes specifically for testing.  I guess I just need to resolve how to deal with the new attributes.

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+TODO  
+
+## Extending Nodes with per instance attributes
+
+Instancing is defined by adding the `KHR_instancing` extension to any glTF mesh node.  The attributes section contains accessor ids for new TRANSFORM4x3 and ID attribute buffers.
```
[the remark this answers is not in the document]
ultrafishotoy: agreed, I fully expected to see these names change.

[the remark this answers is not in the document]
lexaknyazev: > mesh node

Maybe "node with a mesh"?

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
```diff
@@ -0,0 +1,61 @@
...
+```json
+{
+    "nodes": [
+        {
+            "mesh": 0,
+            "name": "teapot",
+            "extensions": {
+                "KHR_instancing": {
+                    "attributes": {
+                        "TRANSFORM4x3": 0,
```
lexaknyazev: The spec should mention in which coordinate space those transforms are defined. Reasonable options are 
- local, like a regular mesh (full chain: instanced transforms, local node's TRS, all parents to the scene root);
- scene, like skinned mesh (all local and parent node transforms are ignored).

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
```diff
@@ -0,0 +1,61 @@
...
+                        "ID": 1
```
lexaknyazev: Are these IDs implementation-dependent, with no spec-defined usage? If so, they shouldn't be included here. glTF 2.0 has an established workflow for supporting app-specific attribute semantics - they just start with an underscore.

Otherwise, the spec should be more sound and include at least:
- allowed datatypes (likely scalar uint8/16/32, uint32 is not available on WebGL 1.0);
- uniqueness within a set of instances, within a scene, or within an asset (otherwise they should not be called IDs).

In shaders, GPU instances have built-in IDs. Should this attribute be somehow related to them?

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
ultrafishotoy: We're only using local space at otoy.  Maybe that could be handled by alternate 'TRANSFORM' attributes that include the space?  i.e. TRANSFORM_LOCAL, TRANSFORM_SCENE, with maybe TRANSFORM defaulting to LOCAL?

lexaknyazev: `TRANSFORM_SCENE` behavior can be achieved with a node positioned at the scene root. Supporting only local (with explicitly described interaction between JSON-stored local TRS and accessor-stored instanced TRS) seems enough to cover both cases from engine perspective.

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
```diff
@@ -0,0 +1,61 @@
...
+```json
+{
+    "nodes": [
+        {
+            "mesh": 0,
+            "name": "teapot",
+            "extensions": {
+                "KHR_instancing": {
+                    "attributes": {
+                        "TRANSFORM": 0,
```
takahirox: Thinking of instancing + animation. Allowed values for `animation.channel.target.path` are 
- translation
- rotation
- scale
- weights

https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#targetpath-white_check_mark

Do we also need `KHR_instancing` extension in `animation.channel` to specify TRANSFORM attribute as target.path?

```json
    "channels": [
        {
            ....,
            "KHR_instancing": {
                "sampler": 0,
                "target": {
                    "node": 0,
                    "path": "transform"
                }
            }
        }
    ]
```

donmccurdy:
> Do we also need KHR_instancing extension in animation.channel to specify TRANSFORM attribute as target.path? ... Or instancing + key frame animation is not popular because of non good performance?

I'm not familiar with best practices for expressing animation of instance batches in a runtime-friendly way. If there are tools, engines, or formats that have set a good precedent on that, please let us know! The closest I can think of would be Point Caches in FBX ([definition](http://download.autodesk.com/us/fbx/2013/3ds_Max_FBX_Plug-in_Help/index.html?url=files/GUID-691D1194-AE6E-4EE9-A6A3-F1797D2F3B27.htm,topicNumber=d30e2947), [example](https://www.html5gamedevs.com/topic/40942-support-for-point-level-animation-aka-pla-point-cache-animation-fbx-vertex-cache/)), and I don't know whether that approach would be appropriate for glTF. In particular, and [similar to a known limitation of morph target animation](https://github.com/KhronosGroup/glTF/issues/1675), I expect users would want the ability to animate a _single_ instance in the batch, rather than having to bake the transforms of each instance at each keyframe collectively.

Because I'm not really confident of the answers on this, I'm inclined to say animation should be left undefined by this extension, but perhaps added later through something like https://github.com/KhronosGroup/glTF/pull/1301.

However, I _do_ think this extension should provide the ability to set application-specific, per-instance attributes on the batch. In addition to the application-specific, per-vertex attributes already allowed on the mesh. For example...

```js
meshes: [{
  primitives: [{
    attributes: {
      POSITION: 0,
      _BARYCENTRIC: 1
    }
  }]
}]
nodes: [{
  name: "tree_person",
  mesh: 0,
  extensions: {
    KHR_instancing {
      mesh: 1,
      attributes: {
        TRANSFORM: 2,
        ID: 3,
        _POSITION_KEYFRAME_1: 4,
        _POSITION_KEYFRAME_2: 5,
        _POSITION_KEYFRAME_3: 6
      }
    }
  }
}]
```

... in this case, the attributes prefixed with `_` are application-specific, and when imported to a generic viewer they are likely to be ignored entirely. However, any user might set up an application to use them as follows:

- `_BARYCENTRIC`: Defines barycentric coordinates of each vertex, useful in applying stylized wireframe effects in custom shaders. These coordinates are the same for all instances.
- `_POSITION_KEYFRAME_1–3`: Defines positions of each instance at three known keyframes. The application is responsible for determining interpolation between the keyframes; these are not associated with any entry in the glTF `animations` list.

**tl;dr** — I'm hoping to allow some flexibility for content authors to apply custom effects to instanced glTF models. Custom attributes and instance IDs enable that without locking in the spec to a particular animation mechanism prematurely. If/when we identify best practices for this sort of animation, which can be implemented in a performant and portable way, suggestions for a formal extension for instancing animation would of course be welcome.

extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing/README.md
donmccurdy: Thanks! Replied on https://github.com/KhronosGroup/glTF/pull/1691#issuecomment-546029405 for wider visibility.

takahirox:
@donmccurdy 

Thanks for the comment. TBH I'm not familiar with, too. I'm happy if someone shares the best practice.

> I expect users would want the ability to animate a single instance in the batch, rather than having to bake the transforms of each instance at each keyframe collectively.

Agreed.

> I'm inclined to say animation should be left undefined by this extension, but perhaps added later through something like #1301.

Oh I didn't know #1301. Yeah I think ok to place animation stuffs out of this extension but somewhere else.

> tl;dr — I'm hoping to allow some flexibility for content authors to apply custom effects to instanced glTF models. Custom attributes and instance IDs enable that without locking in the spec to a particular animation mechanism prematurely. If/when we identify best practices for this sort of animation, which can be implemented in a performant and portable way, suggestions for a formal extension for instancing animation would of course be welcome.

Sounds good to me!

takahirox:
Is per-instance material parameter out of scope of this extension? (Sorry if I'm missing anything.) For example user may want to set different baseColorFactor, roughnessFactor, and other material parameters for each instance.

```
baseColorFactor = baseColorFactor * instanceBaseColorFactor;
roughnessFactor = roughnessFactor * instanceRoughnessFactor;
```

zeux:
Trying to understand the rationale behind this - glTF already allows referencing the same mesh from many nodes in the transform graph. glTF loader can thus gather, for each mesh, a set of nodes that instantiate it and create the transform buffer at runtime - either at load time, if animation is not used / supported, or at runtime.

Does this extension target scenes with thousands of instances of the same mesh and is thus intending to reduce the file sizes? If so, is 4x3 matrix an appropriate storage or should we use independent streams of T/R/S data instead?

lexaknyazev:
> Does this extension target scenes with thousands of instances of the same mesh and is thus intending to reduce the file sizes?

Yes. It also reduces processing time as the transform buffer can be copied to GPU as is.

> If so, is 4x3 matrix an appropriate storage or should we use independent streams of T/R/S data instead?

Currently, it's based on an existing vendor-specific implementation. More efficient storage options (such as a lossless compression pass, sparse arrays, channels separation, etc) will certainly be considered.

donmccurdy:
> glTF loader can thus gather, for each mesh, a set of nodes that instantiate it...

@zeux While a glTF loader could do this, in practice they don't. At least for three.js, doing so would mean reducing the node hierarchy in an opinionated way that feels unexpected without some hint (either from the asset, or from the application) that the meshes are meant to be batched.

From that perspective, I see the rationale here as (1) a way for an asset to identify that meshes can and should be batched, and (2) efficient transmission+parsing of these batches.

I do recognize that specifying GPU batches is somewhat unusual territory for a 3D asset format. Based on the goals of glTF for runtime efficiency, I think this could be a reasonable choice, but would be glad to hear if you'd suggest something else or think this is a bad idea altogether. 🙂

ultrafishotoy:
This extension is intended exactly as Don suggests: a way to identify huge batches meshes that could "never" be rendered, and would be cumbersome to load without the direct to gpu instance transform buffer.

zeux:
> At least for three.js, doing so would mean reducing the node hierarchy in an opinionated way that feels unexpected without some hint (either from the asset, or from the application) that the meshes are meant to be batched.

I am not sure you need to reduce the hiearchy. Just to be clear, I'm not arguing that this extension isn't useful - I just think we should understand the set of optimization criteria more precisely so that we can make the optimal choices.

In my mind, glTF instancing can optimize on three different axes:

- Transmission size
- Loading time
- Rendering performance

I would argue that #3 can and probably should be taken care of by the renderer. The way this can work (we use a system like this at Roblox, and a lot of other game engines like UE4 do as well) is that the renderer can automatically identify meshes that use the same materials, and upload transform buffers dynamically. We go one step further and we support "static" transform buffers - when many objects share the same geometry & material, but have transform that doesn't change, we aggregate objects like this into "clusters" and pre-upload the transform buffer once (and if an object moves, we invalidate the buffer / break the cluster up if necessary).

A system like this has many advantages:

- It transparently supports instancing regardless of how the scene was produced. For example, it doesn't matter if you load a scene that contains a thousand teapots, or if you load a scene that contains a teapot and then clone this a thousand times with JS code.
- It doesn't require redundant specification of transforms for animation or other purposes. With this extension, we either have to declare animation unsupported, or extend the animation spec which is already somewhat painful to deal with to include this usecase.
- It doesn't enforce a specific static grouping of objects into draw calls, which allows independent culling or sorting at the renderer's discretion. I'm not sure what the intent of this extension is in this regard - if the transform buffer is pre-uploaded, culling or sorting become problematic.
- It doesn't enforce a specific criteria of instancing. For example, a question was raised in this thread as to whether material parameters can/should be instanced as well. Indeed, our renderer supports instancing of some material parameters, like base color, but not others, like roughness scale. The lowest common denominator seems to be transforms but this doesn't seem as useful.

So, with this in mind, I feel like this extension should probably focus on supporting GPU instancing for massive amounts of instances, where loading time (as in, JS time involved in processing / deduplicating / JSON parsing) or transmission size of regular glTF specification is prohibitively expensive. Hence my comment about separating position/translation/scale - I'm happy to run some experiments as part of gltfpack as to what the tradeoff there is.

ultrafishotoy:
> This extension is intended exactly as Don suggests: a way to identify huge batches of meshes that could "never" be rendered, and would be cumbersome to load without the direct to gpu instance transform buffer.

I think a bounding box for the whole instance group might be a good idea.  That would allow for culling and picking of the whole group without any extra load-time processing.  Since the extension is part of the Node structure it allows strategic grouping of the same meshes through different instanced nodes which could be leveraged for culling.

vpenades:
I'm surprised nobody paid attention to this proposal: https://github.com/KhronosGroup/glTF/issues/1660 

😢

donmccurdy:
> I'm surprised nobody paid attention to this proposal: #1660

@vpenades Sorry for the delay, I've commented at https://github.com/KhronosGroup/glTF/issues/1660#issuecomment-550447644.

> I think a bounding box for the whole instance group might be a good idea.

@ultrafishotoy From our discussion earlier, I agree it's a good idea, but it can probably be a separate node-level extension. Node-level bounding boxes have been requested for other purposes too: https://github.com/KhronosGroup/glTF/issues/1478.

> when many objects share the same geometry & material, but have transform that doesn't change, we aggregate objects like this into "clusters"...

@zeux is it reasonable to perform that aggregation at runtime? My assumption would have been that most engines do this offline and store some representation of the clusters. I haven't really looked at the possibility of implementing realtime aggregation in an engine like three.js, and so I don't know whether this extension should be aiming to _provide_ pre-aggregated clusters, or just to make realtime aggregation easier (e.g. by marking certain objects as "static").

> It doesn't enforce a specific static grouping of objects into draw calls, which allows independent culling or sorting at the renderer's discretion. I'm not sure what the intent of this extension is in this regard - if the transform buffer is pre-uploaded, culling or sorting become problematic.

@pjcozzi, this sounds like something that would be important to the b3dm/3d-tiles use cases as well, would be curious if your team has opinions on it.

pjcozzi:
>>It doesn't enforce a specific static grouping of objects into draw calls, which allows independent culling or sorting at the renderer's discretion. I'm not sure what the intent of this extension is in this regard - if the transform buffer is pre-uploaded, culling or sorting become problematic.
>
>@pjcozzi, this sounds like something that would be important to the b3dm/3d-tiles use cases as well, would be curious if your team has opinions on it.

Thanks for bringing this up.  @lilleyse will take a look.

ultrafishotoy:
> (Just a note: The path currently is
> `extensions/2.0/Khronos/KHR_instancing/extensions/2.0/Khronos/KHR_instancing`
> while it should be
> `extensions/2.0/Khronos/KHR_instancing/`
> ... just in case it wasn't noticed yet)

fixed now.  sorry about that!

donmccurdy:
Here's a prototype implementation in three.js:

https://github.com/mrdoob/three.js/compare/dev...donmccurdy:feat-gltfloader-instancing

**Simple:**
Screen Shot 2019-11-07 at 4 59 24 PM


**Teapots Galore:**
Screen Shot 2019-11-07 at 4 57 27 PM

A couple issues worth mentioning, found in the process:

- Both examples should include `"extensionsUsed": [ "KHR_instancing" ]`
- We should document the new `MAT4x3` type, if only for clarity while we discuss whether to keep it
- The smaller mesh in the simple example has a negative scale. Per https://github.com/KhronosGroup/glTF/issues/1697, the spec recommends (as a non-normative implementation note) that index order should be reversed in this case, to support mirrored geometry. That isn't possible while keeping the instance in the same batch, to my knowledge, so I think that expectation would have to be dropped in this case.

ultrafishotoy:
> Here's a prototype implementation in three.js:

Wow!  Thanks Don!!

All the issues you mention make perfect sense.  I'll add the extensionsUsed block.  I'm not sure how to add the documentation for MAT4x3 at this point.  Non-normative matrices could be an issue, it's pretty easy to do in octane.

ultrafishotoy:
We need to think a bit more about instance ids and perhaps other per instance data.  

I'm thinking that the built-in gpu ids probably provide enough data to do anything that custom ids would, it's just a matter of properly grouping the data that the id's would reference.  Also custom attributes can be added with the existing underscore mechanism (i.e. _ID).

zeux:
@donmccurdy I don't really see why three.js can't do this at runtime but at the same time there's still space for an extension like this for load / transmission efficiency reasons. So I guess my point is that advanced renderers don't really need to rely on an extension like this to provide meaningful performance gains for rendering efficiency. If one wants to encode a scene with tens of thousands of instances of a mesh however, this extension seems worthwhile.

I've looked a bit into the size impact of 4x3 matrix data vs a componentized storage. It looks like if all 3 components (T/R/S) are required, with the data set presented there's no significant difference in transmission size. This might be because the scene contains a lot of single-axis rotations around a cardinal axis where the quaternion storage and matrix storage aren't that different after deflate, and the scales used in this scene are somewhat irregular and non-uniform. It would be interesting to look at this on real scenes.

I am assuming that the nodes that use instancing are part of the hierarchy, so they can have a parent node (for coordinate system transforms)? Or do nodes that are instanced have to be rooted?

vpenades:
Stupid question here: is instancing expected to work along with skinned or morphed meshes?

zeux:
One more note; it would be nice to have an option to explicitly designate the target buffer with instance data as GPU-ready. Today one can use ARRAY_BUFFER / ELEMENT_ARRAY_BUFFER target for bufferViews - the attribute is optional (which would be nice to fix in glTF next...), but when it's present it allows the loader to directly upload the data without inspecting the rest of the scene for references. We should probably do something similar here, and especially make sure that the view with instance data can't be shared by any non-instance accessors.

lilleyse:
> @pjcozzi, this sounds like something that would be important to the b3dm/3d-tiles use cases as well, would be curious if your team has opinions on it.

Definitely +1 on supporting custom underscored attributes - we would make heavy use of those. 

Instead of using `TRANSFORM` we would define custom attributes for translation, rotation, scale (and quantized / oct-encoded versions of those). A 4x3 transform matrix is overkill if instances are just differentiated by position, so I hope the extension would not require `TRANSFORM`.

[i3dm](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification/TileFormats/Instanced3DModel) has the full set of use cases for 3D Tiles instancing. I like that this extension is thin though, future extensions can easily build on top of it.

ultrafishotoy:
At this point it is intended only for non-animated meshes.

ultrafishotoy:
Summary of current discussion points:

1.  Can't the aggregating just be done at run-time?  This extension is specifically for gpu instancing with huge numbers of instances that would be hard to render otherwise.  Having gpu ready buffers is good for both load and transmission times.

2.  Should the attributes be non-specified and use only the custom underscore mechanism, or should we require TRANSFORM and/or ID?   The idea is that the flexibility of all custom attributes helps with both file size and use cases.  I think this is a good idea.

3.  Clean up the schema: add ARRAY_BUFFER, document MAT4x3, note about negative scale? 

4.  Bounding boxes for the entire instance group should be handled by another optional extension.  Useful for culling and picking.

donmccurdy:
> Should the attributes be non-specified and use only the custom underscore mechanism, or should we require TRANSFORM and/or ID? The idea is that the flexibility of all custom attributes helps with both file size and use cases. I think this is a good idea.

I think we're all in agreement that custom attributes should be allowed, but if the suggestion here is to remove any specification of TRANSFORM or ID semantics, I worry that might leave too little portable information to justify the extension.

Listing possibilities as I see them:

TRANSFORM:
(a) Make TRANSFORM optional. In its absence, application is left to define transform.
(b) Omit TRANSFORM entirely. In its absence, application is left to define transform.
(c) Omit TRANSFORM entirely, and add (optional) POSITION, ROTATION, SCALE semantics. Any semantic omitted will default to the equivalent component of an identity transform.
(d) Keep TRANSFORM, and require it.

I'd lean toward (c).

ID:
(a) Make ID optional. In its absence, GPU IDs are assumed to be the default.
(b) Omit ID entirely. In its absence, GPU IDs are assumed to be the default.
(c) Keep ID, and require it.

I'd lean toward (a) or (b), with a slight preference for (a).

donmccurdy:
> ...it would be nice to have an option to explicitly designate the target buffer view with instance data as GPU-ready.

It feels like a gamble to assume anything we pick here is actually going to be uploaded directly to the GPU in a majority of implementations. 😕 How many engines are using 4x3 layout? Three.js uses 4x4 for instance transforms, but that's a very new API and presumably could change. No idea about other engines. Would be great to design something that's at least suitable for GPU upload, but I'm nervous about over-specifying it.

> I am assuming that the nodes that use instancing are part of the hierarchy, so they can have a parent node (for coordinate system transforms)?

That would be my assumption, yes – each batch of instanced objects is attached to a single node in the hierarchy, but that node could have any number of parents.

> I don't really see why three.js can't do this at runtime...

We probably should consider that, yes, regardless of the outcome on this extension proposal. If you/others are aware of resources on best practice for that sort of optimization I'd be curious to read up.

donmccurdy:
Another suggestion, based on today's call and https://github.com/KhronosGroup/glTF/pull/1706 — consider renaming to *KHR_mesh_instancing*.

donmccurdy:
Wrote a prototype implementation for the glTF Blender addon. Details in https://github.com/KhronosGroup/glTF-Blender-IO/issues/766#issuecomment-569784530. It's currently based on an outdated version of this proposal, using a 4x3 TRANSFORM matrix instead of the TRS attributes.

extensions/2.0/Khronos/KHR_mesh_instancing/README.md
```diff
@@ -0,0 +1,63 @@
...
+```json
+{
+    "nodes": [
+        {
+            "mesh": 0,
+            "name": "teapot",
+            "extensions": {
+                "KHR_mesh_instancing": {
+                    "attributes": {
+                        "POSITION": 0,
```
donmccurdy: minor: Would `TRANSLATION` be a more appropriate semantic name? That would match [TRS property naming for nodes](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#transformations), which seems nearer to the intention here than the vertex `POSITION` semantic.

ultrafishotoy: Agreed.  Change is pushed.

zeux:
The extension draft says that the attributes specify the world transforms, and parent transforms are ignored. Is this accurate? Is this motivated by reducing the GPU-side transform cost (to remove one extra matrix transform per vertex)?

With that specification TRS (translation-rotation-scale) isn't quite accurate and a perhaps more accurate set of names is position-orientation-size. I'm not necessarily suggesting to change the names, just noting the pretty significant difference.

It's unfortunate that this also precludes various existing methods to share data, for example:

- For dequantization, it could be useful to specify the dequant transform just once, but this forces to specify additional scaling factors per instance
- For coordinate system transforms, it's commonplace to export this using extra top-level glTF nodes, whereas here it would need to be baked into each object's transform
- This makes it impossible to share transform buffers between various clones of a single compound object in the scene, e.g. 10 houses with 12 windows in each either need a single mesh with 120 instances for windows, or 10 mesh nodes with a unique 12-entry buffer in each, even if the houses are perfect clones other than the parent xform
- In practice, if the object that uses instancing doesn't support modification of top-level scene graph position it might be too hard to use this extension (what if I want to load a single house from a .glb file many times and move individual copies?) so I suspect many renderers will want to transform by the world matrix as well.

There's precedent for this in skinning transforms although there the transforms are still hierarchical, it's just that the hierarchy is that of bones, and the mesh position in the hierarchy is ignored.

donmccurdy:
Ah, per https://github.com/KhronosGroup/glTF/pull/1691#issuecomment-555282859 I had been assuming that parent transforms _would_ apply to the instances.

Copying the section of the extension spec we're discussing:

> ...node can exist anywhere within the node tree no parent transforms are applied to the attributes (because the attributes are uploaded directy to the gpu).

For three.js, we upload the instance transforms as attributes and — in addition — upload the model matrix derived from the world transform of the node to which the instances are attached, as a uniform. This allows the instances to be re-positioned or animated (as a group).  If GPU upload is the only concern here, I believe it would be OK to remove the "no parent transforms" requirement.

(Although in practice three.js uses 4x4 matrices for instance transforms, so there's still some pre-upload processing. But at least moving the entire instance collection does not require changing the individual instance transform data.)

The requirement that parent transforms of skinned meshes are ignored comes from engines where the skinned mesh is not necessarily part of the node hierarchy at all, and from the need to reconcile the mesh transform with that of the skeleton.

ultrafishotoy:
I realized after your last 2 posts that it was a mistake for me to add that bit to the spec, it really only applies to my custom _TRANSFORM4X3 implementation.   I assume most engines will always pass along  the node's world transform that can be multiplied with the instance transform in the shader, as Don mentioned for three.js, and I believe that addresses all of Arseny's concerns.  I will remove that from the spec.  Thanks guys.

rsahlin:
Hi and sorry for joining the discussion so late - I have just started covering the 3D Formats group.

I wonder if this extension could use the Node's matrix or TRS instead of defining it's own?
The reason is that if a viewer does not support the instancing extension the transform will be wrong (since it resides within the extension). 
As will any spatial calculations that use the node transform. 
It would be great if this extension will not make geometry disapear if it's not supported by the viewer (since the instancing may be done internally anyway)

I guess this is related to @zeux latest comments.

rsahlin:
How shall the FB material extension be handled when used in an instanced Mesh?
https://github.com/KhronosGroup/glTF/pull/1681

As normal or any reasons why it should not be allowed together?

rsahlin:
> This extension is intended exactly as Don suggests: a way to identify huge batches of meshes that could "never" be rendered, and would be cumbersome to load without the direct to gpu instance transform buffer.
@ultrafishotoy Could you please explain in detail what this means?

From my perspective using OpenGL / ES or Vulkan it is a matter of either:
A: GL/GLES Use instanced drawing (drawArraysInstanced or drawElementsInstanced) 
B: Vulkan set command buffers as needed for instancing.
- in either case matrix transform needs to be stored in attribute  buffer (VBO) - unless the instance count is low enough to fit in a UBO.

Is this what you mean by 'direct to gpu instance'?

Instanced drawing is a problem very much related to graphics API - as you say not realistic on OpenGL/ES without drawArraysInstanced/drawElementsInstanced.
As such I have a  feeling this does not match super well for a KHR extension - I see it more as an  implementation / graphics API specific extension. 
As I see it it's not really extending the functionality as such.....

lexaknyazev:
> in either case matrix transform needs to be stored in attribute buffer (VBO) - unless the instance count is low enough to fit in a UBO

An array of matrix transforms could also be copied as-is to SSBO and accessed based on `gl_InstanceID`.

lexaknyazev:
And with quite limited UBOs (16kB at worst), we're still talking about hundreds of instances per drawcall. Storing the same transforms in JSON and building buffers on the fly would bloat transmitted asset size and put unnecessary burden on engines.

rsahlin:
I guess the point I'm trying to make is that I see this as an implementation specific way of getting the transform (matrix) into the shader.

The normal behavior is likely to take the node transform and upload (together with view/projection) as uniform - then render the primitives (ie the Mesh)

The instanced behavior needs to store the transform for each (instanced) object in a buffer (likely attribute) that is available in the shader - and indexed using an id.

I'm sorry but I do not see how this is directly related to glTF.

donmccurdy:
> I wonder if this extension could use the Node's matrix or TRS instead of defining it's own?

The extension uses the transform of the parent node to transform the entire batch. The transforms included within the extension are applied on top of that.

>  The reason is that if a viewer does not support the instancing extension the transform will be wrong (since it resides within the extension).

The extension provides a `.mesh` property within the extension, in addition to the `.mesh` property on the node, so that the asset author can decide how fallbacks will work for viewers that don't support the extension. Viewers that support the extension must prefer the mesh in the extension. Viewers that don't will always use the mesh on the node, which could be a single instance, or a lower LOD mesh representing all instances, or complete geometry for all instances stored in an external `.bin` file — options are available to suit the use case.

> How shall the FB material extension be handled when used in an instanced Mesh?

This extension does not define mechanisms by which specific materials would be applied to specific instances... Other than that, I see no incompatibility between the two. Runtime management of instances (animation, etc.) is intentionally left out of scope and seems likely to be application-specific.

> The normal behavior is likely to take the node transform and upload (together with view/projection) as uniform - then render the primitives (ie the Mesh)

> I'm sorry but I do not see how this is directly related to glTF.

An intended use case is for a scene that might have 10,000 rivets, trees, or buildings, etc. Defining 10,000 nodes with 10,000 transforms in the JSON will slow file loading and parsing considerably. In addition, the viewer will need to parse through all nodes in the graph — at runtime — searching for which nodes can potentially be batched. The intention of this extension is to provide a way for the asset itself to **(1)** more efficiently transmit these batches, and **(2)** signal to the runtime which things are safe/appropriate for batching.

I agree with you that there is no guarantee the transform buffer is in the format any given application would prefer for GPU batching. Short of allowing multiple different options within the extension — which would increase complexity for engines that want to support arbitrary assets — I don't think that can be fixed, but ideas welcome. Regardless, benefits (1) and (2) above seem compelling for cases like those mentioned in https://github.com/KhronosGroup/glTF/issues/1699, to me.

rsahlin:
Hi Don and thanks for your reply

Ok, I did not catch the `.mesh` property. 

Looking at the issue from this direction makes me wonder if it could be combined into a LOD extension?

> Defining 10,000 nodes with 10,000 transforms in the JSON will slow file loading and parsing considerably.

Could you please elaborate?
I have not seen a direct comparison with regards to actual JSON filesize (eg 1000 instanced meshes vs 1000 nodes) - could you please provide one?
As for in runtime searching for nodes that can be instanced - I think this would be quite straightforward as these 10000 nodes would be children to the same node. 
ie meaning that the node hierachy would not become more complex as it would be one parent node containing all instanced meshes.

> The intention of this extension is to provide a way for the asset itself to **(1)** more efficiently transmit these batches, and **(2)** signal to the runtime which things are safe/appropriate for batching.

This to me sounds like some sort of metadata hint - all childnodes use the same mesh and can be instanced/batched.
Would that work in your usecase?

vpenades:
So, if this extension allows rendering 10.000 instances of a mesh using a hard boiled list of matrices.... in terms of performance, or feature wise, what's the difference with having a single mesh with 10.000 copies of the same geometry?

I understand this saves a lot of memory, but that's it, it does not add any additional feature other than saving memory. Is that right?

rsahlin:
@vpenades Since the node contains the transform (matrix) then all copies of the same geometry (primitives) would appear in the exact same position. 
A transform is needed to make the mesh appear in a different position.

The feature, as I see it, is to enable optimization for an engine implementation (eg three.js)

I think the discussion is largely with regards to where this transform shall be:
A: Added with a node referencing the same mesh - 1000 instances = 1000 nodes
B: Using the KHR_mesh_instancing which adds a transform for each use of the mesh - 1000 instances will still be one node.

lexaknyazev:
> What's the difference with having a single mesh with 10.000 copies of the same geometry?

> I understand this saves a lot of memory, but that's it, it does not add any additional feature other than saving memory. Is that right?

- On transmission side, this extension would replace multitude of `node` objects having JSON-stored transforms with binary matrix (or TRS) array that does save a lot of memory.

- On asset loading side, this extension would greatly reduce the amount of JSON processing (because per-instance data would be stored in binary form).

- On rendering side, this extension would give enough information for runtimes to efficiently batch draw calls.

So yes, this extension addresses data transmission, memory consumption, and rendering performance rather some additional feature.

Since "efficient transmission and loading of 3D scenes and models" is the primary goal of glTF, the working group believes that it's the right direction.

/cc @pjcozzi

rsahlin:
Hi @lexaknyazev 

* If the 2 first points are the most relevant (reducing JSON footprint)
Then I would propose to do an extension solely for the purpose of allowing binary data in nodes - as this would achieve the same goal without changing node hierarchy.

* If the third point is most relevant (instancing) I would propose to add metadata that hints to the engine that it could do runtime optimizations.

This is so that the glTF spec can be kept as focused as possible.

lexaknyazev:
> an extension solely for the purpose of allowing binary data in nodes

Thousands of `node` objects would still bloat JSON even without transforms, moreover each of them would need to include a separate pointer to its binary data making the runtime processing even more convoluted.

> metadata that hints to the engine that it could do runtime optimizations

Such separately-defined metadata would need to list all nodes that form the same instancing group or to be included into every node itself, thus again leading to unnecessary JSON bloat.

Anyway, feel free to open a pull request with an alternative extension design that would provide similar gains in the efficiency of transmission, processing, and rendering of thousands of instances of the same mesh.

rsahlin:
> Such separately-defined metadata would need to list all nodes that form the same instancing group or to be included into every node itself, thus again leading to unnecessary JSON bloat.

I don't think so as you would have 1000 children to one node - the metadata would be on the same node that you now put the instance extension on.

rsahlin:
> Thousands of `node` objects would still bloat JSON even without transforms, moreover each of them would need to include a separate pointer to its binary data making the runtime processing even more convoluted.

The point I was trying to make is that if the goal is to reduce JSON footprint I think there are other ways that are more generic.
For instance intruducing a binary node structure - or something similar. 
All for the reason of keeping the glTF specification focused and "clean"

The goal of this PR as far as I can tell is still:
> This extension is specfically designed to enable GPU instancing which renders many copies of a single mesh at once using a small number of draw calls.

rsahlin:
> This extension does not define mechanisms by which specific materials would be applied to specific instances... Other than that, I see no incompatibility between the two.

@donmccurdy I see the two as not compatible with eachother.
The FB_material_variants tags a primitive as using one of a set of materials (texture/program/state).
A likely usecase is that textures and maybe even the shader program changes between the materials.
As far as I can tell, this will be a problem when instancing.

@zellski What's your take on this?

donmccurdy:
> As for in runtime searching for nodes that can be instanced - I think this would be quite straightforward as these 10000 nodes would be children to the same node.

@rsahlin This is rarely the case in real models. This extension joining all instances under a single node is a technical optimization, and not a pre-existing indicator one would expect to find in the wild.

Nevertheless, it is possible for a runtime to find nodes that can technically be batched. What is more difficult is expressing _intent_ for the engine to do batching, which this extension provides. Without that, we (threejs) have to err on the side of loading the data exactly as it was given, because we don't know how the developer plans to use the data. Other engines' approaches may vary — I acknowledge that if threejs had more advanced features for runtime batching, I would probably be approaching this extension purely as a storage optimization rather than discussing hints for batching.

> I understand this saves a lot of memory, but that's it, it does not add any additional feature other than saving memory. Is that right?

@vpenades granting the assumption that a renderer will use GPU instancing when this extension is present, and will be unlikely to do so otherwise, the extension also reduces draw calls dramatically.

> I see the two as not compatible with each other. The FB_material_variants tags a primitive as using one of a set of materials (texture/program/state).

@rsahlin GPU instancing inherently requires a single shader be used for all instances — some tricks may be used to vary the appearance of instances (beyond the scope of this extension) but those tricks are much less flexible than FB_material_variants, e.g. different instances cannot use different textures. In that sense, different variants are very likely to require separate draw calls, and unlikely to benefit from this extension.

donmccurdy:
> I have not seen a direct comparison with regards to actual JSON filesize (eg 1000 instanced meshes vs 1000 nodes) - could you please provide one?

Here are two comparisons, one with 1,000 instances and another with 10,000 instances:

[Suzanne_Instancing.zip](https://github.com/KhronosGroup/glTF/files/4106698/Suzanne_Instancing.zip)

The test is contrived and simply scatters spheres, with random transforms, on the surface of a Suzanne model. I generated this with a Blender script written against a previous version of the `KHR_mesh_instancing` spec, so it's a bit out of date, but that shouldn't affect filesize.

Screen Shot 2020-01-23 at 7 27 21 PM

File sizes:


| count | extensions | raw size | gzip size |
|-------|------------|---------|----------|
| 1,000 | — | 831K     | 212K |
| 10,000 | —  | 2.9M | 806K |
| 1,000 | `KHR_mesh_instancing` | 241K | 71.3K |
| 10,000 | `KHR_mesh_instancing` | 593K | 226K |

I've specifically made sure the `mesh` entry is reused for all instances in the "default" version, so the increased filesize is entirely a result of JSON representing the nodes and their transforms. For the sake of fairness I've also included gzipped sizes, since repeated JSON keywords seemed likely to benefit from that.

There is probably some analysis worth doing on the time required to _parse_ all that JSON, but I haven't done so at this point.

rsahlin:
Great @donmccurdy - thanks for the filesize comparison!
So for a usecase with 1000 instanced meshes (which is a lot) - the reduction in size is some 500K

I think it would be super great if we could find a generic solution for this:
A: The binary storage that can benefit any model and not just those with instancing
For this to work the scope needs to be increased (not only transform)
B: Mesh instancing hint to engines to runtime optimize referenced meshes for instancing.

rsahlin:
> @rsahlin This is rarely the case in real models. This extension joining all instances under a single node is a technical optimization, and not a pre-existing indicator one would expect to find in the wild.

So then I guess in order for this extension to  be effective it is necessary to add the support to exporters as well.
Ie the exporter, either by way of user decision or some calculations, determines that a number of nodes share the same mesh.
Then I guess it could be just as feasible for the exporter to add them as child nodes with a hint that they can be instanced. (eg you _will_ get 1000 childnodes)
I guess if we also support some sort of binary extension for reducing nodes size it could be possible from a performance point of view (for three.js)?

If possible I think this solution would help keeping the format focused.

rsahlin:
After todays 3D formats call it is clear to me that there are applications requesting this feature.
I think making sure the extension is as intended by these apps (tiling?) should be our main focus.

As it's yet not clear how we could handle the binary representation of more data than the transform -  and the apps requesting/implementing this seems to be webgl based - maybe we should consider changing this to a vendor extension (webgl)?

donmccurdy:
> I think making sure the extension is as intended by these apps (tiling?) should be our main focus.

Agreed. Implementations showing a good future-proof foundation for  use cases like batch scenes in [3D Tiles](https://github.com/AnalyticalGraphicsInc/3d-tiles) and [highly complex scenes in AEC](https://github.com/KhronosGroup/glTF/issues/1699) might be a good bar to set for ourselves.

> As it's yet not clear how we could handle the binary representation of more data than the transform...

The mechanism provided here would be:

```js
    "nodes": [
        {
            "mesh": 0,
            "name": "teapot",
            "extensions": {
                "KHR_mesh_instancing": {
                    "attributes": {
                        "TRANSLATION": 0,
                        "ROTATION": 1,
                        "SCALE": 2,
                        "_CUSTOM_INSTANCE_VELOCITY": 3
                    },
                }
            }
        }
    ]
```

Similar to custom semantics for vertex attributes, custom instance attribute semantics are prefixed with an underscore, and defining their behavior is left to the application.

> maybe we should consider changing this to a vendor extension (webgl)?

I would not be opposed to doing a vendor or multi-vendor extension for now, especially if we don't think we have enough information for a really future-proof extension definition yet. We don't currently treat WebGL as a vendor, but an `EXT_` prefix would be applicable.

rsahlin:
I would then propose to move forward using the EXT_ prefix and postpone future-proofing the extension until a later date as this would allow feedback from the apps using the extension.

ultrafishotoy:
new samples are posted.

rsahlin:
The usd format uses instancing.
However, this instancing is quite different from what is meant by instancing in this PR.

In usd instancing means something along the lines of 'creating a new instance of....' and override some data properties.
This is done by tagging a prim (element) with metadata and then reference this element with some changed property data:
https://graphics.pixar.com/usd/docs/USD-Glossary.html#USDGlossary-Instancing

glTF Already has some of this ability to 'instance' geometry by means of indexing meshes - however glTF does not have the option to 'override' data in the referenced mesh.
The mesh and all containing primitives (attributes, indices, material) will be exactly the same in all uses.

Another thing that glTF is lacking is a true binary format - the .glb format is a self contained file but json parsing still needs to be done.
Which is from what I have heard one major motivation for this PR.

* I still feel that the glTF mesh instancing shall not be an KHR_ extension - transform is changed in a non-standard way and hierarchy is 'hidden'.
* I would also like us to be influenced by the flexibility and simplicity of instancing in usd - I know they are not exactly the same thing. 
Set 'instanceable = true' and the element will be a candidate for instancing.
* A true binary glTF format could probably solve the mesh instancing problem, without 'breaking' node hierarchy and introducing a seperate transform scenegraph as this PR does.

extensions/2.0/Khronos/KHR_mesh_instancing/samples/teapots_galore_TRS/teapots_galore_TRS.gltf
```diff
@@ -0,0 +1,565 @@
...
+            "bufferView": 3,
+            "componentType": 5126,
+            "count": 958,
+            "extras": {
+                "arrayIndex": 3
+            },
+            "type": "VEC3"
+        },
+        {
+            "bufferView": 4,
```
Drigax: This is our accessor for the instance rotation buffer, right?
Can we change this to be a Quaternion vec4 instead of an Euler angle vec3?

extensions/2.0/Khronos/KHR_mesh_instancing/samples/teapots_galore_TRS/teapots_galore_TRS.gltf
```diff
@@ -0,0 +1,565 @@
...
+            "componentType": 5126,
+            "count": 958,
+            "extras": {
+                "arrayIndex": 4
+            },
+            "type": "VEC3"
```
Drigax: This is our accessor for the instance rotation buffer, right?
Can we change this to be a Quaternion vec4 instead of an Euler angle vec3?

ultrafishotoy: We definitely can change Euler angles to a Quaternion.  Is that what most engines would expect?

donmccurdy: I think so, yes — this would match the quaternion rotations we store on nodes: https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#transformations

donmccurdy:
**Responses to open questions:**

> I still feel that the glTF mesh instancing shall not be an KHR_ extension - transform is changed in a non-standard way and hierarchy is 'hidden'.

I'm fine with changing from a KHR to an EXT prefix.

In three.js our concept of a GPU instanced batch is already separate from the scenegraph — we cannot instance things without consolidating them to a single node. Instances can be identified and picked/raycasted, but not in the same way as distinct nodes. That's anecdotal, for sure, and I could certainly imagine ways of building a renderer without that constraint.

> I would also like us to be influenced by the flexibility and simplicity of instancing in usd - I know they are not exactly the same thing. Set 'instanceable = true' and the element will be a candidate for instancing.

That flexibility may be incompatible with GPU instancing — no engine can use distinct materials for distinct instances in the same batch, for example, except in a few very constrained ways. 

I'm afraid that I have no idea, at present, how to robustly combine the concept of instancing used by USD with the concept of instancing required by GPUs. These concepts might be too different to encapsulate in a single extension, and perhaps even mutually exclusive.

> A true binary glTF format could probably solve the mesh instancing problem, without 'breaking' node hierarchy and introducing a seperate transform scenegraph as this PR does.

I understand your reasoning here, but introducing a whole new container format is a much more complex solution than I'm interested in pursuing for this feature. 

***

**Suggested next steps:**

In my responses above I'm (by all appearances) disagreeing with @rsahlin, but I think it's fair to say that we're both partially right, especially as it's quite hard to predict future requirements that might build on instancing. I see a few options to proceed:

- **(A)** Keep the extension in its current (simpler, but more limited) form with an EXT_ prefix.
- **(B)** Continue designing a more flexible solution, including a full node hierarchy and perhaps per-instance materials, to be ratified as an EXT_ or KHR_ extension.

Unfortunately, I don't think I have bandwidth any time soon to design (B). On the other hand, I don't know if we have enough expressed implementor interest to proceed with (A). If three.js will be the only client consuming this extension, it might as well be a THREE_-prefixed vendor extension.

Perhaps we should put this on hold until there are more engines saying, "we need this and it solves ______ for us as it is", or "we need this, but with _______ changed". I appreciate that the BabylonJS team is investigating whether they _can_ implement it, but I'm not sure we can count that as actual interest. 🙃

Another type of useful feedback would be for stakeholders (OTOY? Autodesk? AEC community? artists?) to share examples of models that don't currently work well with the glTF format and glTF tooling, and let us evaluate choices against that.

zeux:
I would note that I had the same initial questions as @rsahlin but I came around after thinking about the transmission size impact and the loader-side complexity. While a binary-only format can seem compelling, the explicit design point of glTF is to split the bulky binary data (vertex, image, animation) from the presumably non-bulky JSON data. Some scenes push the split further into the JSON range, and this extension helps move that back. I don't fully understand the KHR vs EXT distinction.

At the same time, we had feedback re: glTF for the use in large scenes with a lot of instanced content, and right now the only workaround is to use gltfpack or a tool like gltfpack to merge meshes - which isn't always a great solution. Having instancing as an option would seem good to me since it makes it easy for loaders to quickly load and display large scenes.

In my mind the optimal workflow goes like this:

- Exporters don't have to care about this extension unless the time/size impact of generating many nodes is prohibitive
- For uses where using many clones of the same object is vital for rendering performance/transmission size, opinionated pipeline transform tools could take the non-instanced scene and convert it to the instanced scene

Does Babylon.JS plan to implement this? gltfpack certainly will implement this extension as an option if we agree to pursue this path in general.

Drigax:
> Does Babylon.JS plan to implement this? gltfpack certainly will implement this extension as an option if we agree to pursue this path in general.

Yes, I am working on an implementation now, I expect to have it checked in within the week:
image

extensions/2.0/Khronos/KHR_mesh_instancing/README.md
```diff
@@ -0,0 +1,64 @@
...
+        {
+            "mesh": 0,
+            "name": "teapot",
+            "extensions": {
+                "KHR_mesh_instancing": {
+                    "attributes": {
+                        "TRANSLATION": 0,
+                        "ROTATION": 1,
+                        "SCALE": 2,
+                    },
```
Drigax: Is there a reason we do not specify the instance count in the extension? I can foresee this causing security vulnerabilities in native clients due to malformed/mismatched accessor counts for the different attributes.

lexaknyazev: This follows existing convention from the core spec: `mesh.primitive.attributes` relies on `accessor.count` being the same for all attributes. Having a dedicated property to signal the instance count does not add much value here as implementations would still need to check that it's less than or equal to each `accessor.count`.

Drigax: Makes sense, it wouldn't be wise to break from the convention here.

Drigax:
Thanks @ultrafishotoy, the updated asset loads without any changes to the extension importer:

image

extensions/2.0/Khronos/KHR_mesh_instancing/README.md
donmccurdy: A possible counterargument is that with the core `mesh` definition, the `POSITION` attribute is required. Here, none of the attributes are required, and implementations might need even need to look to the custom attributes to find the actual instance count. In the three.js implementation this becomes:

```js
var template = translation || rotation || scale || otherAccessors[ 0 ];
var count = template.count;
```

That's a little ugly, but not burdensome in our case. But if other implementations think it's useful to know the count more directly I wouldn't be opposed to putting it in the extension.

donmccurdy:
@zeux it sounds like we have general agreement on moving forward, albeit with an EXT_ extension rather than KHR_. I'll open a PR for the three.js implementation soon. If you'd be willing to work on a gltfpack implementation, that would be much appreciated!

donmccurdy:
@ultrafishotoy would it be much trouble to throw a custom attribute, like `_id`, into the sample asset?

donmccurdy:
three.js draft PR opened in https://github.com/mrdoob/three.js/pull/18913.

donmccurdy:
Since there has been some confusion between GPU instancing and the type of data instancing that GPU already supports<sup>1</sup> ... would `EXT_mesh_batching` be clearer than `EXT_mesh_instancing`? 

<sup>1</sup>reusing mesh data on multiple nodes, reusing accessors in multiple meshes, etc.

zeux:
I feel like "instancing" is the correct name here. We could call this EXT_mesh_instancing_gpu or smth like that to highlight the connection, but I feel like "batching" doesn't help (batching doesn't necessarily imply GPU instancing...).

Separately - the accessor types are underspecified in this extension. Would it be reasonable to specify the same restrictions as that for animation accessors, that is vec3 for position/scale *but* support normalized accessors for rotations? Asking because I don't think the three.js draft supports normalized rotations right now, and it's important for transmission size.

rsahlin:
I agree with @donmccurdy that we should find another name.
In a broader sense `instancing` is used to specify `instances of an object` - this is not what we mean with this extension.
Here I think the term batching makes more sense - regardless of the fact that the API call underneath is called `glDrawArraysInstanced`  or `glDrawElementsInstanced`

The difference being that when you instance an object you can give it new fields/parameters - in this case all parameters are already locked. 

Ok, ok - I know - you can provide other attributes but that's it.

A comparison can be made to USD that has instancing in the object sense, ie they can instance primitives and give them new (overloaded) data.
- See my above comment for deatils

donmccurdy:
Sadly it sounds like "instancing" might be the most appropriate word for both the thing we actually mean, and the thing we do not mean. 😕  I guess @zeux's point is that batching often means merging static meshes, or other techniques? `EXT_mesh_gpu_instancing` might do...

zeux:
@donmccurdy Yeah, batching is a general term, and broadly applies to draw call optimizations. Unity for example (rightfully) calls their dynamic and static mesh aggregation technique "batching": https://docs.unity3d.com/Manual/DrawCallBatching.html.

`EXT_mesh_gpu_instancing` seems unambiguous to me.

rsahlin:
Ok, in order not to derail this PR anymore I succumb to the pressure :-)
I guess its ok since it includes the `gpu` 
👍

ultrafishotoy:
> @ultrafishotoy would it be much trouble to throw a custom attribute, like `_id`, into the sample asset?

Nope, I'll post a new sample in a day or so.

ultrafishotoy:
New samples are posted.  @donmccurdy, what is the best way to handle the name change and move to the Vendor folder?  Should I just create a new pull request?

donmccurdy:
@ultrafishotoy thanks! The samples look good here.

Let's just update the PR in place. I've changed the title, and opened a PR to your PR: https://github.com/ultrafishotoy/glTF/pull/1

Drigax:
Is this ready for ratification?

ultrafishotoy:
> Is this ready for ratification?

Since its a EXT_ extension it doesn't actually need ratification, but the current plan is to wait for gltfPack to implement it before merging.  That way there will be a tool that can actually produce gltf's that use the extension.

zeux:
Oh, I didn't realize that we're waiting on gltfpack. I'll try to implement something this week.

One thing I'd like to see in the spec is a precise definition of the accessor component types for pos/rot/scale. I'd recommend matching the animation specification for this.

ultrafishotoy:
> Oh, I didn't realize that we're waiting on gltfpack. I'll try to implement something this week.
> 
> One thing I'd like to see in the spec is a precise definition of the accessor component types for pos/rot/scale. I'd recommend matching the animation specification for this.

Thanks @zeux !  I'll have a look at the anim spec and update schema.

zeux:
One more spec clarification (this one isn't really relevant for gltfpack, but it's good to highlight) - is there a usecase for this extension being optional (used but not required), or does it have to be in the required list, since there's no alternate way to specify instancing information?

ultrafishotoy:
> One more spec clarification (this one isn't really relevant for gltfpack, but it's good to highlight) - is there a usecase for this extension being optional (used but not required), or does it have to be in the required list, since there's no alternate way to specify instancing information?

My feeling is that it should be required if used;  if it was optional it wouldn't confer any of the transmission, loading, and runtime benefits that are specifically what this extension is designed for, and large instancing batches could get unwieldy really quickly.   Anyway, lets try to do a quick discussion about this tomorrow.

donmccurdy:
See https://github.com/KhronosGroup/glTF/pull/1691#discussion_r338251198 for some previous thoughts on optional use. In general it is a good thing if extensions can be designed to enable optional use. In this case, I think that could be accomplished by adding a `mesh` property to the extension which overrides the default `mesh`. Then the default `mesh` property becomes a fallback for clients not supporting the extension, and could refer to (a) a single instance of the object, (b) a joined mesh containing all instances in an external `.bin` buffer, or (c) whatever the author feels is appropriate to show.

That said, few (no?) users have used runtime optimization extensions — like Draco — without requiring them, in practice. `KHR_draco_mesh_compression` supports uncompressed fallback in external buffers, but the fallback feature hasn't really been used. If we think the complexity of a fallback mesh is harmfull, I could be persuaded to skip it.

juj:
Heya, dropping some drive-by ponderings here, hope it is welcomed!

What is the value of having this extension? Isn't having this a bit redundant?

https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/SimpleMeshes/glTF/SimpleMeshes.gltf#L13 already shows how to support instancing by sharing mesh references, unless I am mistaking something?

It seems odd to have the file format specify what vertex attributes would be instanced. This is because engines have different feature sets that they can support in their instanced code paths, that are engine/shader/code layout specific, rather than input asset specific?

Some engines support instancing on static meshes that have the same texture only (i.e. only instance transforms), other engines might support instancing color data as well (modulate per-vertex colors by a color multiplication modifier, or modulate a texture by a color multiplication modifier). Certain engines might instance also animated meshes, etc.

Since typically the feature set of what per-instance attributes an engine is capable to render instanced, and which it can't depends on the shader code paths that are implemented, having a file format that would specify what the structure of the instance buffer should be seems like it would be dictating how a renderer must be implemented?

Also different GLTF files will be authored with different instance formats, and an engine that wanted to support all would be expected to create a permutational combination of rendering shader paths to support all of them? Also note that the full possible combination of instance buffer structures is not being specified here?

May be that I am missing something here, though this extension seems like it does not need to exist? If an engine is capable of collecting compatible nodes to be submitted instanced, it could do that already without this extension?

I would understand that more instancing opportunities would be enabled by having a spec that would help reduce duplicate data, so that e.g. a renderer would not need to identify when two duplicated materials have the identical values (though this is already achieved by a good exporter), but this spec does not seem to improve reducing data duplication/sharing of data either(?)

Does this extension help some renderers implement instancing, when before they would not have been able to? If it does, perhaps the spec might be helpful to have an intro paragraph to motivate why it helps some renderers "solve" instancing for them?

Great work with GLTF on everyone involved, just scratching my head a bit on this one :)

ultrafishotoy:
Hi @juj,  this extension is specifically for really large instancing groups (thousands or even millions of trees, clumps of grass, etc.), with the goal of reducing file size, transmission time, and load time.  Have a look at @donmccurdy's excellent analysis here : https://github.com/KhronosGroup/glTF/pull/1691#issuecomment-577978794.

juj:
> with the goal of reducing file size, transmission time, and load time.

Oh, thanks, gotcha. Missed that this was about conserving disk space! The description rationale at https://github.com/KhronosGroup/glTF/pull/1691/files#diff-d53ba0f3c9191f50ab2501ad1edcc6f1R21 does not mention reducing file size, transmission time, and load time, but makes it read as if it'd be to optimize GPU rendering efficiency.

ultrafishotoy:
> > with the goal of reducing file size, transmission time, and load time.
> 
> Oh, thanks, gotcha. Missed that this was about conserving disk space! The description rationale at https://github.com/KhronosGroup/glTF/pull/1691/files#diff-d53ba0f3c9191f50ab2501ad1edcc6f1R21 does not mention reducing file size, transmission time, and load time, but makes it read as if it'd be to optimize GPU rendering efficiency.

I'll add some language about that.  Thanks!

zeux:
For reducing transmission size, would it be possible to incorporate support for normalized 16-bit integers for quaternion storage?

glTF animation data currently allows 32-bit floating point, 16-bit normalized integer and 8-bit normalized integer values, but in my experience 8-bit quaternions aren't really adequate in terms of precision so we could either specify both 8-bit/16-bit for consistency with animations, or just 16-bit for pragmatic reasons. GPUs support decoding 16-bit normalized integers when the input is bound as an (instanced) vertex stream, so I'm assuming it's reasonably straightforward to implement and it can significantly reduce the transmission size impact.

Noteworthy is that three.js prototype implementation currently doesn't support that IIRC, but because it decodes quaternion data to matrix data I'm assuming it's a simple matter of adding a conversion step in JS code; @donmccurdy could comment on feasibility.

zeux:
Well, "this week" arrived 3 weeks ago but I did get to this eventually :) https://github.com/zeux/meshoptimizer/pull/142 implements support for this extension in gltfpack.

Note that the output uses 16-bit normalized quaternions as per my suggestion above. It was trivial to implement in three.js and it already works in Babylon.JS so I'm assuming it's straightforward to do this; this saves transmission size by using a more efficient and more compressible quaternion encoding.

Attaching a few example models:

- two basic models, simplemeshes (using KHR_mesh_quantization) and simplemeshes_noq (without KHR_mesh_quantization)
- stadium model provided in #1699, stadium.glb (using KHR_mesh_quantization) and stadium_cc.glb (using KHR_mesh_quantization and MESHOPT_compression draft)

Using prototype three.js implementation with a few fixes the stadium model is very quick to download and render; it's pretty slow in Babylon.JS, I believe it's because that implementation decodes the node instances into the scene graph which takes a while - I'll file an issue for that.

gltfpack right now has a limitation in that it outputs all instances in world space, which leads to meshes that had skew transform (from accumulated relative transforms in the node graph) looking differently. At some point I can rework the implementation to preserve that.

[instancing.zip](https://github.com/KhronosGroup/glTF/files/4619489/instancing.zip)

ultrafishotoy:
> Well, "this week" arrived 3 weeks later but I did get to this eventually :) [zeux/meshoptimizer#142](https://github.com/zeux/meshoptimizer/pull/142) implements support for this extension in gltfpack.

Awesome!  Thanks @zeux!

donmccurdy:
> so we could either specify both 8-bit/16-bit for consistency with animations, or just 16-bit for pragmatic reasons.

Either is fine with me. I might vote for consistency with the animation spec, since the implementation burden is low. Your suggested changes on the three.js implementation look good, as well.

> If an engine is capable of collecting compatible nodes to be submitted instanced, it could do that already without this extension? ... Does this extension help some renderers implement instancing, when before they would not have been able to? 

@juj That is what I hope, anyway, in addition to reducing filesize and parsing cost. I understand your perspective — why don't engines just batch everything already? Nothing prevents them from merging, instancing, or otherwise optimizing a scene loaded at runtime. But in the ~9 years that WebGL has been available, I'm not aware of any engine doing so reliably, for any standard format<sup>1</sup>. I'm less familiar with native engines like Unity or Unreal, but my understanding is that they do this optimization offline and store data that is easier to batch than their input sources, e.g. FBX.

Your concern is fair, and others have been raised, which is partly why this is an EXT_ extension rather than an official KHR_ one. It is testing a hypothesis, that pre-processed data can make instancing more accessible for realtime transmission. I hope it is successful, but this is not an area that prior formats have explored particularly thoroughly. It is entirely possible that we'll learn more constraints are necessary to make a fully future-proof standard for GPU instancing, when/if an official KHR_ extension is considered.

<sup>1</sup>There may be exceptions to this in special areas like AEC? In geospatial visualization heavy offline processing is also common.

juj:
> But in the ~9 years that WebGL has been available, I'm not aware of any engine doing so reliably, for any standard format1.

This is surprising to read, working close to two decades with native APIs that support instancing, all of the engines that I ever worked with that supported instancing, use it dynamically rather than statically from the input data files. When porting UE4 and Unity3D to asm.js/wasm, both of them extend that support to WebGL.

This is the first time I read that it would not be possible to do so reliably. What is preventing them from doing so?

This is also the first time I read that anyone would be attempting to store instancing information "pre-solved" into an input file.

The reason that static instancing is a bad idea is because of view frustum dependence. Even if you stored a forest of 100k trees as 100k instances in a file, it would be inefficient to use a single 100k instanced draw call at runtime to render them, because probably only some 1000 of them might fit in the view frustum at once - and which ones do at a given time, depends on the position/orientation of the frustum.

Engines would rather do one of the following:

a) first frustum+occlusion-cull all the instances, and then render whatever remain using instancing, or
b) divide the world into a grid, and for each cell that the view frustum intersects, render all the trees in that cell instanced. (each cell has a presolved list of instances in that cell)

The solution a) would not be able to leverage any instancing data present in an input file. One might argue that b) could, if b) pre-bucketed the instances to such cells, but the cell size will invariably be engine-specific, and then an engine would have a problem of reconstructing the cell boundaries from input instance data, just complicating things.

> I'm less familiar with native engines like Unity or Unreal, but my understanding is that they do this optimization offline and store data that is easier to batch than their input sources, e.g. FBX.

Both engines employ both instancing and batching techniques. In neither engines does one save "pre-solved" instancing data to the optimized/baked input files. Unity utilizes strategy a) from above, and I believe that so does Unreal.

[Unity GPU Instancing](https://docs.unity3d.com/Manual/GPUInstancing.html)
[Unreal Engine Draw Call Merging Dynamic Instancing](https://dq8iqaixvew1d.cloudfront.net/en-US/Programming/Rendering/MeshDrawingPipeline/index.html)

All this being said, if instancing helps improve code size of glTF scene files, then it might be a worthy addition. However I would expect that renderers would as the first thing have to discard the instancing information after loading, i.e. it is just used as a indirection-based domain-specific compression technique.

To me it seems that renderers that would faithfully render the instance information stored in a file like this, would be ones that ignore frustum culling, and behave inefficiently/suboptimally because of that; or ones that would be expecting that the scene serializer would have done a suitable optimization b) above for them. Should glTF serializers be expected to require applying optimization b)?

Also, it does worry me that with this kind of instancing extension, implementing a loader becomes more complicated as one will need to be able to recognize both non-instanced and instanced ways of loading input files?

vpenades:
@juj I fully agree with you, I am not sure if this extensions features are worth the cost of adding more complexity to the loaders.

At first I thought this extension would be really useful to implement things like hair or fur on animated characters, but since it's been limited to static instancing, that possibility is gone.

Then I thought about trees, forests, and I didn't realize about the issue of frustum culling... so, another possibility gone.

So, I would like to ask which are the real, practical use cases of this extension (beyond filling a small volume with thousands of teapots or cutting down file size). Knowing this is important to me because I need to leverage the pros/cons to decide whether implementing this extension is worth or not.

donmccurdy:
> Also, it does worry me that with this kind of instancing extension, implementing a loader becomes more complicated as one will need to be able to recognize both non-instanced and instanced ways of loading input files?

This is a vendor extension — it is prefixed `EXT_` because multiple vendors have expressed interest in implementing the same thing. General-purpose glTF tools are not expected to implement every custom vendor extension — that would certainly be a huge burden. We had discussed an official version, but came to the conclusion that there wasn't enough precedent or information to design it yet. A vendor extension allows tools to opt in, learn from the implementation, and perhaps (or perhaps not) an official extension is made someday.

> ...all of the engines that I ever worked with that supported instancing, use it dynamically rather than statically from the input data files. When porting UE4 and Unity3D to asm.js/wasm, both of them extend that support to WebGL.

I find this surprising as well. My understanding was that [static batching does not usually work this way](https://answers.unity.com/questions/43441/static-batching-at-runtime.html), but I understand dynamic batching is more flexible.

> This is the first time I read that it would not be possible to [implement dynamic batching] reliably. What is preventing them from doing so?

It's not impossible, no. But it's quite complex, and the argument that Unity and Unreal have implemented it, so why can't everyone else, is a bit worrying to me. I don't mean that as a criticism — if anyone has advice about how smaller engines like three.js, babylon.js, et al could implement dynamic instancing with no pre-processing of the input files, I'm certainly interested. 🙂

> b) pre-bucketed the instances to such cells, but the cell size will invariably be engine-specific, and then an engine would have a problem of reconstructing the cell boundaries from input instance data

A goal of this extension is (IMO) to make it *possible* to create these buckets in an application-appropriate way. For a standalone model (e.g. for architecture or retail product preview) perhaps a single batch is fine. For a larger game world you'd certainly want cells. I consider the flexibility to decide the cell structure as a good thing — there is probably not a one-size-fits-all choice here. Reconstructing cell boundaries does not seem overly complex to me, but separate extensions have been proposed to provide world bounds at the node level, and (if we needed that information here) I would prefer to keep the information in those extensions instead of conflating the information with instancing data.

> However I would expect that renderers would as the first thing have to discard the instancing information after loading, i.e. it is just used as a indirection-based domain-specific compression technique.

For a three.js we are able to use the instancing information very nearly as it is provided; I can't comment on other renderers.

***

> At first I thought this extension would be really useful to implement things like hair or fur on animated characters, but since it's been limited to static instancing, that possibility is gone.

Could you say more about what you are looking for here? Do you mean animation of individual instances, like [Houdini's VAT system](https://github.com/keijiro/HdrpVatExample)?

> Then I thought about trees, forests, and I didn't realize about the issue of frustum culling... so, another possibility gone.

Authoring a single large batch will defeat culling, in the same way that merging an entire game world into one large mesh would defeat culling. Creating well-chosen batch cells will enable efficient culling. I see that flexibility as a strength in this extension: it would potentially enable tools like gltfpack to take input options (`e.g. gltfpack --instancing-cell-size 100`) to optimize a model for the application. Without that flexibility, I'm not sure it's realistic to design an extension guaranteeing the data is optimized for any arbitrary application. Do you think that is possible?

> So, I would like to ask which are the real, practical use cases of this extension...

Cases like trees and forests are, indeed, what I see as the practical uses of this extension.

I think use cases like hair and fur may also *benefit* from this extension's data structure, but the extension is not (alone) sufficient to fully enable those things. In the short term I would expect to do custom processing on the loaded data to do this animation. In the long term perhaps glTF should consider procedural materials and procedural animation, similar to VAT, to more fully enable that use case. Or maybe we should really wait until something like that is designed, but I see that as a very large endeavor.

vpenades:
> Could you say more about what you are looking for here? Do you mean animation of individual instances, like [Houdini's VAT system](https://github.com/keijiro/HdrpVatExample)?

I always thought that instancing would be a good opportunity to allow hair/fur on single characters. Think of Monsters INC or Chewbacca. But also for realistic characters where you want to display actual hair over the whole body like in real life.

Crowds simulation is out of the question because the previously mentioned frustum culling problem, the same for trees/forests.

I believe this instancing extension assumes that a gltf-model is going to be culling-unit, that is, it's either rendered in full, or not at all.  The problem with this approach is that I believe most engines preffer to do a full run per mesh. Some years ago I had the opportunity to use SpeedTree... What it did was, for every kind of tree, it selected a specific tree kind (a single mesh), ran a culling over the whole scene (many square kilometers) while filling an instancing buffer dinamically, so all the visible instances of a given mesh were rendered in a single call.

This is different than culling model "cells" and rendering its contents cell by cell.

My point is that, since glTF does not have any concept of culling, this instancing extension is limited to geometries that are assumed to be culled as a whole, or rendered as a whole. And this removes many of the use cases of instancing.

As a side note... can a gltf model using the instance extension... to be instanced? I think an engine would have to check if the model uses internal instancing, and if it's the case, render the models in the scene one by one. So maybe, supporting instancing _per model_ prevents an engine supporting instancing _per scene_.

I did find a use case that might be interesting to some; which is molecule visualization, as in, defining simple colored meshes as atoms, and producing models with millions of atoms. It's an useful use case, specially these days, but I am not sure if it's sooo specific that people working in the bio industry will probably have their own formats and viewers.

rsahlin:
@juj & @vpenades - I agree with your concerns regarding this extension.

However, as @donmccurdy pointed out this is a vendor extension, ie nothing that I forsee ending up in the coming versions of glTF (as included in the core spec) - as such I guess I'm fine with it since there are a couple of clients/viewers asking for this.

My initial proposal was to have this as a named WEB_GL, eg `WEB_GL_mesh_gpu_instancing` extension - but apparently that does not exist :-)

extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md
```diff
@@ -0,0 +1,59 @@
+# EXT\_mesh\_gpu\_instancing
+
+## Khronos 3D Formats Working Group
```
emackey, line 3: Please change this line to `## Contributors`

ultrafishotoy:
@donmccurdy : it looks like I don't have write access so I can't do the merge.  Maybe you can do it when you add your comments?  Thanks!

extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md
ultrafishotoy: Done!  Thanks Ed!

emackey:
I'm merging this.  If there are further comments, please open a new issue.

Thanks @ultrafishotoy!

ultrafishotoy:
> I'm merging this. If there are further comments, please open a new issue.
> 
> Thanks @ultrafishotoy!

Awesome, thanks to you @emackey!

zeux:
I've noted quantized quaternions a few times on this thread; since this is merged now I've opened #1818 to specifically address that.

donmccurdy:
I've tried to clarify the current answers to questions raised on this PR in https://github.com/KhronosGroup/glTF/pull/1821 — feedback welcome.

PR #1694, opened 2019-10-24
Add JglTF-mesh reference

chadj2:
Request to add JglTF-mesh https://github.com/chadj2/jgltf-mesh/

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,10 @@ To compare WebGL-based glTF loaders, see [gltf-test](https://github.com/cx20/glt
       * [jgltf-viewer-lwjgl](https://github.com/javagl/JglTF/tree/master/jgltf-viewer-lwjgl) - glTF viewer based on [LWJGL version 2](http://legacy.lwjgl.org/)
   * [jgltf-browser](https://github.com/javagl/JglTF/tree/master/jgltf-browser) - Application combining the above libraries
   * [jgltf-validator](https://github.com/javagl/JglTF/tree/master/jgltf-validator) - Simple glTF validator, only intended for internal use
+  
+| Tool | Status | Description |
+|------|--------|-------------|
+| [JglTF-mesh](https://github.com/chadj2/jgltf-mesh) | ![status](https://img.shields.io/badge/glTF-2%2E0-green.svg?style=flat) | Generate glTF meshes and geometry in Java |
 
 #### Ada
 


javagl:
There is a dedicated branch for creating models (including meshes) in JglTF: https://github.com/javagl/JglTF/tree/model-creation (it's not merged into master yet - I've been busy with other things recently :-/ ). I'll take a look at what you have done there, maybe it can be integrated directly into JglTF? (This is independent of merging this PR, of course - just intended as a pointer...)

chadj2:
the JglTF-mesh project (https://github.com/chadj2/jgltf-mesh) was originally designed for the purpose of building textured meshes from real world data. It was intended to hide many of the details from the user (e.g. texcoords, normals, triangles) so the user can focus on simply calculating the points on the mesh but it was not designed to provide maximum flexibility.

There are 2 parts to the problem:
1. Geometry needs to be calculated with triangles, vertexes, normals, etc.
2. Primitives need to be serialized to a buffer and metadata created for it.

These need to be separate steps because multiple passes must be made on the data to complete the calculations. Maybe there needs to be a lower level library of objects like those in com.kinetica.mesh.buffer to handle step (2) above? I can see you already did similar work. There could be another library on top of that to hide some of the complexities of calculating geometry.

I am opening to integrating this into JglTF but I will point out that the functionality does not cover the entire glTF 2.0 spec.

javagl:
Indeed, finding the right level of abstraction is the hard thing. 

My experiments for "glTF model creation", specifically the mesh part, are mainly summarized in https://github.com/javagl/JglTF/blob/model-creation/jgltf-model-builder/src/test/java/de/javagl/jgltf/model/creation/example/GltfModelCreationExample.java#L36 : As you can see, it starts at a point where I assume that the geometry data is already available in arrays.

In contrast to that, from a short glance at `MeshBuilder` and related classes, it seems like you focused on the point where you can add individual `Point3f/Vector3f` objects (from the vecmath package), based on some sort of "rules", e.g. to create grids and other geometries. So this is certainly "on top" of any infrastructure that I've been experimenting with. I *think* that things like the `MeshPrimitiveBuilder` might become relevant roughly at https://github.com/chadj2/jgltf-mesh/blob/master/src/main/java/com/kinetica/mesh/GeometryBuilder.java#L297 , but again, I haven't yet looked at the details, and the `model-creation` branch is still rather experimental....

(In fact, 6 years ago, I created a `Mesh` library, with all the infrastructure for HalfEdge-based meshes and all the levels of abstraction that one could imagine, but it never made its way to GitHub :-/)

Maybe the projects could also converge somehow: If you have identified certain requirements or usage patterns (i.e. an API that you'd consider useful to build your classes on top of that) maybe we can talk about that at some point.

chadj2:
This the main handoff from the geometry engine to the buffer serializer is in `com.kinetica.mesh.GeometryBuilder.buildBuffers(GltfWriter, MeshPrimitive)`.

https://github.com/chadj2/jgltf-mesh/blob/master/src/main/java/com/kinetica/mesh/GeometryBuilder.java#L351
```
        // flush all buffers to the primitive
        this._indices.build(_geoWriter, _meshPrimitive);
        _vertices.build(_geoWriter, _meshPrimitive);
        _texCoords.build(_geoWriter, _meshPrimitive);
        _normals.build(_geoWriter, _meshPrimitive);
        _tangents.build(_geoWriter, _meshPrimitive);
        _colors.build(_geoWriter, _meshPrimitive);
```

If I could break the library into 2 parts this is where it would happen. The serializers are each populated with data and passed a meshPrimitve. They add data to the GLTF JSON and buffer and they update the meshPrimitve. 

This is is the part where your work could be replaced or merged.

javagl:
I've added this one. The task to put the JglTF-libraries into a proper table (similar to the existing ones) remains on my TODO list, though...

Issue #1638, opened 2019-07-07
Clarify color space for vertex colors


vilya:
I think the spec needs some clarification about whether per-vertex colors are required to be linear or sRGB values.

There is a line in the "Metallic-Roughness Material" section of the spec which says

> In addition to the material properties, if a primitive specifies a vertex color using the attribute semantic property COLOR_0, then this value acts as an additional linear multiplier to baseColor.

which implies that the color values are supposed to be linear, but linear could also have other meanings in this context. Then there's also the line in the "Coordinate System and Units" section that says

> RGB color values use sRGB color primaries.

which seems to imply that vertex colors should actually be sRGB instead.

If vertex colors are indeed supposed to be linear, then perhaps the wording in the spec should be changed to something like:

"In addition to the material properties, if a primitive specifies a vertex color using the attribute semantic property COLOR_0, then this value acts as an additional multiplier to baseColor. The vertex colors are required to be stored as linear values and do not require conversion from sRGB."

emackey:
Vertex colors are linear, not sRGB.

> > RGB color values use sRGB color primaries.
> 
> which seems to imply that vertex colors should actually be sRGB instead.

No, this is talking about colorspace reproductions, not linear/sRGB transfer functions.  There's a long discussion of this in https://github.com/KhronosGroup/glTF/pull/1606#discussion_r281733808, but it seems it's still a possible source of confusion.

vilya:
Thanks for the clarification @emackey. For me at least I didn't read "linear multiplier" as meaning the vertex colors were linear, I took it to mean that the multiplication was linear (as opposed to polynomial or exponential, I guess?). 

Sketchfab's glTF exporter gets this wrong (it writes out vertex colors in linear or sRGB, depending on a per-model setting) so it seems I'm not the only one to be confused by the wording.

atteneder:
Interesting decision though. A little remark:

To me it's out of question that calculations have to be done in linear space.

But sRGB does a better job at storing colors perceptually uniform, especially at limited bit depth like UNSIGNED_BYTE ( here's a great, interactive explanation of this https://ciechanow.ski/color-spaces )

Of course, converting color space at run-time, prior to uploading the colors to the GPU (or even converting them in the shader) is not not a good idea.

So i guess if you want to have precise vertex colors, export them as FLOAT or at least UNSIGNED_SHORT.

lexaknyazev:
@atteneder 
[...]

PR #1698, opened 2019-10-29
KHR_materials_transmission

MiiBond:
Here's a quick draft of a possible KHR_materials_transmission to handle thin-surface optical transparency.

There are some open questions, especially around how we define refraction. A thin surface is, of course, not physically correct and, without thickness, really shouldn't refract light. However, it seems reasonable to expect micro-facets in a rough surface to refract light and blur the transmitted light. So micro-facet refractions but not large-scale refraction?
Also, in real-time implementations, how do we propose recommended ways to simulate this blurred refraction?


proog128:
> A thin surface is, of course, not physically correct and, without thickness, really shouldn't refract light. However, it seems reasonable to expect micro-facets in a rough surface to refract light and blur the transmitted light. So micro-facet refractions but not large-scale refraction?

Maybe this extensions should be called `KHR_materials_thin_transmission` to make clear that it assumes a thin surface. As a follow-up, we could add a `KHR_materials_volumetric_transmission` extension that defines refraction for closed meshes, including parameters for absorption and scattering.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,126 @@
...
+*f*<sub>*specular*</sub> = *F* * *G* * *D* / (4 * dot(*N*, *L*) * dot(*N*, *V*))
+, where *F* is the Surface Reflection Ratio.
+
+*f* = *f*<sub>*diffuse*</sub> + *f*<sub>*specular*</sub>
+
+Optical transparency does not require any changes whatsoever to the specular term so we only want to modify the diffuse term to account for transmitted light as well. We can do this by blending *f*<sub>*diffuse*</sub> with the transmitted light, **L**<sub>transmitted</sub>, based on the `transmission` values.
+
+*f* = mix(**f**<sub>*diffuse*</sub>, **L**<sub>transmitted</sub> * `baseColor`, **T**) + *f*<sub>*specular*</sub>
+
+Where **L**<sub>transmitted</sub> is the colour of light coming through the material, towards the eye. It is tinted (multiplied) by the `baseColor`, and already has refraction taken into account (as discussed below). *T* is the transmission value defined by this extension.
```
proog128: How is **L**<sub>transmitted</sub> computed? I guess it includes (1-F)?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+<figure>
+  <img src="./figures/ConstantTransmission.png"/>
+<figcaption><em>The baseColor of the material is used to tint the light being transmitted.</em></figcaption>
+</figure>
+
+### Modeling Refraction
+
+Since the surface is considered to be infinitely thin, we end up with a non-physically correct model. One the one hand, the material has no volume so light shouldn't be refracted when passing through. However, it should be expected that microfacets in a rough surface would refract the transmitted light in such a way as to blur the transmitted image. `How do we define this, mathematically??? By its nature, it's non-physically-based.`
```
proog128: Do you consider the effect of the index of refraction on thin refraction? E.g., as the index of refraction approaches 1, does the blur effect disappear?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: I don't have a description of **L**<sub>transmitted</sub> in here yet but, yes, it would have to include 1-F. The IOR for that calculation would come from the base material (or an extension that overrides the base F0 of 0.04).

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Yes, I think IOR should affect the amount of blur but I wasn't sure how to define anything mathematically yet. How much does an infinitely-thin surface refract light? Well, none...

There would be a (1-F) affect from light hitting the micro-facets, I guess, but no refraction.

Unless I'm not thinking about this correctly.

MiiBond: Am I right in assuming that we do want a refractive blurring effect, even with an infinitely thin surface?
What do offline renderers do in this case? That should be our target reference.

For real-time, we can provide a recommended implementation but I think we need a higher-quality target so that real-time renderers are free to set their sights higher.

proog128: > How much does an infinitely-thin surface refract light?

If I am not mistaken, it depends on the orientation of the microfacets on the two sides of the object enclosing the infinitesimal small volume. If the facets on both sides are aligned such that the light ray entering the thin "volume" exists through a surface that has exactly the same orientation, there is no refraction. On the other hand, if the orientation is different, refraction will occur (if IOR != 1). This can be approximated with a microfacet lobe that is centered around the reflection direction flipped to the lower hemisphere wrt. the macro-surface normal. Here is an example how it is implemented in MDL for offline rendering: [libbsdf, line 514](https://github.com/NVIDIA/MDL-SDK/blob/2019.1.4/src/mdl/jit/libbsdf/libbsdf.cpp#L514). Another good source regarding offline rendering is the [Thin Surface BSDF from Imageworks (slide 34)](https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides_v2.pdf). They also show comparisons between actual modelled thin surfaces and their thin BTDF, and derive a roughness mapping that includes the IOR.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: I've cleaned up the math a bit here.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Thank you. I've modified the spec to reflect this.

cesss:
I'm in real need of having transparency with refraction supported in GLTF. However, GLTF has focused in PBR materials (which I consider a very good choice, BTW), so of course I feel this contradiction: what I need is thin-surface transparency, but I want to work with PBR materials definitions.

So, just wondering (and hoping it's not a dumb question), would it be possible to implement a physically valid transmission model, as an extension to PBR materials, but having some parameters that let you adjust the physically-valid model so that it behaves as thin-surface even if there are no real materials with that behaviour in nature?

My contradiction is of course that I'd like to use the refraction model from the 80s raytracers but at the same time keeping everything into a PBR material definition rather than having one definition for PBR materials and another definition for non-PBR materials.

MiiBond:
If I understand you correctly, you’re basically just asking to have refractions supported in this extension?
I’ve considered this but I think it belongs in the KHR_materials_volume_transmission extension (or whatever we call it). I’ve been held up with other things lately that have prevented me from finishing the first draft of that extension but hope to get back to it soon.

cesss:
Well, I was not thinking in what part belongs to each extension, but in being able to encode (in glTF files) a PBR-correct BSDF (BRDF together with BTDF --and BTDF implies refraction). In other words, to encode (in glTF) a parametric definition for physically-correct BSDFs. 

However, because for some applications I'd like to render transmission using the old-fashioned transparency+refraction model as implemented in older raytracers, this creates a sort of contradiction to me: To use a different BSDF for encoding materials that shall be rendered with PBR raytracers and another non-physically-correct BSDF when I'll be rendering with older raytracers. Or (much more preferably) to use always the same BSDF encoding (a BSDF that somehow lets you define old-fashioned transparency+refraction together with physically-correct BTDF, just depending on the way you tune the parameters). I prefer this approach to having two different BSDF models.

Now, this can be implemented in one extension (i.e: KHR_PBR_BSDF), in two extensions (i.e: KHR_PBR_BRDF and KHR_PBR_BTDF), or in a large number of extensions that, when used together, let you define a BSDF. But I wasn't thinking in the extensions (I tend to believe that the lesser extensions, the better, but that's just my opinion).

proog128:
> To use a different BSDF for encoding materials that shall be rendered with PBR raytracers and another non-physically-correct BSDF when I'll be rendering with older raytracers.

Do you have an example implementation for the non-physically-correct BSDF? What makes it non-physical? glTF provides non-physical transparency as alpha coverage with different blend modes. This can be combined with both thin optical transparency (this extension) or volumetric/refractive optical transparency (a potential future extension).

> I’ve considered this but I think it belongs in the KHR_materials_volume_transmission extension (or whatever we call it). I’ve been held up with other things lately that have prevented me from finishing the first draft of that extension but hope to get back to it soon.

@MiiBond I wasn't aware that you are working on a volume extension ([this one](https://github.com/MiiBond/glTF/blob/KHR_materials_volume_v1/extensions/2.0/Khronos/KHR_materials_volume/README.md)?) and also started to work on a draft. In case you are interested, you can have a look [here](https://github.com/DassaultSystemes-Technology/glTF/blob/KHR_materials_volume/extensions/2.0/Khronos/KHR_materials_volume/README.md). I was planning to create a pull request in the next days, but now I am not sure how to proceed. Two separate pull requests? Or one pull request on which we can work together?

MiiBond:
@proog128, yes, that's the one (though I hadn't pushed changes from last month yet). Yours is far more detailed so I think we should just continue editing off of that. We chose a different parameterization but I don't have strong feelings either way about that.

I do think we do need to consider a couple of issues for realtime rendering (since that's the primary market of glTF). For example, I think we should allow for representing volume on non-closed meshes with a thickness texture (with min and max values as other params). But this conversation can wait until you open a PR.

proog128:
@MiiBond OK, I will create a pull request for the extension. I like the detailed explanations in your proposal, we should them somehow.

I don't have strong feelings for the parameterization either. In the Enterprise PBR material, we chose parameters that are more user-friendly (attenuation (extinction) color and distance, multi-scattering albedo), but for the extension I thought something more "basic", closer to quantities used in physics, is better for compatibility with other material models (Standard Surface, MDL, ...). In the end there is a lossless conversion between them anyway. At least if the conversion is computed on spectral color values (RGB triples are problematic). For this reason, I also thought about allowing different parameterization, so that renderers working on spectral data can do the conversion with spectrums. Let's discuss this also in the pull request.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Contributors
+
+* Mike Bond, Adobe, [@miibond](https://github.com/MiiBond)
+
+## Status
+
+Work in Progress
+
+## DependenciesI
```
emackey: `DependenciesI` -> `Dependencies`

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,138 @@
...
+The amount of light that is transmitted by the surface rather than diffusely re-emitted. This is a percentage of all the light that penetrates a surface (i.e. isn’t specularly reflected) rather than a percentage of the total light that hits a surface. A value of 1.0 means that 100% of the light that penetrates the surface is transmitted through. This is the default.
+
+<figure>
+  <img src="./figures/ConstantTransmission.png"/>
+<figcaption><em>Defaulting to 1.0 allows making a material transparent by simple defining the extension.</em></figcaption>
+</figure>
+
+### transmissionTexture 
+
+A greyscale texture that defines the amount of light that is transmitted by the surface rather than absorbed and re-emitted. A value of 1.0 means that 100% of the light that penetrates the surface (i.e. isn’t specularly reflected) is transmitted through. The value is linear and is multiplied by the transmissionFactor to determine the total transmission value. Note that use of transmission values does not disallow usage of alpha coverage (via `baseColor.a`).
```
jackcaron: Could it be a single channel selected for this? If it's red, we could pair it with the metallic roughness texture.

extensions/2.0/Khronos/KHR_materials_transmission/examples/transmission_example.gltf
```diff
@@ -0,0 +1,1117 @@
...
+                },
+                "metallicFactor": 0.0,
+                "roughnessFactor": 0.0
+            },
+            "alphaMode": "BLEND",
+            "name": "BlueTransWithMask",
+            "doubleSided": true,
+            "extensions": {
+                "KHR_materials_transmission": {
+                    "ior": 1.7999999523162842,
```
jackcaron: is `ior` part of this?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Thanks. Finally addressed this (only 3 months late).

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: The red channel of the metallic-roughness texture can already be used for ambient occlusion.
However, I had thought that we could share this texture with the KHR_materials_volume (#1726) extension which could use one of the other channels for thickness information (if we go the route of allowing the thickness param).

MiiBond: Dispersion (the prism effect) is specifically not part of this extension to keep things simple. It could certainly be included in another extension though.

extensions/2.0/Khronos/KHR_materials_transmission/examples/transmission_example.gltf
MiiBond: Oops. That's left over from some previous work. No, this extension will use the surface IOR defined by either the core spec or another extension.

MiiBond: Thanks for pointing that out. I've removed them.

rsahlin:
I think this extension is well thought out and would probably suit the needs of Ikea very well - thanks for all your work!

I wonder why the IOR is not included in this extension?
My assumption would be that as soon as you have transmission (light entering the surface) the light would refract due to change in the medium?
With that said I would think it makes sense to either include IOR in this extension or declare dependency to #1718

MiiBond:
Thanks, @rsahlin.
IOR will be included as a separate extension as other extensions will depend on it. This extension, by itself, can't refract as it has no thickness. Both sides of the thin shell are parallel with no thickness in between so any bending of the light is reversed immediately on exit. My current idea is to tie refraction to the extension that adds volume/thickness to the material. I'm working on a proposal for the next PBR-focused working group call.

rsahlin:
Ok, thanks for the clarification.

I'm just wondering how we deliver multiple PR/extensions in a way that makes it understandable and easy for implementers?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,147 @@
...
+The amount of light that is transmitted by the surface rather than diffusely re-emitted. This is a percentage of all the light that penetrates a surface (i.e. isn’t specularly reflected) rather than a percentage of the total light that hits a surface. A value of 1.0 means that 100% of the light that penetrates the surface is transmitted through. This is the default.
+
+<figure>
+  <img src="./figures/ConstantTransmission.png"/>
+<figcaption><em>Defaulting to 1.0 allows making a material transparent by simple defining the extension.</em></figcaption>
+</figure>
+
+### transmissionTexture 
+
+A greyscale texture that defines the amount of light that is transmitted by the surface rather than diffusely re-emitted. A value of 1.0 means that 100% of the light that penetrates the surface (i.e. isn’t specularly reflected) is transmitted through. The value is linear and is multiplied by the transmissionFactor to determine the total transmission value. Note that use of transmission values does not disallow usage of alpha coverage (via `baseColor.a`).
```
rsahlin, line 70: If alpha coverage is not needed could we have a solution where the transmissionTexture  scalar is taken from baseColor.a?

This would optimize use of samplers and bandwidth for many usecases.

rsahlin, line 70: Multiplication of texture and scale value is standard for glTF - however, do we see the need to keep this.
Ie, will the texture be used with transmissionFactor != 1.0?

UX3D-nopper: If one want to use transmission without a texture, one can use the `transmissionFactor` to set the value per material. In this case, someone does not need to store the `transmissionTexture`, which does save a lot of resources.
Using the `transmissionTexture` without a factor makes sense from your side, so no need to deep dive into this case.
To have both in combination does allow to have e.g. variations. Same texture, the scalar is doing the variation. Also, if both defined, the calculation somehow needs to make sense.

In general, this is glTF style and I do not see any reason to change this. It always have been a `logical or` and there is no need to switch to an `either or`. @lexaknyazev

The effort to implement this in a shader is minimal and the effort grows linearly and not exponentially.
Finally, maybe the scalars/vectors are animated in the future.

So, at this point of time, this discussion is out of scope.

emackey: I agree, this discussion feels out of scope here.  @rsahlin, you can detect `*Factor == 1.0` for any glTF factor at the app level, and use a shader compiler directive to get rid of the extra multiply, if you believe it's problematic.  But in practice I don't think it's something we need to shift strategy on for future extensions.

emackey: I'm a little torn on this.  If there were a good fallback path to be had here, where non-transmission-aware renderers could do the best fallback by finding the transmission values mixed into the alpha coverage channel, maybe that would be a reason to try this?

But for non-fallback behavior, this seems like a problem.  For example Blender keeps these values separate for a reason, and it's possible to use both on the same material.

rsahlin: > The effort to implement this in a shader is minimal and the effort grows linearly and not exponentially.
> Finally, maybe the scalars/vectors are animated in the future.

@UX3D-nopper My question is who uses the scale factor != 1.0 together with a texture.
Not how it is declared in the specification or how to implement in shader.

So, the answer is:
>In general, this is glTF style and I do not see any reason to change this.

That's fine!

rsahlin: @emackey Yes, I do know how to handle shader variants using one super-shader :-)

>But in practice I don't think it's something we need to shift strategy on for future extensions.

Sure, that's fine.

emackey:
@MiiBond 

> This extension, by itself, can't refract as it has no thickness. Both sides of the thin shell are parallel with no thickness in between so any bending of the light is reversed immediately on exit.

Can this description be included in the README, to explain why IOR is not used here?

MiiBond:
It's stated in the section called *Modeling Refraction*.
> Since the surface is considered to be infinitely thin, we will ignore macroscopic refraction caused by the orientation of the surface.

However, I agree that it could certainly be more explicit. I think I had originally started writing this as a fully-featured transparency extension rather than a thin-surface one so there's a lot in there about refraction that is probably a bit misleading.

IOR should still be used for surface reflection.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
donmccurdy: I agree with @jackcaron that we should pick a single channel, for two reasons:

1. By requiring that three channels be filled with the same value, we are losing any packing flexibility that might (intentionally or not) have been available. The fact that the <kbd>r</kbd> channel in the ORM texture could be used for either AO *or* transmission is a good thing.
2. If in the future we have a general-purpose swizzle extension, it will be awkward that it's required to provide a 3-channel output for a 1-channel value here.

Either <kbd>r</kbd> or <kbd>a</kbd> seem like reasonable choices, to me.

donmccurdy:
Draft implementation in Blender:

https://github.com/KhronosGroup/glTF-Blender-IO/pull/1094

extensions/2.0/Khronos/KHR_materials_transmission/README.md
donmccurdy: Yes. Use of both alpha blend and transmission seems sketchy to me, but there may be reasons to do so. Use of both alpha *mask* and transmission seems totally reasonable. I think it would be unsafe for this extension to try to alter its behavior based on the alphaMode (and as @emackey points out, Blender avoids that too) so I think the extension is probably correct as written.

donmccurdy: You _could_ make an argument that the `transmissionTexture`'s single channel should be `a` instead of `r`? But I'm having trouble convincing myself whether that's a good idea... using `baseColorTexture` and `transmissionTexture` with the same texture would cause problems in some engines — it is difficult to really ensure that a 0% alpha channel doesn't have leak-through side effects even when the material is marked as opaque. Especially if that texture is being reused elsewhere _with_ the alpha channel...

donmccurdy:
We may need to state something about implementing (or not implementing) fallbacks. Here are my thoughts, but let's discuss a little before writing it into the extension?

1. Exporters should only specify both `alphaMode != OPAQUE` and `transmission > 0` if they actually want both of those behaviors to be applied. Alpha blending cannot be used as a 'fallback' to transmission.
2. Clients supporting the extension must respect both `alphaMode` and `transmission` independently. As stated above, transmission does not disable alpha.
3. Clients that cannot fully implement physically-based transmission may approximate it with alpha blending. We encourage clients to study the provided implementation guidelines and aim for full optical transparency, but fallback to alpha blending is preferable to ignoring the extension entirely.
4. Clients that completely fail to recognize the transmission extension will show no transparency at all; this is the expected behavior.

I think all of this follows from the section on "Blending" already, but want to confirm this is everyone's understanding?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: I think these two things are very different concepts and we shouldn't do anything to confuse them for the user. Besides, using baseColor.a for transmission would require taking the inverse of the value which further confuses things.

MiiBond:
1. Yes.
2. Yes.
3. Yes.
4. They could certainly still render opacity via baseColor.a but, yeah, the rest would be opaque. I've been trying to avoid the word "transparency" to avoid confusion between transmission and opacity (i.e. baseColor.a).

donmccurdy:
(4) Yes, good point and I should clarify. The most common use of transmission will be to enable this extension and leave alphaMode=OPAQUE, this is what most exporters will and should do. A client that is completely unaware of the transmission extension will then render the material as opaque. A client that is aware of the extension but unable to implement optical transparency may choose to approximate transmission with alpha blending.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,147 @@
...
+  <img src="./figures/TransmissionTexture.png"/>
+<figcaption><em>Controlling transmission amount with transmissionTexture.</em></figcaption>
+</figure>
+
+## Implementing Transmission ##
+
+From the core [glTF BRDF](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#appendix-b-brdf-implementation), we have:
+
+*f* = *f*<sub>*diffuse*</sub> + *f*<sub>*specular*</sub>
+*f*<sub>*diffuse*</sub> = (1 - *F*) * *diffuse*
```
abwood, line 105: There is some odd formatting here in the rendered result (newlines are missing).

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,147 @@
...
+*f*<sub>*specular*</sub> = *F* * *G* * *D* / (4 * dot(*N*, *L*) * dot(*N*, *V*))
+, where *F* is the Surface Reflection Ratio.
+
+We will now add an additional term for the transmitted light:
+
+*f* = *f*<sub>*diffuse*</sub> + *f*<sub>*specular*</sub> + *f*<sub>*transmission*</sub>
+
+*f*<sub>*transmission*</sub> = (1 - *F*) * *T* * *D<sub>T</sub>* * *baseColor*
+where *T* is the transmission percentage defined by this extension's `transmission` and `transmissionTexture` properties and *D<sub>T</sub>* is the distribution function for the transmitted light. The distribution function is the same Trowbridge-Reitz model used by specular reflection except sampled along the view vector rather than the reflection. The *baseColor* factor causes the transmitted light to be tinted by the surface.
```
abwood, line 114: I personally would benefit from clarity here with regards to `Dt`.  This states that `Dt` is the Towbridge-Reitz distribution function, but sampled along the view direction rather than the reflected direction.  The microfacet distribution function, D, is calculated using dot(n, h).  Is it correct that this should instead use dot(n, -v) in this calculation for when we deal with transmission? Is it that simple?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+Light that penetrates a surface and is transmitted will not be diffusely reflected so we also need to modify the diffuse calculation to account for this.
+*f*<sub>*diffuse*</sub> = (1 - *F*) * (1 - *T*) * *diffuse*
+
+Optical transparency does not require any changes whatsoever to the specular term. So this can all be rearranged so that the transmitted light amounts to a modifaction of the diffuse lobe.
+*f* = (1 - *F*) * (*T* * *D<sub>T</sub>* + (1 - *T*) * *diffuse*) + *f*<sub>*specular*</sub>
+
+
+
```
abwood: Should there be guidance on how transmission works with IBL?  A hint here would be useful: When dealing with light contributions from an IBL source, your IBL environment map is sampled in the same manner as with the surface reflection, but for transmission we use the view vector rather than the reflected view vector.  

I assume that is how it works.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,147 @@
...
+### Refractive Blurring
+Some rasterizer implementations may opt to sample prefiltered IBL or dynamic light probes to simulate refractive blurring due to surface roughness. Others may choose to sample from a blurred version of the rendered scene behind a surface. Raytracers are expected to use more physically correct refraction with a BSDF supporting multi-scattering like the modified [Smith](https://eheitzresearch.wordpress.com/240-2/) model.
+
+<figure>
+  <img src="./figures/Roughness.png"/>
+<figcaption><em>Refraction due to surface roughness.</em></figcaption>
+</figure>
+
+**TODO** - *This is fine for raytracers or for rasterizers that just sample the existing IBL for refracted light. However, what about rasterizers that want to cheaply render transmission in screen-space (I think this is pretty common)? Typically, the refractive blurring is done by sampling mips or manually-blurred versions of the background scene. This can be eye-balled by an implementation but is there a slightly more mathematically-rigorous version that we could recommend?*
```
abwood: I do think it would be really helpful here to include some best practices on how rasterizers can produce plausible background blurring through rough surfaces.  Are there good references that you can provide here?  What are other rasterizers doing for this?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Blending
+
+The glTF `blendMode` is used for alpha-as-coverage, NOT for physically-based transparency (i.e. this extension). If alpha-as-coverage is not being used, the blend mode of the material should be set to "OPAQUE" even though it is transparent. Again, it's helpful to think of alpha-as-coverage as whether the physical surface is there or not. `transmission` applies to the surface material that exists.
```
abwood: I did notice that in the sample model, all materials are flagged as "BLENDED".  While on the topic of the sample model, it requires the KHR_texture_transform extension; that dependency should be removed.

abwood:
I think we should also include links to any reference materials that were used to derive this specification.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+Work in Progress
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+Many optically transparent materials are impossible to represent in a physically plausible manner with the core glTF 2.0 PBR material. This is because the core specification only includes the concept of "alpha as coverage" (exposed via the alpha channel of `baseColorTexture`). Many common materials like glass and plastic require significantly different modeling.
```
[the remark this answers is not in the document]
donmccurdy: > Also, I believe this has already been discussed, but should this extension be renamed to thin_transmission?

I think I prefer `KHR_materials_transmission`. While this extension _alone_ models only thin surfaces, it is also the foundation for future volume transparency extensions, which will use information in this extension. In other words, this extension does not indicate that a surface is thin: that is the default state. Other extensions are required to make a surface "thick."

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,147 @@
...
+]
+```
+
+## Properties 
+
+Only two properties are introduced with this extension and combine to describe a single value; the percentage of light that is transmitted through the surface of the material. These properties work together with the existing properties of the material to define the way light is modified as it passes through the substance. 
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**transmissionFactor** | `number` | The base percentage of light that is transmitted through the surface.| No, default:`1.0`|
```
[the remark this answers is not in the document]
emackey: Somehow this review came out of sync with the latest version, but the default has been updated to the correct value.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,147 @@
...
+|**transmissionTexture** | [`textureInfo`] | A greyscale texture that defines the transmission percentage of the surface. This will be multiplied by transmissionFactor. | No |
+
+### transmissionFactor 
+
+The amount of light that is transmitted by the surface rather than diffusely re-emitted. This is a percentage of all the light that penetrates a surface (i.e. isn’t specularly reflected) rather than a percentage of the total light that hits a surface. A value of 1.0 means that 100% of the light that penetrates the surface is transmitted through. This is the default.
```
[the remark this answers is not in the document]
emackey: This part didn't get updated, and still shows the old default value.  So this line still needs a correction.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Good catch. Only the ones with opacity should have had that. This is fixed now.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: I actually removed all of this. I wonder if it belongs in the spec. or perhaps in a different doc?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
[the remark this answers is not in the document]
MiiBond: Thanks. Fixed.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: I think the `dot(n, h)` term shouldn't come into play at all since the transmitted light isn't being reflected, but I could be wrong. The incoming light direction should be sampled along the view direction because you're looking through the material and are seeing completely unrefracted light. The transmitted light is then scaled by the Fresnel to account for the portion of light reflected off the inside surface of the material (and, therefore, not reaching your eye).
Not sure at all if this is correct so it would be good to have others (like @proog128) look at this.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
emackey: Newlines are strange in markdown.  You actually need two trailing spaces at the end of a line to force a newline.

Alternately, you could place a blank line between the two formulas.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
emackey: Since the sample model has appeared in https://github.com/KhronosGroup/glTF-Sample-Models/pull/265, can the sample files associated with this PR be removed?  I'd like for there not to be two sources of sample models (although I realize some other extensions already violate this).

extensions/2.0/Khronos/KHR_materials_transmission/README.md
bsdorra: To my understanding, the dot product is relevant as we still evaluate the microfacet distribution. There's a brief section on the topic in the DSPBR spec https://bit.ly/2NpmJVn (eq. 5 and fig. 4). We just use a modified half-vector, which uses the flipped light direction.

MiiBond: Yes, you're right. You'd flip it across the plane defined by the surface normal and then it would be centered around the same direction as the view.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Oh, weird. You're right. I was previewing it in VSCode with a plugin and it displayed it fine.
I'll add some newlines.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Will do.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
echadwick-wayfair: I would propose this extension use the r channel of a texture separate from the ORM, as I would not want to have to omit occlusion in order to take advantage of per-pixel transmission. Pre-calculated ambient occlusion has a strong visual benefit.

Perhaps we could use a channel of another texture, and reserve the green and blue channels for other PBR Next features? Transmission (r), clearcoatTexture (g), and clearcoatRoughnessTexture (b) for example.

donmccurdy: We don't need to choose a specific texture to pair transmission with — as long as the channels don't conflict, users can mix and match. Any of (transmission; occlusion/rough/metal), (occlusion; transmission/rough/metal), (transmission; occlusion; ____/rough/metal) would be allowed.

MiiBond: We've been discussing using the 'G' channel for thickness over in the [KHR_materials_volume](https://github.com/KhronosGroup/glTF/pull/1726) thread.

rsahlin: I agree that the texture should be one channel (prism not being part of this extension).
As for which channel to choose I would propose either R or A.
Picking A would make it easy to have occlusion/roughness/metallic/transmission (without swizzle)
Picking R could have benefit if occlusion is less likely to be used with this extension.

Either way is fine to me, I will solve by internally swizzling the textures if needed.

MiiBond:
I just updated the implementation note about minimum expectations regarding transparency layering. Also, added a figure to demonstrate what I mean.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,129 @@
...
+<figure>
+  <img src="./figures/TransmissionWithMask.png"/>
+<figcaption><em>Alpha coverage and optical transparency can be used at the same time so that some areas of a surface are transparent while others disappear entirely.</em></figcaption>
+</figure>
+
+## Transparent Metals
+Metals effectively absorb all refracted light (light that isn't reflected), preventing transmission.
+The metallic parameter of a glTF material effectively scales the `baseColor` of the material toward black while, at the same time scaling the F0 (reflectivity) value towards 1.0. This makes the material opaque for metallic values of 1.0 because transmitted light is attenuated out by absorption. Therefore, for a material with `metallicFactor=1.0`, the value of `transmissionFactor` doesn't matter.
+
+## Implementing Transmission ##
```
rsahlin, line 101: Please make it clear that this is an implementation note:

> **Implementation Note:**

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+where *T* is the transmission percentage defined by this extension's `transmission` and `transmissionTexture` properties and *D<sub>T</sub>* is the distribution function for the transmitted light. The distribution function is the same Trowbridge-Reitz model used by specular reflection except sampled along the view vector rather than the reflection. The *baseColor* factor causes the transmitted light to be tinted by the surface.
+
+Light that penetrates a surface and is transmitted will not be diffusely reflected so we also need to modify the diffuse calculation to account for this.
+*f*<sub>*diffuse*</sub> = (1 - *F*) * (1 - *T*) * *diffuse*
+
+Optical transparency does not require any changes whatsoever to the specular term. So this can all be rearranged so that the transmitted light amounts to a modifaction of the diffuse lobe.
+*f* = (1 - *F*) * (*T* * *D<sub>T</sub>* + (1 - *T*) * *diffuse*) + *f*<sub>*specular*</sub>
+
+## Implementation Notes
```
rsahlin, line 124: I'm leaning towards that we should require a painter algorithm (depth sorting) of triangles - in my opinion this should be normative.  
Making depth sorting of triangles normative will make a huge step towards consistent result across viewers (bar blurring and possibly other effects)
Without it whole objects may pop in and out of existence......

However, I'm not sure how to approach blurring since I think this will be the most 'heavy' effect to implement for multiple layers/objects.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+This extension aims to address the simplest and most common use cases for optical transparency: infinitely-thin materials with no refraction, scattering, or dispersion. Dealing exclusively with “thin” materials (i.e. materials where only the surface is considered and not the volume) allows many simplifications to be made when calculating things like refraction and absorption.
+
+## Extending Materials
+
+A transparent material is defined by adding the `KHR_materials_transmission` extension to any glTF material. When present, the extension indicates that a material should be rendered as a transparent surface and be blended as defined in the following spec. Note that, as rendering transparent objects presents many difficult-to-solve issues with primitive ordering, this extension does not dictate rendering algorithms.
```
rsahlin, line 35: I think that we should make a painters (depth sorting) algorithm normative - otherwise we risk objects 'popping' in and out.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Depth sorting of triangles is not going to be reasonable on all platforms due to the number of passes required (or having the ability to use linked-lists of fragments which isn't possible in WebGL). Besides, I think rendering techniques are out of scope of glTF. Yes, there will be differences between renderers but as long as they do the blending in the way described, I think that's enough.
Blurring of the transmission is pretty cheap if they just sample the mips of the background texture. The background render texture will probably be required anyway to do the appropriate blending.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: This section is normative. The non-normative "Implementation Notes" are below this section.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
emackey: Mike is correct that specific rendering techniques are out of scope for normative glTF spec..  This particular suggestion would prohibit many real-time implementations, which is not what we want.

Separately, 3D Commerce may require something along these lines for certification, but beware that you may be excluding more clients than you bargained for here.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
donmccurdy: I agree that sorting triangles is prohibitively expensive. Maybe a dedicated model viewer (note: model viewers are a subset of glTF clients) can do that when the camera and object are stationary, but it's not a requirement we should have in the extension spec.

But on a similar point — the core specification says this about alpha blending:

> Whether depth value is written or whether to sort is up to the implementation.

In retrospect, I think we could have made a normative requirement that depth _not_ be written for alpha blend materials. Does that same decision apply to transmission? Or is this not an issue, because transmissive objects are treated as opaque (or rendered in an entirely separate pass)?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
rsahlin: I think it could pose a performance problem for a particular type of real-time implementations - mostly webGL / script based ones.
The cost would be performance - however the gain would be predictable results.
Depending on implementation (language and platform) you would get varying degrees of performance penalty, for instance a C based engine could do it on the CPU without too much overhead.
A Vulkan based engine could do it on the GPU.

I would not want us to limit this due to current viewer implementation bottlenecks - it could be a good motivator to push development forward.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
rsahlin: > Depth sorting of triangles is not going to be reasonable on all platforms due to the number of passes required (or having the ability to use linked-lists of fragments which isn't possible in WebGL).

I agree that if all layers have to support blurring an engine could require multiple passes (render to texture, minify then sample) which could become a performance bottleneck.
However - 'just' sorting the triangles of transmission objects could be done in a single pass if effects needing to sample the objects behind are left out.
This would leave us with viewers that would at least display transparent objects in the correct order and visibility.

Yes, the blurring itself is quite cheap - it's having to render out to texture, forcing multiple passes, that will be mostly heavy.

I also do not like us to be so tied to WebGL - just because something can not be done in WebGL should, in my opinion, not mean that we leave it out.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
UX3D-nopper: The rendering should not be normative.
Depending on the rendering method (forward vs. deferred plus it's variations), different and better algorithms do exist.
E.g. pixel sorting (aka order independent transparency) should not be excluded.
From my point of view, we should point to non-normative approaches.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
UX3D-nopper: Hope "blurring" is just in short we do it physically - as far as possible - correct.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
emackey: @rsahlin Also consider that many glTF implementations exist within rendering libraries that are used by 3rd-party applications.  The library cannot dictate how the application structures its render passes or loops.  That's part of what makes this extension particularly challenging, as it imposes some nontrivial best practices there.  These should be as clearly spelled out as we can manage, but not marked normative.  Little more than the glTF file structure and its meaningful interpretation is normative to glTF.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
rsahlin: > Hope "blurring" is just in short we do it physically - as far as possible - correct.

Sure, that's what I mean anyway. :-)

The point I'm trying to make is that we should require a minimum of visibility through triangles using transmission.
Meaning that not only the background should show through - any other primitives (that may also use transmission) should also be visible.
However, what effects to apply to each of these primitives is optional - for instance the roughness that will result in blurring (since will require multiple passes on todays hardware)

extensions/2.0/Khronos/KHR_materials_transmission/README.md
rsahlin: I think it is a simplification calling this the 'rendering ' - to me it boils down to what is visible through the primitives using transmission.

If it's the sorting algorithm you object to then I would suggest a wording similar to:
'The background and other objects, including primitives using transmission, should be visible through the material (that is using transmission)'

abwood: As discussed on today's call, Sketchfab's refraction material has a similar baseline as what is proposed here as an implementation note: we only require the opaque scene to be sampled by the transmission material.  From their [documentation ](https://help.sketchfab.com/hc/en-us/articles/202602073-Transparency):
- A refractive object can not refract itself.
- A refractive object can not refract other refractive objects.
- There can be sorting issues when mixing refractive and transparent objects.
- There can be artifacts when the camera moves.
- Objects outside of the field of view are not refracted.

And for convenience, here are some [sample models](https://sketchfab.com/Sketchfab/collections/refraction-examples) that demonstrate their refraction material.

Sketchfab has established this as an acceptable baseline for their community and I would hope that our spec would be structured in such a way that their implementation is conformant.  It seems like a reasonable baseline.

donmccurdy: Seems to be a similar situation for Blender's realtime renderer (Eevee). It cannot show refractive objects behind other refractive objects, but can display transparent and opaque objects behind refractive objects. https://blenderartists.org/t/realistic-glass-in-eevee/1149937/29

extensions/2.0/Khronos/KHR_materials_transmission/README.md
bghgary: > I also do not like us to be so tied to WebGL - just because something can not be done in WebGL should, in my opinion, not mean that we leave it out.

I agree with this in principle but WebGL will have a huge influence since we want glTF to be viewable on the web.

donmccurdy: I don't think this is a WebGL problem. Requiring that implementations sort triangles, in realtime, is extremely burdensome on any platform. Yes, there are cases where it may be possible (i.e. few enough triangles), but we really cannot expect this at the data format level.

Is there a reason these requirements cannot be in the 3D Commerce Certification process, instead?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,212 @@
...
+Alpha-as-coverage can basically be thought of as describing whether the surface exists or not. When alpha is 0, nothing is rendered while alpha=1 means that the material is solid. It's most useful for materials such as gauze or burlap that only partially cover the area in which the material is rendered. These materials contain lots of small holes where the light can pass through so, strictly speaking, the light isn't actually entering the material itself (i.e. the surface isn't there). For other materials, like clear glass, a photon of light might actually enter the surface and be transmitted through. In this case, the light can be reflected, refracted, absorbed or scattered. Simple alpha-as-coverage cannot represent any of these effects and so we need very different logic to render them correctly.
+
+The most important of these features is surely reflection. Even a completely transparent, infinitely thin, glass material is still visible due to its reflection of light. However, a material using purely alpha as coverage is not visible at all when alpha = 0 (because the surface effectively doesn't exist).
+
+<figure>
+<img src="./figures/OpacityComparison.png"/>
```
abwood, line 42: I just noticed that there appears to be some refraction in this reference image.  Does this need to be updated?

donmccurdy:
Proposed implementation note:

> **Implementation Note:** Various techniques are available to trade off physical accuracy against realtime performance requirements in client implementations. This extension does not mandate any particular technique. However, be aware that the simplest such technique — transmitting only the IBL and no scene objects — will often fall short of what a model's author intended. We recommend that client implementations aim to display at least opaque objects through a transmissive material. When that is not feasible, it may be preferable to fall back to simple alpha blending with `0 < alpha < 1`.

prideout:
Should this spec have an "exclusions" section, similar the clearcoat extension?  I'm not sure if this extension makes sense when combined with the `unlit` extension.  I would also prefer to disallow combining it with `pbrSpecularGlossiness`, to prevent bloat in ubershader-style implementations.

donmccurdy:
@prideout Agreed, I think all of the PBR Next extensions should have the [exclusions](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_clearcoat#exclusions) provided with Clearcoat.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,212 @@
...
+|---|----|-----------|--------|
+|**transmissionFactor** | `number` | The base percentage of light that is transmitted through the surface.| No, Default: `0.0 ` |
+|**transmissionTexture** | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A texture that defines the transmission percentage of the surface, stored in the `R` channel. This will be multiplied by `transmissionFactor`. | No |
+
+### transmissionFactor 
+The amount of light that is transmitted by the surface rather than diffusely re-emitted. This is a percentage of all the light that penetrates a surface (i.e. isn’t specularly reflected) rather than a percentage of the total light that hits a surface. A value of 1.0 means that 100% of the light that penetrates the surface is transmitted through.
+
+<figure>
+  <img src="./figures/ConstantTransmission.png"/>
+<figcaption><em>baseColor or yellow with transmissionFactor = 1.0.</em></figcaption>
```
prideout, line 80: I think "baseColor of" is meant instead of "baseColor or". 

I also noticed that the example model specifies a magenta baseColorFactor for the model that appears yellowish, which is presumably due to the phrasing in this spec that says:

    rather than absorbed light, we can talk about its inverse: transmitted light

So, I think this needs to be clarified.  Should we subtract baseColor from (1,1,1) ?  Should the caption really say "magenta"?

MiiBond: Good catch. Thank you.
It's fixed now.

MiiBond:
> Should this spec have an "exclusions" section, similar the clearcoat extension?

Yes. Done

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Updated. Thanks for noticing that.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+<figure>
+  <img src="./figures/TransmissionTexture.png"/>
+<figcaption><em>Controlling transmission amount with transmissionTexture.</em></figcaption>
+</figure>
+
+## Tinting
+You may notice that the transparent materials shown above are tinted. i.e. they aren't actually transmitting 100% of the non-reflected light. This is because `KHR_materials_transmission` specifies that the `baseColor` of the material be used to model the absorption of light by the surface. This allows materials like stained glass to be easily represented with this extension.
+Absorption is usually defined as an amount of light at each frequency that is absorbed over a given distance through a medium (usually described by Beer’s Law). However, since this extension deals exclusively with infinitely thin surfaces, we can treat absorption as a constant. In fact, rather than absorbed light, we can talk about its inverse: transmitted light. The `baseColor` of the material serves this purpose as it already defines how the light that penetrates the surface is colored by the material. In this model, the transmitted light will be modulated by this color as it passes through.
```
prideout, line 99: Normally, the `baseColorTexture` uses the sRGB transfer function and must be converted to linear space before it is used for any computations.  When the baseColorTexture is used tinting, should we assume that this sRGB transformation still applies?  This should perhaps be clarified.

MiiBond: Yes, the transform should definitely still apply. I can clarify in the specs. Thanks.

emackey: Let's not confuse `baseColor` with `baseColorTexture` or `baseColorFactor` here.  Only `baseColorTexture` is supplied as sRGB, the `baseColorFactor` is supplied already as linear.  The final `baseColor` is a product of the two linear values, after conversion.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,221 @@
...
+</figure>
+
+## Tinting
+You may notice that the transparent materials shown above are tinted. i.e. they aren't actually transmitting 100% of the non-reflected light. This is because `KHR_materials_transmission` specifies that the `baseColor` of the material be used to model the absorption of light by the surface. This allows materials like stained glass to be easily represented with this extension.
+Absorption is usually defined as an amount of light at each frequency that is absorbed over a given distance through a medium (usually described by Beer’s Law). However, since this extension deals exclusively with infinitely thin surfaces, we can treat absorption as a constant. In fact, rather than absorbed light, we can talk about its inverse: transmitted light. The `baseColor` of the material serves this purpose as it already defines how the light that penetrates the surface is colored by the material. In this model, the transmitted light will be modulated by this color as it passes through. Note that because `baseColor` is defined as sRGB, it must be converted to linear before being used in this calculation.
```
emackey, line 99: Actually can I propose a bit of a rewrite?  I'm concerned we've buried the lede here.

```suggestion
## Tinting
The `baseColor` of the material, as defined in the [Metallic-Roughness Material](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#metallic-roughness-material)
section of the glTF specification, defines the amount of light at each frequency
that is transmitted (not absorbed) by the material. Absorption is usually defined
as an amount of light at each frequency that is absorbed over a given distance
through a medium (usually described by Beer’s Law). However, since this
extension deals exclusively with infinitely thin surfaces, we can treat
absorption as a constant. Therefore, its inverse, `baseColor` transmission,
can also be constant.  This provides the color tinting effect in the materials
shown above.
```

donmccurdy: I like Ed's the rephrasing, but two nits — in the last two sentences could you replace the words "its" and "This" with the terms they refer to? I had a bit of trouble tracking the pronouns at the end.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+*f*<sub>*diffuse*</sub> = (1 - *F*) * *diffuse*
+
+*f*<sub>*specular*</sub> = *F* * *G* * *D* / (4 * dot(*N*, *L*) * dot(*N*, *V*)), where *F* is the Surface Reflection Ratio.
+
+We will now add an additional term for the transmitted light:
+
+*f* = *f*<sub>*diffuse*</sub> + *f*<sub>*specular*</sub> + *f*<sub>*transmission*</sub>
+
+*f*<sub>*transmission*</sub> = (1 - *F*) * *T* * *D<sub>T</sub>* * *baseColor*
```
proog128, line 144: Isn't this missing the G term and the normalization, like in f_specular?

f_transmission = (1-F) * T * baseColor * D_T * G_T / (4 * abs(dot(N, L)) * abs(dot(N, V)))

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+<figure>
+  <img src="./figures/TransmissionTexture.png"/>
+<figcaption><em>Controlling transmission amount with transmissionTexture.</em></figcaption>
+</figure>
+
+## Tinting
+The `baseColor` of the material, as defined in the [Metallic-Roughness Material](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#metallic-roughness-material) section of the glTF specification, defines the amount of light at each frequency that is transmitted (not absorbed) by the material. Absorption is usually defined as an amount of light at each frequency that is absorbed over a given distance through a medium (usually described by Beer’s Law). However, since this extension deals exclusively with infinitely thin surfaces, we can treat absorption as a constant. Therefore, its inverse, `baseColor` transmission, can also be constant.  This provides the color tinting effect in the materials shown above and is useful for real-world materials like stained glass and tinted plastics. Note that because `baseColor` is defined as sRGB, it must be converted to linear before being used in this calculation.
```
emackey, line 106: The note at the end, "baseColor is defined as sRGB", is not correct.

Don posted a handy diagram explaining baseColor here:  https://github.com/KhronosGroup/glTF/issues/1638#issuecomment-668918360

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Yes, quite right. Thank you.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: Oh, good catch. I'll just remove that statement completely.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
MiiBond: I rephrased this a bit. Maybe it's clearer now?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+*f*<sub>*diffuse*</sub> = (1 - *F*) * *diffuse*
+
+*f*<sub>*specular*</sub> = *F* * *G* * *D* / (4 * dot(*N*, *L*) * dot(*N*, *V*)), where *F* is the Surface Reflection Ratio.
+
+We will now add an additional term for the transmitted light:
+
+*f* = *f*<sub>*diffuse*</sub> + *f*<sub>*specular*</sub> + *f*<sub>*transmission*</sub>
+
+*f*<sub>*transmission*</sub> = (1 - *F*) * *T* * *baseColor* * *D<sub>T</sub>* * *D<sub>G</sub>* / *(4 * abs(dot(N, L))* * *abs(dot(N, V)))*
```
abwood, line 144: Minor, but I think you meant G<sub>T</sub>.  May want to also mention something about G<sub>T</sub> below.  There's a related conversation about this term on the [KHR_materials_volume PR](https://github.com/KhronosGroup/glTF/pull/1726#discussion_r465386060)

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+where *T* is the transmission percentage defined by this extension's `transmission` and `transmissionTexture` properties and *D<sub>T</sub>* is the distribution function for the transmitted light. The distribution function is the same Trowbridge-Reitz model used by specular reflection except sampled along the view vector rather than the reflection. The *baseColor* factor causes the transmitted light to be tinted by the surface.
+
+Light that penetrates a surface and is transmitted will not be diffusely reflected so we also need to modify the diffuse calculation to account for this.
+*f*<sub>*diffuse*</sub> = (1 - *F*) * (1 - *T*) * *diffuse*
+
+Optical transparency does not require any changes whatsoever to the specular term. So this can all be rearranged so that the transmitted light amounts to a modifaction of the diffuse lobe.
+*f* = (1 - *F*) * (*T* * *D<sub>T</sub>* + (1 - *T*) * *diffuse*) + *f*<sub>*specular*</sub>
```
abwood, line 152: Typo: modifaction.

I think this section needs to get updated to include the additions of G<sub>T</sub> / (4.0 * NdotL * NdotV) from above.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Implementation Notes
```
donmccurdy, line 154: Consider adding a sub-header...

> *This section is non-normative.*

... similar to what we've done in the [appendices of the core spec](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#appendix-b-brdf-implementation).

extensions/2.0/Khronos/KHR_materials_transmission/schema/glTF.KHR_materials_transmission.schema.json
```diff
@@ -0,0 +1,24 @@
...
+            "description": "The base percentage of light transmitted through the surface.",
+            "default": 0.0,
+            "minimum": 0.0,
+            "maximum": 1.0,
+            "gltf_detailedDescription": "The base percentage of non-specularly reflected light that is transmitted through the surface. i.e. of the light that penetrates a surface (isn't specularly reflected), this is the percentage that is transmitted and not diffusely re-emitted."
+        },
+        "transmissionTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "A greyscale texture that defines the transmission percentage of the surface. This will be multiplied by transmissionFactor.",
+            "gltf_detailedDescription": "The percentage of non-specularly reflected light that is transmitted through the surface. i.e. of the light that penetrates a surface (isn't specularly reflected), this is the percentage is transmitted and not diffusely re-emitted. This will be multiplied by the transmissionFactor."
```
lexaknyazev, line 19: Please sync the descriptions here with the README (wrt use of red channel and linear transfer function).

emackey:
@MiiBond Any chance these small fixes can be done in time for Monday?  👼  Let us know if you want patch contributions for any of them, thanks!

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -0,0 +1,228 @@
...
+- [glTF.KHR_materials_transmission.schema.json](schema/glTF.KHR_materials_transmission.schema.json)
+
+## Reference
+
+[Autodesk Standard Surface - Specular Transmission](https://autodesk.github.io/standard-surface/#closures/speculartransmission)  
+[Blender Transparent BSDF](https://docs.blender.org/manual/en/latest/render/shader_nodes/shader/transparent.html#transparent-bsdf)  
+[Enterprise PBR Shading Model - Transparency](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2020x.md.html)  
+[Filament Material models - Transmission](https://google.github.io/filament/Materials.md.html#materialmodels/litmodel/transmission)  
+[Unreal Engine 4 Material - Refraction](https://docs.unrealengine.com/en-US/Engine/Rendering/Materials/MaterialInputs/index.html#refraction)
+[Adobe Standard Material - Interior Properties](https://helpx.adobe.com/dimension/using/standard-materials.html#InteriorProperties)
```
abwood, line 175: Minor: The final MD puts the UE4 and Adobe links on the same line.

extensions/2.0/Khronos/KHR_materials_transmission/README.md
abwood: In this sentence you mention Trowbridge-Reitz by name, but we only suggest that model in Appendix B as non-normative.  Should it be mentioned here?  I would consider rewriting this as:
```suggestion
where *T* is the transmission percentage defined by this extension's `transmission` and `transmissionTexture` properties and *D<sub>T</sub>* and *G<sub>T</sub>* are the microfacet distribution and shadowing functions for the transmitted light. These *D<sub>T</sub>* and *G<sub>T</sub>* functions are the same models used by specular reflection, except sampled along the view vector rather than the reflection. The *baseColor* factor causes the transmitted light to be tinted by the surface.
```

extensions/2.0/Khronos/KHR_materials_transmission/schema/glTF.KHR_materials_transmission.schema.json
emackey: @lexaknyazev I pushed a change here, can you check it?  Thanks!

extensions/2.0/Khronos/KHR_materials_transmission/README.md
emackey: Fixed.

extensions/2.0/Khronos/KHR_materials_transmission/schema/glTF.KHR_materials_transmission.schema.json
lexaknyazev: LGTM for now. We should probably do another pass on all PBR Next extensions to align the language with the core spec as we did in #1606.

Issue #1684, opened 2019-10-10
Spec clarification: Buffer views without buffers, buffers without URIs and GLB


zeux:
This requires some context so bear with me.

# Original use case

I'm working on a new extension that allows glTF/GLB files to specify compressed buffer views, using a custom compression technique optimized for typical geometry/animation data. It's really important that the compression is applied at the bufferView level - applying it at the buffer level conflates data with multiple strides / meaning and necessitates decompressing the entire buffer into memory at once serially; applying it at the accessor level makes it impossible to upload bufferViews to the GPU efficiently, and inflates the JSON blob.

There are two important use cases I'd like to cover with this extension:

- Data is unconditionally compressed; loading the glTF file requires the extension support
- Data is provided in both compressed and raw form; loaders that don't support the extension can load the raw data from a separate buffer

To achieve this, I'm trying to use the usual technique of extension specifying "override" data for bufferView - here's an example for usecase 2:

```json
   "extensionsUsed":["MESHOPT_compression"],
   "buffers":[ 
      { 
         "uri":"BrainStem.bin",
         "byteLength":394128
      },
      { 
         "uri":"BrainStem.fallback.bin",
         "byteLength":1166032
      }
   ],
   "bufferViews":[ 
      { 
         "buffer":1,
         "byteOffset":0,
         "byteLength":136336,
         "byteStride":4,
         "target":34962,
         "extensions":{ 
            "MESHOPT_compression":{ 
               "buffer":0,
               "byteOffset":0,
               "byteLength":2646,
               "byteStride":4,
               "mode":0,
               "count":34084
            }
         }
      },
```

As you can see, there are two buffers - main buffer with compressed data and fallback buffer with uncompressed data. The bufferView refers to buffer 1 (fallback), but compression extension overrides this - if the loader supports the extension, there will not be a need to load the fallback buffer at all. Commonly used GLTF JS loaders use lazy buffer loading, loading buffers as they parse buffer views that refer to them, so this works great.

When data like this is stored in the GLB file, the JSON blob is the same except that the first, main,  buffer doesn't have a URI - this also works well.

What doesn't work quite as well is the "data is unconditionally compressed" workflow - I'd like to be able to generate a file without fallback buffers (with extension marked as required).

# Dummy buffers

[...]

Issue #1560, opened 2019-02-11
Replace GL enum values (eventually)


donmccurdy:
Mentioned briefly in https://github.com/KhronosGroup/glTF/pull/1302, and filing here to keep track. I suggest we consider it a longer-term goal to replace GL enums in the format with something API-agnostic. This is a breaking change, and can't happen in the 2.X lifecycle. Prior discussion that lead to the current design (https://github.com/KhronosGroup/glTF/issues/96, https://github.com/KhronosGroup/glTF/issues/148) predates glTF 2.0 and our support for other graphics APIs. While GL enums certainly don't _prevent_ anyone from rendering a glTF 2.0 model on DirectX, Metal, or Vulkan, they do give an impression that probably is no longer in the long-term interest of the format. 

In the meantime, new extensions (like `KHR_blend`) should aim to be API-agnostic and not rely on GL enums. Presumably `KHR_techniques_webgl` is the exception here.

My preference would be for human-readable string enums (like `OPAQUE`, `BLEND`, `MASK`), but also open to other suggestions.

emackey:
I'm not sure if this change would provide enough benefit for the amount of churn involved.  I feel like the API-agnostic message has been fairly well received, with new Vulkan and Direct3D implementations popping up every so often.  The numeric enums can be easily viewed and edited in VSCode, for the few people who want to manipulate them by hand.

I do agree that new extensions, when not named explicitly after WebGL, should avoid adding new WebGL-derived enums.

donmccurdy:
>  I feel like the API-agnostic message has been fairly well received...

Yes, but between the enums and the name of the format there's definitely still room for confusion, and there have been some negative reactions to that.

I wouldn't want to change the enums in isolation, definitely. But _eventually_ there will be a 3.0 release with other breaking changes, and this is something I'd want to reconsider (perhaps in light of how much other churn we have, or lack thereof) when that happens.

javagl:
The ratio between "incompatiblity" and "benefit" does not looks soooo good for me right now - to some extent, because the latter is "zero": We could *rename* the constants, but for what purpose? The primary target for glTF still is GL, so I don't see a direct advantage in such a drastic change....

donmccurdy:
> The ratio between "incompatiblity" and "benefit"...

I'd argue that both technical incompatibility and technical benefit are 'zero' here. 😅 By and large, glTF parsers are not issuing WebGL commands internally, and these enums are being mapped to engines' own enums.

[...]

PR #1702, opened 2019-11-12
Draft: MESHOPT_compression

zeux:
This is a draft of MESHOPT_compression extension. I'm sharing this early to get feedback that might drastically change the JSON structure here or get extra input on the rest of the proposal.

[Rendered version](https://github.com/KhronosGroup/glTF/blob/0ab6547de9216d3094f7673904b06e274705fb1d/extensions/2.0/Vendor/MESHOPT_compression/README.md) (Updated 4/15/2020)

This extension is used by gltfpack to implement a substantially different type of compressor vs existing offering. Some of this is covered in the overview, but briefly:

- Compression is performed on a buffer view level - this makes it easy to support all types of glTF data (this includes geometry, morph targets and animation) with minimal JSON complexity
- Compressed buffer views can be decompressed directly into GPU/CPU target memory - this makes decoding faster and makes implementations of this extension very simple, here's a Babylon.JS example https://github.com/zeux/meshoptimizer/blob/master/demo/babylon.MESHOPT_compression.js
- Decompression is very fast - with WASM SIMD it runs at ~1 GB/s on modern desktop CPUs. Native implementations run at ~2+ GB/s.
- Decompression code is small. WASM decoder is 5-7 KB binary; two WASM decoders bundled into a .js file and gzip-compressed are a 7 KB download.
- Encoding is gzip-friendly. The compression method provides a moderate size gain over uncompressed storage, with further size gains possible when the binary is gzip-compressed.

Quick examples: (please note that this just two files, I will need to build a larger comparison table later)

BrainStem.glb, original: 3.1 MB
BrainStem.glb, quantized: 1.1 MB (this also does other types of optimization on the file) 
BrainStem.glb, quantized + deflate: 560 KB
BrainStem.glb, with this extension: 440 KB
BrainStem.glb, with this extension + deflate: 330 KB
BrainStem.glb, with Draco: 1.4 MB (animations aren't compressed, unfair example)

Corset.glb, original: 660 KB
Corset.glb, quantized: 340 KB
Corset.glb, quantized + deflate: 180 KB
Corset.glb, with this extension: 145 KB
Corset.glb, with this extension + deflate: 103 KB
Corset.glb, Draco: 112 KB

The BrainStem model above has animation data; the Corset.glb is just a static mesh but it has PBR-ready data set. In my experience, Draco tends to win noticeably on very large meshes with just position + normal data - I'm working on making my codec stronger in this case.

However, all comparisons above are unfair in one way or another to Draco - for example I am not matching quantization settings here and using whatever Draco model comes with glTF-Sample-Models - and in general Draco *will* win vs this extension in terms of pure size comparison when Draco fully supports the models. So the examples above are intended to highlight the wins that come from this extension, and Draco numbers are just ballpark because I know the question will come up. This extension is designed to complement Draco by providing a full compression solution aimed at high decoding speeds and specification simplicity, not to replace Draco fully in cases where mesh compression ratio is the only important criteria.

Note that the details of the bitstream are still in flux. It is my intention to update this extension with the bitstream encoding as it gets finalized; alternatively, the bitstream can be hosted in a separate document. It's much simpler than Draco's but a formal definition may still take significant space? Unsure.

I am hoping to be able to finalize the bitstream details either by the end of the year, or early next year. Any and all feedback welcome.

### DIFF
--- a/extensions/2.0/Vendor/MESHOPT_compression/README.md
+++ b/extensions/2.0/Vendor/MESHOPT_compression/README.md
@@ -173,7 +173,60 @@ The following sections specify the format of the bitstream for compressed data f
 
 ## Mode 0: attributes
 
-TODO
+Attribute compression exploits similarity between consecutive elements of the buffer by encoding deltas. The deltas are stored for each separate byte which makes the codec more versatile since it can work with components of various sizes. Additionally, the elements are stored with bytes deinterleaved, which means that sequences of deltas are more easily compressible by some general purpose compressors that may run on the resulting data.
+
+To facilitate efficient decompression, deinterleaving and delta encoding are performed on attribute blocks instead of on the entire buffer; within each block, elements are processed in groups of 16.
+
+The encoded stream structure is as follows:
+
+- Header byte, which must be equal to `0xa0`
+- Attribute blocks, detailed below
+- Tail block, which consists of a baseline element stored verbatim, padded to 32 bytes
+
+Each attribute block stores a sequence of deltas, with the first element in the first block using the deltas from the baseline element stored in the tail block. The attribute block always stores an integer number of elements, with that number computed as follows:
+
+```
+count = min((8192 / byteStride) & ~15, 256)
+```
+
+The attribute block structure consists of `byteStride` blocks (one for each byte of the element) with the following structure:
+
+- Header bits, 2 bits for each group of 16 elements (count/16 2-bit values), padded to a byte
+- Delta blocks, with variable number of bytes stored for each group of 16 elements
+
+Header bits are stored from least significant to most significant bit - header bits for 4 consecutive groups of 16 elements are packed in a byte together as follows:
+
+```
+bits0 | (bits1 << 2) | (bits2 << 4) | (bits3 << 6)
+```
+
+The header bits establish the delta encoding mode (0-3) for each group of 16 elements that follows:
+
+- bits 0: All 16 byte deltas are 0; the size of the encoded block is 0
+- bits 1: Deltas are using 2-bit sentinel encoding; the size of the encoded block is [4..20]
+- bits 2: Deltas are using 4-bit sentinel encoding; the size of the encoded block is [8..24]
+- bits 3: All 16 byte deltas are stored verbatim; the size of the encoded block is 16
+
+Byte deltas are stored as zigzag-encoded differences between the byte values of the element and the byte values of the previous element in the same position; the zigzag encoding scheme works as follows:
+
+```
+encode(uint8_t v) = ((v & 0x80) != 0) ? ~(v << 1) : (v << 1)
+decode(uint8_t v) = ((v & 1) != 0) ? ~(v >> 1) : (v >> 1)
+```
+
+When using the sentinel encoding, each delta is stored as a 2-bit or 4-bit value in a single 4-byte or 8-byte block, with deltas stored from most significant to least significant bit inside the byte (e.g. two 4-bit deltas are stored as `0xAB` where 0xA is the zigzag encoding of the first delta, and 0xB is the zigzag encoding of the second delta). The encoded value of the delta that has all bits set to 1 (corresponds to `3` for 2-bit deltas and `15` for 4-bit deltas, which, when decoded, represents `-2` and `-8` respectively) indicates that the real delta value is outside of the 2-bit or 4-bit range and is stored as a full byte after the bit deltas for this group. For example, for 4-bit deltas, the following byte sequence:
+
+```
+0x17 0x5F 0xF0 0xBC 0x77 0xA9 0x21 0x00 0x34 0xB5
+```
+
+Encodes 16 deltas, where the first 8 bytes of the sequence specifies the 4-bit deltas, and the last 2 bytes of the sequence specify the explicit delta values encoded for elements 3 and 4 in the sequence, so the decoded deltas (after zigzag decoding) look like
+
+```
+-1 -4 -3 26 -91 0 -6 6 -4 -4 5 -5 1 -1 0 0
+```
+
+Finally, note that the deltas are computed in 8-bit integer space with wrap-around two-complement arithmetic; for example, if the values of the first byte of two consecutive elements are 0x00 and 0xFF, the byte delta that is stored is `-1` (`1` after zigzag encoding).
 
 ## Mode 1: triangles
 
--- a/extensions/2.0/Vendor/MESHOPT_compression/README.md
+++ b/extensions/2.0/Vendor/MESHOPT_compression/README.md
@@ -87,7 +87,7 @@ Compression mode specifies the bitstream layout and the algorithm used to decomp
 
 In all three modes, the resulting compressed byte sequence is typically noticably smaller than the buffer view length, *and* can be additionally compressed by using a general purpose compression algorithm such as Deflate for the resulting glTF file (.glb/.bin).
 
-The format of the bitstream is specified in [supplementary specification](Bitstream.md).
+The format of the bitstream is specified in Bitstream section.
 
 When using attribute encoding, for some types of data exploiting the redundancy between consecutive elements is not enough to achieve good compression ratio; quantization can help but isn't always sufficient either. To that end, when using mode 0, this extension allows a further use of a compression filter, that transforms each element stored in the buffer view to make it more compressible with the attribute codec and often allows to trade precision for compressed size. Filters don't change the size of the output data, they merely improve the compressed size by reducing entropy; note that the use of a compression filter restricts `byteStride` which effectively prohibits storing interleaved data.
 
@@ -98,7 +98,7 @@ Filter specifies the algorithm used to transform the data after decompression, a
 - Filter 2: quaternion. Suitable for storing rotation data for animations or instancing as 8-byte values with variable precision max-component encoding.
 - Filter 3: exponential. Suitable for storing floating point data as 4-byte values with variable mantissa precision.
 
-The filters are detailed further in [supplementary specification](Bitstream.md).
+The filters are detailed further in Bitstream section.
 
 When using filters, the expectation is that the filter is applied after the attribute decoder on the contents of the resulting bufferView; the resulting data can then be used according to the referencing accessors without further modifications.
 
@@ -166,3 +166,31 @@ Additionally it's important to identify tracks with the same output value and us
 To reduce the size of each keyframe, rotation data should be quantized using 16-bit normalized components; for additional compression, the use of filter 2 (quaternion) is recommended. Translation/scale data can be compressed using filter 3 (exponential) with the same exponent used for all three vector components.
 
 After pre-processing, both input and output data should be stored using mode 0 (attributes).
+
+# Appendix: Bitstream
+
+The following sections specify the format of the bitstream for compressed data for various modes and filters.
+
+## Mode 0: attributes
+
+TODO
+
+## Mode 1: triangles
+
+TODO
+
+## Mode 2: indices
+
+TODO
+
+## Filter 1: octahedral
+
+TODO
+
+## Filter 2: quaternion
+
+TODO
+
+## Filter 3: exponential
+
+TODO
--- a/extensions/2.0/Vendor/MESHOPT_compression/README.md
+++ b/extensions/2.0/Vendor/MESHOPT_compression/README.md
@@ -232,7 +232,102 @@ Finally, note that the deltas are computed in 8-bit integer space with wrap-arou
 
 ## Mode 1: triangles
 
-TODO
+Triangle compression compresses triangle list indices by exploiting similarity between consecutive triangles. Given a triangle stream that has been optimized for locality, very often subsequent triangles share an edge with the recently encoded triangle. The encoder uses a few other techniques to try to encode most triangles in optimized triangle lists into a single byte.
+
+The encoded stream structure is as follows:
+
+- Header byte, which must be equal to `0xe1`
+- Triangle codes, referred to as `code` below, with a single byte for each triangle
+- Extra data which is necessary to decode triangles that don't fit into a single byte, referred to as `data` below
+- Tail block, which consists of a 16-byte lookup table, referred to as `codeaux` below
+
+There are two limitations on the structure of the 16-byte lookup table:
+
+- The last two bytes must be 0
+- The remaining bytes must not contain any nibbles equal to `0xf`.
+
+During the decoding process, decoder maintains four variables:
+
+- `next`: an integer referring to the expected next index (also known as high-watermark), starts at 0
+- `last`: an integer referring to the last encoded index, starts at 0
+- `edgefifo`: a 16-entry FIFO with two vertex indices in each entry; initial contents is undefined
+- `vertexfifo`: a 16-entry FIFO with a vertex index in each entry; initial contents is undefined
+
+To decode each triangle, the decoder needs to analyze the `code` byte, read additional bytes from `data` as necessary, and update the internal state correctly. The `code` byte encoding is optimized to reach a single byte per triangle in most common cases; the resulting data can often be compressed by a general purpose compressor running on the resulting .bin/.glb file.
+
+When extra data is necessary to decode a triangle and it represents an index value, the decoder uses varint-7 encoding, which encodes an integer as one or more bytes, with the byte with the 0 most significant bit terminating the sequence:
+
+```
+0x7F => 0x7F
+0x81 0x04 => 0x201
+0xFF 0xA0 0x05 => 0x1FD005
+```
+
+Instead of using the raw index value, a zigzag-encoded 32-bit delta from `last` is used:
+
+```
+uint32_t decodeIndex(uint32_t v) {
+	int32_t delta = (v & 1) != 0 ? ~(v >> 1) : (v >> 1);
+
+	last += delta;
+	return last;
+}
+```
+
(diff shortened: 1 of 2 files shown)


extensions/2.0/Vendor/MESHOPT_compression/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+Morph targets can be treated identically to other vertex attributes, as long as vertex order optimization is performed on all target streams at the same time. It is recommended to use quantized storage for morph target deltas, possibly with a narrower type than that used for baseline values.
+
+When storing vertex data, mode 0 (attributes) should be used; for index data, mode 1 (indices) should be used instead. Mode 1 only supports triangle list storage; indices of other topology types can be stored uncompressed or using mode 0 (attributes). The use of triangle strip topology is not recommended since it's more efficient to store triangle lists using mode 1.
+
+TODO: normal & tangent unit filters
+
+## Compressing animation data
+
+To minimize the size of animation data, it is recommended that encoders resample animations before compression. This serves two purposes:
```
donmccurdy: There are some nuances to consider in this recommendation, such as a file containing multiple animations with different durations. Presumably resampling would happen on a per-animation basis, then? Or share input accessors across all channels of the same duration? I don't know that this needs to be fully specified, but just mention that resampling is not a one-size-fits-all recommendation.

extensions/2.0/Vendor/MESHOPT_compression/README.md
```diff
@@ -0,0 +1,132 @@
+# MESHOPT\_compression
```
donmccurdy: Consider `MESHOPT_mesh_compression`. 😇

donmccurdy:
> I'm sharing this early to get feedback that might drastically change the JSON structure here or get extra input on the rest of the proposal.

The JSON structure looks good to me, given the context and longer discussion in https://github.com/KhronosGroup/glTF/issues/1684. I'll probably have comments on various details later, but nothing that seems important yet.

> It's much simpler than Draco's but a formal definition may still take significant space? Unsure.

I would suggest an appendix in README.md for now, but a separate document would also be fine.

> Quick examples: (please note that this just two files, I will need to build a larger comparison table later)

A larger comparison table would be very helpful — would it be possible to include decode times as well?

extensions/2.0/Vendor/MESHOPT_compression/README.md
zeux: Note that this extension is more general and allows compressing animation data as well. Maybe `view_compression` would be appropriate?

extensions/2.0/Vendor/MESHOPT_compression/README.md
zeux: Yeah that's fair - it's probably worthwhile to outline several approaches. gltfpack currently resamples each animation in isolation, although it's possible to actually go further and share time tracks between all animations. Regardless, the compression efficiency can probably be gained in other ways - I'll expand/reword this section.

zeux:
> A larger comparison table would be very helpful — would it be possible to include decode times as well?

Yeah, sure thing. Brief non-scientific experiment (will include decode numbers in the future) on Buggy.gltf (500K triangles, 400K vertices):

- Draco: 190ms
- meshopt: decodeVertexBuffer 15 ms + decodeIndexBuffer 4.5 ms = 20ms
- meshopt simd: decodeVertexBuffer 9 ms + decodeIndexBuffer 4.5 ms = 15ms (normally SIMD version is ~2.5x faster but the perf increase on the single mesh load is generally smaller... need to profile streaming scenario where meshes are being continuously decoded)

On models of a massive scale the difference can be more pronounced especially with the SIMD variant, here's Thai Buddha statue (https://sketchfab.com/3d-models/thai-buddha-cba029e262bd4f22a7ee4fcf064e22ee, 6M triangles / 3M vertices):

- Draco: 9720 ms
- meshopt: 181ms vertex + 90ms index = 270ms
- meshopt simd: 72ms vertex + 92ms index = 165ms

"simd" above refers to WASM SIMD variant that is supported on Chrome Canary when experimental SIMD support is enabled.

extensions/2.0/Vendor/MESHOPT_compression/README.md
donmccurdy: Ah, good point. In that case `view_compression` sounds reasonable, or I would be open to considering this an exception to the `<scope>_<feature>` guideline too.

zeux:
I've updated the extension spec to the latest version. Will work on bitstream specification in the next few days; comments welcome otherwise. We'll probably need a better extension name but that can wait.

The new changes added since last update make the extension more competitive from the transmission size perspective. Here's a comparison table in terms of encoded size (see performance numbers above, the ballpark is about the same) for models from glTF-Sample-Models, excluding texture data:

```
2CylinderEngine                          gltfpack 496864               gltfpack.gz 261707               draco 499784               draco.gz 402837               gltfpack/draco 0.994                gltfpack.gz/draco.gz 0.650
AntiqueCamera                            gltfpack 166612               gltfpack.gz 115040               draco 130240               draco.gz 114831               gltfpack/draco 1.279                gltfpack.gz/draco.gz 1.002
Avocado                                  gltfpack 8232                 gltfpack.gz 5705                 draco 6616                 draco.gz 5300                 gltfpack/draco 1.244                gltfpack.gz/draco.gz 1.076
BarramundiFish                           gltfpack 29924                gltfpack.gz 23982                draco 23368                draco.gz 21069                gltfpack/draco 1.281                gltfpack.gz/draco.gz 1.138
BoomBox                                  gltfpack 47012                gltfpack.gz 37594                draco 37128                draco.gz 34118                gltfpack/draco 1.266                gltfpack.gz/draco.gz 1.102
BrainStem                                gltfpack 370512               gltfpack.gz 256496               draco 380872               draco.gz 247774               gltfpack/draco 0.973                gltfpack.gz/draco.gz 1.035
Buggy                                    gltfpack 2230596              gltfpack.gz 1047707              draco 2733216              draco.gz 2243675              gltfpack/draco 0.816                gltfpack.gz/draco.gz 0.467
CesiumMan                                gltfpack 50356                gltfpack.gz 34063                draco 45860                draco.gz 36237                gltfpack/draco 1.098                gltfpack.gz/draco.gz 0.940
CesiumMilkTruck                          gltfpack 29664                gltfpack.gz 15927                draco 29640                draco.gz 16732                gltfpack/draco 1.001                gltfpack.gz/draco.gz 0.952
Corset                                   gltfpack 133356               gltfpack.gz 85150                draco 113724               draco.gz 105913               gltfpack/draco 1.173                gltfpack.gz/draco.gz 0.804
DamagedHelmet                            gltfpack 136528               gltfpack.gz 105357               draco 102016               draco.gz 91620                gltfpack/draco 1.338                gltfpack.gz/draco.gz 1.150
Duck                                     gltfpack 27636                gltfpack.gz 23031                draco 21096                draco.gz 19471                gltfpack/draco 1.310                gltfpack.gz/draco.gz 1.183
FlightHelmet                             gltfpack 685856               gltfpack.gz 498758               draco 519572               draco.gz 473156               gltfpack/draco 1.320                gltfpack.gz/draco.gz 1.054
GearboxAssy                              gltfpack 2209512              gltfpack.gz 1271787              draco 2212596              draco.gz 1845696              gltfpack/draco 0.999                gltfpack.gz/draco.gz 0.689
Lantern                                  gltfpack 46200                gltfpack.gz 29917                draco 35548                draco.gz 32146                gltfpack/draco 1.300                gltfpack.gz/draco.gz 0.931
Monster                                  gltfpack 40800                gltfpack.gz 23526                draco 53216                draco.gz 31308                gltfpack/draco 0.767                gltfpack.gz/draco.gz 0.751
ReciprocatingSaw                         gltfpack 804880               gltfpack.gz 457507               draco 869368               draco.gz 702841               gltfpack/draco 0.926                gltfpack.gz/draco.gz 0.651
RiggedFigure                             gltfpack 12744                gltfpack.gz 6973                 draco 10728                draco.gz 6538                 gltfpack/draco 1.188                gltfpack.gz/draco.gz 1.067
SciFiHelmet                              gltfpack 670708               gltfpack.gz 537936               draco 182420               draco.gz 173450               gltfpack/draco 3.677                gltfpack.gz/draco.gz 3.101
Sponza                                   gltfpack 1970516              gltfpack.gz 843461               draco 1678516              draco.gz 1453378              gltfpack/draco 1.174                gltfpack.gz/draco.gz 0.580
Suzanne                                  gltfpack 23588                gltfpack.gz 18963                draco 17912                draco.gz 16285                gltfpack/draco 1.317                gltfpack.gz/draco.gz 1.164
WaterBottle                              gltfpack 32952                gltfpack.gz 20436                draco 24384                draco.gz 21490                gltfpack/draco 1.351                gltfpack.gz/draco.gz 0.951
```

These were generated as follows:

```
gltfpack -i scene.gltf -o scene_noq.glb -noq
gltfpack -i scene_noq.glb -o scene_glp.glb -cc
gzip -9 -k scene_glp.glb
gltf-pipeline -i scene_noq.glb -o scene_drc.glb -d -t --draco.quantizeNormalBits 8
gzip -9 -k scene_drc.glb
```

The use of `noq` (no-quantization) packed GLB as a starting point is important since it establishes a level playing field - gltfpack merges the meshes and filters out redundant attributes/triangles/etc. as a result, so that Draco compression can work off the same data set.

From the numbers above it seems like this extension is an excellent all-rounder alternative to Draco. With recent additions it supports all glTF features (e.g. now supports efficient encoding of line lists and other topologies), and is future-proof - for example, the parts of this extension that allow compressing animation data (notably the quaternion/exponential filters) apply in the exact same way to EXT_mesh_gpu_instancing without any schema changes because of the orthogonal extension design.

My goal over the next week  is to document the bitstream and filters, correct any inconsistencies that arise, and possibly add a color filter which I haven't done in the specification yet. Any and all feedback is welcome.

zeux:
The spec has been updated with detailed documentation on encoding for all modes and filters.

donmccurdy:
Thanks for the detailed numbers here! For easier review I've [formatted them in a sheet](https://docs.google.com/spreadsheets/d/1V0jls9QSb7DRE3-JseCHRcnAzIAp-uSBssnEvx2dQxI/edit?usp=sharing).

three.js will very likely support this extension: the improved decoding speed certainly fills a gap that our users have repeatedly reported. By that I mean, in addition to it being possible to implement the extension externally with the [upcoming extension API](https://github.com/mrdoob/three.js/pull/19144), I think we'd support including the extension in the repo and npm package if that's of interest.

***

A couple minor spec comments. So far, I think all enum values in the glTF spec fall into one of two cases:

1. GL enum value, integer.
2. glTF-specific enum value, string.

From that perspective, the `.mode` and `.filter` enums would be more consistent with the rest of the spec if they were given string types. Am I correct in understanding that the value of `.mode` is not closely associated with `bufferView.target`? E.g. use of the `attribute` mode does _not_ necessarily imply `bufferView.target === gl.ARRAY_BUFFER`?

zeux:
> So far, I think all enum values in the glTF spec fall into one of two cases:

Primitive mode is a small integer, which was part of the motivation here, another part being a slight reduction in glTF size. It's not particularly meaningful so this could be changed to a string.

> .g. use of the attribute mode does not necessarily imply bufferView.target === gl.ARRAY_BUFFER?

Yeah that's correct. `attribute` mode is useful when compressing animation data as well as instance transform data, neither of which would have ARRAY_BUFFER target.

zeux:
> if they were given string types

Is upper case or lower case more consistent with recent additions? It seems that upper case is perhaps a touch more prevalent, but e.g. camera types and node paths use lower case.

donmccurdy:
[Light types](https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_lights_punctual/README.md#defining-lights) are lowercase as well; I think that's the only extension to define an enum-like property since the core spec. @lexaknyazev you'd mentioned some opinions on this before (https://github.com/KhronosGroup/glTF/issues/1560#issuecomment-462932630), do you have a preference moving forward?

zeux:
Separate question: I'd like to rename this to EXT-prefixed extension as it makes more sense to do so - the encoding format/support comes from meshoptimizer/gltfpack but with the spec and existing implementations for three.js and Babylon.JS (both are currently within meshoptimizer repo but can be contributed upstream once this extension lands) this seems to be a cross-vendor extension.

With that in mind, what's a good name? Options OTOH

- EXT_compression_meshopt
- EXT_meshopt_compression // similar to KHR_draco_mesh_compression
- EXT_buffer_compression_meshopt // imprecise - extension works on buffer views - but maybe good enough?
- EXT_view_compression_meshopt // imprecise but I don't like buffer_view prefix as it's really long and two-word...

Other variants?

donmccurdy:
If the implementations will land upstream — i.e. be accepted by the three.js and/or babylon.js projects — then I agree that the `EXT_` prefix is appropriate. I'm optimistic about three.js given the compression numbers here, but can't answer for the project just yet. To clarify for future readers (I think this is what you're aiming for anyway, of course) I believe the expectation for a multi-vendor extension is that it have the support of multiple vendors, not that a single vendor provide third-party plugins to multiple vendors' tools.

On names, I'd be tempted to use `bufferview` rather than `buffer_view`, I dislike the two-words part much more than I dislike the length. In that vein, perhaps one of:

- `EXT_meshopt_bufferview_compression`
  - Maybe a bit long, but consistent with Draco and arguably<sup>1</sup> consistent with the naming guidelines.
- `EXT_bufferview_meshopt_compression`
  - Maybe a bit long, but consistent with the naming guidelines. Not really consistent with Draco.
- `EXT_meshopt_compression`
  - Not quite in line with the naming guideline, but I'd be OK with an exception given the broad scope of the extension and the relative obscurity of the 'bufferview' concept.  Consistent with Draco.

<sup>1</sup> I say "arguably" because it somewhat resembles the double prefixes used in WebGL, like `WEBKIT_OES_vertex_array_object`. That isn't really covered by our naming guidelines today.

zeux:
> I believe the expectation for a multi-vendor extension is that it have the support of multiple vendors, not that a single vendor provide third-party plugins to multiple vendors' tools.

Right, that's what I meant, although this kinda goes both ways - if this extension is EXT_ it's more likely to be upstreamable, and if this isn't it's less likely.

My main motivation for EXT_ is that this extension fills large gaps in functionality. Without it, entire classes of valid glTF models don't have a valid compression solution in glTF ecosystem. (already mentioned earlier, but this includes point clouds, lines, morph targets for geometry, and animation data)

The multi-vendor implementation is, I suppose, more of a consequence.

I like `EXT_meshopt_compression` the most, since it doesn't use the bufferview (which is obscure indeed... all other prefixes read like "extension provides extra data for X", but here bufferView is an implementation detail so to speak, the actual goal is making files smaller), and is in line with Draco naming (Draco has _mesh_ but it only works on meshes...)

bghgary:
+1 to `EXT_meshopt_compression`

It might be just me, but I have an aversion to `bufferview` with a lowercase v.

zeux:
`EXT_meshopt_compression` it is. Renaming is a good opportunity to make the other incompatible changes here, I'm considering:

- Removing `count` property as it's redundant and can be computed from parent buffer view length. I believe I included it to try to keep the extension blob self-sufficient but it doesn't seem like a strong enough reason, since for validation purposes parent length would need to be checked anyway.
- Changing the enums to strings, if this is the preferred direction. It might make sense to keep the indexing in the bitstream spec (e.g. "mode 0") in case a future glTF version has dual indexed & named enums.

The open questions are:
- Should enums be changed to strings?
- Should strings be lower or upper case?

Unsure how to answer these atm.

With some other extensions, I've seen renaming done in-place by editing the PR, or renaming lead to opening a new PR. Which one makes more sense in this case, given that I'd like to move this from "draft" to an actual proposal?

donmccurdy:
> With some other extensions, I've seen renaming done in-place by editing the PR, or renaming lead to opening a new PR.

I think this is up to the preference of the author. When a PR is created in GitHub's web UI it might be easier to restart than to rename folders. It's fine to edit the PR, though.

I think my preference is https://github.com/KhronosGroup/glTF/pull/1702#issuecomment-629942038 but I'm hoping others may have thoughts on this... I think it's less a question of a "right" answer than of having a rough guideline for the future. @emackey @bghgary @lexaknyazev ?

emackey:
I don't think I have a strong opinion.  The core enums are all legacy from when glTF 1.0 and prior were targeting the WebGL API directly.  Was there ever a reason to prevent non-WebGL enums once that link was broken in 2.0?  I'm not sure.

In general, consistency is good for the ecosystem.  So that would suggest either following the pattern of previous extensions, or explicitly establishing a better pattern for future extensions.

lexaknyazev:
The existing suboptimal situation reflects three possible enum sources (it's all glTF 1.0 legacy):
- Integers for values compatible with GL entrypoints.
- Lowercase strings for values that could be used as JSON keys in other places (see `camera.type` / `camera.perspective`, same for punctual lights).
- Uppercase strings otherwise.

To follow the existing design, this extension's JSON should use uppercase strings. Bitstream is not affected by any of this, of course.

I'd refrain from "establishing a better pattern for future extensions" unless we're talking about glTF Next.

zeux:
@lexaknyazev Thanks! The distinction is clear, I'm going to amend this to use upper-case strings then.

zeux:
Since the numbers in the initial post for this PR are out of date and the extension name is going to change, I'm going to close this PR and open a new PR with up-to-date information and new name/spec.

Issue #1442, opened 2018-09-05
What would you like to see in next gen materials?


sbtron:
There have been some discussions lately around an updated glTF material model to support additional capabilities. Please consider this a requirements gathering exercise to understand what kind of capabilities everyone would like to see in the next gen material model for glTF.

Below are just some suggestions in no particular order
- Better Transparency (Refractions etc)
- Subsurface scattering 
- Anisotropy
- Coating

Are there additional capabilities anyone would like to see? Any preferences or priorities on what we should go after next from the list above?
Without getting into details on how exactly we would implement this (for now) our high level goal would still be to ensure any new capability is implementable across all graphics APIs.

n1ckfg:
Visibility keyframes--they make possible small cache-style animations of individual frames.

Ben-Mack:
Please add official support for Lightmap 👍

Popular 3D engines all have lightmap in their PBR material. 

It's an essential feature to achieve much better lighting quality and performance for static objects (benefit not just for game industry but also product visualization, arch viz...)

stevenvergenz already created an extension for it here: https://github.com/KhronosGroup/glTF/pull/1017

UX3D-nopper:
- Thin film
- Multi-layer

see https://blog.selfshadow.com/publications/s2017-shading-course/drobot/s2017_pbs_multilayered.pdf

romainguy:
- Backscattering (cloth)

romainguy:
If that's useful we have a document with real-time definitions for clear coat, anisotropy and cloth which can be added to existing PBR engines: https://google.github.io/filament/Filament.md.html

pjcozzi:
Photogrammetry users of Cesium have a slight request for **refraction** and **reflection**.  The later, I _think_ is already covered enough in today's glTF PBR.

emackey:
See also:

- Comments on clear coat from #1040 
- Comments on IOR from #1079

vyv03354:
* Toon shaded materials
* Outlines

for non-photorealistic rendering.

mlimper:
[...]

PR #1717, opened 2019-12-03
Proposal: More detailed description of PBR in glTF 2.0

proog128:
This is a draft that extends Appendix B (BRDF Implementation) of the glTF 2.0 specification. It describes the metallic-roughness material in an abstract way, derives a physically realistic BRDF (positive, reciprocal, energy conserving) for offline ray tracing, and gives some hints how to approximate this in real-time renderers. The old equations are still valid after this change.

It's work in progress, there are still a lot of TODOs. I am sharing this to get feedback on the approach. Maybe this could be a template for PBR Next (#1442) including extensions (#1677, #1688, #1698) that allows to have consistent implementations of the material in different rendering engines, from mobile, real-time VR or web to offline, unbiased GI.


javagl:
I really appreciate that. Getting started with (i.e. trying to implement) PBR from scratch can be hard, and a detailed description like that can certainly help here. But note that there might be some overlap with things like https://github.com/KhronosGroup/glTF-Sample-Viewer#physically-based-materials-in-gltf-20 . We should either make sure that this is really an *overlap* (meaning that they are describing the same thing, and not refer to different papers with different formulas), or figure out "the one and only place" where this could be described in detail (and then just link to that).

proog128:
Of course, changing this section must not make existing implementations like the sample viewer incompatible to the glTF 2.0 spec. If the proposal here is incomaptible to what has been in this section before, then it is a mistake. I have to admit that I didn't check if Appendix B in the spec agrees with the glTF-Sample-Viewer documentation.

To give a bit more context: We want to have a glTF PBR (Next) compatible material model in a GI renderer. Unfortunately, the equations given in the spec are not well-suited for global illumination, as they break some fundamental properties of physical light transport. At the moment, Appendix B is non-normative, so we could just ignore it. However, I thought this is a good opportunity to improve it.

There are three key aspects of the proposal:
* It defines the material model in an abstract, textual way, inspired by MDL.
```
material = mix(
  fresnel_mix(
    diffuse_brdf(baseColor),
    microfacet_brdf(roughness^2),
    ior = 1.5),
  metal_brdf(baseColor, roughness^2),
  metallic)
```
* It describes the math of the BRDFs such that they are positive, symmetric and energy conserving, based on the [Enterprise PBR Material](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2021x.md.html). This makes it possible to have a clean implementation in an unbiased path tracer which could serve as ground-truth.
* It explains how other implementations can comply to the spec: they have to follow the mixing rules, offer the same parameters and be as close as possible to the ground-truth. They may deviate because of hw/sw constraints (ie., simplify the model) or to give more realistic renderings (ie., complicate the model).

specification/2.0/README.md
```diff
@@ -3900,67 +3888,144 @@ Application-specific data.
...
-Below are common implementations for the various terms found in the lighting equation.
+The Schlick Fresnel is computed via the reflection half vector `H = normalize(V + L)`.
 
-### Surface Reflection Ratio (F)
+```
+fresnel = baseColor + (1 - baseColor) * (1 - HdotV)^5
+```
 
-**Fresnel Schlick**
+The multi-scattering approximation uses the directional albedo `E_m` and average albedo `E_mavg` of the single-scattering model. These values are typically precomputed and fetched from a lookup table during rendering. The average Fresnel `F_mavg` for Schlick can be computed analytically.
```
abwood: This is a bit of a stumbling point for me.  Should we link to [Enterprise PBR Shading Model](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2021x.md.html#appendix/energycompensation) for background? Can we provide these precomputed lookup tables?  I see that your github provides these here:
[GGX_E](https://github.com/DassaultSystemes-Technology/EnterprisePBRShadingModel/blob/master/res/GGX_E.exr)
[GGX_E_avg](https://github.com/DassaultSystemes-Technology/EnterprisePBRShadingModel/blob/master/res/GGX_E_avg.exr)

It might also be worth mentioning the numerical fit described in [Energy Preservation](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2021x.md.html#components/core/in-depthlookatthebsdfs) for those who are willing to sacrifice energy conservation error in exchange for one less texture lookup into floating point texture lookups.

specification/2.0/README.md
```diff
@@ -3900,67 +3888,144 @@ Application-specific data.
...
 
-Simplified implementation of Fresnel from [An Inexpensive BRDF Model for Physically based Rendering](https://www.cs.virginia.edu/~jdl/bib/appearance/analytic%20models/schlick94b.pdf) by Christophe Schlick.
+```
+                                (1 - E_m(VdotN)) * (1 - E_m(LdotN))
```
abwood: Should `E_m(VdotN)` be `E_m(VdotN, alpha)` and `E_mavg` be `E_mavg(alpha)`?

specification/2.0/README.md
abwood: Worth noting that I came across this when trying to see how a new gltf material would be written, using this appendix as reference.  I really appreciate the well defined abstractions and diagrams that have been developed thus far; looking forward to seeing the final writeup!

specification/2.0/README.md
proog128: Yes, `E_m` and `E_mavg` depend on `alpha`. I omitted it here to improve readability, but it is confusing. I will add it. Thanks!

specification/2.0/README.md
proog128: Not sure if we should link to specific implementation in the spec, is it possible to do that? I was also thinking about describing how to build the tables (similar to [this)](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2021x.md.html#appendix/energycompensation), but maybe that is too much detail and we should link to the [slides from Imageworks](https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides_v2.pdf) or [other implementation-independent sources](https://blog.selfshadow.com/2018/05/13/multi-faceted-part-1/). We could also put the tables into the glTF repo.

specification/2.0/README.md
```diff
@@ -3900,67 +3888,144 @@ Application-specific data.
...
-Implementation of microfacet distrubtion from [Average Irregularity Representation of a Roughened Surface for Ray Reflection](https://www.osapublishing.org/josa/abstract.cfm?uri=josa-65-5-531) by T. S. Trowbridge, and K. P. Reitz
+The `fresnel_mix` operation weights the (multi-scatter) microfacet BRDF with the Fresnel term, and the diffuse BRDF  with `diffuse_weight`.
 
-![](figures/lightingD.PNG)
+```
+dielectric_brdf =
+  diffuse_brdf() * diffuse_weight(VdotN, LdotN, f0) +
+  microfacet_brdf(alpha) * fresnel(HdotV, f0) +
+  multiscatter_microfacet_brdf(alpha) * multiscatter_fresnel(f0)
+```
```
abwood: Is this equation correct? I don't see where baseColor factors into dielectric materials.  Is this f0 (0.04) or the full reflectance color (normal incidence angle)?

proog128: In dielectric materials, baseColor defines the color of the diffuse component. It's missing in this equation, fixed it.

specification/2.0/README.md
```diff
@@ -3900,67 +3888,144 @@ Application-specific data.
...
-Simplified implementation of Fresnel from [An Inexpensive BRDF Model for Physically based Rendering](https://www.cs.virginia.edu/~jdl/bib/appearance/analytic%20models/schlick94b.pdf) by Christophe Schlick.
+```
+                                (1 - E_m(VdotN, alpha)) * (1 - E_m(LdotN, alpha))
+multiscatter_microfacet_brdf =  --------------------------------------------------
+                                            pi * (1 - E_mavg(alpha))
 
-![](figures/lightingF.PNG)
+           1                 20
+F_mavg = ---- * baseColor + ----
+          21                 21
```
abwood: Is this equation correct or should it be 1/21 + 20/21 * baseColor?  Same below in the dielectric section.

proog128: Fixed, thanks!

emackey:
@proog128 Please bump with a comment here when this is ready for the next review, thanks!

specification/2.0/README.md
```diff
@@ -3900,13 +3888,70 @@ Application-specific data.
 
 # Appendix B: BRDF Implementation
 
+The metallic-roughness material model is a linear blend of two BRDFs, a metallic BRDF and a dielectric BRDF. The blending factor `metallic` describes the metalness of the material. The BRDFs share the parameters for roughness and base color, following [Burley (2012): Physically-Based Shading at Disney](https://disney-animation.s3.amazonaws.com/library/s2012_pbs_disney_brdf_notes_v2.pdf).
+
+```
+material = mix(dielectric_brdf, metal_brdf, metallic)
+```
+
+The metal BRDF is based on a microfacet model which describes the orientation of microfacets on the surface as a statistical distribution. The distribution is controlled by a parameter called roughness, varying between 0 (smooth surface) and 1 (rough surface). As most microfacet distributions the blending between the two extremes does not behave linear, the material's `roughness` parameter is squared before using it as the distribution's roughness.
```
bghgary, line 3897:
```suggestion
The metal BRDF is based on a microfacet model which describes the orientation of microfacets on the surface as a statistical distribution. The distribution is controlled by a parameter called roughness, varying between 0 (smooth surface) and 1 (rough surface). As the blending between the two extremes does not behave linearly, like most microfacet distributions, the material's `roughness` parameter is squared before using it as the distribution's roughness.
```

specification/2.0/README.md
```diff
@@ -3900,13 +3888,70 @@ Application-specific data.
...
+    base = diffuse_brdf(baseColor),
+    layer = specular_brdf(roughness^2),
+    ior = 1.5)
+```
+
+The BRDFs and mixing operators used in the metallic-roughness material are summarized in the following image.
+
+![](figures/pbr.png)
+
+The glTF spec is designed to allow applications to choose different lighting implementations based on their requirements. Some implementations may focus on an accurate simulation of light transport, others may choose to deliver real-time performance. Therefore, any implementation that adheres to the rules for mixing BRDFs is compliant to the glTF spec.
```
bghgary, line 3925:
```suggestion
The glTF spec is designed to allow applications to choose different lighting implementations based on their requirements. Some implementations may focus on an accurate simulation of light transport while others may choose to deliver real-time performance. Therefore, any implementation that adheres to the rules for mixing BRDFs is comformant to the glTF spec.
```

specification/2.0/README.md
```diff
@@ -3900,13 +3888,70 @@ Application-specific data.
...
+
+In a physically-accurate light simulation, the BRDFs have to follow some basic principles: the BRDF has to be positive, reciprocal and energy conserving. This ensures that the visual output of the simulation is independent of the underlying rendering algorithm, as long as it is unbiased. The specification will provide a mathematical model that allows implementations to achieve an exact result in an unbiased renderer. Note that unbiased renderers may still decide to deviate from the specification to achieve better visual quality.
+
+The unbiased light simulation with physically realistic BRDFs will be the ground-truth for approximations in real-time renderers that are often biased, but still give visually pleasing results. Usually, these renderers take short-cuts to solve the rendering equation, like the split-sum approximation for image based lighting, or simplify the math to save instructions and reduce register pressure. However, there are many ways to achieve good approximations, depending on the platform (mobile or web applications, desktop applications on low or high-end hardware, VR) different constraints have to be taken into account.
```
bghgary, line 3929:
```suggestion
The unbiased light simulation with physically realistic BRDFs will be the ground-truth for approximations in real-time renderers that are often biased, but still give visually pleasing results. Usually, these renderers take shortcuts to solve the rendering equation, like the split-sum approximation for image based lighting, or simplify the math to save instructions and reduce register pressure. However, there are many ways to achieve good approximations, depending on the platform (mobile or web applications, desktop applications on low or high-end hardware, VR) different constraints have to be taken into account.
```

specification/2.0/README.md
```diff
@@ -3900,13 +3888,70 @@ Application-specific data.
...
-An implementation sample is available at https://github.com/KhronosGroup/glTF-Sample-Viewer/ and provides an example of a WebGL implementation of a standard BRDF based on the glTF material parameters.
+```
+     1            alpha^2
+D = -- --------------------------------
+    pi   (HdotN^2 * (alpha^2-1) + 1)^2
 
-As previously defined
+                                           2 * NdotV * NdotL
+G = ----------------------------------------------------------------------------------------------
+    NdotL * sqrt(NdotV^2 * (1-alpha^2) + alpha^2) + NdotV * sqrt(NdotL^2 * (1-alpha^2) + alpha^2))
```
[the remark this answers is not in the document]
proog128: Somehow I messed up the PR during an earlier rebase or merge. The parts following from here were broken. Should be fixed now.

specification/2.0/README.md
```diff
@@ -3900,13 +3888,70 @@ Application-specific data.
...
+
+```
+metal_brdf =
+  microfacet_brdf(alpha) * fresnel(HdotV, baseColor) +
+  multiscatter_microfacet_brdf(alpha) * multiscatter_fresnel(baseColor)
+```
+
+The microfacet BRDF uses the GGX distribution and Smith height-correlated masking-shadowing term ([Heitz 2014: Understanding the Masking-Shadowing Function in Microfacet-Based BRDFs](http://jcgt.org/published/0003/02/03/paper.pdf)).
 
-An implementation sample is available at https://github.com/KhronosGroup/glTF-Sample-Viewer/ and provides an example of a WebGL implementation of a standard BRDF based on the glTF material parameters.
```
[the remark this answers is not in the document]
proog128: You are right, the link to the sample viewer should definitely be kept. I added it back.

specification/2.0/README.md
proog128: Thanks a lot for the review and all the suggestions!

proog128:
Thanks for all the reviews! So far I used the formulas from the Enterprise PBR Material in the non-normative implementation section. However, I understand that it is probably too complicated for a simple example, especially when considering real-time WebGL renderers. Therefore, I will change it back to the implementation that was described before (and, as far as I know, is used in the glTF sample viewer).

specification/2.0/README.md
rsahlin: I think it should be 'conformant'

bghgary: Oops :)

emackey:
@proog128 When you get a chance, please resolve the merge conflicts here.  Thanks!

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -61,29 +61,58 @@ All implementations should use the same calculations for the BRDF inputs. Implem
...
+# layer coating on top of base material
+coated_material = fresnel_coat(
+  layer = clearcoat_brdf,
+  base = material,
+  ior = 1.5,
+  weight = clearcoatTexture.r * clearcoatFactor,
+  normal = clearcoatNormalTexture)
+```
+
+The `fresnel_coat` operator adds a BRDF as a layer on top of another BSDF according to `weight` and a Fresnel term. The layer is weighted with `weight*fresnel(ior)`. The base is weighted with `1-(weight*fresnel(ior)`. `normal` is a float3 vector that affects the top layer, but not the base.
```
abwood, line 84: Minor: `1-(weight*fresnel(ior)` is missing closing parenthesis.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -61,29 +61,58 @@ All implementations should use the same calculations for the BRDF inputs. Implem
...
+# coating BRDF
+clearcoat_brdf = specular_brdf((clearcoatRoughnessFactor * clearcoatRoughnessTexture.g)^2)
+
+# layer coating on top of base material
+coated_material = fresnel_coat(
+  layer = clearcoat_brdf,
+  base = material,
+  ior = 1.5,
+  weight = clearcoatTexture.r * clearcoatFactor,
+  normal = clearcoatNormalTexture)
```
abwood, line 81: I'm a bit confused as to why normal is called out as an input into the fresnel_coat, but is not mentioned in specular_brdf.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
abwood: According to the implementation below, layer is weighted with just `weight`.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -61,29 +61,58 @@ All implementations should use the same calculations for the BRDF inputs. Implem
...
 clearcoatRoughness = clearcoatRoughnessFactor * clearcoatRoughnessTexture.g
 ```
-
 The following abstract code describes how the base and clearcoat layers should be blended together:
   
 ```
 clearcoatBlendFactor = clearcoatTexture.r * clearcoatFactor
 clearcoatFresnel = fresnel(0.04, NdotV)
-
 color = (f_emissive + f_diffuse + f_specular) * (1.0 - clearcoatBlendFactor * clearcoatFresnel) +
```
abwood, line 112: I'm wondering if this implementation section should get updated to look more like an implementation of the pseudo-code above. E.g.:
```
vec3 fresnel_coat(baseLayer, coatLayer, ior, weight)
{
    float fresnelWeight = fresnel(ior, NdotV) * weight;
    return mix(baseLayer, coatLayer, fresnelWeight);
}
```

```
vec3 baseMaterial = mix(dielectric_brdf, metal_brdf, metallic) + emission
vec3 clearcoatMaterial = specular_brdf((clearcoatRoughnessFactor * clearcoatRoughnessTexture.g)^2);  // F*Vis*D, see Appendix B
vec3 coatedMaterial = fresnel_coat(baseMaterial, clearcoatMaterial, ior, weight);
```

abwood:
Something that has always bugged me in Appendix B is that we define Geometric Occlusion (G) in terms of Vis.  While we are making changes, should we rename that section to "Microfacet Shadowing (Vis)"?

abwood:
I feel like there is a disconnect between the brdf pseudo code and the implementation. I wonder if the implementation section would be more cohesive if we renamed f<sub>diffuse</sub> to diffuse_brdf, f<sub>specular</sub> to specular_brdf, defined fresnel_mix, and so on.  We could then reference these building block from other extensions, as we are doing in the clearcoat extension.

specification/2.0/README.md
```diff
@@ -3901,9 +3889,49 @@ Application-specific data.
...
 # Appendix B: BRDF Implementation
 
-*This section is non-normative.*
+The metallic-roughness material model is a linear blend of two BRDFs, a metallic BRDF and a dielectric BRDF. The blending factor `metallic` describes the metalness of the material. The BRDFs share the parameters for roughness and base color, following [Burley (2012): Physically-Based Shading at Disney](https://disney-animation.s3.amazonaws.com/library/s2012_pbs_disney_brdf_notes_v2.pdf).
+
+```
+material = mix(dielectric_brdf, metal_brdf, metallic)
+```
+
+The metal BRDF is based on a microfacet model which describes the orientation of microfacets on the surface as a statistical distribution. The distribution is controlled by a parameter called `roughness`, varying between 0 (smooth surface) and 1 (rough surface). As the blending between the two extremes does not behave linearly, like most microfacet distributions, the material's `roughness` parameter is squared before using it as the distribution's roughness.
```
elalish, line 3898: I would describe the microfacet model separately since it is used by both metals and dielectrics. It is the Trowbridge-Reitz model (also known as GGX) where the microfacets are distributed on infinitesimal oblate ellipsoids, whose half-height in the normal direction is the `roughness^2` times the radius in the tangent plane. Roughness == 1 gives spheres, which result in uniform reflection in all directions, absent geometric shadowing. Geometric shadowing and multiscattering can be derived from this microfacet distribution, with common approximations given in the implementation section.

devshgraphicsprogramming:
@Crisspl maybe you'd like to get involved?

devshgraphicsprogramming:
I would completely support something along the lines of Nvidia MDL (which would be banal to express in json) except that the BRDF blocks that can be composed in a DAG are already predefined, much like Mitsuba's `<bsdf>` plugins.

Such a solution is feasible, my team has a material compiler from a DAG Intermediate Representation into a GPU state machine which optimizes out states that can never be transferred into, right now its only frontend is Mitsuba XML and backend is GLSL. However we aim to support Nvidia MDL as a front end, while CUDA, CUDA+OptiX (leveraging SBT) and GLSL closestHit shaders as extra backends.

Futhermore it would unify #1717 , #1718 and #1719 into a single solution instead of fragmented extensions.

emackey:
@devshgraphicsprogramming We value outside perspectives greatly, but I'd like to be careful not to derail this particular PR with ideas that run in the opposite direction.  This group did carefully consider MDL, including a full presentation on the language from some of the key members of NVIDIA's MDL team.  Some future version of glTF should ideally support either MDL or some other configurable shader type of arrangement.

That said, this particular PR is in support of shoring up the core glTF PBR definition, with a short-term goal of enabling the next wave of PBR parameters to arrive via extensions.  We've chosen to keep the extensions separate during their development, but with a careful eye on the end goal of all of the new "PBR Next" material properties working together in harmony, for example as part of a client's core material definition when this is done.

MDL, or something like it, will probably re-appear on the roadmap at some future point.  But in the short term, "PBR Next" will appear more similar to systems such as [Dassault Systèmes Enterprise PBR](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/) and [Autodesk Standard Surface](https://autodesk.github.io/standard-surface/).

donmccurdy:
@emackey perhaps we should close https://github.com/KhronosGroup/glTF/issues/1442, now that the PBR Next effort is well underway, and open a similar (long-term) survey issue for what users hope to see, or not see, in glTF with respect to node-based material systems like MDL and MaterialX?

proog128:
I pushed an update with the following major changes:
* Rebase onto current master (affects mostly the clearcoat extension).
* A lot more details about microfacet surface.
* Refactored non-normative implementation section: The code is still the same as before (= as in the sample viewer), with all its advantages (small, simple, performance) and quirks (mainly the non energy-conserving layering). But now it links to the normative section, i.e., it is an implementation for the functions used there.
* There are now a couple of small details in the formulas in the implementation section that were missing before (abs, ...). I hope it's correct, please have a close look at it.
* There is now a discussion section that explains some of the design choices in the implementation and mentions alternatives.
* Clearcoat now uses the same terms and structure as in the main spec.

I hope that this separation of normative and non-normative parts now makes clear which parts can be changed by an implementor/are allowed to be approximated and which parts have to stay as in the spec. In comparison to the original document (master branch), it's more complicated, because I had to introduce another layer of abstraction. I think there is still potential to simplify it, but I had to keep some complexity to avoid having to rewrite all the other extensions (because they refer to some terms in Appendix B).

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
proog128: Fixed.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
proog128: Fixed, now the clearcoat normal is passed to specular_brdf. It's still not perfect, the function as defined in Appendix B doesn't have this parameter. I am a bit afraid of making the normal a parameter everywhere, might be bad for readability.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
proog128: Done. That's a good suggestion.

specification/2.0/README.md
```diff
@@ -3901,9 +3889,49 @@ Application-specific data.
...
 
-The glTF spec is designed to allow applications to choose different lighting implementations based on their requirements.
+Metallic surfaces reflect back most of the illumination, only a small portion of the light is absorbed by the material. This effect is described by the Fresnel term with the wavelength-dependent refractive index and extinction coefficient. To make parameterization simple, the metallic-roughness material combines the two quantities into a single, user-defined color value `baseColor` that defines the reflection color at normal incidence, also referred to as `f0`.
+
+```
+metal_brdf =
+  conductor_fresnel(
```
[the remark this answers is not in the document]
proog128: All functions used in the pseudo-code are now defined in the non-normative implementation section.

specification/2.0/README.md
```diff
@@ -3901,9 +3889,49 @@ Application-specific data.
 
 # Appendix B: BRDF Implementation
 
-*This section is non-normative.*
+The metallic-roughness material model is a linear blend of two BRDFs, a metallic BRDF and a dielectric BRDF. The blending factor `metallic` describes the metalness of the material. The BRDFs share the parameters for roughness and base color, following [Burley (2012): Physically-Based Shading at Disney](https://disney-animation.s3.amazonaws.com/library/s2012_pbs_disney_brdf_notes_v2.pdf).
```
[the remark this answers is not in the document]
proog128: That's a very good idea, I added a few sentences that describe the motivation behind the blending and how to use it.

specification/2.0/README.md
proog128: Thanks a lot for that! I added it to the text.

specification/2.0/README.md
```diff
@@ -3903,67 +3891,230 @@ Application-specific data.
...
+
+The diffuse reflection `diffuse_brdf(color)` is a Lambertian BRDF
+
+<img src="https://render.githubusercontent.com/render/math?math=\displaystyle \text{LambertianBRDF} = \frac{1}{\pi}">
+
+multiplied with the `color`.
+
+```
+function diffuse_brdf(color) {
+  return (1/pi) * color
```
elalish, line 4015: This looks like only the D term; I believe you still need the G term multiplied here. Otherwise your diffuse is independent of light direction, and I don't think you're getting any diffuse reflection from light behind the tangent plane.

devshgraphicsprogramming: diffuse BRDF is `albedo/PI` the NdotL factor is to account for projected solid angle and is not part of the BRDF itself.

If however you are referencing with G to the microfacet masking function (i.e. Smith) then that would also be incorrect because G is only defined for a specific microfacet with a specific normal, which is only the case for consideration in specular BRDFs.

The only thing that makes sense here is a Heaviside function of `NdotL` but thats usually omitted because when you render in realtime by evaluating the BRDF at specific points, you take care not to evaluate for points outside the hemisphere (behind the tangent plane), and when you render with Monte Carlo or sampling methods, your sample generator should never produce such samples anyway.

elalish: Hmm, not sure I agree. I think the microfacet and shadowing model applies just as much to the diffuse as the specular, since shadowing stops grazing light from entering the surface at the place you expect. If you use a Heaviside (step) function instead of a smooth G, all that means is that a point light will produce a hard rather than soft shadow line for the colored reflection, which I don't think is a more accurate render. In any case, you said yourself that at least a Heaviside function is required here (which I would call an approximation of G), so either way I think it should be mentioned.

proog128: `NdotL` is not part of the BRDF. The BRDF is defined as reflected differential radiance over differential incident irradiance. 

`NdotL` is part of the integral over all incoming light directions (upper hemisphere around the surface normal) which a renderer needs to solve to compute the outgoing radiance. I don't think it should be described here.

The Lambertian BRDF used here is the most simple BRDF possible, it's just a constant. A more realistic diffuse-like BRDF that fits into the microfacet framework is the Oren-Nayar BRDF. In this model, the facets are assumed to be Lambertian (as opposed to perfectly specular in the "typical" glossy microfacet models).

elalish: Ah, okay, I see my mistake now. So `material = ` means the material's BRDF. I think because I saw some `LdotH` terms in there that we were closer to light output. I think it might be nice to show the BRDF integral somewhere just to make it clear how to get all the way from input light to viewer pixel color. I looked up at the material section that references this appendix, and it doesn't define BRDF at all either.

devshgraphicsprogramming: > NdotL is not part of the BRDF. The BRDF is defined as reflected differential radiance over differential incident irradiance.

What I've said but more formally.

> The Lambertian BRDF used here is the most simple BRDF possible, it's just a constant. A more realistic diffuse-like BRDF that fits into the microfacet framework is the Oren-Nayar BRDF. In this model, the facets are assumed to be Lambertian (as opposed to perfectly specular in the "typical" glossy microfacet models).

Also Oren-Nayar assumes the V-cavities model, not GGX's NDF.

AFAIK, Oren-Nayar is not a closed form solution but rather a numerical fit of a simulation of such a microfacet model, correct me if I'm wrong

> Hmm, not sure I agree. I think the microfacet and shadowing model applies just as much to the diffuse as the specular, since shadowing stops grazing light from entering the surface at the place you expect. If you use a Heaviside (step) function instead of a smooth G, all that means is that a point light will produce a hard rather than soft shadow line for the colored reflection, which I don't think is a more accurate render. In any case, you said yourself that at least a Heaviside function is required here (which I would call an approximation of G), so either way I think it should be mentioned.

Like I said, if you assume microfacets to be anything other than perfectly smooth, your average microfacet reflection is not 1/PI anymore...

A perfectly smooth surface has no geometric shadowing, except by the tangent plane.

We can mention a heaviside function, but its not required in the reference implementation because of practicality:
1) No one will call the BRDF evaluation function when the observer is below the tangent plane (you'd be seeing the back face of a triangle, VdotN is negative) * actually for smooth shaded triangles, a `if (NdotV>0.0)` check would make sense.
2) No one will call the BRDF evaluation function when the surface is backfacing towards the light, people usually use a `max(NdotL,0.0)` outside the BRDF or an `if (NdotL>0.0)` before evaluating lighting

Changing the microfacet distribution to V-cavities necessitates you use Oren-Nayar, which already has the geometric shadowing baked into the fit AFAIK.

Using the same GGX distribution as specular introduces its own problems, mainly that such a diffuse BRDF has no closed solution and needs to be stochastically simulated with inifnite random walks like in Heitz's papers. and it seems there is no standardised numerical fit.

Also you need to decide whether the specular and diffuse distributions are in a coating or a fuzed substrate, basically:
1) is there a layer of coating (varnish) that has different microsurface on top and on bottom (uncorrelated microfacets from two layers)
2) are the individual microfacets all the same and exhibit specular and diffuse behaviour (correlated microfacet from pretty much a single layer)

For practicality, all the models I've seen even in offline rendering use option (1) where the two BRDFs are treated as separate physical layers.

specification/2.0/README.md
```diff
@@ -3903,67 +3891,234 @@ Application-specific data.
...
+function fresnel_mix(ior, base, layer) {
+  f0 = ((1-ior)/(1+ior))^2
+  fr = f0 + (1 - f0)*(1 - abs(VdotH))^5
+  return mix(base, layer, fr)
+}
+```
+
+### Metal BRDF and Dielectric BRDF
+
+Applying the functions we arrive at the metal BRDF and dielectric BRDF:
```
abwood, line 4049: My understanding is that you are trying to build a bridge here between the old Appendix B content and this new formalization.  I think we are missing some details in this translation of the new representation of the brdf from the old representation.  For example, it's not clear here that 0.04 is the f0 when ior is 1.5.  Do we need to elaborate a bit more on why this section is here, and expand a bit more on how we arrive at the metal_brdf and dielectric_brdf?

proog128: My idea was to first describe the material as a normative node graph (Section "Complete Model"), then give a non-normative (sample) implementation for the nodes (Sections "Specular BRDF", "Diffuse BRDF", "Fresnel"), and finally substitute all nodes in the graph for their implementations. This is how we arrive at the old (shader code) representation. I added some notes to the beginning of the "Metal BRDF and Dielectric BRDF" section to make the transition more smooth (hopefully). Does that help?

emackey:
This has been a long time in the making, and it's now in a state where I think merging it is a big improvement over what's been here.

I know some people want further tweaks or fine-tuning.  Let's open new PRs or new issues for any such remaining concerns.

PR #1718, opened 2019-12-03
KHR_materials_ior

proog128:
An extension for the metallic-roughness material that makes the index of refraction configurable. At the moment, there is a fixed value of 1.5, corresponding to F0 = 0.04.

This extension is similar to KHR_materials_specular proposed in #1677, but should work better in combination with KHR_materials_transmission (#1698). It allows to set IOR>2, which is not uncommon for transparent materials.

The extension uses some notation introduced in #1717.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_ior/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_ior/README.md
@@ -76,9 +76,11 @@ dielectricSpecular = ((ior - 1)/(ior + 1))^2
 
 ## Interaction with other extensions
 
-If `KHR_materials_ior` is used in combination with `KHR_materials_specular`, f0 computed from IOR is multiplied by f0 computed from the specular parameters. See [`KHR_materials_specular`](../KHR_materials_specular/README.md) for more details.
+If `KHR_materials_specular` is used in combination with `KHR_materials_ior`, f0 computed from IOR is multiplied by f0 computed from the specular parameters. See [`KHR_materials_specular`](../KHR_materials_specular/README.md) for more information.
 
 If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `ior` affects not only the Fresnel term, but also determines the refractive index of the volume below the surface. Light rays passing through the transmissive surface are bent according to the index of refraction of the outside and inside medium (refraction). As the volume cannot be parametrized with a 2-dimensional texture in UV space, the index of refraction (IOR) is a scalar, uniform value (`ior`).
+
+If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `ior` affects the refraction effect. See [`KHR_materials_volume`](../KHR_materials_volume/README.md) for more information.
 
 ## Schema
 
--- a/extensions/2.0/Khronos/KHR_materials_ior/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_ior/README.md
@@ -78,7 +78,7 @@ dielectricSpecular = ((ior - 1)/(ior + 1))^2
 
 If `KHR_materials_specular` is used in combination with `KHR_materials_ior`, f0 computed from IOR is multiplied by f0 computed from the specular parameters. See [`KHR_materials_specular`](../KHR_materials_specular/README.md) for more information.
 
-If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `ior` affects not only the Fresnel term, but also determines the refractive index of the volume below the surface. Light rays passing through the transmissive surface are bent according to the index of refraction of the outside and inside medium (refraction). As the volume cannot be parametrized with a 2-dimensional texture in UV space, the index of refraction (IOR) is a scalar, uniform value (`ior`).
+If `KHR_materials_transmission` is used in combination with `KHR_materials_ior`, the `ior` affects the strength of the transmission effect according to the Fresnel term. See [`KHR_materials_transmission`](../KHR_materials_transmission/README.md) for more information. If `KHR_materials_specular` is used in addition, the rules for multiplying f0 from `ior` and f0 from the specular parameters apply as defined in `KHR_materials_specular`.
 
 If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `ior` affects the refraction effect. See [`KHR_materials_volume`](../KHR_materials_volume/README.md) for more information.
 
--- a/extensions/2.0/Khronos/KHR_materials_ior/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_ior/README.md
@@ -88,7 +88,7 @@ If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `
 
 ## Schema
 
-- [glTF.KHR_materials_specular.schema.json](schema/glTF.KHR_materials_specular.schema.json)
+- [glTF.KHR_materials_ior.schema.json](schema/glTF.KHR_materials_ior.schema.json)
 
 ## Appendix: Full Khronos Copyright Statement
 
--- a/extensions/2.0/Khronos/KHR_materials_ior/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_ior/README.md
@@ -76,17 +76,7 @@ dielectricSpecular = ((ior - 1)/(ior + 1))^2
 
 ## Interaction with other extensions
 
-If `KHR_materials_ior` is used in combination with `KHR_materials_specular`, f0 computed from IOR is multiplied by f0 computed from the specular parameters.
-
-```
-dielectricSpecularF0 = ((ior - outside_ior) / (ior + outside_ior))^2 * specularFactor * specularTexture.a * specularColorFactor * specularColorTexture.rgb
-dielectricSpecularF90 = specularFactor * specularTexture.a
-
-F0  = lerp(dielectricSpecularF0, baseColor.rgb, metallic)
-F90 = lerp(dielectricSpecularF90, 1, metallic)
-
-F = F0 + (F90 - F0) * (1 - VdotH)^5
-```
+If `KHR_materials_ior` is used in combination with `KHR_materials_specular`, f0 computed from IOR is multiplied by f0 computed from the specular parameters. See [`KHR_materials_specular`](../KHR_materials_specular/README.md) for more details.
 
 If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `ior` affects not only the Fresnel term, but also determines the refractive index of the volume below the surface. Light rays passing through the transmissive surface are bent according to the index of refraction of the outside and inside medium (refraction). As the volume cannot be parametrized with a 2-dimensional texture in UV space, the index of refraction (IOR) is a scalar, uniform value (`ior`).
 
--- a/extensions/2.0/Khronos/KHR_materials_ior/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_ior/README.md
@@ -38,6 +38,10 @@ The index of refraction of a material is configured by adding the `KHR_materials
 }
 ```
 
+| |Type|Description|Required|
+|-|----|-----------|--------|
+| **ior** | `number` | The index of refraction. | No, default: `1.5`|
+
 Typical values for the index of refraction range from 1 to 2. In rare cases values greater than 2 are possible. The following table shows some materials and their index of refraction:
 
 | Material     | Index of Refraction |
@@ -49,30 +53,39 @@ Typical values for the index of refraction range from 1 to 2. In rare cases valu
 | Sapphire     | 1.76                |
 | Diamond      | 2.42                |
 
-In case `KHR_materials_volume` is enabled, the `ior` determines the refractive index of the volume below the surface. Light rays passing through the transmissive surface are bent according to the index of refraction of the outside and inside medium (refraction). As the volume cannot be parametrized with a 2-dimensional texture in UV space, the index of refraction (IOR) is a scalar, uniform value (`ior`).
-
-| |Type|Description|Required|
-|-|----|-----------|--------|
-| **ior** | `number` | The index of refraction. | No, default: `1.5`|
-
-## Implementation
-
-The index of refraction affects the Fresnel term in the dielectric BRDF:
+The reflectance at normal incidence (F0) of dielectric materials is computed from the IOR as follows:
 
 ```
-dielectric_brdf =
-  fresnel_mix(
-    diffuse_brdf(baseColor),
-    microfacet_brdf(roughness^2),
-    ior)
-```
-
-[Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation) defines the `fresnel_mix` operation as
-
-```
-fresnel_mix(base, layer, ior) = base * (1 - fr(ior)) + layer * fr(ior)
-fr(ior) = f0 + (1 - f0) * (1 - cos)^5
 f0 = ((ior - outside_ior) / (ior + outside_ior))^2
 ```
 
-with `ior = 1.5`, corresponding to `f0 = 0.04`. The `outside_ior` determines the index of refraction of the medium on the outside. If the renderer does not track the IOR when traversing nested dielectrics, it can assume `outside_ior = 1`.
+The `outside_ior` determines the index of refraction of the medium on the outside. If the renderer does not track the IOR when traversing nested dielectrics, it can assume `outside_ior = 1`.
+
+For `ior = 1.5` and `outside_ior = 1`, the result is `f0 = 0.04`. This is the fixed value used in the glTF 2.0 metallic-roughness material. The extension makes this value configurable.
+
+
+## Implementation
+
+*This section is non-normative*
+
+The extension changes the computation of the Fresnel term defined in [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation) to the following:
+
+```
+dielectricSpecular = ((ior - 1)/(ior + 1))^2
+```
+
+## Interaction with other extensions
+
+If `KHR_materials_ior` is used in combination with `KHR_materials_specular`, the constant `0.04` is replaced by the value computed from the IOR.
+
+```
+dielectricSpecularF0 = ((ior - outside_ior) / (ior + outside_ior))^2 * specularFactor * specularTexture.a * specularColorFactor * specularColorTexture.rgb
+dielectricSpecularF90 = specularFactor * specularTexture.a
+
+F0  = lerp(dielectricSpecularF0, baseColor.rgb, metallic)
+F90 = lerp(dielectricSpecularF90, 1, metallic)
+
+F = F0 + (F90 - F0) * (1 - VdotH)^5
+```
+
+If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `ior` affects not only the Fresnel term, but also determines the refractive index of the volume below the surface. Light rays passing through the transmissive surface are bent according to the index of refraction of the outside and inside medium (refraction). As the volume cannot be parametrized with a 2-dimensional texture in UV space, the index of refraction (IOR) is a scalar, uniform value (`ior`).
--- a/extensions/2.0/Khronos/KHR_materials_ior/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_ior/README.md
@@ -89,3 +89,7 @@ F = F0 + (F90 - F0) * (1 - VdotH)^5
 ```
 
 If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `ior` affects not only the Fresnel term, but also determines the refractive index of the volume below the surface. Light rays passing through the transmissive surface are bent according to the index of refraction of the outside and inside medium (refraction). As the volume cannot be parametrized with a 2-dimensional texture in UV space, the index of refraction (IOR) is a scalar, uniform value (`ior`).
+
+## Schema
+
+- [glTF.KHR_materials_specular.schema.json](schema/glTF.KHR_materials_specular.schema.json)
--- a/extensions/2.0/Khronos/KHR_materials_ior/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_ior/README.md
@@ -31,7 +31,6 @@ The index of refraction of a material is configured by adding the `KHR_materials
             "extensions": {
                 "KHR_materials_ior": {
                     "ior": 1.4,
-                    "iorTexture": 0
                 }
             }
         }
@@ -39,12 +38,35 @@ The index of refraction of a material is configured by adding the `KHR_materials
 }
 ```
 
-The index of refraction typically is a value between 1 and 2, in rare cases it may be larger than 2. In contrast to other texturable values in glTF, `ior` and `iorTexture` are not multiplied: if no texture is given, `ior` is used, otherwise the texture overrides it.
+Typical values for the index of refraction range from 1 to 2. In rare cases values greater than 2 are possible. The following table shows some materials and their index of refraction:
+
+| Material     | Index of Refraction |
+|--------------|---------------------|
+| Air          | 1.0                 |
+| Water        | 1.33                |
+| Eyes         | 1.38                |
+| Window Glass | 1.52                |
+| Sapphire     | 1.76                |
+| Diamond      | 2.42                |
+
+The index of refraction (IOR) can be defined either as a scalar value (`ior`) or as a texture (`f0Texture`). If both are given at the same time, `f0Texture` overrides `ior`.
 
 | |Type|Description|Required|
 |-|----|-----------|--------|
 | **ior** | `number` | The index of refraction. | No, default: `1.5`|
-| **iorTexture** | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A greyscale texture that defines the index of refraction as 1/ior. | No |
+| **f0Texture** | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A 1-channel luminance texture that defines the index of refraction in terms of the reflectance at normal incidence f0. | No |
+| **maxF0** | `number` | When using `f0Texture`, a texture value of 1.0 is interpret as `maxF0`. Values between 0 and 1 are linearly scaled between 0 and `maxF0`. | No, default: `0.08` |
+
+The texture stores the IOR in terms of the reflectance at normal incidence, a value in the range 0..1. The mapping works as follows:
+
+```
+f0 = ((ior - 1) / (ior + 1))^2
+ior = 2 / (1 - sqrt(f0)) - 1
+```
+
+As f0 usually is a small value (IOR in range 1..2 corresponds to f0 in range 0..0.08), an optional scale factor can be applied to the texture via `maxF0`.
+
+In case `KHR_materials_volume` is enabled, the `ior` affects the bending of light rays (refraction) passing through the transparent surface and it determines the refractive index of the volume below the surface. As the volume cannot be parametrized with a 2-dimensional texture in UV space, `f0Texture` and `maxF0` are ignored.
 
 ## Implementation
 
@@ -58,4 +80,20 @@ dielectric_brdf =
     ior)
 ```
 
-In case `KHR_materials_transmission` is enabled, the index of refraction also affects the refraction of light rays passing through the transparent surface.
+[Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation) defines the `fresnel_mix` operation as
+
+```
+fresnel_mix(base, layer, ior) = base * (1 - fr(ior)) + layer * fr(ior)
+fr(ior) = f0 + (1 - f0) * (1 - cos)^5
+f0 = ((ior - 1) / (ior + 1))^2
+```
+
+with `ior = 1.5`, corresponding to `f0 = 0.04`. The following pseudo-code shows how to compute `f0` from `ior`, `f0Texture` and `maxF0`:
+
+```
+if (f0Texture is enabled) {
+  f0 = fetch(f0Texture, uv) * maxF0;
+} else {
+  f0 = ((ior - 1) / (ior + 1))^2
+}
+```
--- a/extensions/2.0/Khronos/KHR_materials_ior/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_ior/README.md
@@ -20,6 +20,11 @@ Experimental
 
 Written against the glTF 2.0 spec.
 
+## Exclusions
+
+* This extension must not be used on a material that also uses `KHR_materials_pbrSpecularGlossiness`.
+* This extension must not be used on a material that also uses `KHR_materials_unlit`.
+
 ## Overview
 
 The dielectric BRDF of the metallic-roughness material in glTF uses a fixed value of 1.5 for the index of refraction. This is a good fit for many plastics and glass, but not for other materials like water or asphalt, sapphire or diamond. This extensions allows users to set the index of refraction to a certain value.
--- a/extensions/2.0/Khronos/KHR_materials_ior/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_ior/README.md
@@ -49,24 +49,11 @@ Typical values for the index of refraction range from 1 to 2. In rare cases valu
 | Sapphire     | 1.76                |
 | Diamond      | 2.42                |
 
-The index of refraction (IOR) can be defined either as a scalar value (`ior`) or as a texture (`f0Texture`). If both are given at the same time, `f0Texture` overrides `ior`.
+In case `KHR_materials_volume` is enabled, the `ior` determines the refractive index of the volume below the surface. Light rays passing through the transmissive surface are bent according to the index of refraction of the outside and inside medium (refraction). As the volume cannot be parametrized with a 2-dimensional texture in UV space, the index of refraction (IOR) is a scalar, uniform value (`ior`).
 
 | |Type|Description|Required|
 |-|----|-----------|--------|
 | **ior** | `number` | The index of refraction. | No, default: `1.5`|
-| **f0Texture** | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A 1-channel luminance texture that defines the index of refraction in terms of the reflectance at normal incidence f0. | No |
-| **maxF0** | `number` | When using `f0Texture`, a texture value of 1.0 is interpret as `maxF0`. Values between 0 and 1 are linearly scaled between 0 and `maxF0`. | No, default: `0.08` |
-
-The texture stores the IOR in terms of the reflectance at normal incidence, a value in the range 0..1. The mapping works as follows:
-
-```
-f0 = ((ior - 1) / (ior + 1))^2
-ior = 2 / (1 - sqrt(f0)) - 1
-```
-
-As f0 usually is a small value (IOR in range 1..2 corresponds to f0 in range 0..0.08), an optional scale factor can be applied to the texture via `maxF0`.
-
-In case `KHR_materials_volume` is enabled, the `ior` affects the bending of light rays (refraction) passing through the transparent surface and it determines the refractive index of the volume below the surface. As the volume cannot be parametrized with a 2-dimensional texture in UV space, `f0Texture` and `maxF0` are ignored.
 
 ## Implementation
 
@@ -85,15 +72,7 @@ dielectric_brdf =
 ```
 fresnel_mix(base, layer, ior) = base * (1 - fr(ior)) + layer * fr(ior)
 fr(ior) = f0 + (1 - f0) * (1 - cos)^5
-f0 = ((ior - 1) / (ior + 1))^2
+f0 = ((ior - outside_ior) / (ior + outside_ior))^2
 ```
 
-with `ior = 1.5`, corresponding to `f0 = 0.04`. The following pseudo-code shows how to compute `f0` from `ior`, `f0Texture` and `maxF0`:
-
-```
-if (f0Texture is enabled) {
-  f0 = fetch(f0Texture, uv) * maxF0;
-} else {
-  f0 = ((ior - 1) / (ior + 1))^2
-}
-```
+with `ior = 1.5`, corresponding to `f0 = 0.04`. The `outside_ior` determines the index of refraction of the medium on the outside. If the renderer does not track the IOR when traversing nested dielectrics, it can assume `outside_ior = 1`.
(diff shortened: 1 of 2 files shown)


bhouston:
This is an amazing extension.  We should adopt it.  It is relatively simple.  I highly support IOR rather than specifying F0 directly as it is much more flexible and reusable in other BRDFs.

extensions/2.0/Khronos/KHR_materials_ior/README.md
```diff
@@ -0,0 +1,61 @@
...
+    ]
+}
+```
+
+The index of refraction typically is a value between 1 and 2, in rare cases it may be larger than 2. In contrast to other texturable values in glTF, `ior` and `iorTexture` are not multiplied: if no texture is given, `ior` is used, otherwise the texture overrides it.
+
+| |Type|Description|Required|
+|-|----|-----------|--------|
+| **ior** | `number` | The index of refraction. | No, default: `1.5`|
+| **iorTexture** | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A greyscale texture that defines the index of refraction as 1/ior. | No |
```
rsahlin: Could you please clarify that you mean a texture holding one (RGB) channel, as opposed to a greyscale image.
For instance luminance texture or pick the R, G, B or a component as carrier for the 1 / ior value.

extensions/2.0/Khronos/KHR_materials_ior/schema/glTF.KHR_materials_ior.schema.json
```diff
@@ -0,0 +1,23 @@
...
+            "type": "number",
+            "description": "The index of refraction.",
+            "default": 1.5,
+            "minimum": 1.0,
+            "gltf_detailedDescription": "The index of refraction (IOR) is a measured physical number usually in the range between 1 and 2 that determines how much the path of light is bent, or refracted, when entering a material. It also influences the ratio between reflected and transmitted light, calculated from the Fresnel equations."
+        },
+        "iorTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "A greyscale texture that defines the index of refraction.",
+            "gltf_detailedDescription": "The texture stores the inverse of the index of refraction (1/ior) to keep the values in the range 0 to 1."
```
rsahlin: Dont you mean (ior - 1) ?
If we mandate that the ior value must be >= 1 <= 2 then this should work as normalizing the value in the 0 to 1 range.
If not then I think we are wasting precision since 1 / ior values will be in the 1 - 0.5 range.
Could you please elaborate?

extensions/2.0/Khronos/KHR_materials_ior/README.md
rsahlin: I guess what I am after is number of bits and format for the ior.

extensions/2.0/Khronos/KHR_materials_ior/schema/glTF.KHR_materials_ior.schema.json
rsahlin: Maybe I will understand if you clarify the calculations in:
dielectric_brdf =
  fresnel_mix(
    diffuse_brdf(baseColor),
    microfacet_brdf(roughness^2),
    ior)

proog128: The ior varies between 1 (air) and up to 3 for some gemstones (see Wikipedia's [List of refractive indices](https://en.wikipedia.org/wiki/List_of_refractive_indices)).

I picked the 1/ior mapping because it covers the full range of physical IORs (1 to 3) and it is available in [Substance Painter](https://docs.substance3d.com/spdoc/creating-export-presets-98959398.html#Creatingexportpresets-Assigningchannelstooutputmaps) and supported by [Corona](https://docs.substance3d.com/integrations/vray-157352154.html) and [Vray](https://docs.substance3d.com/integrations/vray-157352154.html).

The following is a simple implementation for KHR_materials_ior:

```
fresnel_mix(base, layer, ior) = base * (1 - fr(ior)) + layer * fr(ior)
fr(ior) = f0 + (1 - f0) * (1 - cos)^5
f0 = lerp( ((1-ior)/(1+ior))^2, baseColor, metallic)
```

The notation is introduced in #1717. I was hoping to merge all PBR Next PRs together, so I used the notation in #1718 (KHR_materials_ior) and #1719 (KHR_materials_specular). In case we merge it separately, I will adapt the descriptions.

extensions/2.0/Khronos/KHR_materials_ior/README.md
proog128: I don't have a strong preference here. A single-channel (luminance) texture is sufficient.

extensions/2.0/Khronos/KHR_materials_ior/schema/glTF.KHR_materials_ior.schema.json
rsahlin: Thanks for the clarification!
#1717 Looks like a very good suggestion.
However, I wonder if it would be possible to store F0 value in textures?  (instead of ior , or 1 / ior) 
- Realtime renderers would want to have as much pre-calculated values as possible.
This is especially true for mobile devices.

What is the upside of storing the IOR as opposed to the F0 value from your point of view?

proog128: Storing F0 directly in the (byte) texture will have bad precision. IORs in range 1 to 3 correspond to F0 in range 0 to 0.25, which is 0 to 63 in integer.

rsahlin: In that case I would propse adding a (texture F0) scale value.
This value would be used to expand the range when creating the texture and (1 / scale) when sampling the texture.
For instance a material (texture) with max F0 of 0.25 would use 4 as a multipler when creating the texture and 0.25 as a multipler when fetching the F0 from the texture.
Another material may have a max F0 of 0.1 where the multipler would be 10 (and 0.1 in the shader)
This way the overhead would be one (scalar) multiplication in the fragment shader and we could get good usage of the (limited) number of bits.

proog128: I think a scale value works, but I am a bit afraid of the complications this will add in other areas:

Depending on the implementation, an IOR value is more useful than F0.
* Use full Fresnel equation instead of Schlick approxmation.
* Considering IOR of clearcoat layer (1.5) for base layer Fresnel.
* Refraction direction (KHR_materials_transmission, KHR_materials_volume) needs IOR.
Conversion from IOR to F0 is possible via `ior = 2 / (1-sqrt(F0))-1`.

Apart from that, I am wondering: is it really cheaper to fetch 1 uniform value (textureMaxF0) and 1 varying value (f0) and multiply it instead of fetching only 1 varying value (1/ior) and doing a bit more math to compute f0?

And my last thought on it: There are already different encodings and we will introduce yet another. I know of the following encodings:
* ior = 1/tex (V-Ray, Arnold)
* f0 = 0.08 * tex (Unreal, Cycles; limits IOR to [1,1.7])
* f0 = 0.16 * tex^2 (Frostbite, Filament; limits IOR to [1, 2.4])

We would introduce:
* f0 = textureMaxF0 * tex

Authoring tools like Substance and importers/exporters will have to handle it. Moreover, depending on how authoring tools are going to expose it in the UI, I can easily imagine that it is harder to use, artists need to know how f0 converts to IOR. Exchanging textures only without the scale factor makes them useless.

In conclusion, I think scale + f0 is perfectly fine, but only if the increased complexity in tooling and (some) implementation(s) is worth the performance gain.

rsahlin: Ok, thanks for the clarification regarding usage of IOR for other features.

> Apart from that, I am wondering: is it really cheaper to fetch 1 uniform value (textureMaxF0) and 1 varying value (f0) and multiply it instead of fetching only 1 varying value (1/ior) and doing a bit more math to compute f0?

Yes I would definately say so. Especially for mobile devices. 
In a fragment shader all cycles count**
I think that making it as 'lightweight' (calculation wise) as possible is what we should aim for.

> And my last thought on it: There are already different encodings and we will introduce yet another. I know of the following encodings:
> 
> * ior = 1/tex (V-Ray, Arnold)
> * f0 = 0.08 * tex (Unreal, Cycles; limits IOR to [1,1.7])
> * f0 = 0.16 * tex^2 (Frostbite, Filament; limits IOR to [1, 2.4])

Very good point.

I guess that if we think we could increase the quality by better utilization of the bits we should consider pushing for this solution anyway. :-)

**
In some cases calculations may be hidden or shadowed by other slower operations such as texel fetches. 
Eg, in this case calculations prior to using the texel (sampler) value could be for 'free' if shader is stalling in the sampler.
Similarly, if there is another texture being sampled directly after the ior -> F0 calculation the impact could also be invisible.
However I would not like to count on it, instead try to keep calculations to a minimum.

proog128: Thanks for the input regarding performance. I added some more details to the extension and changed the mapping from 1/ior to the one you proposed. It adds some complexity, but I hope this is justified by the better performance. A nice side-effect of this change is that with an maxF0 of 0.08 the texture becomes compatible to the Unreal and Blender `specular` parameter, decreasing complexity in the exporters/importers.

Btw., as you mentioned performance, there are actually good reasons to _not_ provide a texture for ior/f0 at all. See [this comment](https://github.com/KhronosGroup/glTF/pull/1677#issuecomment-545831193) for more details. I think that even without a texture for f0 or ior, all practical use-cases are already covered by the textures in #1719. In addtion, not offering a texture would also improve performance...

proog128: I just realized that it is much easier to use specularFactor and specularColor of #1719 to support Unreal "specular" textures. This would mean we don't need IOR textures at all.

With #1718 and #1719, f0 and f90 are computed like this:

```
f0 = ((1-ior)/(1+ior))^2 * specularFactor * specularTexture * specularColorFactor * specularColorTexture
f90 = specularFactor * specularTexture
```

Unreal "specular" uses the following formula:

```
f0 = 0.08 * specular_unreal
f90 = 1
```

If we now set ior to 1.5 (the default; no need to use KHR_materials_ior) and specularFactor to 2, specularColorTexture is identical to the Unreal specular texture (because ((1-1.5)/(1+1.5))^2 * 2 = 0.08). Much easier than `maxF0` and `f0Texture`.

@rsahlin Would that work for you? If yes, I would remove `maxF0` and `f0Texture`. The IOR will become uniform (can be converted host-side to an f0 multiplier).

bhouston: > I just realized that it is much easier to use specularFactor and specularColor of #1719 to support Unreal "specular" textures. This would mean we don't need IOR textures at all.

This is a great idea/compromise.  I support this idea.

So specularFactor affects F90, but specularColor does not.  I am okay with that.

Q: Is this how Unreal works?  (I think that in the Autodesk Standard Surface model specularColor does affect F90.)

UX3D-nopper: > * ior = 1/tex (V-Ray, Arnold)
> * f0 = 0.08 * tex (Unreal, Cycles; limits IOR to [1,1.7])
> * f0 = 0.16 * tex^2 (Frostbite, Filament; limits IOR to [1, 2.4])

The limits are wrong, as in e.g. Blender or Unreal the input specular value
allows a range `[0 .. ∞[`. If the specular scalar is in range `[0 .. ∞[` and the specular texture is in range `[0 .. 1]`, one doese have a range of `specular * specularTexture = [0 .. ∞[`

UX3D-nopper: Also, there is no reason **not** to allow textures, as Blender or UE4 do allow them.

bhouston: > Also, there is no reason not to allow textures, as Blender or UE4 do allow them.

In Barcelona you argued one reason to not go with IOR, and instead dircet F0 specification, was because if there was a iorTexture one would have to calculate the Snell's law on a per pixel basis and it was an unnecessary cost.  Thus this idea to have a specularFactor texture and specularColor texture that is outside of the ior/Snell equation -- their work on the result of the Fresnel equations.  I haven't seen any physical situations where you want to have a varying IOR anyhow -- but maybe that could happen if one bakes together a lot of unrelated materials into one.

UX3D-nopper: Yes, I do not want to have an IOR Texture, as I do not want to have these calculations per fragment. But I do want to have a F0 Texture (and related scalar) - and exactly the same encoding like e.g. in Blender and UE4 are using and e.g. can be created with Substance Painter.
Using this encoding - for reflection - no conversion calculation is needed and the final F0 is just a multiplication of the sampled value with the scalar and a constant.
Regarding reflection, both the Metallic-Roughness and Specular-Glossiness are already using varying F0 - just with a different encoding and an additonal texture.
And this new calculated F0 can directly be passed to the current Metallic-Roughness and even (if want) Specular-Glossiness workflow functions.

see and comapre to https://academy.substance3d.com/courses/the-pbr-guide-part-2

bhouston: The specularColor texture that Tobias is proposing that would be equivalent to F0 texture btw mathematically.  Thus I think we should have ior (defines F0), specularFactor (affects both F0 and F90) and specularColor (affects F0 only.)  This achieves the flexibility you want Nopper and achieves the parameterization that I want.

proog128: > Also, there is no reason not to allow textures, as Blender or UE4 do allow them.

Depends. Textures for IOR/F0 are ok if the material is not refractive. But as soon as a renderer supports volumes and refraction, the behavior of a 2D IOR texture that affects refraction is undefined. Yes, you can certainly implement it somehow. But for a render-independent material standard we have to be more precise.

The same reasoning btw. applies to separating IOR/F0 for specular BRDF and specular BTDF. Yes, you can somehack hack it, but it's bad for a renderer-independent standard.

> So specularFactor affects F90, but specularColor does not. I am okay with that.

Yes. Besides the compatibility to Unreal, this is also needed for backwards-compatibility to glTF Spec/Gloss material. (And we don't do anything wrong by defining it in this way, it is physically-correct: in the limit F90 is always 1 for all real-world materials). I summarized the details in [this document](https://docs.google.com/document/d/1SvXWa6XGnFgsRXa1z4xw6QR1_jMmNSUTdRAmalCgoeI/edit?usp=sharing).

UX3D-nopper: That is the reason, why we want to split up this parameter for reflection and refraction.
Also, we are not looking for an independent material standard, we are looking for glTF physically based rendering next - that's a little but important difference.
Also, even if this becomes a material standard, then a texture should already be defined. If a renderer can already handle this texture (e.g. Cycles, Eevee, UE4, ...) that is good, if some renderers can't then these companies should revisit their renderers.

proog128: > That is the reason, why we want to split up this parameter for reflection and refraction.

Blender also _does not_ split up IOR/F0 parameters for reflection (specular BRDF) and refraction (specular BTDF) (for good reasons!). For details, see the document I linked before.

> Also, we are not looking for an independent material standard, we are looking for glTF physically based rendering next - that's a little but important difference.

I am interested in a file/material format that is not ambiguous and can be implemented in rasterizers and ray tracers, with consistent visual results.

bhouston: @UX3D-nopper I can not tell if you support ior, specularFactor, specularColor or not.

To me this proposal seems to align well with Dassault Enterprise PBR as well as Autodesk Standard Surface, it is also compatible with Unreal Engine's model, and it meets your requirement of fast.

I think that others can adopt this model if they do not already.  It makes a lot of sense.

UX3D-nopper: image
Blender does split it up. Relection is Specular, Refraction is IOR. see also here https://docs.blender.org/manual/en/latest/render/shader_nodes/shader/principled.html

UX3D-nopper: Yes, me too, and the file and material format are two separate things. But as I said, we will discuss endless. Let's hear what the others are saying.

proog128: Yes, there are two parameters. They behave as follows:
* "transmission" is a linear blend between a non-transparent "plastic" BSDF and a transparent "glass" BSDF.
* "specular" affects IOR/F0 of _only_ the plastic BSDF
* "IOR" affects IOR/F0 and outgoing direction of _only_ the glass BSDF

This avoids the problematic case, different IORs/F0s for outgoing direction/refraction and reflection in the glass BSDF.

Works the same as the proposed extensions #1718 and #1719 (and Enterprise PBR). #1719 gives a little bit more freedom, so that the glTF Spec/Gloss and Unity Spec/Gloss materials are also supported, and compatibility to Standard Surface is better.

UX3D-nopper: > @UX3D-nopper I can not tell if you support ior, specularFactor, specularColor or not.
> 
> To me this proposal seems to align well with Dassault Enterprise PBR as well as Autodesk Standard Surface, it is also compatible with Unreal Engine's model, and it meets your requirement of fast.
> 
> I think that others can adopt this model if they do not already. It makes a lot of sense.

Standard Surface does allow a texture. Enterprise PBR does not allow a texture. So, if we would use Standard Surface, we would not have this discussion. Only how to encode the values.

UX3D-nopper: > @UX3D-nopper I can not tell if you support ior, specularFactor, specularColor or not.
> 
> To me this proposal seems to align well with Dassault Enterprise PBR as well as Autodesk Standard Surface, it is also compatible with Unreal Engine's model, and it meets your requirement of fast.
> 
> I think that others can adopt this model if they do not already. It makes a lot of sense.

Seems, but it does not. Does not take into account, what other engines and content creation tools already solved for a good reason.

bhouston: > So, if we would use Standard Surface, we would not have this discussion. Only how to encode the values.

I think we do not need a texture for IOR.  I think it is an edge case and I've never been asked for by our artists.  Also if we did adopt a texture for it, I believe you would argue against a unified IOR here and in favor of specifying F0 directly for speed reasons (a need to calculate Snell equation on a per pixel basis.)  Thus I believe this proposal in this fashion threads the needle between a few opposing camps into a near-perfect solution.

It is okay if glTF has less degrees of freedom compared to other more complex models -- in this case an iorTexture.  Not all degrees of freedom are equal in terms of their usefulness or practicality -- and this is one of those cases in my opinion.

UX3D-nopper: image
So your artists are not allowed to do this?

bhouston: They can plug in a monochrome texture to specularColor.  It works great.  It is more intuitively understandable as well because it just modulates the F0 value rather than a more complex non-linear response, which honestly is extremely hard to explain to artists.

Remember it works like this as already mentioned above @proog128:

```
f0 = ((1-ior)/(1+ior))^2 * specularFactor * specularTexture * specularColorFactor * specularColorTexture
f90 = specularFactor * specularTexture
```

UX3D-nopper: And it gets closer and closer to the suggestion I made.

rsahlin: > @rsahlin @MiiBond Would that work for you? If yes, I would remove `maxF0` and `f0Texture`. The IOR will become uniform (can be converted host-side to an f0 multiplier).

I guess so - however we are introducing quite a few heavy texture maps. 
An RGB texture vs one channel F0 texture.
My question would be - does the added texture 'weight' give us more flexibility or would it be enough with F0?

Again, I'm asking from a performance perspective to try and keep impact down.

proog128: > however we are introducing quite a few heavy texture maps.

`specularColorTexture` in #1719 can either be 3-channel (RGB) or 1-channel (luminance) (not yet, I have to change the description). In the end, if we combine #1718 and #1719 it is now only 2 textures (specularTexture + specularColorTexture) instead of 3 (f0Texture + specularTexture + specularColorTexture).

rsahlin: Great - that answers my question and will give us a lot of flexibility and freedom if I understand it correctly :-)

Another thing we should try to solve is the different packings of the textures - for instance packing 1-channel specularColor + specularTexture with for instance the MR texture.
Ideally we want as few samplers (texture sources) as possible

UX3D-nopper:
Please remove the `KHR_materials_volume` entry, otherwise we do have a cyclic dependency.
`KHR_materials_volume` could reference this extension.

proog128:
I refactored large parts of the extension. No change in behavior, but I hope it is now easier to understand and it better aligns with the notation in the core spec.

I also rephrased the interaction with `KHR_materials_volume` to make it clear that there is no dependency, just informal information ("README"). Do you like it better now? Or would you prefer to not mention other extensions at all if there is no explicit dependency? I liked the idea of providing additional information for implementors on how the extensions interact with each other.

extensions/2.0/Khronos/KHR_materials_ior/README.md
```diff
@@ -0,0 +1,95 @@
...
+```
+dielectricSpecular = ((ior - 1)/(ior + 1))^2
+```
+
+## Interaction with other extensions
+
+If `KHR_materials_ior` is used in combination with `KHR_materials_specular`, f0 computed from IOR is multiplied by f0 computed from the specular parameters.
+
+```
+dielectricSpecularF0 = ((ior - outside_ior) / (ior + outside_ior))^2 * specularFactor * specularTexture.a * specularColorFactor * specularColorTexture.rgb
```
rsahlin, line 82: Maybe move the calculation of dielectricspecular to the `KHR_materials_specular` extension to avoid duplication?
I think it's good to keep a reference, saying something along the lines of:
> If `KHR_materials_ior` is used in combination with `KHR_materials_specular`, f0 is computed according to the specification in that extension

extensions/2.0/Khronos/KHR_materials_ior/README.md
```diff
@@ -0,0 +1,95 @@
...
+dielectricSpecularF90 = specularFactor * specularTexture.a
+
+F0  = lerp(dielectricSpecularF0, baseColor.rgb, metallic)
+F90 = lerp(dielectricSpecularF90, 1, metallic)
+
+F = F0 + (F90 - F0) * (1 - VdotH)^5
+```
+
+If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `ior` affects not only the Fresnel term, but also determines the refractive index of the volume below the surface. Light rays passing through the transmissive surface are bent according to the index of refraction of the outside and inside medium (refraction). As the volume cannot be parametrized with a 2-dimensional texture in UV space, the index of refraction (IOR) is a scalar, uniform value (`ior`).
```
rsahlin, line 91: Maybe move the definition of how the two extensions interact to `KRH_materials_volume` to avoid duplication?
A reference to `KHR_materials_volume` where the exact usage is specified would be great :-)

UX3D-nopper:
If we want to ratify IOR quite soon, we need to kick out the references to the not yet ratified ones.
Also, KHR_materials_volume and KHR_materials_specualr can later reference KHR_materials_ior.
Please no cyclic dependencies.

extensions/2.0/Khronos/KHR_materials_ior/README.md
proog128: That's a good idea, I will update it. Thanks!

proog128:
@UX3D-nopper Thanks for clarifying, now I understand your concerns: mentioning other (non-ratified) extensions prevents fast ratification. I am still a bit hesitant to remove the references, because it makes it harder to understand the big picture... Or we track them in a separate document? I once created an overview in #1738, maybe this could be a starting point?

Meanwhile, I will remove the duplication as suggested by @rsahlin.

In general, I think we have to define what "dependency" means in terms of extensions.
* Does it mean that A _requries_ B?
* Or that A _is related to_ B? (ie. ior and specular are related, because they are multiplied)
* Or that A _is compatible to_ B? (ie. clearcoat is compatible to glTF base, but not to pbrSpecGloss)
* Or that A _mentions_ B? (ie. sheen extension describes how it interacts with clearcoat; order of layers)

What I am trying to do (also in #1738) is to create a unified material model that integrates all effects (sheen, clearcoat, ior, refraction, specular, specular color, ...). Each parameter has an appropriate default value, so that leaving out a parameter just removes one degree of freedom, but does not change the interaction between the effects. That means that glTF 2.0 base is just a _special case_ of the more general unified material model. The nice consequence of this is that if parameters are put into several extensions, all of them are compatible and there are no dependencies in the sense of "A requires B".

extensions/2.0/Khronos/KHR_materials_ior/README.md
```diff
@@ -0,0 +1,87 @@
+# KHR\_materials\_ior
+
+## Khronos 3D Formats Working Group
```
emackey, line 3: Please change this line to `## Contributors`.  If possible, add a contributor list similar to [the one in Clearcoat](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_clearcoat#contributors).

Please also copy the [Appendix from Clearcoat](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_clearcoat#appendix-full-khronos-copyright-statement).

extensions/2.0/Khronos/KHR_materials_ior/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+If `KHR_materials_specular` is used in combination with `KHR_materials_ior`, f0 computed from IOR is multiplied by f0 computed from the specular parameters. See [`KHR_materials_specular`](../KHR_materials_specular/README.md) for more information.
+
+If `KHR_materials_transmission` is used in combination with `KHR_materials_ior`, the `ior` affects the strength of the transmission effect according to the Fresnel term. See [`KHR_materials_transmission`](../KHR_materials_transmission/README.md) for more information. If `KHR_materials_specular` is used in addition, the rules for multiplying f0 from `ior` and f0 from the specular parameters apply as defined in `KHR_materials_specular`.
+
+If `KHR_materials_volume` is used in combination with `KHR_materials_ior`, the `ior` affects the refraction effect. See [`KHR_materials_volume`](../KHR_materials_volume/README.md) for more information.
+
+## Schema
+
+- [glTF.KHR_materials_specular.schema.json](schema/glTF.KHR_materials_specular.schema.json)
```
abwood, line 91: Once a schema is available, this will need to get updated.

extensions/2.0/Khronos/KHR_materials_ior/README.md
```diff
@@ -0,0 +1,144 @@
...
+        }
+    ]
+}
+```
+
+| |Type|Description|Required|
+|-|----|-----------|--------|
+| **ior** | `number` | The index of refraction. | No, default: `1.5`|
+
+Typical values for the index of refraction range from 1 to 2. In rare cases values greater than 2 are possible. The following table shows some materials and their index of refraction:
```
abwood, line 49: For the specification, I think we should be more explicit on the allowed range for ior.  If a value of -1 is specified in this extension, most implementations will result in nan values all throughout the material.  If this is a defined constraint, we could have the validator at least flag this value as illegal.

Minimum value: 1.0 ?
Maximum value: 3.0 ?

extensions/2.0/Khronos/KHR_materials_ior/README.md
proog128: The schema is available, but the link was broken. Fixed.

extensions/2.0/Khronos/KHR_materials_ior/README.md
proog128: The schema defines a minimum of 1.0 (although the math would work for any value >0). There is no maximum, because it will make conversions from other material models (e.g. KHR_materials_pbrSpecularGlossiness) more complicated.

emackey: Do we want to allow materials less dense than air?  Also, for an underwater scene, the air bubbles in the scene would be less than 1.0 IOR due to light exiting the water, right?  Could the set the limit as >0 instead of >= 1.0?

emackey: I think I'll withdraw this comment.  After a bit more research, it looks like a pure vaccuum is 1.0, and air is actually 1.0002.  So presumably values less than 1.0 are not physically possible.

extensions/2.0/Khronos/KHR_materials_ior/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+
+## Implementation
+
+*This section is non-normative*
+
+The extension changes the computation of the Fresnel term defined in [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation) to the following:
+
+```
+dielectricSpecular = ((ior - 1)/(ior + 1))^2
```
rsahlin, line 78: Perhaps worth to mention that implementations are free to keep track of IOR for objects/materials that reside inside some other medium than air?
Thus replacing the 1 with the ior for that medium.
One such example could be a (submerged) submarine with a glass window.

emackey: Specifically, when the camera itself resides in a non-air medium, such as water.  I think this is the same concern that my comment was struggling to reach.

proog128: This is mentioned in the "Extending Materials" section that describes how the parameters behave:

> The `outside_ior` determines the index of refraction of the medium on the outside. If the renderer does not track the IOR when traversing nested dielectrics, it can assume `outside_ior = 1`.

It is also mentioned in KHR_materials_volume (#1726), because it is relevant only for volumetric objects:

> Incident and transmitted index of refraction have to be correctly set by the renderer, depending on whether light enters or leaves the object. An algorithm for tracking the IOR through overlapping objects is described in [Schmidt C., Budge B. (2002): Simple Nested Dielectrics in Ray Traced Images](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.125.5212&rep=rep1&type=pdf).

I didn't mention it in the "Implementation" section because so far I tried to keep this section as simple as possible, focused on real-time implementations and with little interactions between extensions as possible (i.e., as the default glTF 2.0 metallic-roughness material is thin-walled, we cannot reside in another medium than air; except if the camera is placed in something else, but that seems to be a rather rare use-case).

UX3D-haertl:
Shouldn't it be written down in the README.md that this extension can not be used together with KHR_materials_pbrSpecularGlossiness?

One could add it in a similar way how it is done in the [KHR_materials_sheen PR](https://github.com/KhronosGroup/glTF/pull/1688):

> ## Exclusions
> * This extension must not be used on a material that also uses `KHR_materials_pbrSpecularGlossiness`.

UX3D-haertl:
How should ior be used if a metallic factor between 0 and 1 is used?

proog128:
@UX3D-haertl The ior has no effect on the metal BRDF:

```
dielectricSpecular = ((ior - 1)/(ior + 1))^2
cdiff = lerp(baseColor.rgb * (1 - dielectricSpecular.r), black, metallic) 
F0 = lerp(dieletricSpecular, baseColor.rgb, metallic) 
```

emackey:
In some software, including Blender, the IOR has an effect on how "blurry" the transmission is, when roughness is nonzero.  This can happen even for thin-walled transmission.

Borrowing a line of thought from Emmet, one can imagine that on a thin-walled piece of geometry, the microfacets on the front and back are not correlated, and so on a rough surface the front and back are not parallel, forming a series of little prisms that can scatter the light.  The base roughness essentially controls the width of these prisms (or how "non-parallel" the front vs. back has become).  But if the IOR is exactly 1.0, the prisms do nothing to refract the light, so no blurring of the transmission.  Conversely, for high values of IOR, even small amounts of roughness cause a lot of blurring, due to steep refraction angles.

Can/should we define this effect in the IOR extension here?

proog128:
Good point. Yes, the microfacet distributions on front and back surface in the thin-walled model are assumed to be uncorrelated. We can create a simple scene to verify that the blurring effect is correct: render a thin box (closed mesh, not thin-walled) with a path tracer; the objects behind the box/the background become blurry. What we can also observe in this simple scene is that the IOR affects the blurriness: when it approaches zero, the blurriness gets smaller and smaller, and for an IOR of 1 the box appears to be clear.

There is a section in [the slides from Imageworks](https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides_v2.pdf) (slides 34-41) that deals with this issue. It derives an IOR dependent factor that needs to be multiplied to the transmission roughness of the thin transmission BSDF. It should be easy to adopt this for PBR Next.

If we decide to include this correction factor, I would prefer to have this in KHR_materials_transmission rather than KHR_materials_ior. 
As far as I understand it only affects transmission only, not reflection, so I think it belongs into the transmission extension.

emackey:
> I think it belongs into the transmission extension

Yes, but that extension is under review already, so modifications there could reset the clock.  Would it be OK to put it in IOR instead?  Later these will all fold up into a unified future version of glTF anyway.

proog128:
I am afraid that if we take the mapping from Imageworks, we have to put it into KHR_materials_transmission. The reason is that the default IOR of 1.5 causes a remapping of the roughness. So if KHR_materials_ior is not present, we still would have to remap the thin roughness.

This is the mapping:
image

I thought a bit more about it, and what consequences it would have if we add it. We have to decide whether we apply it to the full *thin dielectric BSDF* (i.e., reflection and transmission lobes of the transmissive portion of the material) or to the *dielectric BTDF* only. I cannot find a clear statement about that in the slides. You could probably come up with good reasons for either. If it's applied to the full BSDF (BRDF and BTDF), we need to separate the thin glass BSDF from the remaining components. If it's applied to the BTDF only, we suddenly have different roughnesses for reflection and transmission lobes, which may require some special handling (albedo scaling) for energy conservation.

I was also trying to find out whether there is (dedicated) support for this in other renderers/material models. Most don't seem to have a special treatment for thin materials, like for example Blender Cycles. So the IOR-dependent blurriness of glass in Cycles comes only due to the fact that the material is assumed to be volumetric, and you cannot change that (or I missed the setting?). Standard Surface probably behaves similar, I also couldn't find a mentioning of a special thin BTDF in the documentation. Enterprise PBR has a special thin case (Boolean switch), but there is no roughness mapping yet (we are considering it for a future version). Regarding node-based materials: MDL and MaterialX ShaderGen both have a thin flag. But as far as I can tell both do not have a roughness remapping in any of their BSDFs.

The reason for not having a roughness remapping might be simple: it's easy to bake (in almost all cases). And if thin roughness is set by an artist based on intuition, to make it look realistic, the mapping is not needed. It's only needed if it's required to switch between thin and non-thin cases and if then expect a match between the two. (And, of course, to make the model more physically-plausible).

All that makes me a bit hesitant to rush this into the extension now...

extensions/2.0/Khronos/KHR_materials_ior/schema/glTF.KHR_materials_ior.schema.json
```diff
@@ -0,0 +1,18 @@
...
+    "title": "KHR_materials_ior glTF extension",
+    "type": "object",
+    "description": "glTF extension that defines the index of refraction of a material.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "ior": {
+            "type": "number",
+            "description": "The index of refraction.",
+            "default": 1.5,
+            "minimum": 1.0,
```
emackey, line 12: We talked offline today about allowing zero as a special-case value.  This snippet of schema would implement that allowance.

```suggestion
            "oneOf": [
                {
                    "minimum": 0.0,
                    "maximum": 0.0
                },
                {
                    "minimum": 1.0
                }
            ],
```

extensions/2.0/Khronos/KHR_materials_ior/schema/glTF.KHR_materials_ior.schema.json
emackey: I'm marking this sub-thread resolved, because there's no `iorTexture` anymore.  Specular concerns should be discussed on the other PR.

extensions/2.0/Khronos/KHR_materials_ior/schema/glTF.KHR_materials_ior.schema.json
proog128: @emackey Thank you! I updated the ior and specular extensions.

emackey:
The special IOR value of zero is explained in the Specular specification, but not mentioned here in the IOR spec.  Can it be mentioned here too, at least briefly?

proog128:
It's mentioned in the "Extending Materials" section:

> Valid values for `ior` are numbers greater than or equal to 1. In addition, a value of 0 is allowed. This value gives full weight to `layer`, i.e., the Fresnel term evaluates to 1 independent of the view or light direction. It is useful in combination with `KHR_materials_specular` to seamlessly support the specular-glossiness workflow.

emackey:
Merging this.  Status will remain "Draft" for at least a few more weeks, but it will be hosted in the main glTF repository.

PR #1721, opened 2019-12-04
Draco Bitstream Specification

UX3D-kanzler:


lexaknyazev:
Pushed a few language updates for the first sections of the bitstream spec.

Now, they contain all the information needed to decode header, metadata, and sequential connectivity encoding without even looking into "more formal" later parts that are much less readable, imo. Maybe remove them completely?

In the source-code driven parts, there are still errors of various severity from the original document reported earlier in the old Draco bitstream thread. For convenience, here they are:

---

- [ ] Since local vars lack types, it's not obvious in some cases when division operator should be integer or float. Sure, this could be deduced from context, but I'd say that the spec should be explicit about that somehow.

- [ ] `PosOpposite` and `SetPosOpposite` refer to undefined `opposite_corners_` array. What length should it have? Draco source suggests that `opposite_corners_` consists of int32 elements and its length is `num_faces * 3`.

- [ ] `MapCornerToVertex` assigns to uninitialized `corner_to_vertex_map_[0]`.

- [ ] `att_dec_data_id` from `ParseAttributeDecodersData` isn't used in the spec. Moreover it's declared as signed integer in Draco sources and seems to be used for some purpose.

- [ ] Values of `att_connectivity_seam_opp` are never used.

- [ ] `AttrOpposite` declares two arguments but uses only one:
image

- [ ] `face_id` and `local` assigned in `DecodeAttributeSeams` are never used.

- [ ] `att_dec_decoder_type` array is zero-based:

  image
however its usage may imply otherwise. E.g. in `IsOnBoundary`:

  image

  Here, the value of `att_dec_decoder_type[0]` will never be read. Moreover, `IsOnBoundary` will never be called with `att_dec == 0` because it's called only from `EdgeBreakerAttributeTraverser_ProcessCorner`:

  image

  which can be called only when `curr_att_dec` is not `0`:

  image

  Similar concerns could be raised about `Opposite`:

  image

donmccurdy:
Checking in on this — what else is needed? Are we just waiting for more feedback?

lexaknyazev:
@donmccurdy 
I'm doing another pass on the spec, should finish by the end of the week.

PR #1722, opened 2019-12-05
Added glTFast to the list of tools/libs for glTF.

atteneder:
Hi!

as author of glTFast I'd be very glad to have it added to this list.

Thanks for consideration!


emackey:
Hi, we've been adding things to lists in chronological order this year.  Can you move the new entry to the bottom?  Thanks!

atteneder:
Hi,
I thought it makes sense to group the Unity related libs, but okay. I put it to the end.

emackey:
Someday we'll replace this with a better system.  Thanks!

PR #1726, opened 2019-12-12
KHR_materials_volume

proog128:
An extension for the metallic-roughness material that adds volumetric parameters for effects like refraction and subsurface scattering. It requires `KHR_materials_transmission` (#1698) and should be combined with `KHR_materials_ior` (#1718). At the moment, it lacks a suitable model and parameters for rasterizer implementations.


MiiBond:
Very nice start. 

For realtime rasterizers, I feel that it is necessary for this extension to support a thickness texture. It’s very difficult, in practice, to make rasterizers efficiently render volumetric properties of a closed mesh. It’s not impossible, of course, and some viewers (like ours at Adobe) may opt to go this route if that is a critical feature but, in general, performance-minded game engines aren’t going to support this. 

Thickness textures allow you to very efficiently define thin volumes like panels of glass. For example, the window panes of an old building where the glass at the bottom is thicker and rippled. Modeling this (and rendering this) as a volume is far more difficult and costly. 

So, I suggest a dual-mode extension where either thickness is defined or the true volume of the mesh (must be closed) is used. The thickness texture would be accompanied by min/max scaling factors to scale the 0-1 texture values. A constant thickness could be defined by just supplying the min factor and no texture. A boolean flag (or maybe just the presence of the min/max values) could indicate that the thickness mode should be used. 

Here are some more thoughts related to this: 

1. The thickness texture could probably be read from a channel of the transmissionTexture. This would save a sampler. Is it weird to have a single texture include info for properties in two different extensions? 
1. What is the expected behaviour of the volume when scaling the object? Presumably, scene units are used so the appearance changes as the mesh gets larger. This would have to apply to the min/max thickness values too, right? Real-world scale would be critical for correct rendering. Should there be an option not to scale volume properties? 
1. What requirements do we define for rasterizers to be compliant with this extension? Are they required to render more than one layer of transparency with volumetric features? This is definitely non-trivial and I wouldn’t expect rasterizers, in general, to be able to do this. SketchFab, for example, doesn’t render refractive objects through other refractive objects. https://help.sketchfab.com/hc/en-us/articles/202602073-Transparency-Opacity#refraction. 
1. If the true volume mode is intended to be used for the asset, I suggest that rasterizers be allowed to use a constant thickness rather than the actual volume. Otherwise, I fear support for this extension will be very low. 
1. We should specify how renderers should treat overlapping volumes. How do the properties combine? Or, since rasterizers aren’t going to be able to handle this, in general, maybe we should disallow overlapping volumes or at least discourage it in the spec?

proog128:
> For realtime rasterizers, I feel that it is necessary for this extension to support a thickness texture. [...] Thickness textures allow you to very efficiently define thin volumes like panels of glass. [...]

For this reason a thickness texture would also make sense for ray tracers. We have to be a bit careful how the thickness is interpreted so that it works for path tracing, light tracing, and rasterizers. The following definition should work:

The material can be either thin-walled or volumetric, the user has to choose via a Boolean parameter. **Thin-walled** means that the mesh geometry defines two surfaces: the actual surface and a "virtual" surface underneath that is not modeled, but implicitly given by the thickness texture. This applies to both closed and non-closed meshes. Non-closed meshes could be rippled windows, leaves, ... Closed meshes could be light bulbs, ...

thin-walled

**Volumetric** means that the mesh geometry defines a surface boundary. The mesh has to be closed. It is possible to have nested closed objects, which could make them hollow like thin-walled closed meshes, but I guess that's a rare case.

volume

In the volumetric case, a thickness texture can (or must) be provided to help rasterizers render the volumetric properties. Ray tracers will ignore the texture. That means that the texture must not be used for artistic effects, it is basically an automatically generated texture, the "baked" volume thickness at each surface point.

Please tell me what you think about it, I will update the document accordingly.

> What is the expected behaviour of the volume when scaling the object? [...]

That's a good question. sigmaT should probably be given in world space, because I am not sure if applying non-uniform scaling to the distance works well...

> We should specify how renderers should treat overlapping volumes. How do the properties combine? [...]

I'd suggest attaching priority values to shapes/nodes in the scene graph that determine which of several overlapping objects "wins". The technique is described [in this paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.125.5212&rep=rep1&type=pdf), and I think a lot of ray tracers implement it like this. Not sure if this is really a problem we can (or have to) solve for rasterizers.

I am afraid I don't have answers to the other questions, maybe others with more experience in rasterization and glTF have thoughts on it.

MiiBond:
Okay, so the two modes would toggled with a boolean and would work like this?
`isThinWalled == false`:
- mesh *must* be closed.
- thickness texture *must* be defined and is meant to represent the approximate thickness at each point on the surface
- renderers can optionally ignore thickness texture if they can calculate the true volume of the mesh
`isThinWalled == true`:
- mesh can be open or closed.
- thickness is used by all renderers and calculated from the thickness texture and min/max values.

Is that right?

MiiBond:
Do you think we could add some more information about how the single-scatter albedo and attenuation coeff are used to calculate the colour of transmitted light?

proog128:
> [...] Is that right?

Yes. It is just an idea at the moment, I don't know it if works well. The volumetric part with participating media inside the volume boundary should be easy. The tricky part is the thin configuration: the combination of thin, rough transmission with the dipole approximation.

> Do you think we could add some more information about how the single-scatter albedo and attenuation coeff are used to calculate the colour of transmitted light?

Of course. I think this will mostly be important for thin materials. The implementation for volumes/homogeneous participating media should use brute-force SSS as reference solution. Maybe we can shortly describe the algorithm (sample distance according to attenuation coeff, weight with albedo, sample phase function and spawn scatter ray), but there is probably enough literature out there that we don't have to go into details for ray tracing. If you already have ideas, it would be great if you could add it to the PR.

proog128:
The more I think about the idea, the less I like it. Maybe we should concentrate first on the volumetric part and keep thin-walled for later. I looked at a few ray tracers (V-Ray, Corona, Arnold, Maxwell, Iray, ...) and they all seem to require properly modeled volumes for random walk SSS. Although we might be able to come up with a good solution for thin-walled with thickness textures, it won't help if the renderers out there do not support it...

MiiBond:
I understand the concern. The problem is, glTF is mainly used (and intended) for real-time rendering. An extension that only supports closed volumes might not get much support.
We could leave open the option for the runtime to fallback to a constant thickness rather than adhere to the actual volume. But, in that case, maybe just going ahead and including the thickness texture is the way to proceed?

A related question about SSS; I know that this extension can be used for SSS but, in realtime, there are usually SSS approximation techniques that don't involve complex transparency or refraction. Do we leave it up to the runtime to fallback to these approximations when the scattering parameters hit a certain threshold? Or do we define a separate SSS extension for the explicit purpose of real-time SSS?

proog128:
> [...] But, in that case, maybe just going ahead and including the thickness texture is the way to proceed?

Yes, I agree. To sum up:
* We include the thickness texture and make closed meshes mandatory.
* We add a thickness factor (or minThickness, maxThickness) for the texture, which can be used as a constant thickness if texture is not available.
* We describe that the thickness texture has to match the volume of the mesh, i.e., it has to contain the baked thickness of the volume.

Renderers can now freely choose between the two representations. Ray tracing uses the mesh. Real-time could use the mesh (some depth map and shadow map based volume approximations) or the thickness map. This will give us consistent renderings in real-time and ray tracing.

Did I miss anything? If not, I will go on and add the parameters to the document.

> A related question about SSS [...]

I am also unsure about that. A separate extension which limits the range of supported effects to SSS would certainly help real-time renderers, but on the other hand it adds redundancy.

MiiBond:
I'm not sure about the "make closed meshes mandatory" part. For commerce-related assets it might be fine (assuming those assets are CAD or not created for realtime initially anyway) but, for game assets, I'm not sure that it's reasonable to expect every pane of glass to have two sides and an appropriate thickness. If we don't want an explicit flag for this, perhaps having a closed mesh is a suggestion and not a requirement in the spec?

proog128:
Maybe this use-case can be covered by KHR_materials_transmission extension alone? If KHR_materials_volume is not enabled, the surface is automatically thin-walled and, therefore, doesn't require closed meshes. The volumetric effects in a thin glass pane can be squeezed into effects on the surface. Absorption becomes the surface color (transmission only). Roughness can be used to build ground glass. Milk glass can be achieved with a combination of these effects and, if needed, a clearcoat layer. A thickness texture is not needed in these cases: thickness will influence surface color and roughness, so it can be baked into the base color texture and the roughness texture (saving texture lookups and computations in the shader).

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Dependencies
+
+Written against the glTF 2.0 spec. Needs to be combined with `KHR_materials_transmission`.
+
+## Overview
+
+This extension adds material parameters to configure the volume beneath the surface of closed meshes, providing effects like refraction, absorption and subsurface scattering. The transparency of the surface is controlled by `KHR_materials_transmission`.
+
+Light hitting an object will first interact with the surface according the BSDF. The BSDF determines how much and in which direction the incident light is reflected and transmitted. Light that is transmitted enters the volume, where it is absorbed or scattered by particles inside the medium. After bouncing several times inside the object, the light will eventually hit the surface again. The BSDF is evaluated a second time, the light may now leave the object or be kept inside it if total internal reflection occurs.
```
koiava: > light may now leave the object or be kept inside it if total internal reflection occurs <
I think here it should say 'if internal reflection occurs' because there might be reflection event which is not necessarily TIR. TIR is called an event when angle between incident ray direction and surface normal is larger then a critical angle.

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,192 @@
...
+albedo = sigmaS / (sigmaS + sigmaA)
+```
+
+Sometimes the mean free path length **mfp** is given:
+
+```
+sigmaT = 1 / mfp
+```
+
+The attenuation color is a 3-channel value that goes up to infinity. By splitting it into two values, an RGB attenuation color and a distance, the color is easier to set up. In the following equation, **attenuationColor** defines the remaining, wavelength-dependent energy of a ray of white light after travelling a certain distance (**attenuationDistance**) through the medium.
```
koiava: As I understood AttenuationColor' is actually transmittance of the ray after traveling attenuationDistance through the medium. I think Transmittance is a better name because from 'AttenuationColor' it's bit ambiguous if it is attenuated or remaining part of the energy.

UX3D-nopper:
Need to merg with https://github.com/KhronosGroup/glTF/pull/1766

MiiBond:
About the thickness texture, should we define that the thickness data is stored in the G channel so that users can save a sampler and pack transmission/translucency data and thickness data together?
Or, better yet, maybe we should force this by requiring that the thickness data is read from the `G` channel of the `transmissionTexture` (or `translucenceTexture`)?

MiiBond:
How do we feel about the parameterization of `attenuationCoefficient` and `volumeAlbedo`?
Dassault's Enterprise PBR uses `attenuationColor`, `attenuationDistance` and `scatterColor`, which I find far more intuitive.
Now, obviously, the defined parameters convert easily enough and glTF is more about loading and transmission efficiency than it is about being intuitive. However, I think that there's something to be said for being easily understandable and for mapping closely with existing material models.

proog128:
Using the `G` channel for thickness is a good idea. About translucency, we could fetch translucency from the `B` channel. It would mean that the full material needs only a single texture for transparency (`R`), thickness (`G`) and translucency (`B`).

donmccurdy:
> ... should we define that the thickness data is stored in the G channel ...

+1.

> Or, better yet, maybe we should force this by requiring that the thickness data is read from the G channel of the transmissionTexture (or translucenceTexture)?

I wouldn't go this far — let's choose a single channel that gives reasonable packing options ("G" works for me), but still let the author decide how to use that. We may want the option of _not_ combining the textures when using Basis.

***

I think I'll have some comments on parameter names. Is this a good time to make those suggestions, or are the parameters themselves still in flux?

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,172 @@
...
+| **attenuationCoefficient** | `number[3]` | Attenuation coefficient (also called extinction coefficient) in inverse meters. | No, default: `[0.0, 0.0, 0.0]` |
+| **volumeAlbedo** | `number[3]` | Single-scattering albedo inside the volume. | No, default: `[0.0, 0.0, 0.0]` |
+| **subsurfaceAnisotropy** | `number` | Anisotropy of the phase function in range [-1, 1] | No, default: `0` |
+| **thicknessTexture** | `textureInfo` | A texture that defines the thickness of the volume beneath the surface. The values are sampled from the G channel. May be used by renderers that are not able to derive the thickness from the actual geometry, otherwise it should be ignored. | No |
+| **minThickness** | `float` | When using a thickness texture, a texture value of 0.0 is interpreted to be `minThickness`. When a constant thickness is used, the thickness will be `minThickness`. | No, default: `0` |
+| **maxThickness** | `float` | When using a thickness texture, a texture value of 1.0 is interpreted to be `maxThickness`. Values between 0.0 and 1.0 are linearly scaled between minThickness and `maxThickness`. | No, default: `1` |
```
donmccurdy, line 61: Elsewhere, glTF materials allow a texture or a factor for inputs. I realize we may not want textures for `attenuationCoefficient`, `volumeAlbedo`, and `subsurfaceAnisotropy` for complexity reasons. But is there a reason to exclude a factor for `thickness`? I think you could parameterize it in the usual way, and remove the need for `thinWalled` boolean at the same time:

- `thicknessFactor`: Thickness of volume beneath the surface, 0-1, used to lerp between `minThickness` and `maxThickness`. Default is 0.
- `thicknessTexture`: Thickness of volume beneath the surface, 0-1, used to lerp between `minThickness` and `maxThickness`. Linear multiplier against `thicknessFactor`.
- `minThickness`: No change.
- `maxThickness`: No change.

Note that because the default is 0, the transmission is "off" by default (which we want) without the need for a separate property. If we did keep the separate property for that purpose, I think I somewhat prefer the terminology that Filament uses: `refractionType: 'thin' | 'volume'`.

Also — should `minThickness` and `maxThickness` specify units, e.g. meters?

MiiBond: I like this.
I think you mean the **volume** is off my default. Transmission shouldn't be affected.
Also, +1 for metres as the default unit for thickness. That matches the default unit of the rest of the glTF file.

proog128: That would mean the thickness is computed as follows, right?

```
thickness = lerp(minThickness, maxThickness, thicknessFactor * thicknessTexture)
```

I like it, it's much more similar to the usual texture parameters used everywhere else.

However I didn't get why thicknessFactor=0 would disable the volume. Wouldn't it mean that a volume is still present, just with constant thickness = minThickness?

Maybe `surfaceType: 'thin' | 'volumeBoundary'` instead of `refractionType`? Probably also not optimal, but it would stress the fact that not only refraction is affected by the type, but also attenuation and subsurface scattering. Nitpicking, I know, but having to choose those names was the reason for me switching to a Boolean instead of an enum :-)

I will add information about the unit, it should definitely be meters for min and max thickness.

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
```diff
@@ -0,0 +1,66 @@
...
+            "minimum": 0.0,
+            "maximum": 1.0,
+            "default": 0.0,
+            "description": "Minimum thickness of the volume, corresponds to a texture value of 0.0.",
+            "gltf_detailedDescription": "When using a thickness texture, a texture value of 0.0 is interpreted to be minThickness. When a constant thickness is used, the thickness will be minThickness."
+        },
+        "maxThickness": {
+            "type": "number",
+            "minimum": 0.0,
+            "maximum": 1.0,
```
emackey, line 60: This line should be removed.

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
```diff
@@ -0,0 +1,66 @@
...
+        },
+        "thicknessTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "A texture that defines the thickness of the volume beneath the surface in the G channel.",
+            "gltf_detailedDescription": "A texture that defines the thickness of the volume beneath the surface. The values are sampled from the G channel. May be used by renderers that are not able to derive the thickness from the actual geometry, otherwise it should be ignored."
+        },
+        "minThickness": {
+            "type": "number",
+            "minimum": 0.0,
+            "maximum": 1.0,
```
emackey, line 52: This should be removed.

extensions/2.0/Khronos/KHR_materials_volume/README.md
donmccurdy: > I think you mean the volume is off my default.

Oops, yes.

> Wouldn't it mean that a volume is still present, just with constant thickness = minThickness?

Discussed offline – the default for both `thicknessFactor` and `minThickness` will be zero, so the volume is disabled by default. But you're correct that increasing either of these values will cause volume to be enabled.

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
emackey: Also, somewhere note that max must be > min

emackey:
I suggest we pull subsurfaceAnisotropy from this extension, and add to another extension in the future if needed.

extensions/2.0/Khronos/KHR_materials_volume/README.md
donmccurdy: Another suggestion from @UX3D-nopper that sounds useful — we could omit `minThickness` and `maxThickness`, having only two parameters here:

- `thicknessFactor`
- `thicknessTexture`

The product of the two would be defined in units of meters. When `thicknessFactor=0` (default), volume is disabled. Note that this simpler parameterization loses precision _if_ the minimum thickness of the volume is >> 0. Is it likely for minimum thickness to be non-zero in real use cases? If we need that case, the addition of a `thicknessOffset` parameter would restore the precision while keeping a simpler parameterization. In that case the thickness would be computed as:

```
thickness = thicknessFactor * thicknessTexture + thicknessOffset
```

donmccurdy: There are several tools that already support loading or baking thickness maps:

- [Substance Painter](https://docs.substance3d.com/spdoc/thickness-142213479.html) (both)
- [Marmoset Toolbag](https://marmoset.co/posts/toolbag-baking-tutorial/) (bake)
- [Sketchfab](https://help.sketchfab.com/hc/en-us/articles/204429595-Materials-PBR-#sss) (load)
- [Blender](https://docs.blender.org/manual/en/latest/render/shader_nodes/input/ao.html) (bake, via AO "inside")

I could only find the units for Substance Painter, which has two settings:

1. units used in "Max Frontal Distance" et "Max Rear Distance" are the ones defined when exporting your mesh (m, cm or whatever you did set up in your scene export)
2. `units are relative to the bounding box of the object, 1.0 being the diagonal length of the bounding box`.

If there is some consistency among these tools, that might take precedence over some of the suggestions above, for me.

proog128: That's very good information, thanks for sharing! I had a look at [refraction in Babylon.js](https://doc.babylonjs.com/how_to/physically_based_rendering_master) and [Filament](https://google.github.io/filament/Materials.html#materialmodels/litmodel/absorption) to see which of the two modes they consume. Documentation is a bit sparse, but from what I understand they both take the thickness in meters. Babylon.js uses factor and offset as suggested in your previous comment: `minimumThickness + thicknessTexture.r * maximumThickness`

emackey:
Tobias, I'm still reading and digesting, but my first impression is, today's updates look AMAZING.  Just fantastic!

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+ρ<sub>ss</sub> = 1 - (4.09712 + 4.20863 ρ<sub>ms</sub> - sqrt(9.59217 + 41.6808 ρ<sub>ms</sub> + 17.7126 ρ<sub>ms</sub><sup>2</sup>))<sup>2</sup>
+
+The single-scattering albedo is the color at each particle in the medium. Light that is scattered by a particle will be tinted with this color. Note that scattering will happen many times in the medium until the light leaves the volume, so the overall color ρ<sub>ms</sub> will differ significantly from the albedo of the particles ρ<sub>ss</sub>. Via the mapping users can control the albedo of surface, i.e., the color that an object appears to have in the final rendering. The mapping was introduced in [1].
+
+Now that we have computed σ<sub>t</sub> and ρ<sub>ss</sub>, we can finally derive σ<sub>a</sub> and σ<sub>s</sub> as follows:
+
+σ<sub>a</sub> = σ<sub>t</sub> (1 - ρ<sub>ss</sub>)
+
+σ<sub>s</sub> = σ<sub>t</sub> ρ<sub>ss</sub> = σ<sub>a</sub> - σ<sub>s</sub>
```
MiiBond, line 128: Is this right? Shouldn't σ<sub>s</sub> = σ<sub>t</sub> - σ<sub>a</sub>?

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,255 @@
...
+                  abs(LdotN) * abs(VdotN)      (ior_i * VdotH + ior_o * LdotH)^2
+
+```
+
+`ior_i` and `ior_o` denote the index of refraction of the incident and transmitted side of the surface, respectively. `V` is the vector pointing to the camera, `L` points to the light. In a path tracer that starts rays at the camera, `V` corresponds to the incident side of the surface, which is the side of the medium  with `ior_i`. See [Walter et al. (2007)](#Walter2007) for more details.
+
+Using Snell's law, the half vector is computed as follows:
+
+```
+H = -normalize(ior_i * VdotH + ior_o * LdotN)
```
abwood, line 189: I'm missing something here.  H is a vec3 but ior and dotProducts are all scalars.  

From EnterprisePBR, it appears that half vector should be calculated as:
`H = normalize(-V * ior_i - L * ior_o)`

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,255 @@
...
+*This section is non-normative.*
+
+The microfacet BTDF f<sub>transmission</sub> defined in `KHR_materials_transmission` now takes refraction into account.
+
+```
+                  abs(LdotH) * abs(VdotH)                 ior_o^2 * G * D
+f_transmission = ------------------------- * ----------------------------------------
+                  abs(LdotN) * abs(VdotN)      (ior_i * VdotH + ior_o * LdotH)^2
+
+```
```
abwood, line 182: It's not too clear while reading this that `f_transmission` must use the modified H value defined below.  Perhaps we could define the modified half vector first? 

I would also recommend that `D` be renamed to `Dt` to be consistent with KHR_material_transmission notation.  `Dt` is more clear in that D that uses the modified half vector.

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+Incident and transmitted index of refraction have to be correctly set by the renderer, depending on whether light enters or leaves the object. An algorithm for tracking the IOR through overlapping objects is described by in [Schmidt and Budge (2002)](#SchmidtBudge2002).
+
+The microfacet BTDF also makes use of the modified half vector and the indices of refraction.
+
+f<sub>transmission</sub> = *T* * *baseColor* * (1 - *F<sub>R</sub>*) * (abs(dot(*L*, *H*)) * abs(dot(*V*, *H*))) / (abs(dot(*L*, *N*)) * abs(dot(*V*, *N*))) * (*ior<sub>o</sub>*<sup>2</sup> * *G<sub>R</sub>* * *D<sub>R</sub>*) / (*ior<sub>i</sub>* * dot(*V*, *H*) + *ior<sub>o</sub>* * dot(*L*, *H*))
+
+where *T* is the transmission percentage defined by `KHR_materials_transmission`. The F<sub>R</sub>, D<sub>R</sub>, and G<sub>R</sub> terms are the same as in the specular reflection except using the modified half vector *H* calculated from the refraction direction. See [Walter et al. (2007)](#Walter2007) for details.
```
abwood, line 187: The `G` term does not depend on the half vector, so can we just use `G` rather than G<sub>R</sub>? Or should G<sub>R</sub> use a modified L?

proog128: The description is a bit fuzzy and I am unsure how many details we should have here... G depends on the half vector to check if V and L are on the same side of the microsurface regarding H. See [eq. 34 in Walter 2007](https://www.cs.cornell.edu/~srm/publications/EGSR07-btdf.pdf). Maybe it's good enough to link the paper? Or would you prefer a more detailed explanation?

abwood: So this is the positive characteristic function, where G returns 0 if V and H are on opposite sides of the surface? `clamp(sign(VdotH / NdotV), 0.0, 1.0);`
Appendix B does not mention this, so I think it would be helpful to include here if it is now important for transmission.

proog128: Yes. If I am not mistaken, it is important for reflection and transmission, and it is part of the G term in either case. I am wondering why it's not part of Appendix B? Could be that Appendix B is just not precise in this regard, or it is assumed to be handled "outside" of G. [libbsdf for example](https://github.com/NVIDIA/MDL-SDK/blob/master/src/mdl/jit/libbsdf/libbsdf.cpp#L678) has an early out if VdotH or LdotH are on the "wrong" sides of the surface (were "wrong" depends on reflection/transmission and H).

In the end, it is just important to use the modified H when dealing with refraction, and the "normal" H when dealing with reflection.

abwood: Perhaps it is worth a mention in #1717?  It doesn't appear that we are applying this term in the [sampleviewer](https://github.com/KhronosGroup/glTF-Sample-Viewer/blob/88b18169541ad039efcf370ca168ea69b6b048e3/src/shaders/brdf.glsl#L35) either.

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
```diff
@@ -0,0 +1,49 @@
...
+            "description": "Average distance in meters that light travels in the medium before interacting with a particle. The value is given in world space. When undefined, the value is assumed to be infinite."
+        },
+        "attenuationColor": {
+            "type": "array",
+            "items": {
+                "type": "number",
+                "minimum": 0.0
+            },
+            "description": "Color that white light turns into due to absorption when reaching the attenuation distance.",
+            "default": [ 1.0, 1.0, 1.0, 1.0 ],
```
emackey, line 32: Array length of the default value should be 3, here and also in `subsurfaceColor` below.

extensions/2.0/Khronos/KHR_materials_volume/README.md
proog128: The step function is now part of #1717.

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
proog128: Fixed.

donmccurdy:
I'm having a difficult time understanding the difference between `attenuationColor` and `subsurfaceColor`. Looking at DCC tools for conventions:

- [Blender Principled BSDF](https://docs.blender.org/manual/en/latest/render/shader_nodes/shader/principled.html)
- [Houdini Principled Shader](https://www.sidefx.com/docs/houdini/nodes/vop/principledshader.html)
- [Substance Painter Subsurface Parameters](https://docs.substance3d.com/spdoc/subsurface-parameters-172818973.html)

None of these seem to offer anything quite like `attenuationColor`, which may be a challenging for authoring workflows. Perhaps the closest is Blender's Subsurface Radius:

> **Subsurface Radius**: Average distance that light scatters below the surface. Higher radius gives a softer appearance, as light bleeds into shadows and through the object. The scattering distance is specified separately for the RGB channels, to render materials such as skin where red light scatters deeper. The X, Y and Z values are mapped to the R, G and B values, respectively.

This matches the [subsurface radius of Standard Surface](https://autodesk.github.io/standard-surface/#closures/subsurfacescattering), as well. Does `attenuationColor` have a well-defined relationship to these parameters?

donmccurdy:
Based on [Skin Shader & Subsurface Scattering – Getting to Know Toolbag 3. Ep. 6](https://marmoset.co/posts/getting-to-know-toolbag-3-ep-6-subsurface-scattering/), it sounds like Marmoset Toolbag uses a per-channel "scatter depth" as well.

proog128:
@donmccurdy Scattering in the Blender Principled BSDF (and I believe in the others as well) is mostly used for dense media with a "large amount of" subsurface scattering. Such materials can easily be handled with a BSDF, or to be more precise, a BSSRDF.

But Blender actually differentiates between BS(SR)DF and volume parameters:

Blender Principled

As you can see, the Principled BSDF node on the left is connected to the "Surface" input of the material node on the right. But the material node has a second input: "Volume". Here we can connect a volume shader that defines the parameters closely related to `attenuationColor` and `subsurfaceColor`. Blender has for example "Principled Volume", "Volume Absorption", "Volume Scatter".

Now there is a strange disambiguity: It is possible in Blender to define different scattering behavior in the BSSRDF and the volume shader! While it might give interesting effects in some cases, I think we should avoid it to keep the model simple.

There is a very similar concept in Standard Surface. [Subsurface Scattering](https://autodesk.github.io/standard-surface/#closures/subsurfacescattering) has a set of parameters corresponding to a BSSRDF. And [Specular Transmission](https://autodesk.github.io/standard-surface/#closures/speculartransmission) has another set of very similar parameters (transmission_color, transmission_depth, transmission_scatter), corresponding to the volume.

So how does it map to glTF?
* The subsurface scattering originating from the Blender Principled BSDF maps to translucency>0 (glTF KHR_materials_translucency == Blender Principled Subsurface, glTF KHR_materials_volume = Blender Principled SSS Radius/Color)
* The subsurface scattering and absorption from the Blender Volume Shader maps to transmission>0 (glTF KHR_materials_transmission == Blender Principled Transmission, glTF KHR_materials_volume == Blender Volume Shader)

This has the following advantages:
* No ambiguity, only one set of Absorption/SSS parameters.
* No need to specify a certain BSSRDF diffusion profile; random walk is the ground-truth. In fact, Blender Principled and Standard Surface both provide it as an option.

MiiBond:
What does it mean to have a `subsurfaceColor` of black? I ask because this is what we have for the default value and I don't see a way to turn off scattering in the current design. Is the intent for a `subsurfaceColor` of black to turn off the scattering? I would think that this would make the material tend towards black rather than to decrease scattering.
Perhaps we need a parameter for the scatter density or mean-free-path?

proog128:
A black `subsurfaceColor` disables scattering, it results in a single-scattering albedo of 0. Therefore, only absorption will happen in the medium, taking into account the mean free path length. Attenuation distance and color are the inverse of mean free path length:

```
attenuation_coefficient = 1 / mean_free_path = -log(attenuation_color) / attenuation_distance
```

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,268 @@
...
+## Properties
+
+The extension defines the following parameters to describe the volume.
+
+| | Type | Description | Required |
+|-|------|-------------|----------|
+| **thicknessFactor** | `number` | The thickness of the volume beneath the surface in meters in the local coordinate system of the node. If the value is 0 the material is thin-walled. Otherwise the material is a volume boundary. The `doubleSided` property has no effect on volume boundaries. | No, default: `0`. |
+| **thicknessTexture** | `textureInfo` | A texture that defines the thickness, stored in the G channel. This will be multiplied by `thicknessFactor`. | No |
+| **attenuationDistance** | `number` | Density of the medium given as the average distance in meters that light travels in the medium before interacting with a particle. | No, default: +Infinity |
+| **attenuationColor** | `number[3]` | The color that white light turns into due to absorption when reaching the attenuation distance. | No, default: `[1, 1, 1]` |
```
rsahlin, line 82: Is this parameter here to model the effect on absorption at different wavelengths, ie similar to what happens when light is absorbed in water and blue/violet light reaches further - making a sea appear blue-ish?

If so - maybe this could be mentioned?
If not, could you please explain the physicality of this parameter?

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Properties
+
+The extension defines the following parameters to describe the volume.
+
+| | Type | Description | Required |
+|-|------|-------------|----------|
+| **thicknessFactor** | `number` | The thickness of the volume beneath the surface in meters in the local coordinate system of the node. If the value is 0 the material is thin-walled. Otherwise the material is a volume boundary. The `doubleSided` property has no effect on volume boundaries. | No, default: `0`. |
+| **thicknessTexture** | `textureInfo` | A texture that defines the thickness, stored in the G channel. This will be multiplied by `thicknessFactor`. | No |
+| **attenuationDistance** | `number` | Density of the medium given as the average distance in meters that light travels in the medium before interacting with a particle. | No, default: +Infinity |
```
rsahlin, line 81: Is this the average distance between particles in the medium that will affect how many times the light will either scatter or be absorbed?

extensions/2.0/Khronos/KHR_materials_volume/README.md
proog128: Yes, exactly, attenuation color is the wavelength-dependent part of the attenuation coefficient. The description in the table here is just a small teaser. The detailed relationship between attenuation color and distance is explained in section "Absorption and Subsurface Scattering" (which describes the relationshop between the parameters and the physical quantities) and, to some extend, also in section "Base Color and Absorption".

rsahlin: Thanks @proog128 
While I read the `Absorption and Subsurface Scattering` section I was not able to deduce the source of the parameter.

extensions/2.0/Khronos/KHR_materials_volume/README.md
proog128: Maybe not exactly how many times (the light bounces infinitely), but more the average distance light travels in the medium until it hits a particle. The particle will then either scatter the light or absorb it. How much of it is absorbed or scattered is given by the absorption coefficient (not to be confused with the attenuation coefficient!) and the particle's albedo (i.e., the single scattering color derived from the subsurface color). 

If subsurface color is 0, the single scattering color becomes 0. Thus, there is no scattering, and only in this case (!) absorption coefficient = attenuation coefficient.

If subsurface color is > 0, single scattering color also becomes > 0. Now the particles scatter and absorb the light, according to some ratio.

Note that by parameterizing with attenuation coefficient and particle albedo we cannot construct the case of no absorption, scattering only.

The attenuation distance is the scalar part of the attenuation coefficient. The attenuation coefficient is the inverse of the mean free path length.

attenuation_coefficient = 1 / mean_free_path = -log(attenuation_color) / attenuation_distance

extensions/2.0/Khronos/KHR_materials_volume/README.md
proog128: I will try to improve the connection between the section and the table. Something like (additions in bold):

> To provide convenient parameterization for users this extension **provides three user-friendly parameters** attenuation color a, attenuation distance d and subsurface color ρms (see section "Properties"). **We will now describe how to derive the physical parameters for absorption and scattering in the medium from the user-friendly parameters.**

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,276 @@
...
+| **attenuationColor** | `number[3]` | The color that white light turns into due to absorption when reaching the attenuation distance. | No, default: `[1, 1, 1]` |
+| **subsurfaceColor** | `number[3]` | The overall color perceived at the surface due to subsurface scattering. | No, default: `[0, 0, 0]` |
```
rsahlin, line 83: I would propose to remove _subsurfaceColor_ and rely on _baseColor_ - mainly for two reasons:
- I don't think there is a realistic usecase where we need both _baseColor_ AND _subsurfaceColor_
- It could be argued that we are already (sort of) modelling this effect with the diffuse contribution.

I think it would simplify the solution to only have _attenuationDistance_ and _attenuationColor_ without sacrificing material representation.

rsahlin: To follow up on yesterdays discussion on the pbr tsg call:
I am happy with the naming as it is since it is a combination of several physical properties that we do not need to model separately (scattering + absorption)

rsahlin: I would also like to add a non linear factor for the absorption rate (attenuation coefficient) - a lot of our translucent materials use this.
What I mean by this is an exponential function for the attenuation as the light interacts with particles within the volume - a value of 1.0 means linear attenuation, larger values will make the light absorb faster and vice versa.

proog128: > I would propose to remove subsurfaceColor and rely on baseColor [...]

There are a lot of similarities between *subsurfaceColor* and *baseColor*, but there is one important difference: *baseColor* can be textured (2D), but *subsurfaceColor* must be constant (or a 3D texture). Thus we need to split it up.

> I would also like to add a non linear factor for the absorption rate (attenuation coefficient)

Do you mean providing an *attenuationCoefficient* parameter instead of *attenuationColor* and *attenuationDistance*? Note that *attenuationCoefficient* = -log(*attenuationColor*)/*attenuationDistance*.

rsahlin: > There are a lot of similarities between _subsurfaceColor_ and _baseColor_, but there is one important difference: _baseColor_ can be textured (2D), but _subsurfaceColor_ must be constant (or a 3D texture). Thus we need to split it up.

Hmm, I don't think I understand the reasoning for this.
As defined in the transmission extension the baseColor is used to tint the light that enters through the medium.
I think the same should be the case for this extension - or am I missing something?

Conceptually, I see it something like this:
Light enters through the medium and is colored by the baseColor, it travels _attenuationDistance_ before interacting with a particle (inside the volume).
If light escapes the volume after interacting with a particle the entered light will be colored by the attenuationColor and affected by the attenuationDistance.
If light escapes the volume without interacting with a particle it will be the same as the entered light.

Is this in line with what you have in mind?

rsahlin: > Do you mean providing an _attenuationCoefficient_ parameter instead of _attenuationColor_ and _attenuationDistance_? Note that _attenuationCoefficient_ = -log(_attenuationColor_)/_attenuationDistance_.

I think so - sorry for not grasping all of your calculations correctly :-)

The effect I am after is a non-linear falloff of escaping light based on the object thickness - as the above calculations does not take thickness into account I cannot see how that would be done.....

MiiBond: Hi @rsahlin. This extension is all about thickness. The parameters combine to do exactly what you're describing. The light is attenuated based on thickness. The attenuation coefficient is calculated as @proog128 said but it is applied using the thickness and Beer's Law.
`baseColor` is a constant absorption only at the surface.

rsahlin: Again, sorry for not being able to fully follow your calculations.

I think that I need to spend some time understanding this and how it relates to the properties we want our materials to exhibit....

emackey:
@proog128 I think the direction we're headed now indicates that subsurface-related parameter(s) should be split out from `KHR_materials_volume` and migrated to their own draft extension.  This would allow an earlier release of the volume extension, where it is focused only on the parameters needed for refraction.

Ideally, this extension should be "future compatible" with subsurface scattering, but not support it directly or include such parameters directly.  This is because the specular and IOR extensions both have bi-directional references between their respective specifications and this volume specification, and indeed "macroscopic" refraction cannot be enabled without the thickness parameter offered by this volume extension.  The absorption parameters should remain here as well, and used for refraction.

So, in order for ratification and implementations of refraction to proceed separately from SSS, I think we should migrate the subsurface parameter(s) here to a new draft extension.  Does this sound OK?

proog128:
@emackey Sounds good! I will make the necessary changes to this extension and create a new extension for subsurface scattering. I hope to have a draft ready next Monday!

extensions/2.0/Khronos/KHR_materials_volume/README.md
donmccurdy: Have we resolved this thread? I don't know whether something like `minThickness` and `maxThickness` (or `thicknessOffset`) are really important, but just want to verify that we're OK with omitting them as this extension moves to completion.

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
```diff
@@ -0,0 +1,37 @@
...
+    "type": "object",
+    "description": "glTF extension that defines the parameters for the volume of a material.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "thicknessFactor": {
+            "type": "number",
+            "minimum": 0.0,
+            "default": 0.0,
+            "description": "Thickness of the volume in meters.",
+            "gltf_detailedDescription": "The thickness of the volume beneath the surface in meters. The value is relative to the local coordinate system of the mesh's node. A value greater than 0 turns the mesh into a volume with a homogeneous medium, enabling refraction, absorption and subsurface scattering. The actual value may be ignored by renderers that are able to derive the thickness from the mesh (ray tracer)."
```
emackey, line 13: The first sentence here says "in meters", but in `attenuationDistance` it says "in world space".

Let's be consistent and use one term for both.  Which one is better, world space?  That way some other extension could redefine world distance units to be something other than meters.

extensions/2.0/Khronos/KHR_materials_volume/README.md
emackey: @abwood is this thread resolved?

extensions/2.0/Khronos/KHR_materials_volume/README.md
emackey: The `thicknessFactor` is now specified in meters, although I started a separate thread comparing it to `attenuationDistance`.  I think this old thread can be resolved, as there is no more min/max thickness.

extensions/2.0/Khronos/KHR_materials_volume/README.md
abwood: Yes, this is now covered in Appendix B.

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
abwood: This conversation triggered a thought here.  In many applications, a scale transformation can be applied to models to fit into a scene.  What does a scaled model mean for the volume material? I think we might want to say that thickness and attenuation distance is in model space, implying that these values should also be scaled if the model is also scaled.

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,253 @@
...
+<img src="./figures/transmissive-roughness.png"/>
+<figcaption><em>Transmissive sphere with varying roughness. From left to right: 0.0, 0.2, 0.4.</em></figcaption>
+</figure>
+
+If the extension is combined with `KHR_materials_translucency`, a diffuse transmission BTDF describes the transmission of light through the volume boundary. The object becomes translucent. The roughness parameter only affects the reflection. `KHR_materials_translucency` may be a good choice if the appearance should match a material which exhibits strong subsurface scattering. Examples for these dense materials are skin or candle wax. 
+
+<figure style="text-align:center">
+<img src="./figures/translucent-roughness.png"/>
+<figcaption><em>Translucent sphere with varying roughness. From left to right: 0.0, 0.2, 0.4.</em></figcaption>
+</figure>
```
emackey, line 152: This section needs to move to one of the other draft extensions, such as SSS or translucency. (If the latter, it should get updated screenshots showing thin-walled differences between transmission and translucency).

bsdorra: I'm on it

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
bsdorra: @abwood wouldn't that be what "in world space" means? So when you have the same object in different sizes it still looks identical. Whereas "in meter" would mean that the objects change appearance with size. Two different behaviors, not sure which one to prefer.

bsdorra: ah, didn't read carefully. You're saying this would be "model space". So "meter" and "world space" would be identical.

emackey: Thinking this through further:  Imagine the case of a large wax candle with SSS that allows a bit of light to come through around the edges.  If a copy of this candle model gets scaled down and placed on a birthday cake, the smaller candle should allow light to scatter through the center of the candle in ways that the full-scale model did not allow.

But now consider the difference between "thickness" (from a map) and "attenuationDistance."  When scaling the model node up or down, thickness scales along with the node.  The material's attenuation distance does not scale.

Not sure if there's a perfect solution, but we should consider these effects.

proog128: > But now consider the difference between "thickness" (from a map) and "attenuationDistance." When scaling the model node up or down, thickness scales along with the node. The material's attenuation distance does not scale.

This is how it should behave, we may have to explain it more clearly. As with any other map in the material (2D texture), the "thickness" scales with the node. The "attenuationDistance", on the other hand, is difficult to scale, because the scale might be non-uniform, and this is hard to handle in the renderer (ray length derived from attenuation distance suddenly becomes dependent on the direction due to the non-uniform scaling matrix). Apart from that, its hard to use: when assigning materials after placing objects in the scene (scale, rotate, position), it is hard to reason about the "local size" of objects (it's hidden behind potentially nested transformation matrices, shown in some hidden dialog boxes).

A quick check in Blender Cycles confirms that. This is a screenshot of Blender with two spheres of different size, but the same material sharing the density value (≈ attenuation distance). Both look the same.

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+<figure style="text-align:center">
+<img src="./figures/thin-thick-rendering.png"><br/>
+<img src="./figures/thin-thick.svg">
+<figcaption><em>Renderings of various objects (top) and corresponding top-down slice through the scene (bottom). The solid line represents the mesh. The gray area represents the volume. Thin-walled materials can be applied to open (left) and closed meshes (middle). The dashed line indicates the imaginary bounds of the infinitely thin volume. The volumetric material can only be applied to closed meshes (right), resulting in volumetric effects like refraction.</em></figcaption>
+</figure>
+
+The volume extension has to be combined with `KHR_materials_transmission` or `KHR_materials_translucency`. Light that falls onto the volume boundary may enter the volume, depending on the transmission or translucency of the surface's BSDF. Light traveling through the volume is subject to absorption by the medium's particles. When hitting the surface from inside the volume, the light may either decide to leave the volume or bounce back, depending on the material configuration and the incident direction.
```
abwood, line 44: This section can only refer to KHR_materials_transmission.

```suggestion
In order for light to enter the volume, a non-zero `transmissionFactor` must be defined through the `KHR_materials_transmission` extension. Light traveling through the volume is subject to absorption by the medium's particles. When hitting the surface from inside the volume, the light will leave the volume, depending on the incident direction.
```

extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -0,0 +1,253 @@
...
+>
+> This extension does not define the scattering part of the volumetric light transport. For all further definitions we assume the scattering coefficient σ<sub>s</sub> to be zero for all wavelength, and therefore σ<sub>t</sub> = σ<sub>a</sub>.
+
+The infinite range of the coefficient makes it rather unintuitive to control by users. To provide a convenient parameterization, this extension exposes two derived parameters: *attenuation color c* and *attenuation distance d* (see [Properties](#Properties)). The relation between the two parameters and the attenuation coefficient σ<sub>t</sub> is defined as
+
+σ<sub>t</sub> = -log(*c*) / *d*
+
+For rendering, we are interested in the change of light when traversing the medium. So we need to integrate the attenuation coefficient along a path of a certain length.
+
+In a homogenous medium, σ<sub>t</sub> is constant, and we can compute the fraction of light (radiance) transmitted after traveling a distance x via Beer's law:
```
abwood, line 115: Nitpick: missing formatting around `x`.  It should stand out as a variable in this sentence

extensions/2.0/Khronos/KHR_materials_volume/README.md
emackey: I'd prefer to avoid language that says KHR_materials_transmission is the *only* way for light to enter a volume.  Future extensions and 3rd-party extensions can define means for light to enter volumes.

bsdorra:
Added suggested changes and removed references to KHR_materials_translucency

extensions/2.0/Khronos/KHR_materials_volume/README.md
emackey: Also I think we ruled out "min" thinkness, on the thought that any sharp edge, such as along the edge of a cube, the thickness approaches zero the closer you get to the edge.  So "min" didn't seem useful in that context, but maybe the Babylon folks have different thoughts about that.

For "max", it was just included in the thicknessFactor by allowing values higher than 1.0.

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
emackey: To disassociate this with the issue of what units are used in the glTF file itself, can we avoid saying "meter" and use "world space" or "node space" instead?  I suspect that AttenuationDistance happens in "world space" but the thicknessFactor is actually "node space", because thickness scales with the node.  Would you agree?

extensions/2.0/Khronos/KHR_materials_transmission/README.md
```diff
@@ -134,34 +134,81 @@ Metals effectively absorb all refracted light (light that isn't reflected), prev
...
+    ior = 1.5,
+    base = mix(
+      diffuse_brdf(baseColor),
+      specular_btdf(α = roughness^2) * baseColor,
+      transmission),
+    layer = specular_brdf(α = roughness^2))
+```
 
-*f*<sub>*transmission*</sub> = (1 - *F*) * *T* * *baseColor* * *D<sub>T</sub>* * *G<sub>T</sub>* / *(4 * abs(dot(N, L))* * *abs(dot(N, V)))*
+where `transmission` is the transmission percentage defined by this extension's `transmissionFactor` and `transmissionTexture` properties and `specular_btdf` is a bidirectional transmission distribution function (BTDF) based on microfacet theory. The microfacet distribution function is the same Trowbridge-Reitz model used by specular reflection except sampled along the view vector rather than the reflection. The *baseColor* factor causes the transmitted light to be tinted by the surface.
```
abwood, line 161: It would be helpful to write out the definition of `specular_btdf` (using the D<sub>T</sub> G<sub>T</sub> notation).

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
```diff
@@ -0,0 +1,37 @@
...
+        },
+        "thicknessTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "Texture that defines the thickness of the volume, stored in the G channel.",
+            "gltf_detailedDescription": "A texture that defines the thickness of the volume, stored in the G channel. Will be multiplied by thicknessFactor."
+        },
+        "attenuationDistance": {
+            "type": "number",
+            "minimum": 0.0,
```
emackey, line 22: It looks like if `attenuationDistance` is exactly 0.0, it results in a divide by zero error, is that correct?  If so, we should add an `exclusiveMinimum` parameter here.

```suggestion
            "minimum": 0.0,
            "exclusiveMinimum": true,
```

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
bsdorra: Changes added with f616d7b

extensions/2.0/Khronos/KHR_materials_volume/README.md
donmccurdy: > The thicknessFactor is now specified in meters,

I can't find this in the specification, is this still current? The language in the extension specification seems unclear to me, and could be interpreted several ways:

>  The value is relative to the local coordinate system of the mesh's node.

donmccurdy:
@emackey @proog128 Just to flag that while this extension appears to be nearing consensus, there are still several un-resolved comment threads in this PR. I say "unresolved" in the sense that the GitHub "Resolve Conversation" button has not been pressed; whether the questions were resolved elsewhere I'm not sure. 

The definition of "thicknessFactor" still does seem ambiguous (i.e. no units are mentioned), that may require a bit more wording.

proog128:
@donmccurdy Thanks for bringing this up! From my point of view all conversations are resolved and the button could be pressed. Not sure what the others think?

The wording may be improved though. If I remember correctly, we settled on the following behavior:
* Thickness Factor: Given in scene units. Scales with the transformation of the parent node(s), i.e., is given in "node space" (not sure if "node space" is a familiar term in glTF).
* Attenuation Distance: Given in scene units. Does not scale with the transformation of the parent node(s), i.e., is given in world space.
Scene unit is meters in glTF (defined in the main spec) and as such we perhaps shouldn't repeat it here again.

So maybe the following is less ambigous:
* Thickness Factor: 
  * Description: Thickness of the volume ~~in meters~~.
  * Detailed Description/Readme: The thickness of the volume beneath the surface. The ~~value is relative to~~ thickness scales with the ~~local coordinate system~~ transformation matrix of the mesh's node. [...]
* Attenuation Distance:
  * Description: Average distance that light travels in the medium before interacting with a particle.
  * Description/Detailed Description/Readme: Density of the medium given as the average distance that light travels in the medium before interacting with a particle. ~~The value is given in world space.~~ The attenuation distance is defined in world space and does not scale with the transformation matrix of the mesh's node.

echadwick-wayfair:
I'm trying to resolve the appropriate attenuation settings for the hanging amber beads in the [StainedGlassLamp asset](https://github.com/KhronosGroup/glTF-Sample-Models/pull/292).

[glTF Sample Viewer](https://github.khronos.org/glTF-Sample-Viewer-Release/) screenshot:
2021-04-03 09_28_49-glTF Sample Viewer

Dimensions of the beads model, in z-up coordinate space. The node has no additional transform in glTF, it should be 1-to-1.
2021-04-03 09_22_52-2021-03-25_StainedGlassLamp_EmissiveGems max - Autodesk 3ds Max 2020

Dimensions of a single bead (although all the beads are actually in a single mesh in the glTF):
2021-04-03 09_25_25-2021-03-25_StainedGlassLamp_EmissiveGems max - Autodesk 3ds Max 2020

The current material. I'm just not sure what to use for thicknessFactor and attenuationDistance.
2021-04-03 09_34_46-StainedGlassLamp gltf - Visual Studio Code

So if I want the beads to show some attenuation on the thickest beads, would I use something like this?
thicknessFactor: 1
attenuationDistance: 0.013 (thickest depth across a single bead)

I'm not using a thickness texture, thus I don't have per-texel values to scale, so I would want the mesh itself to determine thickness. In other words, I want the thickness to be per-bead. Is that right?

proog128:
@echadwick-wayfair The thicknessFactor tells the renderer how large the geometry is, so the 0.013 must go into the thicknessFactor. A ray tracer would ignore this value, because it can infer it from the mesh. A rasterizer needs a correct thickness. It's crucial that the thicknessFactor matches the actual thickness of the geometry, otherwise the renderings will be inconsistent between rasterizer and ray tracer.

Once thickness is set up, you can adjust the attenuationDistance and attenuationColor such that it has the intended look.

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
```diff
@@ -0,0 +1,38 @@
...
+            "exclusiveMinimum": true,
+            "description": "Average distance that light travels in the medium before interacting with a particle. The value is given in world space. When undefined, the value is assumed to be infinite."
+        },
+        "attenuationColor": {
+            "type": "array",
+            "items": {
+                "type": "number",
+                "minimum": 0.0
```
lexaknyazev, line 30: Why is there no `maximum` value here?

emackey: This one might be an oversight...  @proog128 ?

proog128: Yes, that's an oversight, maximum is 1.

proog128: Fixed.

proog128:
Based on the feedback from yesterday I tried to improve the description of attenuation distance and the section about the thickness texture. Please have another look and let me know what you think.

emackey:
Looks good to me.  Thanks!

echadwick-wayfair:
Thanks for the changes, the thicknessTexture is easier to understand now!

extensions/2.0/Khronos/KHR_materials_volume/schema/glTF.KHR_materials_volume.schema.json
```diff
@@ -0,0 +1,40 @@
...
+            "items": {
+                "type": "number",
+                "minimum": 0.0,
+                "maximum": 1.0
+            },
+            "description": "Color that white light turns into due to absorption when reaching the attenuation distance.",
+            "default": [ 1.0, 1.0, 1.0 ],
+            "minItems": 3,
+            "maxItems": 3
+        }
```
lexaknyazev, line 38: Please add the regular trailing boilerplate here:
```json
"extensions": { },
"extras": { }
```

proog128: Done.

emackey:
We'll add a non-normative note discouraging folks from using nonuniform scales.  That can go in a new PR when it's ready.

Merging this.  Status will remain "Draft" for at least a few more weeks, but it will be hosted in the main glTF repository.

PR #1728, opened 2019-12-16
Register KDAB prefix

emackey:
@jjcasmar Your KDAB prefix should be registered.  Does this entry look OK?


emackey:
via https://github.com/KhronosGroup/glTF-Blender-IO/pull/732

jjcasmar:
That seems perfect. Thanks

PR #1729, opened 2019-12-24
Clarify that vertex colors are in linear space

donmccurdy:
Fixes https://github.com/KhronosGroup/glTF/issues/1638.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -770,7 +770,7 @@ Valid accessor type and component type for each attribute semantic property are 
 |`TANGENT`|`"VEC4"`|`5126`&nbsp;(FLOAT)|XYZW vertex tangents where the *w* component is a sign value (-1 or +1) indicating handedness of the tangent basis|
 |`TEXCOORD_0`|`"VEC2"`|`5126`&nbsp;(FLOAT)<br>`5121`&nbsp;(UNSIGNED_BYTE)&nbsp;normalized<br>`5123`&nbsp;(UNSIGNED_SHORT)&nbsp;normalized|UV texture coordinates for the first set|
 |`TEXCOORD_1`|`"VEC2"`|`5126`&nbsp;(FLOAT)<br>`5121`&nbsp;(UNSIGNED_BYTE)&nbsp;normalized<br>`5123`&nbsp;(UNSIGNED_SHORT)&nbsp;normalized|UV texture coordinates for the second set|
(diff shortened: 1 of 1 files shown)


zeux:
(moving the comment from #1638 here)

Note that when vertex colors are stored as 8-bit, sRGB is going to be substantially more efficient at using the available precision for encoding with small perceptual error. I intuitively expected vertex colors to be in sRGB space just like texture contents for base_color texture is for this reason.

The discussion in the linked issue suggests using wider encoding when precision is important, but this seems unsatisfactory - why should we compromise the transmission/memory size (... yeah I'm biased) here, if sRGB data is typically perfectly adequate?

I am wondering what the reasoning is for using linear vertex colors.

donmccurdy:
Note that this PR is just clarifying the existing spec — allowing sRGB vertex colors would be a breaking change requiring an extension or a future glTF version.

I don't know the original reasoning here, but will take a guess: vertex colors are linear multipliers with the material color and (gamma-corrected) texture color. But, unlike with textures, there is no automatic hardware conversion from sRGB to linear, and it must be done on the CPU or in the shader instead, either of which involve tradeoffs between performance and correctness.

My understanding is that engines may expect different color spaces for vertex colors, as well:

- Unity: sRGB vertex colors
- Unreal: linear vertex colors
- three.js: linear vertex colors

^I couldn't actually find documentation for this on Unity and Unreal, but forum discussions seemed to suggest sRGB and linear, respectively. More datapoints would be welcome. :)

zeux:
It's true that without hardware support for the color conversion, it's a bit harder to support both.

Our (Roblox) renderer uses sRGB colors but of course it's not very general purpose, and this is at least partially motivated by the use of sRGB pipeline in the earlier days. UE4 probably supports anything through shader graphs. Babylon.js looks like it assumes linear vertex color as well.

I guess it is what it is. It's still feasible to encode 8-bit colors, it just doesn't look quite as good in some cases, and future expansions of the format selection can help (e.g. RGB10_2 can store linear colors with precision that should be sufficient to represent sRGB 8-bit colors, so if vertex colors don't have alpha, at some point in the future this can become a reasonable choice).

emackey:
@donmccurdy Was this PR closed intentionally?

donmccurdy:
Yes, but I was just closing a bunch of my older PRs on various repos. Can reopen if we want to merge it.

PR #1738, opened 2020-01-16
PBR Metallic-Roughness Extension Overview

proog128:
This pull request adds a section to the glTF 2.0 specification that summarizes all extensions of the PBR metallic-roughness material:
* KHR_materials_ior (#1718)
* KHR_materials_specular (#1719)
* KHR_materials_transmission (#1698)
* KHR_materials_sheen (#1688)
* KHR_materials_clearcoat (#1677)


specification/2.0/README.md
```diff
@@ -1175,6 +1175,12 @@ The following equations show how to calculate bidirectional reflectance distribu
 
 All implementations should use the same calculations for the BRDF inputs. Implementations of the BRDF itself can vary based on device performance and resource constraints. See [Appendix B](#appendix-b-brdf-implementation) for more details on the BRDF calculations.
 
+#### Extensions
+
+glTF 2.0 defines several optional extensions that make the metallic-roughness material more flexible. The following image shows the complete material including all extensions. The default values for parameters introduced in extensions are chosen in a way that they do not affect the material if the extension is not used.
```
bghgary, line 1180: I'm not sure this section makes sense in the core spec. glTF 2.0 does not define optional extensions. These extensions are based on glTF 2.0, but glTF 2.0 itself does not define them. Maybe this should go in the [extensions README](https://github.com/KhronosGroup/glTF/blob/master/extensions/README.md)?

proog128: That's a good point. Only thing that bothers me a bit is the fact that the information about materials is already scattered across so many places (material chapter, appendix b, several extension readmes), and with this we would introduce yet another place. I probably wouldn't have guessed that the glTF Extension Registry contains information about material semantics, but maybe it's just me :-). Maybe rephrasing to something like "Several optional extensions provide more flexibility to the metallic-roughness material. ....." could be an alternative?

bghgary: I think I'll be okay if we figure out a way to say it such that it's some kind of note (or appendix?) to the spec. The way you have it originally doesn't sound right to me.

PR #1741, opened 2020-01-17
KHR_materials_specular extension branch.

UX3D-nopper:


bhouston:
I've made some minor changes in this fork:
https://github.com/bhouston/glTF/tree/extensions/KHR_materials_specular/extensions/2.0/Khronos/KHR_materials_specular

Specifically instead of using the unreal engine parameterization of specifying F0 directly, I've proposed adopting "specularIOR".  I have also suggested "specularFActor" which allows you to scale the grazing angle max Fresnel result.  And lastly "specularColor" which is a tint to the overall specular reflection.  "specularFactor" and "specularColor" could be combined in an implementation but both Autodesk and Dassault keep them separate for easy of use for artists.

bhouston:
I just noticed that my proposal was nothing more than the existing competing specular proposal: https://github.com/KhronosGroup/glTF/pull/1719  I support that one over this one.

kofi-de-com:
+1 for this extension. Aligns well with Blender's workflow.

donmccurdy:
@kofi-de-com is there something specific about this proposal that you prefer over the version in #1719? Or do you mean a general +1 for either?

kofi-de-com:
@donmccurdy it turned from a +1 for the specific proposal into a +1 for the extension in general.
At first glance it seemed easier to not have a specularColor feature. I've since then changed my mind and would suggest to move forward with #1719.
Background: As far as I can tell Blender does not properly support specular color tinting as suggested in #1719. It's specular tint value defaults to 0 (which behaves something like specularColorFactor(1, 1, 1)). If specular tint in Blender is set to 1, it looks like specularColorFactor(baseColor.r, baseColor.g, baseColor.b). This means Blender's material values can be converted at export time to work with #1719.

UX3D-nopper:
closed in favor for https://github.com/KhronosGroup/glTF/pull/1719

PR #1742, opened 2020-01-17
KHR_materials_thinfilm extension branch.

UX3D-nopper:


bhouston:
Could we simplify the parameters for this to just thinFilmThickness and thinFilmIOR?

I notice this is the parameterization used in the main reference paper here: https://hal.archives-ouvertes.fr/hal-01518344/document

It is also the parameterization used in Autodesk Standard Surface:
```
thin_film_thickness* | float | 0 | thickness of the film (in nanometres
thin_film_IOR* | float | 1.5 | refractive index of the film
```
https://autodesk.github.io/standard-surface/

Why was this more specific parameters chosen rather than the more physical parameterization that is also what everyone else is using?  I highly advocate for the more physical ones as this allows for implementations to improve their physicality and we do not have to change the specification.

UX3D-nopper:
The parametrization - in general for glTF materials - is to feed the GPU directly. So the parameters above are sufficient for a path tracer, but not for the GPU - depending of course which algorithms etc. we use.
Internally, you can use Standard Surface or Enterprise PBR. From these parameters, you need to generate the data for real-time - as glTF is a transmisson format.

UX3D-nopper:
We can also discuss this in the break ;-)

bhouston:
I created a branch here with my very minor proposed changes:
* https://github.com/bhouston/glTF/tree/KHR_material_thinfilm

I can not make a PR against your fork apparently.

donmccurdy:
As this PR is more recent than [KHR_materials_transmission](https://github.com/KhronosGroup/glTF/pull/1698), I have to assume there are reasons you prefer it — would you mind expanding on the motivation? Am I correct in understanding that this extension (if ratified) would provide the features of both `KHR_materials_transmission` and `KHR_materials_volume` (which would then not be ratified)?

UX3D-nopper:
The `KHR_materials_transmission` is a different visual effect and from my point of view do not intersect. The same is for `KHR_materials_volume `.

emackey:
Would it be reasonable to rename this extension to something more like `KHR_materials_iridescence`?

I ask because when I look at the long list of proposed PBR Next extensions, there are many that talk about "thin" and "coat" and IOR and transmission.  I feel that this extension's name should distance itself from all those other proposed extensions.

UX3D-nopper:
http://www.dgp.toronto.edu/~nmorris/CSC2530/Project/morris.pdf

bhouston:
There is still intent from Threekit for pushing this towards a full standard.  It is something that arises often.

UX3D-nopper:
Yes, it is the roadmap for PBR Next. Maybe just join the Monday call.

UX3D-nopper:
image
@bhouston Let's push this, as it is well defined:
https://blog.unity.com/technology/a-practical-extension-to-microfacet-theory-for-the-modeling-of-varying-iridescence
https://hal.archives-ouvertes.fr/hal-01518344/document
https://blog.selfshadow.com/publications/s2017-shading-course/#course_content

UX3D-nopper:
We continue here https://github.com/KhronosGroup/glTF/pull/2027

PR #1748, opened 2020-01-22
Only one node can reference a particular camera.

prideout:
Otherwise, it is ambiguous which transform to use.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -1260,7 +1260,7 @@ This specification does not define size and style of non-triangular primitives (
 
 ## Cameras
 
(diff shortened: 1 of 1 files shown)


scurest:
I've always assumed cameras are instantiated at nodes the same way meshes are

> A mesh is instantiated by `node.mesh` property. The same mesh could be used by many nodes, which could have different transformations.

The discussion about KHR_lights_punctual also seemed to assume cameras/lights are instantiated, although it never says it explicitly in the spec like it does for meshes.

lexaknyazev:
A `camera` object defines only "lens config", so there's no limit on the number of references to it.

prideout:
Okay that makes sense.  I think some language *is* missing, but my PR is obviously wrong.

PR #1749, opened 2020-01-24
Elaborate on skinning weights sum

lexaknyazev:
Following the discussion from https://github.com/KhronosGroup/glTF-Validator/issues/132.

Rationale is borrowed from https://github.com/KhronosGroup/glTF-Validator/pull/124#issuecomment-552758108 by @zeux.

/cc @vpenades

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -942,9 +942,15 @@ The number of joints that influence one vertex is limited to 4 per set, so refer
 
 The joint weights for each vertex must be non-negative. No joint may have more than one non-zero weight for a given vertex.
 
-When the weights are stored using `FLOAT` component type, their linear sum must be as close as possible to `1.0` for a given vertex. When the weights are stored using `UNSIGNED_BYTE` or `UNSIGNED_SHORT` component types, their linear sum before normalization must be `255` or `65535` respectively.
+When the weights are stored using `FLOAT` component type, glTF exporters should produce weights with linear sum as close as reasonably possible to `1.0` for a given vertex.
 
-> **Implementation Note:** Otherwise, vertices would be deformed very significantly because the weight error would get multiplied by the joint position. So an error of `1/255` in the weight sum is actually unacceptably large.
+> **Implementation Note:** The threshold in the official validation tool is set to `2e-7` times the number of non-zero weights per vertex.
+
+When the weights are stored using `UNSIGNED_BYTE` or `UNSIGNED_SHORT` component types, their linear sum before normalization must be `255` or `65535` respectively.
+
+> **Implementation Note:** Since the allowed threshold is much lower than minimum possible step for quantized component types, exporters should just renormalize weight sum after quantization.
+
+Without these requirements, vertices would be deformed significantly because the weight error would get multiplied by the joint position. For example, an error of `1/255` in the weight sum would result in an unacceptably large difference in the joint position.
 
 In the event that of any of the vertices are influenced by more than four joints, the additional joint and weight information will be found in subsequent sets. For example `JOINTS_1` and `WEIGHTS_1` if present will reference the accessor for up to 4 additional joints that influence the vertices. Note that client implementations are only required to support a single set of up to four weights and joints, however not supporting all weight and joint sets present in the file may have an impact on the model's animation.
 
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -935,12 +935,16 @@ The mesh for a skin is defined with vertex attributes that are used in skinning 
 }
 ```
 
-The number of joints that influence one vertex is limited to 4 per set, so referenced accessors must have `VEC4` type and following component formats:
+The number of joints that influence one vertex is limited to 4 per set, so referenced accessors must have `VEC4` type and following component types:
 
 * **`JOINTS_0`**: `UNSIGNED_BYTE` or `UNSIGNED_SHORT`
 * **`WEIGHTS_0`**: `FLOAT`, or normalized `UNSIGNED_BYTE`, or normalized `UNSIGNED_SHORT`
 
-The joint weights for each vertex must be non-negative, and normalized to have a linear sum of `1.0`. No joint may have more than one non-zero weight for a given vertex.
+The joint weights for each vertex must be non-negative. No joint may have more than one non-zero weight for a given vertex.
+
+When the weights are stored using `FLOAT` component type, their linear sum must be as close as possible to `1.0` for a given vertex. When the weights are stored using `UNSIGNED_BYTE` or `UNSIGNED_SHORT` component types, their linear sum before normalization must be `255` or `65535` respectively.
+
+> **Implementation Note:** Otherwise, vertices would be deformed very significantly because the weight error would get multiplied by the joint position. So an error of `1/255` in the weight sum is actually unacceptably large.
 
 In the event that of any of the vertices are influenced by more than four joints, the additional joint and weight information will be found in subsequent sets. For example `JOINTS_1` and `WEIGHTS_1` if present will reference the accessor for up to 4 additional joints that influence the vertices. Note that client implementations are only required to support a single set of up to four weights and joints, however not supporting all weight and joint sets present in the file may have an impact on the model's animation.
 
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -942,15 +942,11 @@ The number of joints that influence one vertex is limited to 4 per set, so refer
 
 The joint weights for each vertex must be non-negative. No joint may have more than one non-zero weight for a given vertex.
 
-When the weights are stored using `FLOAT` component type, glTF exporters should produce weights with linear sum as close as reasonably possible to `1.0` for a given vertex.
+When the weights are stored using `FLOAT` component type, glTF exporters should produce weights with linear sum as close as reasonably possible to `1.0` for a given vertex. When the weights are stored using `UNSIGNED_BYTE` or `UNSIGNED_SHORT` component types, their linear sum before normalization must be `255` or `65535` respectively. Without these requirements, vertices would be deformed significantly because the weight error would get multiplied by the joint position. For example, an error of `1/255` in the weight sum would result in an unacceptably large difference in the joint position.
 
 > **Implementation Note:** The threshold in the official validation tool is set to `2e-7` times the number of non-zero weights per vertex.
 
-When the weights are stored using `UNSIGNED_BYTE` or `UNSIGNED_SHORT` component types, their linear sum before normalization must be `255` or `65535` respectively.
-
 > **Implementation Note:** Since the allowed threshold is much lower than minimum possible step for quantized component types, exporters should just renormalize weight sum after quantization.
-
-Without these requirements, vertices would be deformed significantly because the weight error would get multiplied by the joint position. For example, an error of `1/255` in the weight sum would result in an unacceptably large difference in the joint position.
 
 In the event that of any of the vertices are influenced by more than four joints, the additional joint and weight information will be found in subsequent sets. For example `JOINTS_1` and `WEIGHTS_1` if present will reference the accessor for up to 4 additional joints that influence the vertices. Note that client implementations are only required to support a single set of up to four weights and joints, however not supporting all weight and joint sets present in the file may have an impact on the model's animation.
 


specification/2.0/README.md
```diff
@@ -935,12 +935,16 @@ The mesh for a skin is defined with vertex attributes that are used in skinning
...
 
 * **`JOINTS_0`**: `UNSIGNED_BYTE` or `UNSIGNED_SHORT`
 * **`WEIGHTS_0`**: `FLOAT`, or normalized `UNSIGNED_BYTE`, or normalized `UNSIGNED_SHORT`
 
-The joint weights for each vertex must be non-negative, and normalized to have a linear sum of `1.0`. No joint may have more than one non-zero weight for a given vertex.
+The joint weights for each vertex must be non-negative. No joint may have more than one non-zero weight for a given vertex.
+
+When the weights are stored using `FLOAT` component type, their linear sum must be as close as possible to `1.0` for a given vertex. When the weights are stored using `UNSIGNED_BYTE` or `UNSIGNED_SHORT` component types, their linear sum before normalization must be `255` or `65535` respectively.
+
+> **Implementation Note:** Otherwise, vertices would be deformed very significantly because the weight error would get multiplied by the joint position. So an error of `1/255` in the weight sum is actually unacceptably large.
```
bghgary: This section reads weird with the "Otherwise" in the front as an implementation note. I think you don't need to put implementation note here.

```suggestion
> Without this requirement, vertices would be deformed significantly because the weight error would get multiplied by the joint position. For example, an error of `1/255` in the weight sum would result in an unacceptably large difference in the joint position.
```

vpenades: For the Float case, the "as close as possible to 1", might lead a developer to choose its own arbitrary error tolerance, which might fall within gltf-validator tolerance, or not.

So I would suggest replacing it to "within +-Err to 1" or something like that, where Err is the value used by gltf-validator.

scurest:
I brought up this specific case back when normalized weights were first made required: https://github.com/KhronosGroup/glTF/pull/1352#issuecomment-405137407.

My impression was that some tolerance for this problem was specifically allowed?

specification/2.0/README.md
scurest: There's no way to pick an error tolerance that isn't arbitrary. Also that still wouldn't be enough to fully specify whether it's allowed. There are multiple possible ways you can sum floating point numbers.

lexaknyazev:
@scurest 
It seems that the allowed tolerance could only be observed with float32 weights. This means that with `unorm8` components, exporters should renormalize weights after quantization.

[Implementation example](https://github.com/zeux/meshoptimizer/blob/8cae35792da51e29c1cf18469d6912a956e40d99/tools/gltfpack.cpp#L1473-L1489).

scurest:
Isn't this a breaking change?

lexaknyazev:
> There's probably plenty of code that quantitizes weights by multiplying by 255 and rounding.

There were plenty of reports of models (even in our own Sample-Models repo) behaving incorrectly because of non-unit sum (like `0.7`) or inaccurate weights quantization (see screenshots in the linked thread). The alternative would be to require engines to always preprocess weights on load.

> Isn't this a breaking change?

This PR just clarifies what "unit linear sum" means for specific component types.

vpenades:
I'm surprised with the results seen in the screenshots here: https://github.com/KhronosGroup/glTF-Validator/pull/124#issuecomment-552758108

While it might be true that this is related to weights not having exactly the same value as the originals, I suspect this is happening also because the original model's hands have some undesired influences from unrelated bones, and this has been made more apparent by quantized weights. So maybe in most cases it might not be so apparent.

I also fear what @scurest has said; even if the schema clarifies this case scenario, everybody has been asumming some level of tolerance on all component types; After all, _"if some tolerance is allowed in the Float component, why not allow it in others?"_ , so to me, even if it is a clarification, it's also a breaking change.

So, maybe a compromise solution would be to have TWO tolerances: if the weights sum has an exact sum of 1, all is good, but if it's very close to 1, issue a warning instead of an error. This will allow current implementations to keep working, and give due warning for those that want to walk the extra mile.

In any case, the specification and gltf-validator need to be aligned, so developers following the specification will not hit gltf-validator errors.

zeux:
> While it might be true that this is related to weights not having exactly the same value as the originals, I suspect this is happening also because the original model's hands have some undesired influences from unrelated bones, and this has been made more apparent by quantized weights.

I don't think the result is very surprising. If you have any surface that's skinned to multiple bones, the influences of the bones will gradually decrease/increase. As a result, the sum error after quantization will fluctuate between -1/255, 0/255 and 1/255 (occasionally getting to +-2/255 for 4-weight sources). This should just naturally occur for any regions that have >2 weights. If you have exactly 2 weights, then I think this will be rare since the sum should mostly result in precisely 255. But even if you have 2 weights, and they are both exactly equal to 0.5, naive quantization will introduce an error of 1/255.

When you have *any* error, you will see a problem. Consider the character in that screenshot. The hands look like they are at about +-0.6m from the center of the character along the X axis. The error of 1/255 results in 0.2cm in the steady position, which is 10% of the finger width. I think this is why the fingers are visibly distorted even in the stationary screenshot. Now, if you use world-space skinning (which three.js does, and it is suboptimal for a few reasons but a straightforward implementation of glTF spec would also skin in world space), then moving the character just 5 meters to the right would place the hands at 5.6m from the origin, and so the error becomes 2cm. 50 meters, and it's 20 cm. This is exactly what you see in the screenshots.

The reason why you see the vertices "oscillating" is because you have the error of the sum oscillating. 

For 16-bit weights, the problem is much less severe because the error from naive quantization is 256 times smaller - so you have to move the character much further from the origin to see it. But for 8-bit weights it's severe.

Having said that, it might make sense to change this to be a warning? I am not sure I fully understand the distinction.

vpenades:
> Having said that, it might make sense to change this to be a warning? I am not sure I fully understand the distinction.

Currently gltf-validator issues an error when it finds weights not perfecly normalized. Given that the "naive normalization" approach is what most developers will naturally take, maybe being so strict is not good for adoption.

Bug given that also not many developers are writing gltfs with packed weights, I don't think being strict will affect many developers either...

scurest:
> The alternative would be to require engines to always preprocess weights on load.

Or normalize them in the shader or whatever. CesiumMan has non-normalized weights so any engine that can load it is already doing some kind of normalization.

Having the exporter write round(255\*weight) and then having the engine normalize after de-quantization is the best case for accuracy. Any adjustments you do to make them sum to 255 increases the absolute error.

----

Btw the algorithm linked above is not necessarily the best way to do it, even with just four weights. It concentrates all the error in the maximal element.

Here are some weights quantized with just round(255\*weight). These are best possible errors.

|weight|quantized|error|
|---|---|---|
|0.7|179|0.002|
|0.1|26|0.002|
|0.1|26|0.002|
|0.1|26|0.002|

Adjusting the quantized weights to sum to 255 by modifying only the maximal element increases the absolute error.

|weight|quantized|error|
|---|---|---|
|0.7|177|-0.006|
|0.1|26|0.002|
|0.1|26|0.002|
|0.1|26|0.002|

OTOH, if you defray it over two elements, the absolute error is still the best possible

|weight|quantized|error|
|---|---|---|
|0.7|178|-0.002|
|0.1|26|0.002|
|0.1|26|0.002|
|0.1|25|-0.002|

zeux:
> Btw the algorithm linked above is not necessarily the best way to do it, even with just four weights. It concentrates all the error in the maximal element.

Yes, spreading the error across multiple elements can reduce the per-component error, but this is less important compared to minimizing the error in the sum.

I agree that normalizing in the shader can work around these issues, but do all engines do that? three.js clearly doesn't (it normalizes during load - even that should be unnecessary, it's only doing this because some models were broken)

scurest:
If it's normalizing on load, shouldn't it be adjusting these quantized weights at load? Or does it only normalize floats or something?

vpenades:
@zeux How cesium man looks in threejs? 

I think we're in a sort of crossroads, where we need to choose either requiring renormalization after quantization (which, as @scurest said, it does not guarantee to solve the problem) - or requiring engines to renormalize on load.




> If it's normalizing on load, shouldn't it be adjusting these quantized weights at load? Or does it only normalize floats or something?

I believe it's the floats, so the deviation is evenly distributed across all weights, instead of overshooting one weight.

zeux:
> it does not guarantee to solve the problem

Can you expand on this?

> If it's normalizing on load, shouldn't it be adjusting these quantized weights at load?

It only normalizes floating point weights right now. In general I don't think the specification should assume that all loaders need to normalize weights on load, the only reason why three.js did it is that bugs with broken models were reported and needed to be fixed.

lexaknyazev:
> After all, "_if some tolerance is allowed in the Float component, why not allow it in others?_" , so to me, even if it is a clarification, it's also a breaking change.

Technically, the allowed level of tolerance (`2e-7` per non-zero weight in the validator) does not depend on the component type - it's just too small to be achievable with `unorm8` (that is limited by `1/255` steps).

> even if it is a clarification, it's also a breaking change

At first, the spec didn't mention the weights sum at all. Then, after multiple reports of unexpected behavior, we agreed on "linear unit sum" in general. Now, since this check has been implemented and we have some specific examples of its impact, it's time to make the spec more concrete. 

This change certainly invalidates assets with obviously wrong weights - that's fine. There are also assets that were valid all along - that's great. Everything in the middle has some issues that may be visible only in a particular pose or with exaggerated scale (when engines do not fix that somehow, like CesiumMan in three.js).

Keep in mind, that one of the primary goals of the validator is to improve glTF compatibility with different engines/platforms/environments by subjecting assets to ever-tightening rules. A less picky workflow may knowingly override the corresponding validation severity code or mute it completely.

> Having said that, it might make sense to change this to be a warning? I am not sure I fully understand the distinction.

In my opinion, a warning should not lead to a broken render (in general).

> we're in a sort of crossroads ...

As a rule of thumb, glTF prefers putting the burden on exporters.

lexaknyazev:
Please review the updated spec language.

lexaknyazev:
Final nit. The validation note says:
> The threshold in the official validation tool is set to `2e-7` times the number of non-zero weights per vertex.

It may be unclear from what number of weights threshold accumulation starts, i.e., whether `[0, 0, 0, 0.9999998]` is a valid accessor element. Currently, this case is allowed by the validator.

vpenades:
Additionally, given the exquisitely precise this requirement is, I would also suggest to pick old CesiumMan and other models that have invalid weights and move them to the gltf-generation collection as negative load cases.

donmccurdy:
> It may be unclear from what number of weights threshold accumulation starts, i.e., whether [0, 0, 0, 0.9999998] is a valid accessor element. Currently, this case is allowed by the validator.

The validator's behavior, and considering that case to be one non-zero weight for a total threshold of 2e-7, seems like clearest option to me.

emackey:
I'm not a skinning expert, so @donmccurdy or @bghgary can one of you merge this when ready?

donmccurdy:
@lexaknyazev your comment (https://github.com/KhronosGroup/glTF/pull/1749#issuecomment-578941205) asks about the wording, and to me the wording seems clear. Is this OK to merge, or was there more clarification you feel is needed, like saying "regardless of weight order"?

lexaknyazev:
@donmccurdy 
It's not exactly about the order. To put it simply: when there's only one "active" (i.e. non-zero) weight, should it be strictly `1.0`?

Some implementations read only three weights and use `w4 = 1.0 - w1 - w2 - w3`.

donmccurdy:
Ok, I have no preference on that.

specification/2.0/README.md
donmccurdy: +1 for Gary's suggested wording. In addition, we could include an implementation note:

> **Implementation Note**: The glTF Validator uses an error tolerance of ____ when testing glTF assets for conformance.

donmccurdy:
> To put it simply: when there's only one "active" (i.e. non-zero) weight, should it be strictly 1.0?

Let's skip adding further requirements for now, a single clear requirement for the skinning sum is a good starting point.

specification/2.0/README.md
lexaknyazev: @donmccurdy 
This suggestion has been already addressed in the later commit.

lexaknyazev:
> Let's skip adding further requirements for now, a single clear requirement for the skinning sum is a good starting point.

It's more like a disambiguation rather than an extra requirement.

`2e-7` * `num_non_zero_weights` allows `[0, 0, 0, 0.9999998]` which makes little sense when zeros represent missing joints. At the very least, I'd like the spec to have a clear position on validity of such a case.

donmccurdy:
Sorry, I can't understand why `2e-7 * num_non_zero_weights` would not be sufficient, even when there's just one non-zero weight. It can be extremely difficult to correctly convert various DCC tools' skinning representations into any standard format, and I don't see much benefit to adding another branch in the validation requirements for this case.

emackey:
Having written my own skinning implementation earlier this year, I finally understand certain issues such as the importance of the weights summing to 1.0.  That said, it does make sense to allow some tolerance, particularly for systems that use different floating-point representations and must convert.  With such tolerance allowed, the `[0, 0, 0, 0.9999998]` case should also be allowed, for consistency.  That case doesn't appear any more harmful than any other 2e-7 tolerance.

lexaknyazev:
After going through the original rationale one more time, I agree with allowing `[0, 0, 0, 0.9999998]` case for consistency.

@donmccurdy @emackey
Please take a final look on the PR's language.

specification/2.0/README.md
```diff
@@ -935,12 +935,22 @@ The mesh for a skin is defined with vertex attributes that are used in skinning
...
+
+When the weights are stored using `FLOAT` component type, glTF exporters should produce weights with linear sum as close as reasonably possible to `1.0` for a given vertex.
+
+> **Implementation Note:** The threshold in the official validation tool is set to `2e-7` times the number of non-zero weights per vertex.
+
+When the weights are stored using `UNSIGNED_BYTE` or `UNSIGNED_SHORT` component types, their linear sum before normalization must be `255` or `65535` respectively.
+
+> **Implementation Note:** Since the allowed threshold is much lower than minimum possible step for quantized component types, exporters should just renormalize weight sum after quantization.
+
+Without these requirements, vertices would be deformed significantly because the weight error would get multiplied by the joint position. For example, an error of `1/255` in the weight sum would result in an unacceptably large difference in the joint position.
```
emackey, line 953: Should this line be part of the implementation note immediately above it?

donmccurdy: I think I would merge the non-implementation-note lines into a single paragraph, followed by two implementation notes. No need to interleave them sentence by sentence.

donmccurdy:
Other than the minor layout issue @emackey mentioned, this looks good to me. 👍

shrinktofit:
@lexaknyazev What happened if some vertices are not skinned? e.g they have zero sum weights?

Issue #1457, opened 2018-09-27
Alpha coverage is underspecified


romainguy:
The [glTF 2.0 specification](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0) is very vague on how alpha coverage should be supported. I do agree with what this phrase from the spec:
> There is no perfect and fast solution that works for all cases. 

I however believe that the spec could be more opinionated here. Here are several issues that could (and should IMO) be addressed by the spec:

- There is no mention of premultiplied alpha vs non-premultiplied alpha (it's kind of implied by the mention of Porter-Duff SourceOver blending mode for the `BLEND` case, but it should be more explicitly described).
- There is no mention of whether opacity is applied to specular lighting. In `BLEND` mode I think this should be explicitly disallowed.
- If opacity doesn't affect specular lighting, a new alpha coverage mode ('FADE') should probably be added to have this capability. This can be useful for specific effects like fading an object in & out.

MiiBond:
I believe that the term "alpha coverage" implies a simple over operator so lighting is not considered. i.e. it should currently work as a fade. This is why, I'd like an extension for simple optical transparency. See #1385 
I agree that the spec should be more explicit about this though. I know the implementation in Babylon already doesn't treat this as pure alpha coverage (though, they've added an option to).

emackey:
Some additional comments on blending modes are in #1033 and #822.

emackey:
Specifically, `BLEND` mode is only for coverage, not transparency or translucency:

> Use this mode for materials that simulate geometry such as gauze cloth or animal fur.

A gauze cloth with near-zero coverage would not have much if any specular reflection.

Proper support for transparency/translucency is being discussed in #1385, and generally the future of glTF materials is being brainstormed in #1442.

donmccurdy:
Some of this (premultiplied alpha, FADE mode) could perhaps be covered more explicitly in a possible [KHR_blend extension](https://github.com/KhronosGroup/glTF/pull/1302).

donmccurdy:
> There is no mention of premultiplied alpha vs non-premultiplied alpha

The spec does say, `The stored texels [of baseColorTexture] must not be premultiplied`. I'm not sure if I missed that before or if the clarification was added.

romainguy:
I don't remember seeing it before either. And does it mean the texture is not store premultiplied or that you should not alpha multiply the RGB channels during shading? And if so, why?

lexaknyazev:
[...]

PR #1751, opened 2020-01-30
KHR_texture_basisu: Basis Universal for material textures

lexaknyazev:
_Following up from #1612._

Now, `KHR_texture_basisu` extension is self-contained and should be ready for implementation.

> **Implementation note:** Basis Universal supercompression is optimized for correlated color data, like natural images. It does not perform well on images with independent channels, like normal or some material maps. See the discussion in #1682 for more details.

**Open questions before ratification**
- [x] ~Can alpha values be premultiplied? (It's possible to set this flag in KTX2).~ Not premultiplied, per core spec.
- [X] ~How to handle normal maps?~ UASTC for best results now, possible future work out of scope.
- [X] ~How to handle roughness-metallic data that is expected to use green and blue channels?~ Out of scope.
- [x] ~Does `KHR_texture_basisu` allow zstd, ~zlib~, or rely on file-level compression?~ Allow zstd, but not required.
- [x] ~Texture size restrictions?~ Require 4x4, suggest power of 2.

**Ecosystem action items**

See https://github.com/KhronosGroup/glTF/issues/1771.


extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -0,0 +1,154 @@
...
+### Using KTX2 Images with Basis Universal Supercompression for Material Textures
+
+When a texture referencing a KTX2 image with Basis Universal supercompression is used for glTF 2.0 material maps (such as base color), the KTX2 image must be of **2D** type as defined in the KTX2 Specification, namely:
+
+- `pixelHeight` must be greater than 0.
+- `pixelDepth` must be 0.
+- `layerCount` must be 0.
+- `faceCount` must be 1.
+
+In general, a KTX2 image used for material texture should contain a full mip pyramid.
```
donmccurdy: This line is normative, correct? If so, perhaps phrase this and the line below as:

A KTX2 image must contain a full mip pyramid, unless another extension depending on KHR_texture_basisu imposes different requirements.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds the ability to specify textures using KTX2 images with Basis Universal supercompression. An implementation of this extension can use such images as an alternative to the PNG or JPEG images available in glTF 2.0 for more efficient asset transmission and reducing GPU memory footprint. Furthermore, specifying mip map levels is possible.
+
+When the extension is used, it's allowed to use value `image/ktx2` for the `image.mimeType` property.
```
bghgary: I'm not sure how to word this properly, but I don't think we should simply allow image/ktx2 for images. It should only be allowed when pointed to from a `KHR_texture_basisu` extension.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
lexaknyazev: Do e.g. 2x2 and 1x1 levels bring so much value that the spec needs to _require_ them? I always thought that authors are free to decide where to stop with low-resolution levels.

zeux: How does WebGL 1.0 handle texture completeness requirements? I'm struggling to find this in the WebGL 1 spec - however GLES2 spec declares textures with partial mipchains incomplete, and I'm not sure how a WebGL1-on-top-of-GLES2 implementation would handle this.

lexaknyazev: Sorry, I forgot about GLES2, we need all levels.

If something isn't said in WebGL 1.0 then it's usually implied to follow GLES2.

lexaknyazev: On the other hand, a texture can refer to a sampler object that uses minification without mip maps. Should we allow KTX2 files to have only level 0 in this case?

donmccurdy:
> Can alpha values be premultiplied? (It's possible to set this flag in KTX2).

/cc https://github.com/KhronosGroup/glTF/issues/1457

> How to handle roughness-metallic data that is expected to use green and blue channels?

The only resource I've found was [this one](https://polycount.com/discussion/211152/best-pbr-texture-compression-and-channel-layout), claiming that ORM packed as RGB is plausible with DXT especially given that the metallic channel doesn't require much precision. But realistically, confirming that will require some investigation by an artist.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
donmccurdy: Just confirming your meaning — `image.mimeType = "image/ktx2"` would be allowed ONLY IF that image is referenced by a `texture.extensions.KHR_texture_basisu.source`; the image cannot be referenced by `texture.source`? Or are you saying we need something like `image.extensions.KHR_texture_basisu.mimeType = "image/ktx2"` here?

bghgary: The former.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
lexaknyazev: Please review the updated language.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
lexaknyazev: Please review the updated language.

lexaknyazev:
> ORM packed as RGB is plausible with DXT

Color endpoints have more degrees of freedom in DXT than in ETC1S. Even a simple case of e.g. using only metallic while keeping roughness on a single value may cause severe artifacts with ETC1S. Typical advice is to do 2x upscale before compression. It helps sometimes.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
bghgary: LGTM

extensions/2.0/Khronos/KHR_texture_basisu/README.md
bghgary: LGTM

extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -0,0 +1,153 @@
...
+}
+```
+
+### JSON Schema
+
+[texture.KHR_texture_basisu.schema.json](schema/texture.KHR_texture_basisu.schema.json)
+
+## Restrictions on KTX2 Images with Basis Universal Supercompression
+
+When used with glTF 2.0, KTX2 images with Basis Universal supercompression must conform to these additional restrictions:
```
donmccurdy: While "KTX2 images with Basis Universal supercompression" is conceptually clear, it does imply other required settings. Should we list any of these here?

- supercompressionScheme = 1
- vkFormat = VK_FORMAT_UNDEFINED
- typeSize = 1

Are key/value data and DFD blocks (other than Khronos Basic Data Format Descriptor Block) allowed?

lexaknyazev: `supercompressionScheme`, `vkFormat`, and `typeSize` for BasisU are enforced by the KTX2 spec.

Custom metadata and DFD blocks should be ignored as it's a `KHR` extension.

donmccurdy: I think it's worth explicitly saying that other methods of supercompression are not allowed by this extension – the other details are (I agree) clear enough in the KTX2 spec.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
donmccurdy: Looks good to me.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
lexaknyazev: Done.

polarathene:
> How to handle normal maps

Unless there is comparable compressed size, some might be ok with the lower quality. Two component/channel at least seems worthwhile, and is what basisu officially advises on their github README when dealing with normal maps.

Since the last activity on this discussion, basisu also has released UASTC support trading off compression for better quality textures, which is meant to be similar to BC7.

Allowing for flexibility to deliver lower quality textures for lower bandwidth can be useful if quality doesn't take too big of a dive. 

> Even a simple case of e.g. using only metallic while keeping roughness on a single value may cause severe artifacts with ETC1S.

Could that be worked around the same as is advised for normalmaps with the separation to RGB and A slices option basisu offers? KTX swizzling can then provide the data to the GB format that glTF expects?

> Typical advice is to do 2x upscale before compression. It helps sometimes.

On mobile devices, having a texture that is too large(unsupported by the device / ES version) can result in no texture rendered at all(ends up rendering as black iirc). Upscaling before compression would trigger that same issue at runtime?(I think in my experience it was 8k or 16k textures in sketchfab) Or does KTX/Basis have a way to avoid it?(downscales or uses mipmaps instead of failing?)

donmccurdy:
Whether UASTC and Basis Universal should be bundled into the same glTF extension, or not, is an important question. Some things we know:

1. Basis Universal textures are smaller files. Decreasing the quality settings of a UASTC file does not match Basis Universal in terms of compactness.
2. UASTC is higher quality. Increasing quality settings of Basis Universal encoding will not match UASTC in terms of quality.
3. Given (1) and (2), there are clear benefits with both formats. However, supporting both requires larger transcoders and more client implementation complexity. 

Some things we don't know:

* How large are the transcoders for each? Does it make sense to create a combined transcoder, and if so how large would it be?
* Early estimates on ZSTD decoder put it at ~890kb, although @MarkCallow has suggested that could likely be reduced. That decoder will be necessary for client implementations of UASTC, in addition to the transcoder.
* Will additional universal texture formats appear over the upcoming years, and when?

***

My opinion on this — which I'll readily admit is skewed toward support on the Web — is that Basis Universal and UASTC should be supported by _separate_ glTF extensions, and that we should focus on getting `KHR_texture_basisu` completed first. I believe that web developers will prioritize transmission size over GPU memory footprint, and that Basis Universal strikes a balance that will make it more common on the web. For cases where the quality of Basis Universal is insufficient, I don't want to predict yet when developers will choose UASTC (lower GPU memory cost & upload) vs. uncompressed PNG/JPG (lower transmission size). I'd prefer to allow time for that to become clear, rather than trying to foresee it.

Thoughts?

zeux:
> How large are the transcoders for each? Does it make sense to create a combined transcoder, and if so how large would it be?

Combined etc1s + uastc transcoder including support for BC7 et al is ~230 KB post-deflate; ETC1S only transcoder is ~180 KB post-deflate. I haven't looked closely into the size impact inside ETC1S and how much code is shared (e.g. whether a separate UASTC transcoder is ~50KB or more like ~150 KB).

> Early estimates on ZSTD decoder put it at ~890kb, although @MarkCallow has suggested that could likely be reduced. That decoder will be necessary for client implementations of UASTC, in addition to the transcoder.

I really feel like this is a problem in KTX2 that should be addressed. ZSTD is stronger than deflate but it's not so much stronger that it should be required in contexts where deflate is already available. Previously the KTX2 spec suggested that only Basis support was required, but this has changed and while I understand the reasoning somewhat, it doesn't seem right to me.

MarkCallow:
It is not at all clear the PNG will be smaller than UASTC/zstd. More likely it will be larger. In one experiment I did I took a PNG file, inflated it then deflated it with zstd. The result was smaller than the PNG file, despite lacking any of the scanline adjustments PNG does to improve LZ efficiency. UASTC files will certainly be smaller than the inflated PNG equivalent, giving a better starting point for zstd.

zeux:
> I really feel like this is a problem in KTX2 that should be addressed

Having said this I tried to quickly build zstd's single-file decoder a few weeks ago (available in contrib repo) and the resulting wasm binary was much smaller than 900KB. I didn't try running the resulting file so I'm now doubting my results - how was that estimate obtained?

MarkCallow:
> Early estimates on ZSTD decoder put it at ~890kb

What I said was that the size of libzstd.a (release version) is ~890kb. This will have symbols and I built with the default options which means it has support for older versions included. I have not yet made any attempt to minimize it.

Rich is of the opinion that Zstd will provide much better results with UASTC than Deflate. We don't want to support a large number of supercompression options so we chose zstd. We absolutely need some supercompression with UASTC.

zeux:
This might be relevant, this is using single-slice ETC1S for all textures so naturally the quality of ETC1S is inferior. These are the results:

```
jrat: 12 2kx2k textures (64M pixels, including mipmaps)
etc1s: encode 67s, .glb 4.9 MB, .glb.gz 4.8 MB, .glb.zst 4.8 MB
uastc: encode 54s, .glb 67.5 MB, .glb.gz 37.0 MB, .glb.zst 33.5 MB
uastc rdo: encode 634s, .glk 67.5 MB, .glb.gz 26.6 MB, .glb.zst 24.1 MB
```

So maybe zstd is fine to include. The decoder I've built is 25 KB WASM (10 KB post-deflate), so it's possibly insignificant compared to the gains.

lexaknyazev:
>  in contexts where deflate is already available

I like this approach for UASTC. On the web, HTTP layer can provide gzip/deflate without any JS intervention. KTX2 header would indicate no supercompression in such a case. Generic clients should still support zstd though.

MarkCallow:
> On the web, HTTP layer can provide gzip/deflate without any JS intervention. KTX2 header would indicate no supercompression in such a case. Generic clients should still support zstd though.

This will definitely be an option for app developers. I plan to make selection of zstd compression a separate option from selection of UASTC in `toktx`. Before possibly mandating this approach for glTF extension(s) though we need to gain experience with UASTC/zstd vs UASTC/Deflate.

donmccurdy:
@zeux following up on https://github.com/KhronosGroup/glTF/pull/1751#issuecomment-611153791, is there a zstd decoder you would recommend for the web (js or wasm)?

zeux:
@donmccurdy I believe this was just building the single-file decoder from https://github.com/facebook/zstd/tree/dev/contrib/single_file_libs with Emscripten to wasm using -Os and standalone .wasm mode

MarkCallow:
The "single-file decoder" link is actually a script that combines the sources into a single file. I've already done that and use it in the web build of libktx. You can find it in lib/zstddeclib.c in the KTX-Software repo. On other builds we try to use the platform's libzstd.

This could easily be included in the web builds for msc_basis_transcoder. The missing piece for a JS KTX parser is a JS wrapper so the JS can call it. @zeux what is standalone .wasm mode?

zeux:
@MarkCallow https://github.com/emscripten-core/emscripten/wiki/WebAssembly-Standalone - this effectively means not using any Emscripten-generated JS code. It requires manually implementing a JS interface though; I mentioned this in context of size comparisons, meaning that JS code used for the JS->WASM interface may have additional cost that I wasn't counting.

donmccurdy:
Meta question.... is "KHR_texture_basisu" still the right name for this, with UASTC in the mix? Or is "basisu" generic enough, and BasisLZ / UASTC refer to the two subtypes?

MarkCallow:
Yes. Basis Universal is the umbrella name and covers both the current formats.

donmccurdy:
In the core specification, `mimeType` is optional when the image is not embedded in a bufferView, presumably because MIME type can be provided by the network layer.  I think it's unlikely that network layers will be able to provide accurate MIME type information for KTX 2.0 in the near future, simply because it is so new. Should we consider replacing this language...

> ...it's allowed to use value `image/ktx2` for the mimeType property of images 

... with wording that _requires_ the `mimeType` property for all KTX images (embedded or otherwise)?

donmccurdy:
Some implementation updates:

* Experimental three.js viewer: https://gltf-viewer-experimental.donmccurdy.com/
* Experimental glTF authoring tool: https://github.com/donmccurdy/glTF-Transform/pull/36

To try out the authoring tool, first [install KTX-Software](https://github.com/KhronosGroup/KTX-Software/releases), then run:

```shell
npm install --global @gltf-transform/cli@alpha

# show help
gltf-transform ktx --help

# compress normals with UASTC, everything else with ETC1S
gltf-transform ktx input.glb partial.glb --mode uastc --quality 0.5 --slots "normalTexture" --verbose
gltf-transform ktx partial.glb output.glb --mode etc1s --quality 0.5 --slots "*" --verbose
```

Note that the viewer does not support zstd yet: that significantly affects filesizes with UASTC. I recommend using the `--verbose` output for the CLI tool, to get more insight into the actual settings it's passing to `toktx`. I definitely haven't explored the range of options fully yet, only a 0–1 overall quality is exposed in the CLI so far.

lexaknyazev:
One more note wrt RGB normal maps. At runtime, engines should use ASTC or BC7 transcode targets or decode to RGBA8. Transcoding to other formats may substantially reduce the quality.

MarkCallow:
> I recommend using the --verbose output for the CLI tool, to get more insight into the actual settings it's passing to `toktx`.

Note that I have just made a change to `toktx` (& `ktxsc`) so that it adds a custom metadata item, KHRtoktxScParams, to any files it creates which will show the supercompression parameters used exactly as they were on the command line. It will be in the next beta release. I did this because I got fed up trying to encode in the names of my test files information about the compression parameters that were used. You can view the metadata with `ktxinfo`.

bghgary:
> In the core specification, mimeType is optional when the image is not embedded in a bufferView, presumably because MIME type can be provided by the network layer.

Not sure about this one. Babylon.js uses the file extension. I would imagine native implementations to use file extension or the magic number. I don't know of code that uses the info from the request, but I suppose that's possible.

lexaknyazev:
@MarkCallow 
I like the idea of putting the exact compression parameters (i.e. options that may cause loss of original data) in the file's metadata, let's put that in the spec. Naming nit: isn't "toktx" _within_ a KTX file a bit redundant?

@donmccurdy @bghgary 
> Should we consider replacing this language...
>> ...it's allowed to use value `image/ktx2` for the `mimeType` property of images

>... with wording that requires the `mimeType` property for all KTX images (embedded or otherwise)?

This is not strictly required for BasisU use-case because for now such images could be referred only from the texture extension object, so their MIME type could only be `image/ktx2`. OTOH, requiring this property for all KTX2 resources would simplify conformance rules and would future-proof use of KTX2 files for other purposes.

> Babylon.js uses the file extension. I would imagine native implementations to use file extension or the magic number.

File extensions for other resources (buffers and PNG/JPEG images) are not mandated by glTF spec, so we shouldn't rely too much on them here.

bghgary:
> File extensions for other resources (buffers and PNG/JPEG images) are not mandated by glTF spec, so we shouldn't rely too much on them here.

That's fair. What are the expectations when the file extension is missing in the core spec for png/jpg? Look at the magic number?

lexaknyazev:
> That's fair. What are the expectations when the file extension is missing in the core spec for png/jpg? Look at the magic number?

Yes, opened #1838 with a spec note. That's already standardized on the web.

donmccurdy:
>  Can alpha values be premultiplied? (It's possible to set this flag in KTX2).

Trying to finish up this question ... I think this should be clarified in the specification for each material texture property, not here. For `baseColorTexture` the spec already states, `The stored texels must not be premultiplied`. If it didn't, that would be a problematic lack of clarity even for PNG, which does not have a flag. For future material properties using the alpha channel (like sheen color + sheen roughness?) we should clarify similarly.

donmccurdy:
> Engines (at least 2 of three.js, Babylon, Cesium)

Implementations have landed in both three.js and Babylon.

donmccurdy:
I've published an authoring implementation of KHR_texture_basisu in [@gltf-transform/cli](https://gltf-transform.donmccurdy.com/cli.html):

```
# Compress normal textures with UASTC, everything else with ETC1S:
gltf-transform uastc input.glb output.glb --slots "normalTexture"
gltf-transform etc1s input.glb output.glb --slots "!normalTexture"
```

I'd recommend testing files in https://sandbox.babylonjs.com/ rather than my viewer, which is still on an experimental branch and lacks ZSTD decoding support.

donmccurdy:
@lexaknyazev I've edited your comment (https://github.com/KhronosGroup/glTF/pull/1751#issue-369078075) at the top of the PR with what I think would be the latest status on open questions. Please correct me if anything there seems incorrect.

As for zstd vs zlib vs file-level (arbitrary) compression... how do implementors feel about streaming mip levels individually from KTX files? Presumably that would require at least one compression option _within_ the KTX container?

lexaknyazev:
> Presumably that would require at least one compression option _within_ the KTX container?

Why? A KTX file without supercompression still allows streaming.

> Does KHR_texture_basisu allow zstd, zlib, or rely on file-level compression?

For more context, let's put all three options here (relevant only for UASTC payloads):

- KTX-level zlib/deflate
  - Almost free in Chrome (with [WICG Compression Streams](https://wicg.github.io/compression/)).
  - Other browsers require a bit more client code - e.g. https://github.com/nodeca/pako.
- KTX-level Zstandard
  - May provide different speed/ratio trade-offs compared to zlib.
  - All clients require extra code.
  - Reference C implementation is included in libktx.
- HTTP-level zlib/deflate (assuming that the server has it enabled)
  - Inconvenient for complex transmission/storage workflows.

donmccurdy:
> Why? A KTX file without supercompression still allows streaming.

Good point. Not required, then. But if you've gone to the trouble of implementing streaming, it's probably to make the initial render faster — wouldn't you also want supercompression in that scenario? Admittedly we (three.js) have not yet implemented this for any format, that I'm aware of.

donmccurdy:
For browsers that don't have WICG Compression Streams, the size of the client code is quite similar. 14kb for pako, 17kb for [zstddec](https://github.com/donmccurdy/zstddec). speed/ratio tradeoffs TBD.

lexaknyazev:
> wouldn't you also want supercompression in that scenario?

Assuming that [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams) is used instead of range-requests, compression could be provided on HTTP level as well.

donmccurdy:
Following up on today's discussion — @lexaknyazev could you draft changes to this PR for any zstd and zlib language that would be necessary for either option?

All others, we are looking for feedback on the tradeoffs of zstd, zlib, and arbitrary file-level compression for UASTC. Tradeoffs might include compressed file size, decoding time, or any other concerns. We would like to narrow things down to _at most_ allow one of the two compression methods — zstd or zlib — within the current glTF extension. An argument for omitting both could be made, since any compression method could be applied to a container .gltf, .glb, or .ktx2 file transmitted over the network.

zeux:
I feel like two good options are a) allow zstd as a compression method, b) allow neither.

zlib is a bad option [in my opinion] because:

- It doesn't add value compared to options that are ubiquitous on the web (if your CDN isn't capable of delivering compressed glb/gltf files you're losing on bandwidth big time anyhow)
- Compared to zstd, it provides inferior compression for KTX2 UASTC content (~5% delta between gzip -9 and zstd -9 on a random example)

So in my mind:

- Having no embedded compression options keeps the format simple - simple is good! - while allowing the use of zlib in effectively 100% of cases for web delivery, and allowing zstd when/if support for zstd compression on the web becomes widespread. Non-web consumers can integrate zstd (a fast general purpose compression layer like lz4/zstd over gltf/glb/bin is *always* a good idea).

- Having zstd as a compression option requires a zstd decoder which comes at a complexity cost, but transcoding is already really complex so arguably if we enforce ETC1S/UASTC we might as well enforce zstd, to get state of the art compression beyond what's available on the web

- Having zlib requires a zlib decoder which isn't drastically simpler than a zstd decoder, only to provide subpar compression that is already obtainable on the web for free through the transparent decompression via Accept-Encoding.

Caveat emptor, I haven't done analysis on the decoding time - I am assuming that even with the wasm penalty, in the browser context zstd is competitive wrt that compared to a browser-native zlib decoder, and transcoding time is going to dominate but I haven't timed this to back these claims up so take this with a grain of salt.

MarkCallow:
Measurements need to made with RDO optimized data. In one of my simple experiments, using an rdo level of 5 when encoding a UASTC image resulted in an almost 50% reduction in size of the zstd compressed file. It went from 26MB to 15MB (it was a 2048x2048 mipmapped cubemap - 72 images).

Other questions that need to be resolved are how does using external compression (compressing all the levels together along with the file header information) affect the result in terms of size and does external compression affect the efficacy of the RDO.

An issue with external compression is that it prevents streaming. You need the whole file before you can do anything.

zeux:
> An issue with external compression is that it prevents streaming. You need the whole file before you can do anything.

This is true if the external compressor doesn't support streaming, but I don't think it's true otherwise? Given the order of the mipmap levels in a KTX2 file you can load the texture progressively, including stopping at an arbitrary mip level (within limitations of the streaming decompressor, here the zlib and zstd mechanics is a bit different from what I know). You can't load random mip levels of course, so e.g. to load just the largest mip level you do need to decompress the entire pyramid.

lexaknyazev:
Official [Zstandard benchmarks](https://github.com/facebook/zstd#benchmarks) suggest that it's both faster and gives a better ratio. so embedded zlib is probably not that interesting. We need to confirm that, though.

donmccurdy:
Thanks @zeux, https://github.com/KhronosGroup/glTF/pull/1751#issuecomment-678596658 seems like a pretty solid argument that zstd is preferable to zlib. The zstd decoder is small (17kb WASM) so I don't see much reason to keep zlib as a candidate here.

@lexaknyazev do you know more about the size restriction issues in https://github.com/KhronosGroup/glTF/pull/1751#issuecomment-678593994? I think it's safe to say that libraries like three.js and Babylon.js will need to continue supporting WebGL 1 for quite a while, but I'm not necessarily arguing that means we should impose WebGL 1's restrictions on this extension.

MarkCallow:
Re the size restrictions, I think the right answer for WebGL 1 is to have an option for the transcoder to resample to power-of-2 when BC1[3] on WebGL 1 is the target. I asked Rich if this is possible as he had been planning to do the same for PVRTC targets. He says it should be. He hasn't implemented it for PVRTC due to lack of demand and was unaware of the WebGL 1 limitations until my question.

Due to D3D it may be necessary for this extensions to limit sizes to multiples of 4. Even though the byte length of the image will be a whole number of blocks, you can't just increase the dimensions in pixels because then sampling will pick up undefined colors at the edges of the larger image. It is unfortunate that MS chose to impose this restriction.

zeux:
The combination of limiting sizes to multiple of 4 in this extension + resampling to power of two during transcoding seems ideal. If encoders want to (gltfpack does :D) they are free to provide an offline resampling option if WebGL1 is the dominant target.

This requires the power of two restrictions to actually be relaxed in WebGL2 but I believe this is being worked on...

lexaknyazev:
> I think the right answer for WebGL 1 is to have an option for the transcoder to resample to power-of-2 when BC1[3] on WebGL 1 is the target.

OpenGL ES 2.0 (and therefore WebGL 1.0) does not seem to support mipmapping with npot textures:
> The main texture image has a level of detail number of 0 and is known as the _level zero array_ (or the image array of level zero). If `level` is greater than zero, and either `width` or `height` is not a power of two, the error `INVALID_VALUE` is generated.

Are we sure that such a restriction does not apply to ETC1, ASTC, and BC7 when running on WebGL 1.0?

MarkCallow:
> The main texture image has a level of detail number of 0 and is known as the level zero array (or the image array of level zero). If level is greater than zero, and either width or height is not a power of two, the error INVALID_VALUE is generated.

Where is this text? I've been unable to find it in OpenGL ES 2.0.25 but then search in PDF is pretty broken.

lexaknyazev:
> Where is this text?

Section 3.7.1, page 68 (78).

zeux:
Is there a concern that transcoding to a power of two effectively requires a different transcoding approach that's more expensive and requires a larger transcoder?

That is, transcoders today - maybe with an exception of PVRTC? - work on a block basis, reencoding each block into the appropriate representation. This sometimes requires extra work to figure out the optimal endpoints etc. but rarely (never?) requires decompressing and recompressing the block.

To do a full rescale you'd need to be able to decode UASTC & ETC1S into RGBA, rescale the RGBA image to fit to power of two, and reencode the resulting RGBA image to BCn / ETCn.

donmccurdy:
> To do a full rescale you'd need to be able to decode UASTC & ETC1S into RGBA, rescale the RGBA image to fit to power of two, and reencode the resulting RGBA image to BCn / ETCn.

At least for three.js, I don't think we'd want that behavior in a transcoder. Would prefer to decode to RGBA, perhaps at lower resolution, and use that directly on these devices without re-encoding.

lexaknyazev:
To sum up the today's call, it seems like we've got two options here:
- Require power-of-two dimensions for maximum compatibility.
- Require multiples of 4 and use slow path on WebGL 1.0 (ES 2.0) clients: decode to uncompressed, rescale, regenerate mipmaps, upload as uncompressed.

donmccurdy:
To restate the reasons for each option —

(1) requiring power-of-two dimensions avoids the slow path on WebGL 1.0 (ES 2.0)
(2) requiring multiples-of-four avoids the slow path on D3D hardware

Were there any other hardware cases we are concerned about? I see PVRTC mentioned a few times above — do one or both of those restrictions also apply to any hardware where PVRTC is the only transcode target?

To hopefully strike a balance of supporting modern hardware and not overly restricting future options, I'd be inclined to require (2), but simply include an implementation note about (1).

MarkCallow:
> do one or both of those restrictions also apply to any hardware where PVRTC is the only transcode target?

No. 1 but even Imagination considers PVRTC to be deprecated.

donmccurdy:
Proposals from today's call:

1. Allow zstd (not zlib) as the embedded compression option with UASTC, but do not require its use.
2. Require multiple-of-four dimensions, but not power-of-two. Requiring power-of-two would impose size/quality tradeoffs on modern hardware. Leaving it optional means less capable devices on WebGL 1.0 will sometimes have to fully decode the texture, but this can be avoided when necessary by opting into the power-of-two dimensions.

lexaknyazev:
Lifting the power-of-two restriction leads to an interesting question about required number of mip levels.

GLES 2.0 recap:
- npot textures do not support `REPEAT` or `MIRRORED_REPEAT` wrapping modes (Section 3.8.2);
- mipmapping requires power-of-two dimensions and a full mip pyramid (Sections 3.7.10 and 3.8.2);

This means that there's no reason to require a full mip pyramid for npot textures.

DCC would have two export options:
- Default (ES 3.0+, WebGL 2.0)
  - any non-zero dimensions, any number of mip levels
- Legacy
  - power-of-two textures with all mip levels present

Engines running on ES 2.0 / WebGL 1.0 would need to fall back to uncompressed textures in some cases.

donmccurdy:
> This means that there's no reason to require a full mip pyramid for npot textures.

Or WebGL 1.0 doesn't provide a reason, anyway... Are there other reasons to consider? In ES3.0+ and WebGL 2.0, we still can't generate the mip pyramid at runtime, so if engines need a full pyramid (even though their hardware does not) that may be an argument for requiring the full pyramid. At least for relevant minFilter/magFilter states.

zeux:
> Are there other reasons to consider?

In general the full pyramid should be strongly recommended for encoders; however, there are some fringe use cases where it's important to drop a few smallest mips (at a possible cost of some aliasing) to prevent blending from occurring. For example, in cases where multiple images are combined into an atlas, it may be important to remove all mip levels where different images are represented by the same pixel (or, alternatively, same block). Similarly, when packing lightmaps, to avoid bleeding you would typically need to include a 2^k-px border for k mips, so there's some tradeoff between aliasing vs wasted space in the texture.

So overall as long as ES2-class hardware isn't critical to support, it seems best to recommend the use of a full mipchain pyramid, but allow partial pyramids with an understanding that decoders will *not* generate one (short of ES2 where partial mipchain would be one of the reasons to fall back into slower RGBA decompression mode)

MarkCallow (approved):

emackey (approved):

extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -0,0 +1,173 @@
+# KHR_texture_basisu
+
+## Contributors
+
+* Members of 3D Formats Working Group
```
bghgary, line 5: I don't see this used anywhere else. Is this what we want to do in the future?

extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds the ability to specify textures using KTX v2 images with Basis Universal supercompression. An implementation of this extension can use such images as an alternative to the PNG or JPEG images available in glTF 2.0 for more efficient asset transmission and reducing GPU memory footprint. Furthermore, specifying mip map levels is possible.
+
+When the extension is used, it's allowed to use value `image/ktx2` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_basisu` texture extension object.
```
bghgary, line 20: Would it be useful to add links for Basis Universal and image/ktx2 in case people don't know what they are?

For Basis Universal, maybe the [GitHub page](https://github.com/BinomialLLC/basis_universal)?
For image/ktx2, maybe [this link](https://www.iana.org/assignments/media-types/image/ktx2)?

extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+When a texture referencing a KTX v2 image with Basis Universal supercompression is used for glTF 2.0 material maps (such as base color), the KTX v2 image must be of **2D** type as defined in the KTX v2 Specification, Section 4.1.
+
+Exporters should choose the most appropriate universal texture format based on the texture's usage and content. As a general rule, textures with color data should use ETC1S while textures with non-color data (such as roughness-metallic or normal maps) should use UASTC.
+
+When the high-quality GPU formats (ASTC or BC7) are not available at runtime, engines should transcode UASTC textures with non-color data to uncompressed RGBA to avoid artifacts.
+
+## Known Implementations
+
+None.
```
bghgary, line 167: Should we populate this?

bghgary (approved):
Some minor comments. Looks good otherwise.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
donmccurdy: Starting point in #1771

extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+Regardless of the format used, these additional restrictions apply for compatibility reasons:
+
+- Swizzling metadata (`KTXswizzle`) must be `rgba` or omitted.
+- Orientation metadata (`KTXorientation`) must be `rd` or omitted.
+- Colorspace information in the DFD must match the expected usage, namely:
+  - For textures with color data, `colorPrimaries` must be set to `KHR_DF_PRIMARIES_BT709`; `transferFunction` to `KHR_DF_TRANSFER_SRGB`.
+  - For textures with non-color data, `colorPrimaries` must be set to `KHR_DF_PRIMARIES_UNSPECIFIED`; `transferFunction` to `KHR_DF_TRANSFER_LINEAR`.
+- `KHR_DF_FLAG_ALPHA_PREMULTIPLIED` flag must not be set unless the material's specification requires premultiplied alpha.
+- When a texture refers to a sampler with mipmap minification or when the sampler is undefined, the KTX image should contain a full mip pyramid.
```
donmccurdy, line 152: Consider using uppercase for "MUST" and "SHOULD" throughout — I'm assuming you intend their [normative meanings](https://www.ietf.org/rfc/rfc2119.txt), but that would be easy for readers to miss.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+When the high-quality GPU formats (ASTC or BC7) are not available at runtime, engines should transcode UASTC textures with non-color data to uncompressed RGBA to avoid artifacts.
+
+## Known Implementations
+
+None.
+
+## Resources
+
+[KTX File Format Specification, version 2](https://github.khronos.org/KTX-Specification/)
```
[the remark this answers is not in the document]
MarkCallow: It should be 2.x so  just "v2" is fine. The detailed requirements that have just been added are enough to avoid any issues from new supercompression types or universal formats that may be added in a 2.X.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
lexaknyazev: Ah, I wrote that as a placeholder initially. @pjcozzi, how should we attribute this extension given its long history?

extensions/2.0/Khronos/KHR_texture_basisu/README.md
lexaknyazev: @donmccurdy 
Should we resolve that issue before adding anything here?

extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -0,0 +1,173 @@
...
+- Colorspace information in the DFD must match the expected usage, namely:
+  - For textures with color data, `colorPrimaries` must be set to `KHR_DF_PRIMARIES_BT709`; `transferFunction` to `KHR_DF_TRANSFER_SRGB`.
+  - For textures with non-color data, `colorPrimaries` must be set to `KHR_DF_PRIMARIES_UNSPECIFIED`; `transferFunction` to `KHR_DF_TRANSFER_LINEAR`.
+- `KHR_DF_FLAG_ALPHA_PREMULTIPLIED` flag must not be set unless the material's specification requires premultiplied alpha.
+- When a texture refers to a sampler with mipmap minification or when the sampler is undefined, the KTX image should contain a full mip pyramid.
+  > **Implementation Note**: Depending on the platform features available, engines may need to decompress the texture and generate missing mipmaps at runtime when a full mip pyramid is not available.
+- `pixelWidth` and `pixelHeight` must be multiples of 4.
+  > **Implementation Note**: Exporters may opt to make the dimensions powers of 2. Combined with all mip levels present this would ensure maximum compatibility.
+
+### Using KTX v2 Images with Basis Universal Supercompression for Material Textures
```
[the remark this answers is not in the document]
lexaknyazev: >  this extension might someday be referenced by an extension that is not related to material textures?

That was the original design. I'm not so sure now given that this extension became fully standalone (and any new extension could in theory override anything it needs).

I still think that support for swizzling and orientation metadata would heavily reduce portability regardless of texture's usage.

Our current materials don't use premultiplied alpha so I moved it up to the "generic" section (although the reference to material's spec remained - probably should be removed).

extensions/2.0/Khronos/KHR_texture_basisu/README.md
lexaknyazev: I don't think the main spec uses that style although it would be great to adopt it. Do you think it's fine to start with this extension?

extensions/2.0/Khronos/KHR_texture_basisu/README.md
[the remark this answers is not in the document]
MarkCallow: > I still think that support for swizzling and orientation metadata would heavily reduce portability regardless of texture's usage.

Why? Because some engines might ignore this metadata?

extensions/2.0/Khronos/KHR_texture_basisu/README.md
donmccurdy: Most appearances of "should" in the main spec are in implementation notes (based on a quick scan anyway) and so are fairly clear even if the reader isn't looking for normative language. This bulleted list would be easier to misinterpret, I think, and needs either:

(a) "Implementation Note" language on optional items, or  
(b) "SHOULD" language for optional items.

I think I'd vote for (b), even though we haven't done so in the main spec. This could be changed later, but more likely, we may even want to adopt it in the main spec.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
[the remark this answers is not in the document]
donmccurdy: > Our current materials don't use premultiplied alpha so I moved it up to the "generic" section ...

Current materials explicitly prohibit premultiplied alpha ("stored texels must not be premultiplied"), so it might be worth including an implementation note that premultiplied alpha is prohibited by core spec materials? That is easy to miss.

> I'm not so sure now given that this extension became fully standalone (and any new extension could in theory override anything it needs).

Good point... maybe I'm overthinking it. I would still move the premultiplied alpha item into the material's section, because it explicitly mentions "unless the material's specification..." the others, I think, are fine.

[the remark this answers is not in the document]
lexaknyazev: @MarkCallow
From the KTX spec:
> Where possible, the preferred default is to use the logical upper-left corner of the image as the texture origin.

glTF core spec also uses `rd` image orientation for JPEG/PNG textures. I don't think that we should diverge from that. Arbitrary swizzling would require either patching shaders on the fly or adding extra uniforms because API-level swizzling is not universally available,

extensions/2.0/Khronos/KHR_texture_basisu/README.md
pjcozzi: Would be great to explicitly name those that have contributed significantly and then remove this catchall.  When in doubt, let's be generous with credit.

extensions/2.0/Khronos/KHR_texture_basisu/README.md
donmccurdy: I think that issue is going to track a lot more things than we want to list here, and ultimately the list should end up on the glTF Project Explorer.

In the meantime, I think I would just list 1-2 authoring tools and 1-2 client tools in the extension, which we can do now:

Authoring:

- [gltfpack](https://github.com/zeux/meshoptimizer)
- [glTF-Transform CLI](https://gltf-transform.donmccurdy.com/cli.html)

Viewing: 

- [Babylon](https://www.babylonjs.com/)
- [three.js](https://threejs.org/)

Obviously libktx and Universal-Texture-Transcoders are important for building an implementation, but we can't really claim that they "implement" this extension.  Better to follow up with a separate developer guide?

extensions/2.0/Khronos/KHR_texture_basisu/README.md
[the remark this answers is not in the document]
donmccurdy: In contexts referring to 2.0, do we plan to write KTX v2.0 or KTX 2.0? We should try to be consistent here with KTX v2 or KTX 2.

Maybe it's just me, but "v2" feels ambiguous in a way that 2.x would not... it might be worth specifying? But either way.

donmccurdy (approved):

emackey:
I took a stab at adding the "Contributors" section.  Please let me know if I left anyone out, folks can always be added there without resetting the clock on ratification.

I'm merging this since the vote passed and it's entering ratification review.

PR #1753, opened 2020-02-01
add python library pygltflib to README.md

dodgyville:
Hi, I have a python library for GLTF2 that people seem to like using, so thought I'd add it to your list. Hope it's useful and thanks for creating such an accessible and open format!


emackey:
Thanks!

PR #1740, opened 2020-01-17
KHR_materials_clearcoat extension branch.

UX3D-nopper:

PR #1756, opened 2020-02-07
KHR_materials_clearcoat extension

emackey:
This is a continuation of #1740.  I wasn't able to push updates to UX3D's repo, so I've moved the branch here to the main repo.

The changes here add an Overview and remove the TODOs.  @abwood and I also fine-tuned some descriptions of parameters, just for clarification (it will not require any changes to existing implementations).

Alex and I also considered @proog128 comments in https://github.com/KhronosGroup/glTF/pull/1740#issuecomment-577724563, but until the Appendix B changes are ready to merge, it does not yet make sense to reference them here.  Once Appendix B is updated, the descriptions here that reference Appendix B will also need updating as described in the comment.


lexaknyazev:
Nit: the spec should mention what is the normal map for clearcoat layer when `clearcoatNormalTexture` is undefined.

MiiBond:
I'm sure that this was probably discussed a long time ago but there are other properties that sometimes appear in clear coat parameterizations like anisotropy, ior and tint. I certainly understand the desire to keep our extension simple, both for ease of implementation and for cross-renderer compatibility but I wonder if we're limiting this extension a bit too much. Tint, for example, would be very easy for renders to support.
Ultimately though, it's usage that should drive our decision so perhaps we should reach out to, say, the 3D Commerce group and see what their interest in clear coat tinting is. If there are virtually no real-world products/assets that need tint, then perhaps we can confidently disregard it.

emackey:
@lexaknyazev Done.

@MiiBond Thanks.  I'm concerned about feature-creep here.  I'd like to keep this extension compatible with existing renderers that have a simple clear-coat function.  Perhaps there could be a more advanced `KHR_materials_coating` or `KHR_materials_layered` type of extension to handle these more exotic terms?  As it stands, this extension is relatively low-hanging fruit, and already has working implementations in BabylonJS and other engines.

MiiBond:
Yeah, that's totally fair. An additional extension at a later date is probably the best way to handle additional parameters that are in use by some runtimes and not others.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -0,0 +1,92 @@
...
+```
+clearcoatRoughness = clearcoatRoughnessFactor * clearcoatRoughnessTexture.g
+```
+
+The following abstract code describes how the base and clearcoat layers should be blended together:
+  
+```
+clearcoatWeight = clearcoatTexture.r * clearcoatFactor * fresnel(0.04, NdotV)
+
+f = (f_emissive + f_diffuse) * (1.0 - clearcoatWeight) + mix(f_specular, f_clearcoat, clearcoatWeight)
```
proog128: There was a comment by @emackey in the old pull request (#1766) that suggested to rewrite to

```
f = mix(f_emissive + f_diffuse + f_specular, f_clearcoat, blendFactor)
```

It doesn't change the behavior, but it explains much better what actually happens: based on a Fresnel blending factor, the clear coat is applied on top of emission, diffuse, and specular component.

emackey: Yes, this comment: https://github.com/KhronosGroup/glTF/pull/1677#pullrequestreview-333608693

I'll apply the change, thanks @proog128.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension defines a clear coating that can be layered on top of an existing glTF material definition.  A clear coat is a common technique used in Physically-Based Rendering to represent a protective layer applied to a base material.  See [Theory, Documentation and Implementations](#theory-documentation-and-implementations)
+
+## Extending Materials
+
+The PBR clearcoat materials are defined by adding the `KHR_materials_clearcoat` extension to any glTF material.
```
donmccurdy: The phrase "`to any glTF material`" is probably more open-ended than it should be. Is it reasonable to specify interaction with other material extensions (`KHR_materials_unlit`, `KHR_materials_pbrSpecularGlossiness`) here, or what would be the right place? 

My preference would be that neither the unlit nor spec/gloss extensions be used with clearcoat. The first is fairly obvious. The second is more opinionated, but (in my opinion) the overall goals of [advanced PBR](https://github.com/KhronosGroup/glTF/milestone/2) in glTF would be better served by focusing on a single PBR workflow. Clearcoat — and future advanced PBR extensions — should gradually obviate `KHR_materials_pbrSpecularGlossiness`. 

Whether that means normatively disallowing use of the two extensions together, or just informal guidance, I'm not sure.

emackey: I made some updates, how does it look?

donmccurdy: The change matches my preference, thanks. If others are concerned about this choice, please chime in!

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+The following abstract code describes how the base and clearcoat layers should be blended together:
+  
+```
+clearcoatWeight = clearcoatTexture.r * clearcoatFactor * fresnel(0.04, NdotV)
+
+f = mix(f_emissive + f_diffuse + f_specular, f_clearcoat, clearcoatWeight)
+```
+
+If `clearcoatFactor = 0`, the clearcoat layer is disabled.
```
abwood: Minor correction: replace `clearcoatFactor` here with `clearcoatWeight`.

emackey: `clearcoatWeight` can differ per-pixel.  But if `clearcoatFactor` is zero in the JSON, then the entire extension can be considered disabled.

donmccurdy:
Draft implementation in Blender: https://github.com/KhronosGroup/glTF-Blender-IO/pull/931

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -0,0 +1,96 @@
...
+|**clearcoatTexture**              | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)             | The clearcoat layer intensity texture. | No                   |
+|**clearcoatRoughnessFactor**      | `number`                                                                        | The clearcoat layer roughness.         | No, default: `0.0`   |
+|**clearcoatRoughnessTexture**     | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)             | The clearcoat layer roughness texture. | No                   |
+|**clearcoatNormalTexture**        | [`normalTextureInfo`](/specification/2.0/README.md#reference-normaltextureinfo) | The clearcoat normal map texture.      | No                   |
+  
+The clearcoat formula `f_clearcoat` is computed using the specular term from the glTF 2.0 Metallic-Roughness material, defined in [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation).  Use F0 equal `0.04`, base color white `1.0, 1.0, 1.0`, metallic value `0.0`, and the clearcoat roughness value defined in this extension as follows:
+
+```
+clearcoatRoughness = clearcoatRoughnessFactor * clearcoatRoughnessTexture.g
```
donmccurdy: Given the [challenges](https://github.com/KhronosGroup/glTF/issues/1682) of compressed textures, do we want to continue requiring that clearcoat be packed into `.r` and clearcoat roughness into `.g`? I'm all for the memory savings, but we'll need a workaround in `KHR_texture_basisu` almost right out of the gate, and we haven't established whether that adjustment belongs in this extension or `KHR_texture_basisu`.

Requiring `.r` and `.a` would be one option. Throwing up our hands and dealing with this packing stuff once swizzle is generally available is another, and increasingly tempting. 😞

Finally – I'd missed this line on the initial readthroughs. I would consider mentioning the packing in the table above, in addition to the sample code.

lexaknyazev: What's wrong with using RG? As of today, BC5 is available in Chrome Canary on virtually all desktops.

donmccurdy: Great news, but that's the whole conversation around https://github.com/KhronosGroup/glTF/issues/1682#issuecomment-581608529, isn't it? We can transcode two independent ETC1S slices to `xy01` on desktop, but not on mobile, which would still require `xxxy`?

lexaknyazev: On mobile, two ETC1S slices are used as two ETC1 (two PVRTC on iOS for now) textures or one EAC RG11. There's no need to do any swizzling there.

donmccurdy: Ah I see. We can use two PVRTC textures here because implementations already have to support separated clearcoatTexture/clearcoatRoughnessTexture, whereas loading a normal map's channels from two PVRTC textures would be a new requirement on implementations.

That's fine with me, then.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+The following abstract code describes how the base and clearcoat layers should be blended together:
+  
+```
+clearcoatWeight = clearcoatTexture.r * clearcoatFactor * fresnel(0.04, NdotV)
```
emackey: During code review, it was found I was passing `NdotV` into a `fresnel` function that wants to take `VdotH` as a parameter.  Which is correct?  It looks like the paper says `VdotH` goes here, not `NdotV`.

@UX3D-nopper Do you know where this term comes from?

emackey: /cc @donmccurdy @bghgary @sebavan

UX3D-nopper: Good, that you review this.
To answer your question, NdotV is used e.g. in Subatsance Painter.
Furthermore, I would prefer following formula:
`color = (f_emissive + f_diffuse) * (1.0 - clearcoatBlendFactor) + mix(f_specular, f_clearcoat, clearcoatBlendFactor);`
Which paper are you refering to?

UX3D-nopper: In eevee from Blender, they also use NdotV (see lit_surface_frag.glsl).

UX3D-nopper: E.g. if you compare with Enterprise PBR:
https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2020x.md.html#components/clearcoat
They are using VdotH but then you also need to blend with NdotV and NdotL. In this model, one has to calculate three fresnel terms. As far as I can remember, this was considered to be too expensive. This is the reason, for real-time rendering, NdotL is used in most cases for approximation. However, the blending currently defined is not used in Blender eevee and/or Substance Painter.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+f = mix(f_emissive + f_diffuse + f_specular, f_clearcoat, clearcoatWeight)
```
emackey: Hi Norbert,

Here's the original formula (note that "clearcoatBlendFactor" was renamed "clearcoatWeight" in this PR for clarity.  I'm open to renaming it back if needed.)

```
f = (f_emissive + f_diffuse) * (1.0 - clearcoatWeight) +
    mix(f_specular, f_clearcoat, clearcoatWeight)
```

But `mix` is not a black box, let's expand out what `mix` does, and separate emissive/diffuse too:

```
f = f_emissive * (1.0 - clearcoatWeight) +
    f_diffuse * (1.0 - clearcoatWeight) +
    f_specular * (1.0 - clearcoatWeight) +
    f_clearcoat * clearcoatWeight
```

From here we can see that the base layer's emissive, diffuse, and specular are all equally reduced by the clear coat.

```
f = (f_emissive + f_diffuse + f_specular) * (1.0 - clearcoatWeight) +
    f_clearcoat * clearcoatWeight
```

By now it should be clear that `mix` can handle all of this in one go.

```
f = mix(f_emissive + f_diffuse + f_specular, f_clearcoat, clearcoatWeight)
```

This is the same formula, just optimized a bit.  It yields numerically equivalent answers.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
emackey: The reason I'm asking about `NdotV` here, is because in [Appendix B](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#appendix-b-brdf-implementation) we specify `VdotH` as the parameter for the main Fresnel Schlick (F) for the core material, and we're referencing that appendix here.  I tried rendering with `VdotH` and the results look very similar, actually a little softer on the coating and more of the base layer visible, with less "glowing" at the edges.  I guess I'm OK with either one, I just wanted to double-check this since we do site Appendix B and then promptly use a different term from the one given there.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
UX3D-nopper: lol ... yes u are right

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
emackey: OK, keeping `NdotV` makes sense.  For those using IBL exclusively, there won't be an `H` at all.

PR #1761, opened 2020-02-11
added IBL appendix

elalish:
As requested, here's a first draft of IBL best practices. Feedback welcome, though I'm heading out for vacation, so it'll be a couple weeks before I respond.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -96,7 +96,8 @@ Copyright (C) 2013-2017 The Khronos Group Inc. All Rights Reserved. glTF is a tr
 * [Appendix A: Tangent Space Recalculation](#appendix-a-tangent-space-recalculation)
 * [Appendix B: BRDF Implementation](#appendix-b-brdf-implementation)
 * [Appendix C: Spline Interpolation](#appendix-c-spline-interpolation)
-* [Appendix D: Full Khronos Copyright Statement](#appendix-d-full-khronos-copyright-statement)
+* [Appendix D: Image-based Lighting](#appendix-d-image-based-lighting)
+* [Appendix E: Full Khronos Copyright Statement](#appendix-e-full-khronos-copyright-statement)
 
 # Introduction
 
@@ -4001,7 +4002,17 @@ When the sampler targets a node's rotation property, the resulting ***p***(*t*) 
 
 > **Implementation Note:** The first in-tangent ***a***<sub>1</sub> and last out-tangent ***b***<sub>*n*</sub> should be zeros as they are not used in the spline calculations.
 
-# Appendix D: Full Khronos Copyright Statement
+# Appendix D: Image-based Lighting
+
+Image-based Lighting (IBL) is often used in conjunction with glTF to represent an environment or skybox and to create reflections. Whether this image is packed in the glTF via an extension or delivered on a side channel, it is important to have consistency in how it is interpreted. If represented in KTX2 as a cube map, its orientation is already specified, but if the more common equirectangular format is used, the following guidelines should be adhered to. 
+
+Equirectangular textures are aspect ratio 2, representing the 360 degrees of longitude by the 180 degrees of latitude. The center of the image is where the +X axis points, the top of the image is where the +Y axis points, and moving left in the image is equivalent to a positive rotation about the Y axis.
+
+When converting equirectangular textures to cube maps, it is best to sample an *n*x*n* cube map from a 4*n*x2*n* equirectangular image, e.g. a 1024x512 equirectangular images becomes a 256x256 cube map. 
+
+Note that unlike a skybox, if an image is meant for IBL, properly capturing the high dynamic range of most lighting environments is critical to realistic rendering. Therefore, it is advisable to use RGBE via the .hdr image format, or some equivalent method that ensures the image is not clamped even when the dynamic range is orders of magnitude beyond what an HDR TV is capable of. 
+
+# Appendix E: Full Khronos Copyright Statement
 
 Copyright 2013-2017 The Khronos Group Inc. 
 


specification/2.0/README.md
```diff
@@ -4001,7 +4002,17 @@ When the sampler targets a node's rotation property, the resulting ***p***(*t*)
...
+
+Equirectangular textures are aspect ratio 2, representing the 360 degrees of longitude by the 180 degrees of latitude. The center of the image is where the +X axis points, the top of the image is where the +Y axis points, and moving left in the image is equivalent to a positive rotation about the Y axis.
+
+When converting equirectangular textures to cube maps, it is best to sample an *n*x*n* cube map from a 4*n*x2*n* equirectangular image, e.g. a 1024x512 equirectangular images becomes a 256x256 cube map. 
+
+Note that unlike a skybox, if an image is meant for IBL, properly capturing the high dynamic range of most lighting environments is critical to realistic rendering. Therefore, it is advisable to use RGBE via the .hdr image format, or some equivalent method that ensures the image is not clamped even when the dynamic range is orders of magnitude beyond what an HDR TV is capable of.
```
lexaknyazev: "HDR TV" should be replaced with something more specific. Ideally, we should just mention expected value ranges.

UX3D-nopper:
Needs to be in synch with https://github.com/KhronosGroup/glTF-IBL-Sampler

specification/2.0/README.md
```diff
@@ -4001,7 +4002,17 @@ When the sampler targets a node's rotation property, the resulting ***p***(*t*)
 
 > **Implementation Note:** The first in-tangent ***a***<sub>1</sub> and last out-tangent ***b***<sub>*n*</sub> should be zeros as they are not used in the spline calculations.
 
-# Appendix D: Full Khronos Copyright Statement
+# Appendix D: Image-based Lighting
+
```
rsahlin, line 4006: I think we should consider putting this as normative in an extension while at the same time adding details with regards to how to handle specular and diffuse contribution (eg using spherical harmonics)
This is needed by the certification group in order to get consistent handling of the light setup.

- if that is not an option then I think it should be marked as non-normative (just like the BRDF appendix).

donmccurdy:
What do others think of making this a standalone document, similar to [3DC Asset Creation Guidelines](https://github.com/KhronosGroup/3DC-Asset-Creation/blob/main/asset-creation-guidelines/RealtimeAssetCreationGuidelines.md)?

specification/2.0/README.md
```diff
@@ -4001,7 +4002,17 @@ When the sampler targets a node's rotation property, the resulting ***p***(*t*)
...
+Image-based Lighting (IBL) is often used in conjunction with glTF to represent an environment or skybox and to create reflections. Whether this image is packed in the glTF via an extension or delivered on a side channel, it is important to have consistency in how it is interpreted. If represented in KTX2 as a cube map, its orientation is already specified, but if the more common equirectangular format is used, the following guidelines should be adhered to. 
+
+Equirectangular textures are aspect ratio 2, representing the 360 degrees of longitude by the 180 degrees of latitude. The center of the image is where the +X axis points, the top of the image is where the +Y axis points, and moving left in the image is equivalent to a positive rotation about the Y axis.
```
emackey, line 4009: Note that when glTF is imported into apps that use alternate coordinate systems, the axis names may swap around.  For example, glTF's +Y "up" axis becomes Blender's +Z "up" axis on import.  We should define HDR orientations in terms of "up" and "forward" such that appropriate axis name changes can take place for whatever app loads the glTF.

That said, I'm not sure that the main glTF spec is the correct place to specify HDR orientations at all, given that core glTF does not directly reference any HDR.  As Rickard points out, there could be a glTF extension with an HDR and a defined orientation for certification purposes.  But I don't think the core glTF spec can impose usage guidelines for host apps that may have already implemented HDR support before they implemented glTF support.

As you mention in the above paragraph, KTX2 includes orientation information, and I don't know that it needs any additional help from the glTF spec to be implemented correctly (beyond what this spec already says about "up" and "forward").

UX3D-nopper: I have attached a ZIP file, which contains the Blender file including the HDR map, a glb including the exported camera plus a screenshot made inside Blender. I am using the well known footprint_court HDR panorama image:
image
If you drag and drop the glb into e.g. the Babylon Sandbox (https://sandbox.babylonjs.com/) the reflections are the same as in Blender. Please note, that the IBL information is not stored in the glb and the reflections are the same, as the Babylon Sandbox is using the same environment map. Most of the current implementations are using this convention.
In this case, the center of the panorama image is +X side of the cube map. However, some DCC tools have e.g. +Z as the default setting. And this will probably cause problems, while editing in the DCC tool and later viewing the content in a glTF viewer.
[orientation.zip](https://github.com/KhronosGroup/glTF/files/5492687/orientation.zip)

echadwick-wayfair (approved):
I did a few tests inside 3ds Max 2020, with different bitmap loaders, just for information purposes.

Here are the cube map orientations overlaid on the HDR, for reference.
2020-11-05_latlong_cubemap_directions


1. Bitmap node, the default Autodesk loader. The bitmap has to be flipped horizontally for the correct orientation (text is readable). The front is +Z.
2020-11-05_Orientation_3dsMax2020_BitmapNode


2. VRayHDRI node, the default V-Ray loader. The bitmap has to be flipped horizontally for the correct orientation (text is readable). The front is +X.
2020-11-05_Orientation_3dsMax2020_VRayHDRINode


3. OSL Bitmap Lookup node. No horizontal flipping needed. The front is +Z.
2020-11-05_Orientation_3dsMax2020_OSLBitmapNode

UX3D-nopper:
@echadwick-wayfair Thanks for the evaluation. So in general we do have +X and +Z.
@elalish What do you guys think?
Should we accept the pull request and add a best practice for these tools? Then +X would be the default, correct?

elalish:
Yes, and there's no doubt that there's discrepancy between tools, but this
seems like a good place to start cleaning that up. We also had that thread
this summer where I think 3/4 tools used +X as the middle, so I think
that's a good place to start. It's also nice in the odd case you have a
seam that doesn't quite match up perfectly, since it will be visible
neither in the background nor the central reflection since +X puts the seam
on the side.

I'd still like to know if people would like to merge this PR here or
somewhere else. I don't have a strong preference, so long as developers
notice it.

-Emmett

emackey:
Are we taking "+X" "+Y" and "+Z" to be in the glTF reference frame, I presume?  In other words, +Y is up and +Z is forward?

We need to be clear.  These axis letter names get changed to and from the glTF coordinate system when entering and leaving other tools.  If you just say +X without specifying a reference frame, this will be a mess.

elalish:
Yes, I forgot to include my agreement there. I will change it to front, up,
etc. That makes sense.

-Emmett

UX3D-nopper:
Found on Twitter, which gives implicitly a nice overview:
IMG_20201109_060627

echadwick-wayfair:
I made a quickie diagram to help me understand handedness a bit better. 
2020-11-10 11_23 handedness

Hand model by 3dhaupt 
https://free3d.com/3d-model/rigged-hands-28855.html

donmccurdy:
I'm not sure this information should be in an appendix to the glTF specification; how about adding it to https://github.com/KhronosGroup/3D-Formats-Guidelines?

elalish:
> I'm not sure this information should be in an appendix to the glTF specification; how about adding it to https://github.com/KhronosGroup/3D-Formats-Guidelines?

Okay, so put this in an `IBLDeveloperGuide.md` file in that repo? Works for me.

emackey:
Reading back through this old PR, it sounds like if this were still desired, it would need to be re-done with different wording, in a different document. So, this old PR does not need to remain open here.  If someone wishes to take this up, feel free to open a new PR with a new draft.

PR #1762, opened 2020-02-12
Update extension status table.

donmccurdy:
Updates the various PBR next extensions (https://github.com/KhronosGroup/glTF/milestone/2), and removes extensions we aren't currently working on.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -30,20 +30,18 @@ but may still change before they are complete._
 
 | Extension | Status |
 |-----------|--------|
-| [KHR_texture_basisu](https://github.com/KhronosGroup/glTF/pull/1751) | Ready for implementation and feedback. |
+| [KHR_mesh_instancing](https://github.com/KhronosGroup/glTF/pull/1691) | Early discussions. |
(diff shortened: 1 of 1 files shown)


zeux:
Should the "tabled" status be reflected in the extension files for each individual extension, to discourage implementation?

donmccurdy:
I'll close the `KHR_blend` PR (https://github.com/KhronosGroup/glTF/pull/1302). I'm not sure what to do with `KHR_techniques_webgl`... it's already merged, but hasn't been ratified, and I don't think we're prioritizing it at the moment. I'm inclined to say we should delete the folder (which is still in version history of course), and create a new PR when and if appropriate.

emackey:
I think Cesium uses `KHR_techniques_webgl` internally, as implementation details for accepting both glTF 1.0 and 2.0 (IIRC, the 1.0 models are upgraded to 2.0 with `KHR_techniques_webgl`).  But I don't think the extension needs to stay published in an un-ratified form, maybe it could be parked on a branch instead of master.

Anyway this README update looks great, thanks @donmccurdy.  Merging.

PR #1775, opened 2020-03-02
Fix for clearcoat blending

UX3D-nopper:


emackey:
@romainguy @sebavan @proog128 I've been following the email discussion.  Are you guys OK with this PR?

emackey:
I'm trying to get an A/B screenshot of the difference.  I'm doing something wrong though because all of my clearcoat is washed out & overexposed now...

emackey:
So in my implementation, `f_clearcoat` did not already contain a Fresnel term, and removing it from the formula like this has obvious bad effects.

Also it looks like the sample clearcoat shader in Substance Painter matches what's in master here, not this current change.

Still investigating...

romainguy:
We used a squared Fresnel attenuation for a while in Filament as an attempt to account for loss of energy when entering and exiting the clear coat layer. I believe Unity HDRP does this still. May Substance does the same?

emackey:
It's not square though.  There's no Fresnel term on the IBL being reflected in `f_clearcoat`, as far as I can tell.

There is a Fresnel term on a punctual light reflected in clearcoat, but that's `Fresnel(0.04, VdotH)` for `H` that comes from the light source in question.  That's a very different term from the `Fresnel(0.04, NdotV)` that is used for blending.

For me, this PR makes IBL on the coat come in at full strength, with no Fresnel effect at all.

sebavan:
Is the formula not implying it should be done for both analytical and ibl the same ?

In Babylon we actually use the exact same in IBL :

```
        // IBL clear coat energy conservation
        float fresnelIBLClearCoat = fresnelSchlickGGX(clearCoatNdotV, clearCoatIOR);
        fresnelIBLClearCoat *= clearCoatIntensity;

        float conservationFactor = (1. - fresnelIBLClearCoat);

        environmentDiffuse *= conservationFactor;
        environmentSpecular *= conservationFactor;
```

sebavan:
but as a small diff, we do not apply it to emissive that we consider taking over everything else

emackey:
I found the difference!  My own code was following the Substance Painter example, which uses a clearcoat specular color of `1.0` instead of `0.04` as this extension calls for.  Oddly, pairing this much brighter specular color with the additional Fresnel term yields very similar results to using the darker specular color without the Fresnel term.

emackey:
I think I'm good with this PR.  Let's leave it open until @proog128 and others have had a day to process and comment.

proog128:
Just to make sure: `f_clearcoat` includes the Fresnel term, right? So it's

```
f_clearcoat = fresnel(0.04, VdotH) * G * D / (4 * NdotL * NdotV)

coated = f_base * (1 - clearcoatTexture.r * clearcoatFactor * fresnel(0.04, NdotV)) + f_clearcoat * clearcoatTexture.r * clearcoatFactor
```

This will produce some energy (because the base layer Fresnel is computed with VdotH), but the general structure is OK (matches [this comment](https://github.com/KhronosGroup/glTF/pull/1740#issuecomment-577724563)). Alternatives:
* Compute `f_clearcoat` with NdotV in the Fresnel. This is energy conserving, but less "correct", because the Fresnel should act on the facets. (Although the error is not high, because the facet orientation for small roughness values does not deviate much from the macro-surface normal).
* Apply Fresnel two times for base layer: max(fresnel(0.04, NdotV), fresnel(0.04, NdotL)). This is energy conserving, but loses energy at grazing angles for high roughness (not common for clearcoat).This is used in Enterprise PBR and MDL. Instead of max(...) you could also use (1-fresnel(0.04, NdotV)) * (1-fresnel(0.04, NdotL)), but loses even more energy. This is used in AxF.
* Albedo-scaling technique as in the sheen BRDF (#1688).
* More fancy techniques: [1](https://www.cg.tuwien.ac.at/research/publications/2007/weidlich_2007_almfs/) [2](http://www.cs.cornell.edu/projects/layered-sg14/), ...

emackey:
Yes.  I wonder if the PR needs to capture some of this discussion, particularly that `f_clearcoat` is expected to already have the Fresnel in it.

emackey:
This has enough approvals though, so I'll merge it and make some additional tweaks in a new PR.

emackey:
Small follow-up in #1776.

PR #1776, opened 2020-03-03
Tweak clearcoat wording & formatting a bit.

emackey:
Fixed a word wrap issue, and tried to clarify the essence of what prompted the recent formula change.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
@@ -68,8 +68,6 @@ The clearcoat formula `f_clearcoat` is computed using the specular term from the
 clearcoatRoughness = clearcoatRoughnessFactor * clearcoatRoughnessTexture.g
 ```
 
-The resulting `f_clearcoat` color is expected to have a Fresnel term blended into it as a result of the above.
-
 The following abstract code describes how the base and clearcoat layers should be blended together:
   
 ```
--- a/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
@@ -62,11 +62,13 @@ All implementations should use the same calculations for the BRDF inputs. Implem
 |**clearcoatRoughnessTexture**     | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)             | The clearcoat layer roughness texture. | No                   |
 |**clearcoatNormalTexture**        | [`normalTextureInfo`](/specification/2.0/README.md#reference-normaltextureinfo) | The clearcoat normal map texture.      | No                   |
   
-The clearcoat formula `f_clearcoat` is computed using the specular term from the glTF 2.0 Metallic-Roughness material, defined in [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation).  Use F0 equal `0.04`, base color white `1.0, 1.0, 1.0`, metallic value `0.0`, and the clearcoat roughness value defined in this extension as follows:
+The clearcoat formula `f_clearcoat` is computed using the specular term from the glTF 2.0 Metallic-Roughness material, defined in [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation).  Use specular F0 equal `0.04`, base color black `0.0, 0.0, 0.0`, metallic value `0.0`, and the clearcoat roughness value defined in this extension as follows:
 
 ```
 clearcoatRoughness = clearcoatRoughnessFactor * clearcoatRoughnessTexture.g
 ```
+
+The resulting `f_clearcoat` color is expected to have a Fresnel term blended into it as a result of the above.
 
 The following abstract code describes how the base and clearcoat layers should be blended together:
   
@@ -74,7 +76,8 @@ The following abstract code describes how the base and clearcoat layers should b
 clearcoatBlendFactor = clearcoatTexture.r * clearcoatFactor
 clearcoatFresnel = fresnel(0.04, NdotV)
 
-color = (f_emissive + f_diffuse + f_specular) * (1.0 - clearcoatBlendFactor * clearcoatFresnel) + f_clearcoat * clearcoatBlendFactor
+color = (f_emissive + f_diffuse + f_specular) * (1.0 - clearcoatBlendFactor * clearcoatFresnel) +
+        f_clearcoat * clearcoatBlendFactor
 ```
 
 If `clearcoatFactor` (in the extension) is zero, the whole clearcoat layer is disabled.


UX3D-nopper:
`The resulting `f_clearcoat` color is expected to have a Fresnel term blended into it as a result of the above.` is for me just confusing/obsolete. We are mentioning F, but not G or D or the dot products either.

emackey:
Is there some way we can reword this just to convey why we changed the formula?

This extension uses a different formula than Substance Painter, for example, and yet the results look very, very similar, nearly pixel-perfect similar in certain situations.  In SP clearcoat calls their `pbrComputeSpecular(...)` with a specular color of `vec3(1.0)` instead of our `vec3(0.04)`.  Then they use the Fresnel 0.04 term in a `mix` statement the way we had originally specified here.  It's just a question of where the Fresnel 0.04 comes into clearcoat coloring, either before mixing with the base layer, or during the mix.

During the email chain we had several people ask if `f_clearcoat` already included that term or not, and confusion over that led us to change the formula.  I think we need to say something to clarify that (and it sounds like I didn't accomplish that here, can someone suggest better wording please?)

UX3D-nopper:
Let's asks the others later today.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -62,19 +62,22 @@ All implementations should use the same calculations for the BRDF inputs. Implem
 |**clearcoatRoughnessTexture**     | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)             | The clearcoat layer roughness texture. | No                   |
 |**clearcoatNormalTexture**        | [`normalTextureInfo`](/specification/2.0/README.md#reference-normaltextureinfo) | The clearcoat normal map texture.      | No                   |
   
-The clearcoat formula `f_clearcoat` is computed using the specular term from the glTF 2.0 Metallic-Roughness material, defined in [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation).  Use F0 equal `0.04`, base color white `1.0, 1.0, 1.0`, metallic value `0.0`, and the clearcoat roughness value defined in this extension as follows:
+The clearcoat formula `f_clearcoat` is computed using the specular term from the glTF 2.0 Metallic-Roughness material, defined in [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation).  Use specular F0 equal `0.04`, base color black `0.0, 0.0, 0.0`, metallic value `0.0`, and the clearcoat roughness value defined in this extension as follows:
```
rsahlin: Are we looking for this F0 term from our PBR calculations?
F0 = lerp(dieletricSpecular, baseColor.rgb, metallic)

emackey: The clearcoat is non-metallic, so this extension is using the dielectricSpecular of 0.04.  The baseColor doesn't really factor into it, but in some implementations you must supply a color there to call the PBR functions.

rsahlin: Yes, so the lerp would just pick the dielectric color (currently 0.04) - most shaders would probably calculate it just once then access the value.
Maybe it makes sense to keep the terms the same for sake of being consistent?

emackey: Is this critical?  I'd like us to finish up clearcoat, it's been open since early 2019.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
```diff
@@ -62,19 +62,22 @@ All implementations should use the same calculations for the BRDF inputs. Implem
...
 
 ```
 clearcoatRoughness = clearcoatRoughnessFactor * clearcoatRoughnessTexture.g
 ```
 
+The resulting `f_clearcoat` color is expected to have a Fresnel term blended into it as a result of the above.
```
proog128: I am still confused about how `f_clearcoat` is now computed... Why not write it down explicitly? Something like:

> The clearcoat formula `f_clearcoat` is computed using the specular term from the glTF 2.0 Metallic-Roughness material, defined in [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation).
>
> ```
> f_clearcoat = Fc * G * D / (4 * NdotL * NdotV)
> ```
>
> with Fresnel for dielectrics and index of refraction of 1.5
>
> ```
> Fc = 0.04 + 0.96 * (1 - VdotH)^5
> ```
>
> The roughness value used in the G and D terms is computed from the `clearcoatRoughness`: `alpha = clearcoatRoughness^2`.

emackey: Doesn't this repeat formulas that are specified in Appendix B?  We already reference that appendix at least twice.

I'm not opposed to the idea, but I'm eager to move this proposal out of the "new ideas" phase and onto ratification.  There are several implementations already deployed.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
emackey: I'm not opposed to the idea, but I'm eager to move this proposal out of the "new ideas" phase and onto ratification.  There are several implementations already deployed.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
proog128: I don't have a strong opinion, I just thought that repeating the formula instead of text makes it more clear, avoids misunderstanding.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
rsahlin: Maybe enough with reference to [Appendix B] and:
`Use specular F0 with a metallic value `0.0` and the clearcoat roughness value defined in this extension as follows:`

On a different note I think we need to mark the section as non-normative - to be consistent with the information in Appendix B.

emackey: I'm not sure if part of the main formula can be non-normative.  But additional PRs are still welcome after Khronos ratification so long as they don't change the Intellectual Property, so, anyone can still feel free to open a new PR for this.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
emackey: I pulled out this sentence.  I had hoped to explain a long chain of emails with a quick note, but this clearly wasn't the correct way to do it.

emackey:
The offending sentence was removed, so this PR is mostly error correction and formatting changes now.  I'm merging it time for today's meeting.

As a reminder, new PRs for issues discussed here are still welcome after ratification so long as they don't change the IP.

extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
proog128: I am planning to put [this](https://github.com/KhronosGroup/glTF/pull/1740#issuecomment-577724563) into #1717 (if there is still interest). This separates the normative and non-normative parts. Whether we use the equations from Enterprise PBR or something else (or both) for the non-normative part is up to discussion.

emackey: Yes, I know Alex at least is still interested in #1717.

Issue #1743, opened 2020-01-20
Quality Level of Detail or Quality of glTF asset


raviramachandra:
W3C immersive-web group is planning to use glTF as an asset for Favicons, WebApp Manifest Icons on XR devices.  See this:  https://github.com/immersive-web/spatial-favicons/blob/master/explainer.md

To allow Browsers/Platforms to select from multiple glTF icons appropriately to the use case, a quality/LoD metric is needed. There are many questions pertaining to quality of the asset. 
1. Should this be a single weighted sum of individual quality vectors ?
2. Or should this be a set of metrics which together provide a quality measurement ?

This issue is opened to track this effort and derive at this.

rsahlin:
Hi @raviramachandra 

Have you looked at the Microsoft LOD extension?
https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/MSFT_lod

Does this cover your needs to store the actual LODs?

Could you please explain how you would use the quality vectors?
Do you mean to have a sum of the 'weight' of the model - as total number of nodes/vertices and textures?
 
From your usecase desciption I would guess that your browser/platform would choose the smaller LOD in case of bookmarks and use the heavier ones in other usecases - or am I missing something?

donmccurdy:
The phrase "LOD" might be misleading here — I don't think the W3C question is about storage of the LODs, which would likely be in entirely separate glTF files unlike the Microsoft extension. The difficult question is how choose the metric(s), expressing to the browser the quality of each available glTF file in a way that enables it to choose the appropriate version.

The Microsoft extension uses screen coverage as the single metric for their extension. Which really describes the context into which the model will be loaded, as all LODs presumably have the same dimensions. I don't know whether that metric would be right for W3C? Alone, it leaves no room for flexibility based on factors like filesize, network bandwidth, or device GPU performance. But it is similar to HTML `srcset`.

rsahlin:
Ah, thanks @donmccurdy 
This is also a setup we are looking into at Ikea.

I think we need to make distinction between, at least:
A: Metric for model download size (bytes) - `Download Metric`
B: Performance metric - eg in a very genaralized way describing how demanding the model is to render. - `Runtime Metric`

At least in our case the `Runtime Metric` would be used together with the `Download Metric` to determine which model is downloaded to a target (viewer).

@raviramachandra 
What's your view on these metrics with regards to what you want to achieve (your usecases)?

raviramachandra:
[...]

PR #1779, opened 2020-03-05
glTF performance metrics

rsahlin:
This is the inital commit to get discussion started.
The goal of this PR is to provide:
* Standalone performance metrics - not part of glTF spec
* Command line tool to produce performance metrics

Performance metrics can be used for instance by backend to choose the most suitable model to deliver to a client based on usecase and target capabilitites. 
Backends should have access to the necessary data so that a selection from different LoD's is possible.

The goal is not to provide one number for the complexity but rather to break down usage in terms of data, then provide this data without glTF dependencies.
It is up to the consumer to decide what the metrics mean.

This is non-normative and should not be part of glTF spec.

### DIFF
--- a/specification/2.0/appendix/performance-metrics/README.md
+++ b/specification/2.0/appendix/performance-metrics/README.md
@@ -1,10 +1,10 @@
-# Runtime Performance Metrics 
+# Runtime GLTF Metrics 
 
-The goal of these metrics is to assess the runtime performance needs of glTF assets.  
+The goal of these metrics is to assess the runtime needs of glTF assets.  
 By runtime we mean the cpu/gpu, bandwidth and memory resources needed to render the model in realtime.  
 These performance metrics are divided in two major categories - `complexity` and `memory`
 
-## Performance Metric Categories
+## Metric Categories
 
 ### Complexity  
 
@@ -60,7 +60,7 @@ It is possible to reference different meshes (nodes) using different scenes, thi
 Ie one model could have 3 different scenes, where the same accessors (position, uv) are used but different primitives that reference materials with varying number of texture channels.  
 
 [nodeCount]  
-This represent the number of traversed nodes in a scene.  
+This represent the total number of traversed nodes in a scene.  
 This is calculated by traversing nodes in each scene (depth or breadth first does not matter) - for each node increase the nodecount.  
 
 [drawCount]  
@@ -75,7 +75,9 @@ Each primitive references a material with 0 or more texture sources, has attribu
 [channels]  
 This value represents the texture source channels that are used by a scene.  
 This is the max value from the most 'complex' primitive that will be referenced by a Node in the scene.  
-The goal of this metric is to provide a worst case texture usage and processing cost for rendering a pixel.  
+The goal of this metric is to provide a worst case texture usage and processing cost for rendering a pixel. 
+Please note that the processing cost for the different channels will vary, for instance the EMISSIVE or  
+BASECOLOR channels are quite cheap but other channels may add a signification processing overhead.  
 
 **Note**  
 This value can not be used to know if texture channels are tightly packed according to uniform Sampler usage,  
--- a/specification/2.0/appendix/performance-metrics/README.md
+++ b/specification/2.0/appendix/performance-metrics/README.md
@@ -1,31 +1,42 @@
 
---- Name -----------Property    ----- Description ----------------------------
-[vertexcount]       Scene       Total number of vertices used by a model in a scene
-[nodecount]         Scene       Max nodecount in scene (add upp all nodes in a scene)
-[primitivecount]    Scene       Total number of referenced primitives (per scene)
-                                This figure is the un-batched number of primitives, engines may optimize if primitives and meshes share textures.
-[textures]          Scene       Flags specifying presence of materials and it's texture usage, this is the aggregated max usage
-                                BASECOLOR, METALLICROUGHNESS, NORMAL, OCCLUSION, EMISSIVE, (SPECULARGLOSS)
+|                      |                           |                                           | 
+|----------------------|---------------------------|-------------------------------------------|
+| Name |Property   |Description  |
+|[vertexcount]       |Scene       |Total number of vertices used by a model in a scene|  
+|[nodecount]         |Scene       |Max nodecount in scene (add upp all nodes in a scene)|  
+|[primitivecount]    |Scene       |Total number of referenced primitives (per scene).  This figure is the un-batched number of primitives, engines may optimize if primitives and meshes share textures.  |  
+|[textures]          |Scene       |Flags specifying presence of materials and it's texture usage, this is the aggregated max usage. BASECOLOR, METALLICROUGHNESS, NORMAL, OCCLUSION, EMISSIVE, (SPECULARGLOSS)  
 
+# Scene #
+Metrics that describe properties of a scene.  
+glTF models may contain multiple scenes, only one scene will be visible at a time.  
+To accomodate this the following metrics are calculated on a per-scene basis.  
+It is possible to reference different meshes (nodes) using different scenes, this makes it possible to have the same texture assets but reference different primitives that use alternative materials (textures).  
+Ie one model could have 3 different scenes, where the same accessors (position, uv) are used but different primitives that reference materials with varying number of texture channels.  
+
+[nodecount]  
+This represent the number of traversed nodes in a scene.  
+This is calculated by traversing nodes in each scene (depth or breadth first does not matter) - for each node increase the nodecount.  
+
+[verticecount]  
+This represents the number of drawn vertices for each scene in the model.  
+This is calculated by traversing the nodes in each scene.  
+For each mesh use the POSITION attribute in each primitive, get the Accessor and add up the count field.  
+
+[primitivecount]  
+This represents the number of primitives for each scene  
+Each primitive references a material with 0 or more texture sources, has attributes and accessors.  
+
+
+[textures]  
+This value represents the texture sources that are used by a scene.  
+This is the max value from the most 'complex' primitive that will be referenced by a Node in the scene.  
+The goal of this metric is to provide a worst case texture usage where texturecount and complexity is known.  
+Ie it is known if a texture is BASECOLOR or if part of PBR such as METALICROUGHNESS and now many texture sources that are used.  
+
+
+# JSON #  
+This is how the output would be formatted using JSON  
 
 "scene" : [ { "verticecount" : 4300, "nodecount" : 20, "primitivecount" : 50 ,
-              "textures" : ["BASECOLOR", "METALLICROUGHNESS"]} ]
-
-
-[nodecount]
-This represent the number of traversed nodes in a scene.
-This is calculated by traversing nodes in each scene (depth or breadth first does not matter) - for each node increase the nodecount.
-
-[verticecount]
-This represents the number of drawn vertices for each scene in the model.
-This is calculated by traversing the nodes in each scene.
-For each mesh use the POSITION attribute in each primitive, get the Accessor and add up the count field.
-
-[primitivecount]
-This represents the number of primitives for each scene
-
-[textures]
-This value represents the texture sources that are used by a scene.
-This is the max value from the most 'complex' primitive that will be referenced by a primitive in the scene.
-The goal of this metric is to provide a worst case texture usage where texturecount and complexity is known.
-Ie it is known if a texture is BASECOLOR or if part of PBR such as METALICROUGHNESS
+              "textures" : ["BASECOLOR", "METALLICROUGHNESS"]} ]  
--- a/specification/2.0/appendix/performance-metrics/README.md
+++ b/specification/2.0/appendix/performance-metrics/README.md
@@ -11,6 +11,11 @@ These performance metrics are divided in two major categories - `complexity` and
 The `complexity`category provides metrics to calculate cpu/gpu and memory bandwidth target requirements - larger numbers will mean a more powerful target device is needed.  
 These metrics do not provide a mapping to target device performance, this mapping must be done by users of the metrics.  
 
+### Screen space coverage
+
+The metrics do not include screen space coverage, it is advised that clients shall know how large the render area will be before requesting a model (from the backend)  
+It is up to the backend to take screen coverage into account when providing a suitable model to the client, for instance by taking width + height as parameters.  
+
 ### Memory  
 
 The `memory` category provides metrics to calculate the runtime memory (size) needs of the model.  
@@ -25,6 +30,7 @@ Using these metrics an estimate of target memory can be calculated - however kee
 |[nodeCount]      | Integer     | Complexity  |Scene     |Max nodecount in scene (add upp all nodes in a scene)           |  
 |[primitiveCount] | Integer     | Complexity  |Scene     |Total number of referenced primitives (per scene).  This figure is the un-batched number of primitives, engines may optimize if primitives and meshes share textures. |  
 |[textures]       | Integer     | Complexity  |Scene     |Flags specifying presence of materials and it's texture usage, this is the aggregated max usage. BASECOLOR, METALLICROUGHNESS, NORMAL, OCCLUSION, EMISSIVE, (SPECULARGLOSS) |  
+|[maxNodeDepth]   | Integer     | Memory      | Scene    | The max node depth of the scene, ie the max number of parent/child relations. This number will represent the max stack depth needed when traversing the nodegraph |
 |[attributes]     | Integer     | Memory      | Asset    | The total memory footprint, in bytes, for the models attribute buffer storage |  
 |[textureSizes]  |Dimension    | Memory      | Asset    | The width and height of defined textures in the Asset. |  
 
@@ -79,7 +85,8 @@ This is how the output would be formatted using JSON
             "vertexCount" : 4300,
             "nodeCount" : 20,
             "primitiveCount" : 50,
-            "textures" : ["BASECOLOR", "METALLICROUGHNESS"]
+            "textures" : ["BASECOLOR", "METALLICROUGHNESS"],
+            "maxNodeDepth" : 7
         }
     ],
     "asset" : {
--- a/specification/2.0/appendix/performance-metrics/README.md
+++ b/specification/2.0/appendix/performance-metrics/README.md
@@ -17,6 +17,14 @@ The metrics do not include screen space coverage.
 However, implementations should generally consider the size of the render area in addition to performance metrics of the model.  
 A higher performance budget may be acceptable when screen space coverage is higher.  
 For most usecases it's best to assume a worst case where all or most of the screen space is filled with rendered pixels.  
+Here `screen space coverage` is defined as the number of pixels rendered in the output framebuffer.  
+
+### Bounding volume
+
+A bounding volume, on each scene, is included as a measurement of the world space that the model occupies.  
+The intention of this is to be able to estimate the screen space size of the model in for instance AR/VR usecases.  
+glTF units are equal to metric meters [1.0 = 1.0m] meaning that using the bounding volume the client can estimate the portion of the visible display it will occupy.  
+A volume with dimensions [0.5,0.5,0.5] viewed at 50 cm will cover a large portion of the screen and a volume with dimensions [0.05,0.05,0.05] viewed at 50 cm will cover only a small portion.  
 
 
 ### Memory  
@@ -34,13 +42,14 @@ Using these metrics an estimate of target memory can be calculated - however kee
 |[primitiveCount] | Integer     | Complexity  |Scene     |Total number of referenced primitives (per scene).  This figure is the un-batched number of primitives, engines may optimize if primitives and meshes share textures. |  
 |[textures]       | Integer     | Complexity  |Scene     |Flags specifying presence of materials texture usage, this is the aggregated most complex usage. BASECOLOR, METALLICROUGHNESS, NORMAL, OCCLUSION, EMISSIVE, SPECULARGLOSS, DIFFUSE, CLEARCOAT |  
 |[materials]      | Integer     | Complexity  |Scene     | Flags specifying presence of materials, this is the aggregated most complex usage.  |
+|[bounds]         | Bounds      | Complexity  | Scene    | The static 3D bounds for each scene, can be used to estimate world space size of model. |
 |[maxNodeDepth]   | Integer     | Memory      | Scene    | The max node depth of the scene, ie the max number of parent/child relations. This number will represent the max stack depth needed when traversing the nodegraph |
 |[accessors]     | Accessor     | Memory      | Asset    | Total number and format of vertex accessors, this can be used to calculate vertex buffer memory requirements |  
 |[textureSize]  |Integer    | Memory      | Asset    | The size and format of textures |  
 
 Dimension is an Integer[2] containing width and height  
 Accessor is an object containing count, componentType and type, componentType and type is taken from glTF Accessor  
-
+Bounds is an object containing min and max values.    
 
 # Scene #  
 
@@ -87,6 +96,13 @@ The goal of this metric is to provide a worst case computational cost for render
 1 PBR
 2 SPECULARGLOSS (KHR_materials_pbrSpecularGlossiness)
 2 UNLIT (KHR_materials_unlit) 
+
+
+[bounds]  
+This value represents the static (non animated) bounding volume for the scene.  
+It is calculated by traversing the nodes in each scene, applying node transform to primitive's MinMax values and updating the result.  
+The resulting MinMax will contain the total transformed volume the model occupies, ie concatenated MinMax of all primitives as they are transformed.  
+This will include translation, meaning that the bounds will not be centered.  
 
 
 [accessors]  
@@ -138,6 +154,16 @@ This is how the output would be formatted using JSON
             "primitiveCount" : 50,
             "textures" : ["BASECOLOR", "METALLICROUGHNESS"],
             "materials" : ["PBR", "SPECULARGLOSS"],
+            "bounds" : "min": [
+                -0.9999999403953552,
+                -1.0,
+                -1.0
+             ],
+             "max": [
+                0.9999999403953552,
+                1.0,
+                1.0
+            ]
             "maxNodeDepth" : 7
         }
     ],
--- a/specification/2.0/appendix/performance-metrics/README.md
+++ b/specification/2.0/appendix/performance-metrics/README.md
@@ -1,7 +1,7 @@
 # Runtime Performance Metrics 
 
 The goal of these metrics is to asses the runtime performance needs of glTF assets.  
-By runtime we mean the cpu/gp, bandwitdh and memory resources needed to render the model in realtime.  
+By runtime we mean the cpu/gpu, bandwidth and memory resources needed to render the model in realtime.  
 These performance metrics are divided in two major categories - `complexity` and `memory`
 
 ## Performance Metric Categories
--- a/specification/2.0/appendix/performance-metrics/README.md
+++ b/specification/2.0/appendix/performance-metrics/README.md
@@ -96,7 +96,9 @@ This will include translation, meaning that the bounds will not be centered.
 
 
 [accessors]  
-This value represents the total vertex accessor usage of the asset, it is an indication of the memory requirements for the asset.  
+This value represents the total vertex accessors defined in the asset, it is an indication of the memory requirements for the asset.  
+These values does not take into account if an acessor is unused or only used in some scene.  
+
 Note that this value will not take into account if normal/tangent/bitangent values are defined or shall be runtime computed.  
 It is calculated by adding upp the accessors in the model, storing count, componentType and type.  
 This may give an indication of the runtime memory footprint of the buffers needed for the model.  
--- a/specification/2.0/appendix/performance-metrics/README.md
+++ b/specification/2.0/appendix/performance-metrics/README.md
@@ -1,13 +1,36 @@
+# Runtime Performance Metrics 
 
-|                      |                           |                                           | 
-|----------------------|---------------------------|-------------------------------------------|
-| Name |Property   |Description  |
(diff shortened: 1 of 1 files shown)


specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,31 @@
+
+--- Name -----------Property    ----- Description ----------------------------
+[verticecount]      Scene       Total number of vertices used by a model in a scene
```
emackey: It's `vertexcount`, not `verticecount`.  The singular form of "vertices" is "vertex."

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,31 @@
+
```
raviramachandra: Minor Comment: May be we should have a description saying: The goal of these metrics is to assess "Runtime" performance  of the asset.

rsahlin: Good point - I will add.
I also had some feedback from Don that I will add.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -1,13 +1,36 @@
...
 
-|                      |                           |                                           | 
-|----------------------|---------------------------|-------------------------------------------|
-| Name |Property   |Description  |
-|[vertexcount]       |Scene       |Total number of vertices used by a model in a scene|  
-|[nodecount]         |Scene       |Max nodecount in scene (add upp all nodes in a scene)|  
-|[primitivecount]    |Scene       |Total number of referenced primitives (per scene).  This figure is the un-batched number of primitives, engines may optimize if primitives and meshes share textures.  |  
-|[textures]          |Scene       |Flags specifying presence of materials and it's texture usage, this is the aggregated max usage. BASECOLOR, METALLICROUGHNESS, NORMAL, OCCLUSION, EMISSIVE, (SPECULARGLOSS)  
+The goal of these metrics is to asses the runtime performance needs of glTF assets.  
+By runtime we mean the cpu/gp, bandwitdh and memory resources needed to render the model in realtime.
```
raviramachandra: nit: bandwidth

raviramachandra: nit: gpu

rsahlin: Thanks - fixed.

rsahlin: Thanks - fixed.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -1,6 +1,6 @@
 
 --- Name -----------Property    ----- Description ----------------------------
-[verticecount]      Scene       Total number of vertices used by a model in a scene
+[vertexcount]       Scene       Total number of vertices used by a model in a scene
```
jeymichael: I believe this is intended to be the total number of vertices used by **all models** in a scene.
But it says `used by a model`.  Potentially confusing.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,77 @@
+# Runtime Performance Metrics 
+
+The goal of these metrics is to asses the runtime performance needs of glTF assets.
```
javagl: Nitpick: it's "to assess", not "to asses". Here, the second "s" is more important than in many other cases.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,77 @@
...
+glTF models may contain multiple scenes, only one scene will be visible at a time.  
+To accomodate this the following metrics are calculated on a per-scene basis.  
+It is possible to reference different meshes (nodes) using different scenes, this makes it possible to have the same texture assets but reference different primitives that use alternative materials (textures).  
+Ie one model could have 3 different scenes, where the same accessors (position, uv) are used but different primitives that reference materials with varying number of texture channels.  
+
+[nodecount]  
+This represent the number of traversed nodes in a scene.  
+This is calculated by traversing nodes in each scene (depth or breadth first does not matter) - for each node increase the nodecount.  
+
+[verticecount]
```
javagl: This should also be "vertexcount" (it was already fixed above). 

An aside: I think that glTF uses camelCase by convention (see https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#indices-and-names ). So the names should probably be `primitiveCount`, `nodeCount`, `vertexCount`, `textureSizes` etc.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,77 @@
...
+|-----------------|-------------|----------|----------------------------------------------------------------|
+| Name            | Category    | Property |Description                                                     |
+|[vertexcount]    | Complexity  |Scene     |Total number of vertices used by a model in a scene             |  
+|[nodecount]      | Complexity  |Scene     |Max nodecount in scene (add upp all nodes in a scene)           |  
+|[primitivecount] | Complexity  |Scene     |Total number of referenced primitives (per scene).  This figure is the un-batched number of primitives, engines may optimize if primitives and meshes share textures. |  
+|[textures]       | Complexity  |Scene     |Flags specifying presence of materials and it's texture usage, this is the aggregated max usage. BASECOLOR, METALLICROUGHNESS, NORMAL, OCCLUSION, EMISSIVE, (SPECULARGLOSS) |  
+|[attributes]     | Memory      | Asset    | The total memory footprint, in bytes, for the models attribute buffer storage |  
+|[texturesizes]  | Memory      | Asset    | The sizes of defined textures in the Asset. |  
+
```
javagl: Either the table or the JSON description below should indicate the *type* of the properties. E.g. `integer` or `number`.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,77 @@
...
+This value represents the texture sources that are used by a scene.  
+This is the max value from the most 'complex' primitive that will be referenced by a Node in the scene.  
+The goal of this metric is to provide a worst case texture usage where texturecount and complexity is known.  
+Ie it is known if a texture is BASECOLOR or if part of PBR such as METALICROUGHNESS and now many texture sources that are used.  
+
+[attributes]  
+This value represent the total attribute buffer usage of the model. This value is calculated by adding up the size, in bytes, of all buffer objects in the Asset.  
+This may give an indication of the runtime memory footprint of the buffers needed for the model.  
+
+[texturesizes]
```
javagl: This information is relevant for two aspects: On the one hand, the memory that will be occupied by the texture. But also whether the client can support the texture at all (i.e. whether the size is not larger than what `glGet(GL_MAX_TEXTURE_SIZE)` returns. 

In any case: The information in the example is given as a *string*, like `"128 * 100"`, so it has to be parsed in some way. People will have to rely on the structure for that - namely, the string *MUST* always have the format `<integer> * <integer>` - this could be worth mentioning. 

In some ways, it might be easier/cleaner to specify the sizes using two integer values, like 

    "textureSizes" : [
        [ 1024, 512 ],
        [ 512, 512 ],
        [ 128, 256 ],
     ],

or so, but that's just a thought...

specification/2.0/appendix/performance-metrics/README.md
rsahlin: The glTF asset can contain one or more scenes.
Each scene is a Node hierarchy.
I agree that model is ambigous because there is no json property with that name.
I would say that a glTF asset is usually called the model.
In this scenario I think it could be better to specify:
> Total number of vertices used by the Nodes in a scene.

specification/2.0/appendix/performance-metrics/README.md
rsahlin: Lol - thanks for catching this :-)

specification/2.0/appendix/performance-metrics/README.md
rsahlin: Thanks - ok good point regarding camelcase. 
I will update.

specification/2.0/appendix/performance-metrics/README.md
rsahlin: Added

specification/2.0/appendix/performance-metrics/README.md
rsahlin: Added something similar by adding Dimension that is Integer[2] containing width and height.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+[textures]  
+This value represents the texture sources that are used by a scene.  
+This is the max value from the most 'complex' primitive that will be referenced by a Node in the scene.  
+The goal of this metric is to provide a worst case texture usage where texturecount and complexity is known.  
+Ie it is known if a texture is BASECOLOR or if part of PBR such as METALICROUGHNESS and now many texture sources that are used.  
+
+[attributes]  
+This value represent the total attribute buffer usage of the model. This value is calculated by adding up the size, in bytes, of all buffer objects in the Asset.  
+This may give an indication of the runtime memory footprint of the buffers needed for the model.
```
jeymichael: Given the goal is to break down usage in terms of data, it would be more useful to split the attributes data further, I would say.    We have the attributes POSITION, NORMAL, TANGENT, TEXCOORD_0, TEXCOORD_1, COLOR_0, JOINTS_0, and WEIGHTS_0.  vertexCount is already handling the POSITION attribute.

The normals/tangents/texCoords/colors are very useful data to know, independent on its own.  A system can handle glTF assets in a lot of different ways such as point cloud or wireframe display (only using POSITION attribute), or shaded/textured display (using NORMAL, TANGENT attributes) or only with color attribute, or one set of texture coordinates (at a given time) etc.   So, having the breakdown would be very helpful.
What do you think of the metrics breakdown of these attributes per scene?

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,79 @@
...
+The goal of these metrics is to assess the runtime performance needs of glTF assets.  
+By runtime we mean the cpu/gpu, bandwidth and memory resources needed to render the model in realtime.  
+These performance metrics are divided in two major categories - `complexity` and `memory`
+
+## Performance Metric Categories
+
+### Complexity  
+
+The `complexity`category provides metrics to calculate cpu/gpu and memory bandwidth target requirements - larger numbers will mean a more powerful target device is needed.  
+These metrics do not provide a mapping to target device performance, this mapping must be done by users of the metrics.
```
jeymichael: In the complexity category, another interesting metric would be the "depth" of the scene graph.  Runtime systems typically have stack-limits (while pushing the Transformation matrices) and knowing this tree depth is often helpful.
But, this complexity lies at a higher level of abstraction.  It is worth discussing about how/if it would fit in the metric here.

specification/2.0/appendix/performance-metrics/README.md
rsahlin: I understand your point - however I think the same information could more or less be extrapolated from texture usage - and also that the overhead is very unlikely to be in the vertex stage.
I think that the vertexCount metric is more to give an indication of number of triangles (4000 vs 4 000 000).

I can see how the primitive mode value could perhaps be of interest:
>POINTS
LINES
LINE_LOOP
LINE_STRIP
TRIANGLES
TRIANGLE_STRIP
TRIANGLE_FAN
 
@raviramachandra What's your input on this?

raviramachandra: @rsahlin I agree with your analysis. The direct impact of these (apart from the obvious memory utilization) attributes on runtime is not obvious or in some cases like texcoords, can be extrapolated from the current information available. For now we can skip this info.

Another interesting point that came up during our internal discussion here was this: 
What if there is an asset which has very high vertex count, but it's overall screen space is very small. For example: say a 400000 vertex asset which may take very small screen space say : 5 cubic cm. I believe GPUs do some kind optimization and may save some compute during pixel shading or rasterization stages ?  Would you also recommend adding something like an static bounding box as additional metric ? Something similar to what FB has in their extension ?
https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/FB_geometry_metadata/README.md

rsahlin: Yes, I think the screen space dimension somehow needs to be known for the usecase.
Perhaps the client can send this with the request to the backend?

So the client would ask the backend for an asset to be used for usecase A taking up X by Y pixels screen space.
Do you think this could work?

Apart from that I am not sure how we could incorporate it into these metrics.

specification/2.0/appendix/performance-metrics/README.md
rsahlin: Yes, I think this is valid - I will add something along the lines of 'max node depth'

specification/2.0/appendix/performance-metrics/README.md
rsahlin: Added a section in the beginning that mention screen coverage as not being part of these metrics

specification/2.0/appendix/performance-metrics/README.md
rsahlin: Added metric for each scene called maxNodeDepth

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Complexity  
+
+The `complexity`category provides metrics to calculate cpu/gpu and memory bandwidth target requirements - larger numbers will mean a more powerful target device is needed.  
+These metrics do not provide a mapping to target device performance, this mapping must be done by users of the metrics.  
+
+### Screen space coverage
+
+The metrics do not include screen space coverage, it is advised that clients shall know how large the render area will be before requesting a model (from the backend)  
+It is up to the backend to take screen coverage into account when providing a suitable model to the client, for instance by taking width + height as parameters.
```
donmccurdy: I think I might rephrase this in terms that don't assume a backend server exists. For example, the models may be stored statically (disk, CDN) and fetched by a client, where the client considers screen coverage when deciding how much it is willing to pay in performance metrics. For example:

> The metrics do not include screen space coverage. However, implementations should generally consider the size of the render area in addition to performance metrics of the model. A higher performance budget may be acceptable when screen space coverage is higher.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Screen space coverage
+
+The metrics do not include screen space coverage, it is advised that clients shall know how large the render area will be before requesting a model (from the backend)  
+It is up to the backend to take screen coverage into account when providing a suitable model to the client, for instance by taking width + height as parameters.  
+
+### Memory  
+
+The `memory` category provides metrics to calculate the runtime memory (size) needs of the model.  
+Using these metrics an estimate of target memory can be calculated - however keep in mind that the exact memory requirements will vary with viewer implementations.
```
donmccurdy: A small nit — the metrics below are not exclusively about memory use. Primitive count mainly affects draw call count (and thus CPU time). A distinction could be drawn between CPU and GPU memory, but I don't think it's necessary. Some indication of GPU time might also be helpful... that's hard to measure, but i'll attempt some suggestions below.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,107 @@
...
+|[vertexCount]    | Integer     | Complexity  |Scene     |Total number of vertices used by the Nodes in a scene           |  
+|[nodeCount]      | Integer     | Complexity  |Scene     |Max nodecount in scene (add upp all nodes in a scene)           |  
+|[primitiveCount] | Integer     | Complexity  |Scene     |Total number of referenced primitives (per scene).  This figure is the un-batched number of primitives, engines may optimize if primitives and meshes share textures. |  
+|[textures]       | Integer     | Complexity  |Scene     |Flags specifying presence of materials and it's texture usage, this is the aggregated max usage. BASECOLOR, METALLICROUGHNESS, NORMAL, OCCLUSION, EMISSIVE, (SPECULARGLOSS) |  
+|[maxNodeDepth]   | Integer     | Memory      | Scene    | The max node depth of the scene, ie the max number of parent/child relations. This number will represent the max stack depth needed when traversing the nodegraph |
+|[attributes]     | Integer     | Memory      | Asset    | The total memory footprint, in bytes, for the models attribute buffer storage |  
+|[textureSizes]  |Dimension    | Memory      | Asset    | The width and height of defined textures in the Asset. |
```
donmccurdy: Some suggestions on this list:

- **vertexCount:** LGTM.
- **vertexMemory:** Suggested as an alternative name to `attributes`. 
- **nodeCount:** LGTM.
- **primitiveCount:** LGTM.
- **materialCount:** Suggested addition. This value, combined with primitiveCount, heavily affects draw count and thus CPU time.
- **textures:** I'm not sure I understand this metric, could you explain further? Or see comment below.
- **textureCount:** Suggested addition. 
- **textureMemory:** Suggested alternative to `textureSizes`, which is more future-proof for compressed textures which have lower per-pixel cost.

I haven't yet seen an asset with node depth >5-10, so I don't think max stack depth is a major concern here. The `nodeCount` metric alone should be fine.

One factor that is mostly missing from this list is per-pixel cost of the materials. Notably, materials using `KHR_materials_unlit` are _much_ cheaper when filling the screen than core PBR materials, and PBR materials using more of the future PBR Next extensions will be more expensive yet. I'm not sure how to encode that in a metric, at this point. The `textures` property above sort of helps here, although I think it's less important to know what material properties are using textures, than to know what material properties are in effect at all. For example, subsurface scattering will be expensive on some clients regardless of whether a texture is used for the effect.

donmccurdy: ^correction, I think *both* `textureMemory` and `textureMaxSize` would be useful.

specification/2.0/appendix/performance-metrics/README.md
jeymichael: After reading through the Spatial Favicons background of this PR, and syncing with Ravi, I can now see the use-case for this metric better.  Also, @donmccurdy 's reference to W3C's responsive images is a good parallel (https://github.com/KhronosGroup/glTF/issues/1743).  Given that, my earlier broken down data would not fit those goals.
It made me thinking more on the screen coverage, and I would try to summarize my thoughts separately.

specification/2.0/appendix/performance-metrics/README.md
jeymichael: I wanted to highlight the complexity that server-reference brings in, as well.  But Don's phrasing captures my thinking.  
Only comment I would add is... given that we are talking about Spatial stuff, 'screen' might not be the accurate word choice.  I don't have a concrete alternative, but something to do with 'volume' may be?

specification/2.0/appendix/performance-metrics/README.md
jeymichael: node depths can vary based on the asset types.  Eg. For character models (dragons etc), it is not uncommon to have (20+) deeper tree, to drive a range of skeleton behaviors.   In our case in Magic leap, simple (but animatable) avatars have crossed way beyond 5 level depth.

I do agree that, compared to vertex, textures etc, this node-depth conveys the metric at a relatively different scale (roughly about logarithmic to node count, in typical cases)   
Thats why maxNodeDepth was meant (at least for me) to be on the 'complexity' category.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,107 @@
...
+## Performance Metric Categories
+
+### Complexity  
+
+The `complexity`category provides metrics to calculate cpu/gpu and memory bandwidth target requirements - larger numbers will mean a more powerful target device is needed.  
+These metrics do not provide a mapping to target device performance, this mapping must be done by users of the metrics.  
+
+### Screen space coverage
+
+The metrics do not include screen space coverage, it is advised that clients shall know how large the render area will be before requesting a model (from the backend)
```
jeymichael: Should we exclude the 3d bounds (screen space coverage) completely, though?  I would suggest revisiting this a bit.  
For AR/VR experiences, the unit of meters (though get scaled in practice, for good reasons) provide valuable information about the model.  Coupling the unit with the 3d bounds of the asset, would help set a guideline for the Spatial FavIcon creators, I believe.  
Its hard to predict, but over time, it appears that the industry would rely on the 3d bounds (along with other metrics already listed here) also for spelling out the intent of the asset.   
Curious to hear others perspectives on this.

specification/2.0/appendix/performance-metrics/README.md
raviramachandra: My intention here was to determine the screen bounds of the asset, as used in this extension:
https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/FB_geometry_metadata/README.md
My understanding is that if we had a fairly "detailed" asset with lots of vertices but it's final screen bounds turned out to be small say: 5 cm cube, The GPU would save compute in the final rasterization phase.

donmccurdy: Two components of this:

1. Number of pixels to be rendered is a multiplier of GPU cost, particularly with more expensive PBR materials. An expensive material is more costly to render the more pixels it shades.
2. A client may _want_ to spend more of its performance budget on an asset covering more of the screen/viewport/space.

I think (2) is the more relevant piece here. (1) comes into play when choosing materials for background environments (high coverage, low detail) vs. hero assets (low coverage, high detail) as an artist. These are general graphics best practices and tradeoffs, I'm not sure how much we should specify.

donmccurdy: I see what you mean about spatial extent being a third component, although it mostly affects performance insomuch as it affects the number of pixels shaded, and ultimately only the application can know the relationship between spatial extent and screen space.

rsahlin: I agree with @donmccurdy - I have updated the screen space section according to his comments.

rsahlin: > My intention here was to determine the screen bounds of the asset, as used in this extension:

As far as I can tell the bounds in that extension are boundingboxes.
These can be used, for instance, to choose scene (LoD) depending on the runtime transform of the model.
Whereas in our case I think we should take a 'worst case' scenario approach.
Meaning that if the framebuffer (render area) is 500 * 500 pixels then we should assume that most of these pixels are drawn (maybe 70% or something) - but this is up to the client.

I will try to find a wording for this.

specification/2.0/appendix/performance-metrics/README.md
rsahlin: I meant that the `memory` category of metrics are for memory. 
The `primitiveCount` is in the complexity category - is this not explained properly @donmccurdy

specification/2.0/appendix/performance-metrics/README.md
rsahlin: Thanks for your comments @donmccurdy 
I will update `attributes` to vertexMemory.

The `textures` metric is used to calculate the per pixel cost.
This will store an array of values indicating the texture/material properties of the most _complex_ material.
For a normal type of PBR material this would contain something like [BASECOLOR, METALLICROUGHNESS, NORMAL, OCCLUSION]
In future with the new extensions coming we would add something like [CLEARCOAT, DIELECTRIC_F0] - or whatever they shall be called :-)

I guess what's missing is a flag to indicate presense of PBR material and how to handle `KHR_materials_unlit` in a generic way?

rsahlin: Texture count can be found by looking at the size of the array of textureSizes.
I see your point regarding `textureMemory` - I will update and add a metric for `maxTextureSize`

rsahlin: Ok, after reviewing my changes I have changed back to similar to what it was first:

[textures] - the worst case texture maps/channels - this can be used to estimate worst case per pixel texel cost (bandwidth)

[materials] - the worst case computational cost per output pixel - this can be used to estimate per pixel computaton cost.
Handles pbr, speculargloss and unlit.

[accessors] - the asset vertex accessor usage, with count, componentType and type - this can be used to estimate the runtime memory consumption for vertex buffers.

[textureSizes] - the asset texture sizes and component formats - this can be used to estimate runtime texture memory.

For accessors and texture sizes I would rather store the details (count or size + format) because clients can take platform details into account.
One target may convert SHORT to FLOAT, or the other way around - a target may need to handle texture sources in a special way, eg 3 component textures are almost always stored (in GPU memory) as 4 bytes.
This way is also flexible when it comes to future use of BasisU - the texture will be transcoded to a supported target format. 
Memory consumption for target format can be calculated by width * height * target format bitdepth.

specification/2.0/appendix/performance-metrics/README.md
rsahlin: Ah, ok I see your point @jeymichael
I do think there is a distinction between screen space coverage and 3d boundingbox - the latter is useful for calculating the asset size as compared to rest of the 'world'.

I have assumed that traversing the nodes and calculating the 3d bounds from positional min/max is super easy - however I guess this extension is a relavant place for it.

- Would a resulting box calculated from concatenating node matrices and positional mix/min be enough?
Ie, start in scene with identity matrix, for each node cancat node transform or matrix and apply to min/max - then update bounds min/max.
This would give the outer boundingbox, in meters, for the world space transformed meshes.

donmccurdy: I've been assuming that these performance metrics are meant for choosing among models that are — other than some quality vs. performance tradeoffs — functionally equivalent within the application. While engines use 3D bounds for many things, does that fit here? Given an asset with world bounds of 100km and an asset of 0.01km, how would an engine choose which to show? I don't know that this document should get into defining functional and contextual differences between assets.

specification/2.0/appendix/performance-metrics/README.md
donmccurdy: > The `textures` metric is used to calculate the per pixel cost.

Knowing the purposes of the textures in the material would be an incomplete measure of per-pixel cost, I think. More advanced shading features (clearcoat, sheen, thin transparency) add cost simply by being enabled — whether a texture is used to achieve them or not. By comparison, the cost of a texture lookup is mostly fixed regardless of what that texture is used for.

A list of extensions might be the simplest way to express that, since the advanced PBR features will mostly be enabled by per-feature extensions.

specification/2.0/appendix/performance-metrics/README.md
rsahlin: I agree with what you are saying for a _normal_ usecase @donmccurdy - but I interpret the need to be for AR/VR situation.
The reason this is important is because you cannot beforehand say that the rendertarget area is X * Y pixels - that you would normally use for rendering a model in a 2D canvas.

Instead the model dimension could be used in relation to the total viewing space, the objects distance and clip regions (ie at what distance does the object get clipped) - for instance an object with (real) world dimensions of 0.5,0.5,0.5 and a clip distance of 0.3 meters would potentially take up a large portion of the screen area.
But a 0.05, 0.05, 0.05 object would only cover a fraction of the screen area and possibly being rendered with more fidelity.

This is my guess anyway :-)

jeymichael: > I've been assuming that these performance metrics are meant for choosing among models that are — other than some quality vs. performance tradeoffs — functionally equivalent within the application. While engines use 3D bounds for many things, does that fit here? Given an asset with world bounds of 100km and an asset of 0.01km, how would an engine choose which to show?

The 3D Bounds proposal is not meant for assisting engines in many things, but only to add direct/relevant information about the space the asset occupies (for the AR/VR use-cases)  After going back to the original [Issue](https://github.com/KhronosGroup/glTF/issues/1743) with FavIcons/ManifestIcons that started this, I am convinced with your parallel to [Responsive Images](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images).   
With that in mind, I could see how browser can pick between different assets based on the bounds.  I believe Ravi had some screenshots / examples, for AR/VR W3C use-cases.
 
In short, 3D Bounds help in addressing the **art direction** & **resolution switching** problem.

raviramachandra: I had discussed with W3C community to just use "static bounding box" as analogous to "resolution" in 2D space. Though as a stand alone metric, it was not enough. Here are my analysis: https://github.com/immersive-web/spatial-favicons/blob/master/design_docs/SpatialFaviconSizes.pdf

rsahlin: Thanks @raviramachandra
Looking through the document leads me to believe that a static (non animated) bounding volume for the aggregated min/max (transformed while traversing the nodetree) primitive values would do the trick?

I will add this as a metric on each scene.

specification/2.0/appendix/performance-metrics/README.md
```diff
@@ -0,0 +1,194 @@
...
+"SCALAR"  
+"VEC2"  
+"VEC3"  
+"VEC4"  
+"MAT2"  
+"MAT3"  
+"MAT4"  
+
+
+[textureSize]
```
MarkCallow, line 131: If transmission size is important, there should be separate information about the size of the texture file. There can be a vast difference between the transmission (file) size and the GPU memory requirement, the latter being much larger when the former is .jpg or .png format.

What is the "texture array" referenced here? If it refers to an array texture, there is not much difference between iterating over the images in one of those and iterating over a set of mipmaps to calculate size.

For the purpose of calculating GPU size (& file size) most block-ccompressed textures have 1 "component", a small number of formats have 2. You will need to provide the size of those components (that is the block size) and the dimension (num_pixels_x * num_pixels_y * num_pixels z) of the block for a  size calculation to be possible.

For uncompressed formats you will need to provide the component type just as you are for accessors. Maybe today glTF only uses 8-bit components but I'm sure that will change.

rsahlin: Sure, that's why there is a texture array - this is the textures (images) that are defined in the asset together with the sizes so that the runtime uncompressed memory requirements can be calculated for textures.
This will currently not handle compressed textures or any other component types than are currently supported (8 bit)

rsahlin:
This PR is not active any longer, mostly due to MagicLeap not being active any longer.
We have gone ahead internally with a different solution.

PR #1798, opened 2020-04-28
KHR_materials_anisotropy extension branch.

UX3D-nopper:


flashk:
The schema does not mention which channel `anisotropyTexture` is sampled from, but the example implementation shows it being sampled from the red channel. Should it be sampled from the alpha channel instead? This would allow combing it with `anisotropyDirectionTexture`.

UX3D-nopper:
@UX3D-eckerlein Can you please cross check, if this packing is possible?

extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
```diff
@@ -0,0 +1,140 @@
...
+Sample values:
+
+```json
+{
+    "materials": [
+        {
+            "extensions": {
+                "KHR_materials_anisotropy": {
+                    "anisotropy": 0.6,
+                    "anisotropyDirection": [0.0, 1.0, 0.0]
```
elalish, line 34: Why do we need a separate anisotropy direction vector? We already define a tangent/bitangent direction and if positive anisotropy is in the tangent direction and negative is in the bitangent direction, isn't that enough?

extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
```diff
@@ -0,0 +1,140 @@
...
+|**anisotropy**                 | `number`                                                             | The anisotropy.                   | No, default: `0.0`             |
+|**anisotropyTexture**          | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)  | The anisotropy texture.           | No                             |
+|**anisotropyDirection**        | `array`                                                              | The anisotropy direction.         | No, default: `[1.0, 0.0, 0.0]` |
+|**anisotropyDirectionTexture** | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo)  | The anisotropy direction texture. | No                             |
+
+## Anisotropy
+
+Two new material properties are introduced: an explicit anisotropy parameter and the direction in which the specular reflection elongates relative to the surface tangents.
+The anisotropy parameter is a dimensionaless number in `[-1, 1]` and forms an injective relation to the roughness distribution along two orthogonal directions, one of which is the direction parameter and the other the result of crossing the direction and the geometric normal.
```
elalish, line 52: What space is the direction vector in? Same as the normal map, presumably? And what is the distribution function? Our normal roughness is defined by infinitesimal oblate ellipsoids (Trowbridge-Reitz), so presumably this would map to some more complex ellipsoids? And likewise, how does it interact with the normal roughness parameter?

tksuoran:
Is anisotropy = 0.0 supposed to match non-anisotropic version?

extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
```diff
@@ -0,0 +1,140 @@
...
+To achieve certain surface finishes, it is possible to define the anisotropy and its direction using a texture.
+
+| Anisotropy | Direction |
+|:---:|:---:|
+| ![Fig. 7](figures/fig7.jpg) | ![Fig. 8](figures/fig8.jpg) |
+| ![Fig. 9](figures/fig9.jpg) | ![Fig. 10](figures/fig10.jpg) |
+
+## Implementation
+
+While uniform and textured anisotropy are multiplied, uniform and textured direction defintions are mutual exclusive and the latter overrides the former. (While it is surely possible to add both directions in terms of their complex argument, the computional overhead may not justify the additional editorial convenience.)
```
abbaswasim, line 71: Is there a typo here 'defintions' -> 'definitions'

bhouston:
We have this:
* anisotropy - intensity float
* anisotropyTexture - intensity texture
* anisotropyDirection - 3D direction vector
* anisotropyDirectionTexture - 3D direction texture.

I think we do not need this many parameters and I am confused why we need a 3D texture.  I would suggest the following:

* anisotropy - intensity float
* anisotropyFlowTexture - a 2D direction texture, the magnitude of which controls the strength of the effect.

Thus we can control the direction and the strength of the effect.  This is similar to [Sketchfab's implementation](https://help.sketchfab.com/hc/en-us/articles/360020687812-Anisotropy).

I haven't seen any use of a separate anistropy strength map.  I also do not believe we have to support a singular anisotropy direction vector.

Although in[ previous implementations I've done](https://clara.io/view/a8b929e6-c703-47ff-a85d-44bc44e5a132), I have added an "anisotropyFlowRotation", which was an angle.  It rotated the flow texture.  I only added this because there was inconsistencies in some of the exported flow textures I ran into.  I would suggest with glTF, these flow textures can be rotated into the correct orientation during export.

PS. I've had the idea previously that if you think that the anisotropyFlowTexture has insufficient precision for short directions, one could use a 3D texture and have the directions be (r*b, g*b), so that you get 16 bits of precision instead of just 8.  This way you can use 8 bits on the direction (r,g), and then use the b channel to scale it down.  That said I am unsure if this is truly needed - I suspect it isn't.

echadwick-wayfair:
I like Ben's idea of reducing parameters, but I think we still need an anisotropyDirection 3D vector because I imagine some models will want to set a direction without requiring a texture.

I'm working on a sample model to test the anisotropyDirectionTexture, which has multiple directions for microgrooves across its surfaces (plus a clearcoat layer for isotropic reflections). This is a photo of the real product:
anisotropy_photo

I produced this diagram to help me understand how to generate the right anisotropyDirectionTexture; which colors I might need to use on which areas of the model.
image

- Circle at top left indicates the color to use for zero anisotropy.
- Arrows indicate directions of microgrooves or hairs.
- Ellipses indicate the the directions these colors will elongate the lighting.

I don't have an implementation to test with though, this is just a proposal.

I based it partly on [a diagram for Modo](https://learn.foundry.com/modo/901/content/help/pages/shading_lighting/shader_items/anisotropic_direction.html), though I find their use of arrows a bit confusing (why are they only facing counter-clockwise?) This is also based on the [flowmap example from Sketchfab](https://help.sketchfab.com/hc/en-us/articles/360020687812-Anisotropy), but I flipped the green channel because it seems we might want to use the same OpenGL +Y convention we use for normal bump textures.

BruceCherniak:
I still wonder, though, if a 3D vector is necessary?  Aren't the microgrooves always, effectively, normal to the surface?  In Eric's diagram above, the blue channel is a constant.  And, if we fallback to a 2D vector, then an intensity and direction could represent that, and a 2D texture can modulate both intensity and direction.

The Autodesk Standard Surface (https://autodesk.github.io/standard-surface/) is specified as such.  Our OSPRay Principled material follows the same implementation.

In a model intended for a rasterizer renderer focused on web-delivery, are we making this too difficult?

In short, I guess I'm suggesting exactly what Ben commented above https://github.com/KhronosGroup/glTF/pull/1798#issuecomment-926828406

Bruce

emackey:
I can see pros and cons to having the intensity be a separate texture vs encoding it in the flow texture as Ben describes above.

Either way, I think this draft should not represent direction as 3D.  It should be 2D, and it must clarify which of those two axes corresponds to tangent and which is bitangent along the surface.

emackey:
Filament has now clarified their documentation to specify that the anisotropy vector cannot have a nonzero Z component:

https://github.com/google/filament/commit/de1281dd9fdfa130a9f319f4a36793728e7ff11c

emackey:
Per offline discussions, the changes needed here are:

* `anisotropy` should range from 0 to 1, disallowing negative values.  Default is zero.  I would recommend using the blue color channel for this, when a texture is supplied.

* `anisotropyDirection` should be a 2D vector (not 3D), and can be packed into the red & green channels of a texture in the same manner as a normal map's X and Y vectors (where zero stretches to -1).  The blue and alpha channels should be ignored.  The direction is always in-plane with the material (would it make sense to be the tangent vector?).  The client should always normalize (in 2D) the vector after reading, to get the direction.  What would the default factor be, maybe `[1, 0]`?  Maybe `[1, 1]` in case a texture is used, and the user must manually supply a `[1, 0]` or `[0, 1]` vector (or any 2D direction) when not using a texture here?

How does this interact with base roughness?  Can roughness zero show anisotropic effects, as if the value of `anisotropy` added to the base roughness in one direction?

bhouston:
> How does this interact with base roughness? Can roughness zero show anisotropic effects, as if the value of anisotropy added to the base roughness in one direction?

The base roughness determines the roughness value.  This determines the angle and amount of anisotropy in that angle.  So they are complimentary and orthogonal.

donmccurdy:
By convention, shouldn't these names be...

- anisotropyFactor
- anisotropyTexture
- anisotropyDirectionFactor
- anisotropyDirectionTexture

I have no opinion on 2D vs 3D here, or reducing the parameter space, but elsewhere we've used `*Factor` as a suffix for scalar or vector values paired with a texture.

donmccurdy:
Also — which vertex attributes must be included in the mesh primitive when using anisotropy?

echadwick-wayfair:
Opened a pull request for the sample model: https://github.com/KhronosGroup/glTF-Sample-Models/pull/366

emackey:
Hi all,

Per many conversations here and in the PBR TSG, the names have changed.  This is where we stand at the moment:

- `anisotropyFactor` - Float from [0, 1] for strength.  Cannot be negative.
- `anisotropyDirection` - Now a 2-element array in JSON (no longer 3D), specified as tangent & bitangent, client must normalize.
- `anisotropyTexture` - This is a 2-channel (R, G) texture that overrides `anisotropyDirection` with X & Y vectors in range [-1, 1] (like two channels of a normal map).  The magnitude of the 2D vector is multiplied with `anisotropyFactor` to form the per-pixel anisotropy strength, and the remaining direction must then be normalized (to remove that magnitude) to rotate the anisotropy.

Remaining issues:

- We talked about what "strength" means in terms of what it does to tangent roughness & bitangent roughness.  I think this is using Filament's model for that, but lots of engines use other formulas.  Let's make sure we pick one and get it normatively specified in here.
- @echadwick-wayfair Can you update the sample model to use this new parameter set (if everyone is in agreement now?)
- @elalish Can you update the ThreeJS draft to read from the new parameter set?

Thanks all.

emackey:
@elalish Can you double-check my GLSL math in e0333e7013d55b8?  I added the `mat2` thing where `anisotropyDirection` is now allowed to rotate the vector from the texture, but I don't have an easy way to test it.

extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
```diff
@@ -0,0 +1,155 @@
...
+float D_GGX_anisotropic(float NdotH, float TdotH, float BdotH, float anisotropy, float at, float ab)
+{
+    float a2 = at * ab;
+    vec3 f = vec3(ab * TdotH, at * BdotH, a2 * NdotH);
+    float w2 = a2 / dot(f, f);
+    return a2 * w2 * w2 / M_PI;
+}
+
+float V_GGX_anisotropic(float NdotL, float NdotV, float BdotV, float TdotV, float TdotL, float BdotL,
+    float anisotropy, float at, float ab)
```
elalish, line 139: `anisotropy` is an unused parameter in both these functions - `at` and `ab` already represent its effect.

extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+```glsl
+vec3 BRDF_specularAnisotropicGGX(vec3 f0, vec3 f90, float alphaRoughness,
+    float VdotH, float NdotL, float NdotV, float NdotH, float BdotV, float TdotV,
+    float TdotL, float BdotL, float TdotH, float BdotH, float anisotropy)
+{
+    float at = max(alphaRoughness * (1.0 + anisotropy), 0.00001);
+    float ab = max(alphaRoughness * (1.0 - anisotropy), 0.00001);
+
+    vec3 F = F_Schlick(f0, f90, VdotH);
```
elalish, line 123: Schlick is great, but it's still an approximation and therefore not normative. I would pull the definition of D out as the one normative part of the spec, maybe shown a little more mathy like in Burley's original. The visibility term is derived from this distribution being height-correlated, so again, no need for it to be normative.

elalish:
@emackey Your math looks pretty good to me - sorry, I put a couple of comments in your commit, which makes them hard to find. But we definitely need to update the spec a bit more.

extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
```diff
@@ -0,0 +1,155 @@
...
+        }
+    ]
+}
+```
+
+|                         | Type     | Description               | Required           |
+| ----------------------- | -------- | ------------------------- | ------------------ |
+| **anisotropyStrength**  | `number` | The anisotropy strength.  | No, default: `0.0` |
+| **anisotropyDirection** | `number` | The direction of the anisotropy in tangent, bitangent space, measured in radians counter-clockwise from the tangent. When anisotropyTexture is present, anisotropyDirection provides additional rotation to the vectors in the texture. | No, default: `0.0` |
+| **anisotropyTexture**   | [`textureInfo`](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | The anisotropy texture. Red and green channels represent the anisotropy direction in [-1, 1] tangent, bitangent space. The magnitude of this vector is multiplied by anisotropyFactor to obtain the anisotropy strength. | No |
```
donmccurdy, line 63:
```suggestion
| **anisotropyTexture**   | [`textureInfo`](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | The anisotropy texture. Red and green channels represent the anisotropy direction in [-1, 1] tangent, bitangent space. The vector is rotated by anisotropyDirection, and multiplied by anisotropyStrength, to obtain the final anisotropy direction and strength. | No |
```

extensions/2.0/Khronos/KHR_materials_anisotropy/schema/glTF.KHR_materials_anisotropy.schema.json
```diff
@@ -0,0 +1,30 @@
...
+        "anisotropyDirection": {
+            "type": "number",
+            "description": "Anisotropic direction",
+            "default": 0.0,
+            "gltf_detailedDescription": "The direction of the anisotropy in tangent, bitangent space, measured in radians counter-clockwise from the tangent. When anisotropyTexture is present, anisotropyDirection provides additional rotation to the vectors in the texture."
+        },
+        "anisotropyTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "Anisotropy texture",
+            "gltf_detailedDescription": "The anisotropy texture. Red and green channels represent the anisotropy direction in [-1, 1] tangent, bitangent space. The magnitude of this vector is multiplied by anisotropyFactor to obtain the anisotropy strength."
```
donmccurdy, line 25:
```suggestion
            "gltf_detailedDescription": "The anisotropy texture. Red and green channels represent the anisotropy direction in [-1, 1] tangent, bitangent space. The vector is rotated by anisotropyDirection, and multiplied by anisotropyStrength, to obtain the final anisotropy direction and strength."
```

extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+The anisotropic GGX is defined as follows:
+
+```glsl
+vec3 BRDF_specularAnisotropicGGX(vec3 f0, vec3 f90, float alphaRoughness,
+    float VdotH, float NdotL, float NdotV, float NdotH, float BdotV, float TdotV,
+    float TdotL, float BdotL, float TdotH, float BdotH, float anisotropy)
+{
+    float at = max(alphaRoughness * (1.0 + anisotropy), 0.00001);
+    float ab = max(alphaRoughness * (1.0 - anisotropy), 0.00001);
```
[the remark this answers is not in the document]
emackey: @abwood Found a reference to this formula, it's published.  I'm not convinced it's the best one to use, but I do want to use one that's published somewhere.  The MaterialX one is looking nice to me, perhaps I'll collect some documentation and propose it at the next meeting.

emackey:
Per today's meeting, the anisotropy strength formula has been set.

```
alphaTangent = mix(alphaRoughness, 1.0, anisotropy * anisotropy);
alphaBitangent = alphaRoughness;
```

Using the material's roughness along the bitangent (or along the direction perpendicular to the chosen anisotropy direction) allows rasterizers to continue using material roughness as a base lookup into the IBL, unmodified.  The tangent direction then needs an approximation such as stretching or multisampling, but only along one axis.

extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Contributors
+
+* Mike Bond, Adobe, [@miibond](https://github.com/MiiBond)
+* Ed Mackey, AGI [@emackey](https://github.com/emackey)
+* Alex Wood, AGI [@abwood](https://twitter.com/abwood)
+* Nicolas Savva, Autodesk, [@nicolassavva-autodesk](https://github.com/nicolassavva-autodesk)
+* Tobias Haeussler, Dassault Systemes [@proog128](https://github.com/proog128)
+* Bastian Sdorra, Dassault Systemes [@bsdorra](https://github.com/bsdorra)
+* Don McCurdy, Google [@donrmccurdy](https://twitter.com/donrmccurdy)
```
donmccurdy, line 11: Minor edit – no longer at Google, and no longer using Twitter actively.

```suggestion
* Don McCurdy, Individual Contributor [@donrmccurdy](https://github.com/donmccurdy)
```

emackey (approved):
This was approved today for "Release Candidate" status by the PBR TSG.

PR #1825, opened 2020-06-10
KHR_materials_diffuse_transmission

proog128:
An extension for the metallic-roughness material that adds a translucency effect, i.e., a diffuse lobe on the back-side of the surface. This is useful
* for thin materials like leaves or paper
* for volumetric materials like plastic or candle wax; in combination with `KHR_materials_volume` (#1726) in case of SSS with short scattering distance


emackey:
/ping @abwood New diagram added, looks nice!

cx20:
BTW, This extension is not yet listed in the [list](https://github.com/KhronosGroup/glTF/tree/master/extensions#in-progress-khronos-and-multi-vendor-extensions) of glTF extensions.
It's a good idea to list this extension as an extension to "In-progress".

emackey:
@cx20 Fixed in f287c13f5f, thanks.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,252 @@
...
+```
+diffuseTransmissionColor = diffuseTransmissionColorFactor * diffuseTransmissionColorTexture.rgb
+```
+## Examples
+
+The examples use `diffuseTransmissionColorTexture` from `baseColorTexture`.
+|                No transmission                 | <sub>`diffuseTransmissionFactor: : 0.25`</sub> | <sub>`diffuseTransmissionFactor: 0.25`<br>`diffuseTransmissionColorFactor: [1.0,0.9,0.85]`</sub> |
+| :--------------------------------------------: | :--------------------------------------------: | :----------------------------------------------------------------------------------------------: |
+| ![](figures/teatime_backlit_no_trans.jpg) <br> |        ![](figures/teatime_backlit.jpg)        |                             ![](figures/teatime_backlit_colored.jpg)                             |
+|  ![](figures/teatime_side_no_trans.jpg) <br>   |         ![](figures/teatime_side.jpg)          |                              ![](figures/teatime_side_colored.jpg)                               |
```
donmccurdy, line 79: Minor cleanup — I think if this table were transposed, the images in the same column would have the same size and differences would be easier to compare:

| Case | Backlit | Side |
|------|---------|------|
| No Transmission |  |  |
| <sub>`diffuseTransmissionFactor: 0.25`</sub> |  |  |
| <sub>`diffuseTransmissionFactor: 0.25`<br>`diffuseTransmissionColorFactor: [1.0,0.9,0.85]`</sub> |  |  |

bsdorra:
Created (path-tracing) live demos for the example scenarios. 
(Please note: half-way **decent gpu required**. **Be patient**, webgl shader linking is slooooow)

https://dassaultsystemes-technology.github.io/dspbr-pt/index.html?scene=Potted+Plant&ibl=Studio_05&rotate=-45&unfold=transmission

https://dassaultsystemes-technology.github.io/dspbr-pt/index.html?scene=Tea+Time&ibl=Studio_05&rotate=-45&unfold=transmission

Hope this is useful!

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/schema/glTF.KHR_materials_diffuse_transmission.schema.json
```diff
@@ -0,0 +1,41 @@
...
+          "description": "The color of the transmitted light.",
+          "default": [ 1.0, 1.0, 1.0 ],
+          "minItems": 3,
+          "maxItems": 3,
+          "gltf_detailedDescription": "The color of the transmitted light."
+      },
+      "diffuseTransmissionColorTexture": {
+          "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+          "description": "A texture that defines the color of the transmitted light",
+          "gltf_detailedDescription": "A texture that defines the color of the transmitted light, stored in the RGB channels and encoded in sRGB. This texture will be multiplied by translucencyColorFactor."
```
abwood, line 36:
```suggestion
          "gltf_detailedDescription": "A texture that defines the color of the transmitted light, stored in the RGB channels and encoded in sRGB. This texture will be multiplied by diffuseTransmissionColorFactor."
```

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,260 @@
...
+    layer = specular_brdf(α = roughness^2)
+  )
+```
+
+Since the diffuse BTDF does not have controls for roughness, the roughness parameter acts only on the reflective part of the surface. By decoupling the reflection and transmission parts it is possible to configure materials which have a smooth reflection and a diffuse transmission, as shown in image below (right).
+
+<figure style="text-align:center">
+<p float="left">
+ <img src="./figures/transmission-translucency_thin.jpg" width="40%" />
+ <img src="./figures/transmission-translucency_thin_angle.jpg" width="39.5%" />
```
abwood, line 191: These images will need to get updated to reflect the new name change to diffuse transmission.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,260 @@
...
+</figure>
+
+For comparison, below is the result for the same scene with `KHR_materials_transmission` instead of `KHR_materials_diffuse_transmission`. Here, a refractive microfacet BTDF describes the transmission of light through the volume boundary. The refraction occurs on microfacets, and thus the roughness parameter affects both reflection and transmission.
+
+<figure style="text-align:center">
+<img src="./figures/transmissive-roughness.png"/>
+<sub><figcaption><em>Transmissive sphere with varying roughness. From left to right: 0.0, 0.2, 0.4.</em></figcaption></sub>
+</figure>
+
+### KHR_materials_sss
```
abwood, line 237: Should this discussion move to the SSS extension?

abwood: Just an update to my comment above. On one of our TSG calls, we did confirm that this section belongs in the KHR_materials_sss extension.

bsdorra: Removed all references to KHR_materials_sss with 70ff0da. Will add the relevant infos to KHR_materials_sss.

coreh:
Hey 👋 [Currently implementing this](https://github.com/bevyengine/bevy/pull/8015) for Bevy Engine, and wondering: Why was the A channel chosen for the diffuse transmission texture?

The specular transmission texture from `KHR_materials_transmission` uses the R channel, and the thickness texture from `KHR_materials_volume` uses the G channel. So perhaps the B channel would be a good fit for packing all of these into single RGB texture that's easier to author and visualize, as these values are all somewhat related. Is it currently reserved for something else? 

I could not find anything using the B channel, apart from `ADOBE_materials_clearcoat_specular`

bsdorra:
Hi @coreh, the idea is to have all diffuse transmission values in a single texture, color in the RGB channels and the factor stored in A. 
The glTF group is discussing texture packing on a more general level. With all these new PBR parameters, hard-coding the packing structure doesn't seem to be optimal anymore.

coreh:
Oh, that makes a lot of sense. I was thinking in terms of the existing packed texture, and didn't even think that one would frequently want to use factor and color together. I'll update the docs on my implementation to reflect this.

echadwick-wayfair:
I noticed a typo in the example codeblock in the README.md (see below) and was about to clone the PR locally to fix it, but GithubDesktop reported this error:

image


The typo I mentioned above is a misplaced comma, currently as:
```
                    "diffuseTransmissionTexture": {
                      "index": 0,
                    }
                    "diffuseTransmissionColorFactor": [
``` 
But should instead be:
```
                    "diffuseTransmissionTexture": {
                      "index": 0
                    },
                    "diffuseTransmissionColorFactor": [
```

bsdorra:
@echadwick-wayfair fixed it, thanks!
The error message is weird, the branch exists and is valid. Is your remote url pointing the correct repo?

zeux:
Any updates on the progress on this extension? I'm looking at releasing a new version of gltfpack in the next month or so and unsure if it should include support for this extension even though it's still not merged to avoid a situation where it gets released shortly afterwards and isn't supported for some period of time, or if this extension is still in a significant state of flux and won't get merged any time soon.

emackey:
@zeux That's a great question.  I would like to see this extension ratified mostly as-is, but it may not happen soon.  I've re-started the discussions on the Khronos PBR TSG calls, but I can't guarantee it will suddenly move forward.  There are an assortment of concerns ranging from performance to interoperability to consistency between engines.  If you know of some solid use cases or sample models that we could use as fuel for the burning desire for this, please let me know.

Reon90:
there is a sample model?

echadwick-artist:
I have added three sample assets so far (thanks to @bsdorra for [pathtraced versions](https://github.com/KhronosGroup/glTF/pull/1825#issuecomment-1120813652) of the plant and teacup)

https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/DiffuseTransmissionTeacup 

https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/MandarinOrange (this should be done with SSS instead, not a thin-walled surface shape, but it does demonstrate some odd backscatter behavior)

https://github.com/KhronosGroup/glTF-Sample-Assets/pull/79

echadwick-artist:
We probably need a sample asset grid of flat shapes to stress test combinations of values. Which of these should be represented?

- diffuseTransmissionFactor
- diffuseTransmissionTexture
- diffuseTransmissionColorFactor
- baseColorFactor
- image based lighting with sunlight
- punctual lighting 
- is the extension affected by normal mapping, and thus we should include this in tests?
- what additional factors directly influence diffuse transmission?

I can manufacture a test asset, if I have a set of parameters we’d like to include.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,241 @@
...
+| **diffuseTransmissionTexture**      | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A texture that defines the strength of the diffuse transmission effect, stored in the alpha (A) channel. Will be multiplied by the diffuseTransmissionFactor.                                       | No                       |
+| **diffuseTransmissionColorFactor**  | `number[3]`                                                         | The color of the transmitted light.                                                                                                                                                                 | No, default: `[1, 1, 1]` |
+| **diffuseTransmissionColorTexture** | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A texture that defines the color of the transmitted light, stored in the RGB channels and encoded in sRGB. This texture will be multiplied by diffuseTransmissionColorFactor.                       | No                       |
+
+
+The strength of the effect is controlled by `diffuseTransmissionFactor` and `diffuseTransmissionTexture`, combined via multiplication to describe a single value.
```
rsahlin, line 68: I think that a more suitable wording would be along the lines of:
"The percentage of light that is transmitted through the surface is calculated by multiplication of `diffuseTransmissionFactor` and `diffuseTransmissionTexture` to describe a single value"

Since there is not really any 'strength' :-)

emackey: Recently the OpenPBR effort appears to have adopted the term "weight" where glTF has traditionally used the word "strength."  I guess I like "weight" better, but we should check our existing ratified extensions to see how hard it will be to make the change.  Anisotropy, for example, uses "strength" normatively, and I'm sure others do as well.  We should figure out how important it is to have consistency on this term among glTF extensions.

bghgary: > I'm sure others do as well

Occlusion uses strength also in the core spec. Not that I'm advocating using this term here.

rsahlin: I really think we must approach this from a physical point of view:
The parameter governs the percentage (factor) of light that is transmitted into the surface - to be re-emitted at the other side.

From the side being illuminated - this could be seen as absorbed or internally transmitted factor.
From the side facing away from the light - this could be seen as the re-emitted  factor.

I fully understand that editors/dcc tools call this something akin to 'weight' or 'strength' because that is understandable to content creators - in the glTF spec I think we need to be more accurate and use 'percentage' or 'amount' instead.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,241 @@
...
+                }
+            }
+        }
+    ]
+}
+```
+
+|                                     | Type                                                                | Description                                                                                                                                                                                         | Required                 |
+| ----------------------------------- | ------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------ |
+| **diffuseTransmissionFactor**       | `number`                                                            | The percentage of reflected, non-specularly reflected light that is transmitted through the surface via the Lambertian diffuse transmission, i.e., the strength of the diffuse transmission effect. | No, default: `0`         |
```
rsahlin, line 62: `The percentage of reflected, non-specularly reflected....` 
 I think this is vague and somewhat misleading.
 
 I propose:
 "The percentage of non-specularly reflected....."

rsahlin: I also propose to drop `Lambertian diffuse` since implementations may use some other model.


So it becomes:
`The percentage of non-specularly reflected light that is transmitted through the surface and re-emitted on the other side, i.e., the diffuse transmission effect.`

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,241 @@
...
+| **diffuseTransmissionTexture**      | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A texture that defines the strength of the diffuse transmission effect, stored in the alpha (A) channel. Will be multiplied by the diffuseTransmissionFactor.                                       | No                       |
```
rsahlin, line 63: `A texture that defines the diffuse transmission factor, stored in the alpha (A) channel. Will be multiplied by the diffuseTransmissionFactor. `

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,241 @@
...
+| **diffuseTransmissionColorFactor**  | `number[3]`                                                         | The color of the transmitted light.                                                                                                                                                                 | No, default: `[1, 1, 1]` |
```
rsahlin, line 64: I would say that this is the color of the re-emitted light.
`The color of the re-emitted light.`

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,241 @@
...
+```
+
+<figure style="text-align:center">
+<img src="./figures/bsdf.svg"/>
+<sub><figcaption><em>Left: Diffuse BRDF. Right: Diffuse BTDF.</em></figcaption></sub>
+</figure>
+
+## Combining Diffuse Transmission with other Extensions
+### KHR_materials_transmission
+Both `KHR_materials_diffuse_transmission` and `KHR_materials_transmission` replace the diffuse BRDF with a mix of diffuse BRDF and a BTDF that transmits light onto the opposite side of the surface. In case of `KHR_materials_transmission`, this is a microfacet BTDF that shares its roughness with the microfacet BRDF. In case of `KHR_materials_diffuse_transmission`, on the other hand, this is a diffuse BTDF.
```
rsahlin, line 158: I don't think it makes sense to combine the two extensions KHR_materials_transmission and KHR_materials_diffuse_transmission.
They do both model the same physical phenomena - ie, the transmission of light through the material.
They both have a transmission factor - which is the amount of light that enters into the material.
Where the original transmission extension can adapt the roughness, this value will scatter the light being transmitted - the new extension are for materials where no light will pass through without being scattered/refracted.
Using the original extension with roughness = 1 should produce the same result as this extension.

I see this extension as a great way of simplifying/optimizing shader performance since it is known that the background will never be visible through the material.

Only difference that I can see is that this extension has a parameter to set the color of the light passing through the material. 
However this should be possible by combining KHR_materials_transmission with the volume extension.

emackey: This draft was written by Tobias and Bastien some years ago, and they are describing how the effect works in Enterprise PBR which is being emulated here.  That system is designed for path tracing, where it makes sense to offer both specular transmission (the original glTF transmission) and diffuse transmission (this PR).

I agree that we should strongly discourage realtime systems from applying both at once.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
emackey: Remember our lingo in glTF is a little strange.  A "factor" is what GLSL would call a "uniform".  A factor is specifically a uniform value that can exist in the absence of a texture, or be multiplied by a texture.

So the actual parameter name here is simply `diffuseTransmission`.  It happens to be presented as a factor / texture pair, which is how all texturable glTF parameters are presented.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
emackey: Changing from "strength" to some other terminology is a much larger discussion, one that retroactively touches the glTF core specification and many ratified PBR extensions.  I don't think this PR is the correct venue to propose such a change.

rsahlin: Really?
I don't understand that reasoning.
Just because we use a technical wording in this extension does not mean that we have to make changes to existing spec.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
rsahlin: Well, it's the same when its called `color of the transmitted light` - it still depends on the received light.....
I think that something along the lines of:
`Internal coloring of the light that is transmitted through the surface`

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
rsahlin: >I agree that we should strongly discourage realtime systems from applying both at once.

I think it's more than that - they model different and non-compatible materials.
* This extension models the behavior under full/very high roughness - so high that you never see anything through the material.
* KHR_materials_transmission using varying values for roughness - where most values means that the material is (mostly) transparent.

Just because Enterprise PBR uses the effect together is not, in my mind, a reason for us to do exactly the same.

My view is that the materials that these extension model are not compatible and we should disalow the use of the extensions together.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
bsdorra: How about `The color that modulates the transmitted light`?

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
emackey: > This extension models the behavior under full/very high roughness

Actually no, this is a common misconception.  The surface of KHR_materials_diffuse_transmission is allowed to be smooth, and will produce specular reflections according to surface roughness as usual, including very shiny ones.  The transmitted light is diffuse, fully scattered, but not because of roughness at the surface.

Both this and the original KHR_materials_transmission extension are permitted to be present, and this section of the spec describes the interaction in mathematical detail.  Admittedly such details are primarily of use to path-tracing systems for which this math was originally worked out, and is not much use in a real-time rasterizer.

Perhaps we could add a note discouraging realtime assets from using both together.  But it's far too late to deeply alter such a fundamental aspect of this spec. This draft has been around for four years and has already shipped with stable versions of BabylonJS, for example.  We're not making breaking changes to it at this point, the time for that is long gone.  We're just looking to polish up the wording and get it ratified soon.

rsahlin: >Actually no, this is a common misconception. The surface of KHR_materials_diffuse_transmission is allowed to be

My bad, I was hoping that you would understand what I meant - let me phrase it this way:
This extension models the behavior of full/high roughness of the inside (of the material)
All light that enters into the material is scattered - there can be no translucensy.
It is simply not compatible with KHR_materials_transmission from a physical point of view!

>Perhaps we could add a note discouraging realtime assets from using both together. But it's far too late to deeply alter such a fundamental aspect of this spec. This draft has been around for four years and has already shipped with stable

I really do not understand this statement!
We must seek to release clearly defined extensions - as of now this extension is not very well defined from a physical point of view. Regardless of current implementations or time.

> gone. We're just looking to polish up the wording and get it ratified soon.

Well, extensions shall be ratified when ready - this extension needs more work!

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
rsahlin: I really think we must re-work this extension and word it in a more physically based manner.
The phenomena that is modelled is attenuation of light - we have simplified it so that light is directly attenuated when entering the material.
On the opposite side there will be reflection and transmission as light hits the surface.
Of the transmitted light (from the opposite side) we have a factor controlling the amount that will enter into the surface - all of this light will be attenuated (regardless of distance) and re-emitted.

As a final parameter we can add the so-called thickness texture - in this case it is more of an absorption texture.
This texture controlls the amount of attenuated light that is re-emitted, regardless of thickness, we really do not know if light is absorbed due to distance travelled or density of material.
And so it ought to be called `absorptionTexture` and can work the same way that occlussion texture works (ie higher values let light trough, lower values reduce)

rsahlin (commented):
Maybe I should open up a new PR with my proposed changes in order to get progress on this extension?

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,241 @@
...
+
+Experimental
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds a Lambertian diffuse transmission BSDF to the metallic-roughness material. Thin, dielectric objects like leaves or paper diffusely transmit incoming light to the opposite side of the surface. For optically thick media (volumes) with short scattering distances and therefore dense scattering behavior, a diffuse transmission lobe is a phenomenological plausible and cheap approximation.
```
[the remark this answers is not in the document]
bsdorra: Other extensions suggest that the overview part focuses more on the practical application of the extension than the underlying physical concepts. Property descriptions seem to be a more appropriate location. Let me try to reword and add sections based on your suggestion (and also following the example of KHR_materials_transmission). 

## Overview

This extension models the physical phenomenon of light being diffusely transmitted through an infinitely thin material. Thin dielectric objects like leaves or paper diffusely transmit light due to dense volumetric scattering within the objects surface boundaries. In 3D graphics, it is common to approximate thin volumetric objects as non-volumetric surfaces. 

`KHR_materials_diffuse_transmission` models the diffuse transmission of light through such infinitely thin surfaces. For optically thick media (volumes) with short scattering distances and dense scattering behavior, `KHR_materials_diffuse_transmission` provides a phenomenologically plausible and cost-effective approximation.

.
.
.

## Properties
### diffuseTransmissionFactor

The proportion of light that is diffusely transmitted through a surface, rather than being diffusely re-emitted. This is expressed as a percentage of the light that penetrates the surface (i.e., not specularly reflected), rather than a percentage of the total light incident on the surface. A value of 1.0 indicates that 100% of the light that penetrates the surface is transmitted through it.

### diffuseTransmissionColorFactor

The proportion of light at each color channel that is not attenuated by the surface transmission. Attenuation is usually defined as an amount of light at each frequency that is reduced over a given distance through a medium by absorption and scattering interactions. However, since this extension deals exclusively with infinitely thin surfaces, attenuation is constant and equal to 1.0 - `diffuseTransmissionColorFactor`.

### diffuseTransmissionTexture

The `A` channel of this texture defines proportion of light that is diffusely transmitted through a surface, rather than being diffusely re-emitted. This is expressed as a percentage of the light that penetrates the surface (i.e., not specularly reflected), rather than a percentage of the total light incident on the surface. A value of 1.0 indicates that 100% of the light that penetrates the surface is transmitted through it.

The value is linear and is multiplied by the `diffuseTransmissionFactor` to determine the total diffuse transmission value.

```
diffuseTransmission = diffuseTransmissionFactor * diffuseTransmissionTexture.a
```

### diffuseTransmissionColorTexture

The `RGB` channels of this texture define the proportion of light at each color channel that is not attenuated by the surface transmission.
The values are multiplied by the `diffuseTransmissionColorFactor` to determine the total diffuse transmission color.
```
diffuseTransmissionColor = diffuseTransmissionColorFactor * diffuseTransmissionColorTexture.rgb
```

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
bsdorra: Not sure if I really understand your thickness texture proposal. Isn't it exactly what you control with the intensity of the diffuseTransmissionColorTexture?

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
bsdorra: As the co-author, I may be biased, but I'd say that this extension is very well defined in the context of the current material model and other ratified material extensions. :) There's a clear definition of how it interacts with KHR_materials_transmission (that should also work for rasterizers). You find identical or very similar definitions in other PBR models, i.e. OpenPBR.  

> ...as of now this extension is not very well defined from a physical point of view

What does that even mean? PBR appearance models do not necessarily replicate physics directly. Instead, they create abstract mathematical concepts that enable us to simulate the appearance of the real world in a computationally feasible manner, while adhering to important physical principles such as energy conservation. So, it is always a balancing act between physical correctness, performance, visual quality, artistic control and also inherently non-physical annoyances as infinitely thin surfaces.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
emackey: Just from a consistency point of view, I wouldn't want to use the term "strength" in one extension and then "weight" in another extension, and then have to explain why we used two different terms to mean the same thing.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
emackey: Apologies if my own demos have led to some of the confusion here.  Both transmission and diffuse transmission are surface-only properties.  When used on a thin-walled surface, either of these properties allows light to transmit directly to the opposite side.  But, if the volume extension is present, there may be a thick volume beyond the surface.  In that case, transmission or diffuse transmission is just a gateway allowing light to enter and exit the volume.  The properties of T or DT don't affect the volume itself, as the volumetric properties are covered by its own extension (as thickness, attenuation color, attenuation distance).

I gave a demo of [candle wax](https://github.com/KhronosGroup/glTF-Sample-Viewer/pull/544#issuecomment-2177354713) recently.  In the material for this, diffuse transmission is set to 0.5, untextured white.  All of the wax coloring is established by volumetric properties (thickness and attenuation), and DT is just an enabler for the light to get in and out of that volume.  You can imagine that DT offers a surface-level scattering event, white in this case, paired with a volumetric attenuation towards wax color.

```json
    "materials":[
        {
            "doubleSided":false,
            "extensions":{
                "KHR_materials_diffuse_transmission":{
                    "diffuseTransmissionFactor":0.5
                },
                "KHR_materials_volume":{
                    "attenuationColor":[
                        0.8964201807975769,
                        0.8795840740203857,
                        0.6938217878341675
                    ],
                    "attenuationDistance":0.02,
                    "thicknessFactor":0.11999999731779099,
                    "thicknessTexture":{
                        "index":0
                    }
                },
                "KHR_materials_ior":{
                    "ior":1.4500000476837158
                }
            },
            "name":"Wax Material",
            "occlusionTexture":{
                "index":1
            },
            "pbrMetallicRoughness":{
                "baseColorFactor":[
                    0.8964201807975769,
                    0.8795840740203857,
                    0.6938217878341675,
                    1
                ],
                "metallicFactor":0,
                "roughnessFactor":0.30000001192092896
            }
        },
```

Something similar happens in my dragon diffuse transmission demo, as the dragon's coloring in that demo comes from the volumetric properties as well.  I don't yet have a demo that pairs surface-level coloring with volumetric-imposed coloring, but the two are mathematically separate: The diffuse transmission coloring would be along the surface, and would not change with thickness.  Volumetric attenuation is an un-texturable homogeneous color throughout the volume, that does potentially change with thickness.  Light reaching the eye must survive both.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
emackey: @bsdorra I have a related question about how the light breaks down.  Let's say at a particular point along the surface you have 10% Fresnel reflection, and you have transmission and diffuse transmission both set to 50% each on the same material.  This is how I think it breaks down, is this correct?

1. Fresnel / specular reflection takes some light off the top.  Of the original 100% received at the chosen point, 10% is reflected by this, and 90% is not yet accounted for.
2. Transmission (the specular kind) asks for 50%, but it gets 50% of what's left (90%), so for this point only 45% of the original received light shines through the surface along the specular transmission ray.  The remaining 45% of the original is not yet accounted for.
3. Diffuse transmission also asks for 50% of the remaining 45%.  As a result, 22.5% of the original light transmits through the surface and is released as a diffuse lobe on the opposite side.  An additional 22.5% of the original light is all that remains not yet accounted for.
4. All remaining light (22.5% of the original) is modulated by the base color and released as a diffuse lobe on the same side of the surface where it was received.

Does that sound right?

bsdorra: @emackey that sounds correct to me 👍🏼

rsahlin: >What does that even mean? PBR appearance models do not necessarily replicate physics directly. Instead, they create

It means that the goal of glTF must be to parameterize the physical properties of real world materials and the interaction with light.
Not mimic the behavior of one or more given editing formats.
One of the reasons why glTF is so succesful is that we are agnostic and that the parameters used are a good destillation of physical properties.

Also of great importance - glTF is a `fileformat` - not a PBR model - you cannot compare glTF to existing PBR models.
Our job is to package the physical material and light properties into a dataset that can be distributet as a transmission format.

rsahlin:
There are a couple of design issues with this extension, let me explain:

1: This extension models the scattering behavior of dense, non fully opaque, materials such as leaves or wax.
This phenomena is not the same as what is specified in KHR_materials_transmission and KHR_materials_volume
These two extensions both model the behavior of light passing through a transparent media such as glass or water. 
They both use the Snell equations - attenuation color, viewpoint change due to IOR and total internal reflection are important.

2: This extension is more or less based on an adaptation of the Beer/Lambert law.
Instead of attenuation color, viewchanges due to IOR and total internal reflection, the total light extinction distance is important.
Ie, how far into the material does light travel before it is eliminated.

These two phenomena shall not be mixed into the same extensions - I know that it is super easy to get a shader to execute the code - it's just not a good parameterization of needed data.

Luckily there is a simple solution:

Make this extension into 'scattering transmission'

* Light entering into the surface of the material is scattered, this scattered light may be emitted on the opposite side, colored by the basecolor of the exit side. It is a model of the behavior as light travels through mostly organic materials such as leaves, wax or 

* A factor controlls the amount of transmitted (not reflected) light that enters into the surface. The remaining part is re-emitted and colored by the basecolor of the incident side. Ie, business as usual.

* A total light extinction distance controlls the distance into the surface where light will not travel past.
This is instead of the attenuation color - I know that the same visual effect can be achieved by setting the attenuation color to black - but that is wrong because it is not the same phenomena.

Now we have a self contained simplified extension, with less number of parameters, where the parameters are connected to what they _actually_ control.
One extension that will satisfy the same usecases in less number of parameters and data.

Surely we must strive to have self contained, well parameterized  extensions - we must not keep using the rationale 'we did it like this in extension XYZ'

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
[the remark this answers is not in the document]
rsahlin: In my opinion the overview must clearly outline the goal of the extension.
In the case of a material property extension like this, it makes perfect sense to describe the physical behavior or phenomena that is modelled.
The property description then specifies each parameter.

emackey:
slide 16 from the 2022 PBR webinar

lexaknyazev:
@UX3D-haertl I'll update the object model after this is merged.

rsahlin:
Confusion likely comes from specifying (physically plausible) behavior using (only) surface properties and the concept of 'thin walled' or infinetly thin materials, which I generally regard as unwise. 

For this extension I think we must focus on usecases, leaf and wax candle, and figure out what we need to model this behavior.
Looking at how extensions are used in an environment with tens of thousands of models being converted, QA'ed handled and updated over time.
Also taking animation pointer into consideration, - we are just learning how this may affect specifications and maintenance. 
Keep in mind that complexity will increase when multiple extensions are targetted.

I strongly believe in simplification - less is more - use of two extensions instead of one will increase complexity. 
In this case I would rather not follow the trend of materials transmission/volume.
Instead, for the usecases we have I think a simpler, easier to manage over time, design that models behavior using one extension is to be preferred.

From what I can tell our usecases can be fulfilled by adding two optional parameters:
1: Total light extinction distance
2: Thickness texture 

What, if any, is the benefit of splitting and declaring surface properties and interior/scattering behavior into separate extensions?

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
emackey: I believe this thread was resolved during last week's PBR TSG meeting.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,404 @@
...
+    <td align="center">1.0</td>
+  </tr>
+  <tr>
+    <td colspan="5" align="center">
+      <em>Single-sided plane in a symmetric light setup. <code>baseColorTexture</code> and <code>diffuseTransmissionColorTexture</code> use textures that represent the different sides of the one-dollar bill. Series shows the setup at varying values of <code>diffuseTransmissionFactor</code>.</em>
+    </td>
+  </tr>
+</table>
+
+## Implementation
```
emackey, line 176: This header should say "Material Structure Updates"

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,404 @@
...
+
+*This section is normative.*
+
+This extension changes the `dielectric_brdf` defined in [Appendix B](../../../../specification/2.0/README.md#appendix-b-brdf-implementation)
```
emackey, line 180: This URL should be: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#material-structure

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,404 @@
...
+    base = mix(
+      diffuse_brdf(color = baseColor),
+      diffuse_btdf(color = diffuseTransmissionColor),
+      diffuseTransmission),
+    layer = specular_brdf(α = roughness ^ 2)
+  )
+```
+Increasing the strength of the diffuse transmission effect using the `diffuseTransmissionFactor` parameter takes away energy from the diffuse reflection BSDF and passes it to the diffuse transmission BSDF. The specular reflection BSDF and Fresnel weighting are not affected.
+
+*This section is non-normative.*
```
emackey, line 203: Add a header here called "Implementation"

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
emackey:
```suggestion
## Material Structure Updates
```

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -354,7 +364,10 @@ diffuse_bsdf = mix(
   </tr>
   <tr>
     <td colspan="5" align="center">
-      <em>Dragon with fixed <code>diffuseTransmissionFactor</code> of 1.0 and varying <code>transmissionFactor</code>.</em>
+      <em>Dragon with fixed <code>diffuseTransmissionFactor</code> of 1.0 and varying <code>transmissionFactor</code>.<br>
+        <p style="font-size:0.7em;">
+        (dragon model from <a href="http://graphics.stanford.edu/data/3Dscanrep/">Stanford 3D Scanning Repository</a>)
```
lexaknyazev, line 369: nit: use `https`.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,423 @@
...
+<div align="center">
+<figure style="text-align:center">
+<p float="left">
+<img src="figures/leaves.jpg" height="350" />
+<img src="figures/candle_0.5.jpg" height="350" />
+</p>
+<figcaption>
+<em>Left: (Thin) translucent leaves/foliage, Right: (Thick) translucent candle with colored volume attenuation and translucent surface.
+</em><br>
+<p style="font-size:0.7em;">(candle model by <a href="https://sketchfab.com/lucatorcigliani">lucatorcigliani</a> with modifications by @emackey. Original <a href="https://sketchfab.com/3d-models/candle-in-a-glass-ec6cf358c4dd4cbda7be67f7846cbc9e">source</a> CC-BY 4.0)<p>
```
lexaknyazev, line 37: Leaves image credit?

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,423 @@
...
+| **diffuseTransmissionTexture**      | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A texture that defines the percentage of non-specularly reflected light that is diffusely transmitted through the surface. Stored in the alpha (`A`) channel. Will be multiplied by the diffuseTransmissionFactor.                 | No                       |
+| **diffuseTransmissionColorFactor**  | `number[3]`                                                         | The color that modulates the transmitted light.                                                                                                                                 | No, default: `[1, 1, 1]` |
+| **diffuseTransmissionColorTexture** | [`textureInfo`](/specification/2.0/README.md#reference-textureInfo) | A texture that defines the color that modulates the diffusely transmitted light, stored in the `RGB` channels and encoded in sRGB. This texture will be multiplied by diffuseTransmissionColorFactor. | No                       |
+
```
lexaknyazev, line 77: Since these properties follow the regular `*Factor`/`*Texture` glTF pattern, the four subsequent sections should be refactored as two: "Diffuse Transmission" and "Diffuse Transmission Color" without using property names as headings.

_Similar change is needed for `KHR_materials_transmission` but that's irrelevant here._

lexaknyazev (approved):

emackey (approved):
The Khronos PBR TSG voted today (2024-11-18) to move this extension to "Release Candidate" status.  This means the parameters of the extension will not change prior to ratification unless major problems are found.  All glTF implementers are encouraged to test this extension in their respective implementations and provide feedback on any roadblocks encountered.

extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md
```diff
@@ -0,0 +1,421 @@
...
+* Tobias Haeussler, Dassault Systèmes [@proog128](https://github.com/proog128)
+* Bastian Sdorra, Dassault Systèmes [@bsdorra](https://github.com/bsdorra)
+* Ed Mackey, AGI [@emackey](https://twitter.com/emackey)
+
+TODO
+
+## Status
+
+Experimental
```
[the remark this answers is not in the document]
emackey: I didn't have push access here, we'll update it in the main repo.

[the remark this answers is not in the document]
emackey: Done

Issue #1427, opened 2018-08-22
Should Occlusion apply to IBL and lights or just IBL?


sbtron:
Some background discussion here - https://github.com/BabylonJS/Exporters/issues/291

Should occlusion affect IBL and lights both or just IBL?
Currently Babylon was applying to occlusion to both (with an option to just do it for IBL) but other packages might just be doing it for IBL.  The core spec doesn't specify anything on this but as lights and environment extensions come along we do need to specify what the expected behavior is. 
 
Curious to know what other packages are doing and if there is a standard practice around this already?

//cc @echadwick-wayfair @sebavan

echadwick-wayfair:
Ambient occlusion is meant to occlude only indirect lighting (ambient), not direct lighting (light sources). 

Direct lighting is light emitted from a strong light source, and striking our model's surface directly, with nothing occluding the light (or very little occlusion, for example our atmosphere somewhat attenuates the direct lighting from the sun, but it's still pretty bright). Direct light examples are the sun, a light bulb, a car headlight, etc.

Indirect lighting occurs when direct lighting bounces off surfaces and illuminates our model's surface. These are secondary, tertiary, etc. light bounces. This is also called ambient lighting. 

Ambient occlusion mimics the way a model's crevices reduce the effect of secondary bounces (as the lighting bounces more and more across the creases of a surface, more of the energy is absorbed).

Some early research might help illuminate this (sorry!) 
https://link.springer.com/chapter/10.1007/978-3-7091-6453-2_5

echadwick-wayfair:
But... if the IBL contains imagery of light sources and these are stored in HDR, then the IBL is basically casting direct lighting. Over-bright pixels are casting sun-like lighting into our scene.

So having it as an option seems like it might make the most sense?

emackey:
Our current reference PBR shader [applies the ambient occlusion to both analytic lights and IBL](https://github.com/KhronosGroup/glTF-WebGL-PBR/blob/88eda8c5358efe03128b72b6c5f5f6e5b6d023e1/shaders/pbr-frag.glsl#L297-L298).  This is helpful for engines where self-shadowing is not enabled, as ambient occlusion becomes a form of self-shadowing.

echadwick-wayfair:
Hmm. that could certainly be an option for those who want it. But I believe it shouldn't be the default. 

[...]

PR #1826, opened 2020-06-12
Clarify that AO does not affect direct lighting.

donmccurdy:
Based on https://github.com/KhronosGroup/glTF/issues/915 and https://github.com/KhronosGroup/glTF/issues/1427. 

Another change may still be needed after this, for the second part of https://github.com/KhronosGroup/glTF/issues/915#issuecomment-574413797.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -1181,7 +1181,7 @@ The material definition also provides for additional maps that can also be used 
 
 Materials define the following additional maps:
 - **normal** : A tangent space normal map.
-- **occlusion** : The occlusion map indicating areas of indirect lighting.
+- **occlusion** : The occlusion map indicates areas that receive less diffuse lighting from ambient sources. Direct lighting is not affected.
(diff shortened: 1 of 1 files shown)


rsahlin:
I think that clarifying that it is the diffuse part that is reduced would be a good addition.
Apart from that I belive we should also decide how to handle AO for metal roughness (unreal/substance?)

Is this what you are refering to @donmccurdy ?

donmccurdy:
If this PR's clarifications are correct and useful then I think we should merge them without waiting for additional decisions to be made in https://github.com/KhronosGroup/glTF/issues/915... I'm opening this to ensure that the consensus from https://github.com/KhronosGroup/glTF/issues/1427 is not lost after that issue is closed.

But I have no opinion on the remaining open questions, at this point.

emackey:
This seems small and targeted, so I'm merging it.  Further clarifications are always welcome in new PRs.

PR #1830, opened 2020-06-15
EXT_meshopt_compression extension

zeux:
> Note: this was originally filed as #1702 but was since renamed to EXT_meshopt_compression, hence the new PR

[Rendered version](https://github.com/KhronosGroup/glTF/blob/8ad3a24db1cdd98c3dfa20ead9df68155a203222/extensions/2.0/Vendor/EXT_meshopt_compression/README.md) (Updated 10/2/2020)

This is an extension designed to reduce the transmission size of glTF files. The structure of this extension is very different from KHR_draco_mesh_compression, as it works on a per-bufferView basis instead of a per-mesh basis - with all of the remaining glTF schema in tact.

During loading of a compressed glTF file, loaders are expected to decompress the compressed bufferView data and then proceed with the loading as usual - this, for example, is compatible with decoding the buffer views directly into GPU-visible buffers.

Any type of data can be compressed; however, this extension isn't exactly a general purpose compressor. Instead, there are several algorithms tailored to different kinds of data that may be stored inside the buffer view - for all of the algorithms, they are designed to provide extremely fast decoding (on modern desktop CPUs decoding runs at ~2+ GB/s for native code and at ~1 GB/s in Wasm, using Wasm SIMD to accelerate parts of the processing). The decoders are implemented in meshoptimizer (https://github.com/zeux/meshoptimizer), and files compressed with this extension can be produced by gltfpack (https://github.com/zeux/meshoptimizer/tree/master/gltf) - or, of course, any other tool that is compliant with this specification.

For each bufferView, an appropriate compression technique must be picked to maximize the compression ratio of the data. The extension provides a compression mode for attribute data (suitable for mesh attributes, animation keys or values, instance transform components), triangle indices (suitable for mesh index data when using triangle list primitive) and indices (suitable for mesh index data for other primitives as well as sparse indices for general accessor storage). It's the job of the encoder to split the compressible data into bufferViews as necessary; additionally, for all compression modes preparing data for compression is important to get high compression ratios - this includes finding the optimal order of the data elements, quantizing them with KHR_mesh_quantization, and possibly other kinds of preprocessing such as animation data resampling.

Additionally, for attribute storage (for mesh/animation/etc.), the encoder may decide to use compression filters. These provide extra savings on top of attribute compression provided by this extension at the cost of extra precision loss; for example, instead of storing quantized quaternion values, filters allow storing only 3 components of a quaternion and reconstructing the quaternion from that - this can result in small precision loss. All filters are designed to be variable bit rate - that is, the encoder can pick the optimal number of bits used for data encoded by the filters, and fewer bits of data used will result in the attribute compression using fewer bytes to represent the data stream.

Unlike Draco or Basis ETC1 that have capable entropy coders embedded into the format, this extension doesn't employ Huffman or rANS or similar entropy coders; instead all algorithms are designed to reduce the data size as much as possible while still representing the data as a byte sequence; the expectation is that for maximum compression ratio a general-purpose compressor such as gzip (which is commonplace for asset delivery on the web) can be used as well. Noteworthy is that this extension acts as a pre-processor for gzip/et al in that by itself, gzip can't get anywhere close to the level of compression this extension provides; so the expectation is that this extension can compress data, gzip can compress the resulting compressed result further, but gzip or an equivalent compressor is optional.

For mesh data, this extension usually is slightly less efficient than Draco in terms of the transmission size. However, it's completely general (Draco glTF can't compress point clouds, non-triangle geometry and morph targets), and supports mesh, animation, instance data and general index/attribute compression with a simple specification.


zeux:
In terms of compression ratio, here are the results of compressing data using this extension compared to using Draco:

```
2CylinderEngine                          gltfpack 496864               gltfpack.gz 261707               draco 499784               draco.gz 402837               gltfpack/draco 0.994                gltfpack.gz/draco.gz 0.650
AntiqueCamera                            gltfpack 166612               gltfpack.gz 115040               draco 130240               draco.gz 114831               gltfpack/draco 1.279                gltfpack.gz/draco.gz 1.002
Avocado                                  gltfpack 8232                 gltfpack.gz 5705                 draco 6616                 draco.gz 5300                 gltfpack/draco 1.244                gltfpack.gz/draco.gz 1.076
BarramundiFish                           gltfpack 29924                gltfpack.gz 23982                draco 23368                draco.gz 21069                gltfpack/draco 1.281                gltfpack.gz/draco.gz 1.138
BoomBox                                  gltfpack 47012                gltfpack.gz 37594                draco 37128                draco.gz 34118                gltfpack/draco 1.266                gltfpack.gz/draco.gz 1.102
BrainStem                                gltfpack 370512               gltfpack.gz 256496               draco 380872               draco.gz 247774               gltfpack/draco 0.973                gltfpack.gz/draco.gz 1.035
Buggy                                    gltfpack 2230596              gltfpack.gz 1047707              draco 2733216              draco.gz 2243675              gltfpack/draco 0.816                gltfpack.gz/draco.gz 0.467
CesiumMan                                gltfpack 50356                gltfpack.gz 34063                draco 45860                draco.gz 36237                gltfpack/draco 1.098                gltfpack.gz/draco.gz 0.940
CesiumMilkTruck                          gltfpack 29664                gltfpack.gz 15927                draco 29640                draco.gz 16732                gltfpack/draco 1.001                gltfpack.gz/draco.gz 0.952
Corset                                   gltfpack 133356               gltfpack.gz 85150                draco 113724               draco.gz 105913               gltfpack/draco 1.173                gltfpack.gz/draco.gz 0.804
DamagedHelmet                            gltfpack 136528               gltfpack.gz 105357               draco 102016               draco.gz 91620                gltfpack/draco 1.338                gltfpack.gz/draco.gz 1.150
Duck                                     gltfpack 27636                gltfpack.gz 23031                draco 21096                draco.gz 19471                gltfpack/draco 1.310                gltfpack.gz/draco.gz 1.183
FlightHelmet                             gltfpack 685856               gltfpack.gz 498758               draco 519572               draco.gz 473156               gltfpack/draco 1.320                gltfpack.gz/draco.gz 1.054
GearboxAssy                              gltfpack 2209512              gltfpack.gz 1271787              draco 2212596              draco.gz 1845696              gltfpack/draco 0.999                gltfpack.gz/draco.gz 0.689
Lantern                                  gltfpack 46200                gltfpack.gz 29917                draco 35548                draco.gz 32146                gltfpack/draco 1.300                gltfpack.gz/draco.gz 0.931
Monster                                  gltfpack 40800                gltfpack.gz 23526                draco 53216                draco.gz 31308                gltfpack/draco 0.767                gltfpack.gz/draco.gz 0.751
ReciprocatingSaw                         gltfpack 804880               gltfpack.gz 457507               draco 869368               draco.gz 702841               gltfpack/draco 0.926                gltfpack.gz/draco.gz 0.651
RiggedFigure                             gltfpack 12744                gltfpack.gz 6973                 draco 10728                draco.gz 6538                 gltfpack/draco 1.188                gltfpack.gz/draco.gz 1.067
SciFiHelmet                              gltfpack 670708               gltfpack.gz 537936               draco 182420               draco.gz 173450               gltfpack/draco 3.677                gltfpack.gz/draco.gz 3.101
Sponza                                   gltfpack 1970516              gltfpack.gz 843461               draco 1678516              draco.gz 1453378              gltfpack/draco 1.174                gltfpack.gz/draco.gz 0.580
Suzanne                                  gltfpack 23588                gltfpack.gz 18963                draco 17912                draco.gz 16285                gltfpack/draco 1.317                gltfpack.gz/draco.gz 1.164
WaterBottle                              gltfpack 32952                gltfpack.gz 20436                draco 24384                draco.gz 21490                gltfpack/draco 1.351                gltfpack.gz/draco.gz 0.951
```

These were generated as follows:

```
gltfpack -i scene.gltf -o scene_noq.glb -noq
gltfpack -i scene_noq.glb -o scene_glp.glb -cc
gzip -9 -k scene_glp.glb
gltf-pipeline -i scene_noq.glb -o scene_drc.glb -d -t --draco.quantizeNormalBits 8
gzip -9 -k scene_drc.glb
```

Using gltfpack as a pre-processor in addition to Draco is important to establish a level playing field, as gltfpack does a lot of high level scene processing as well.

Results as a Google Spreadsheet: https://docs.google.com/spreadsheets/d/1V0jls9QSb7DRE3-JseCHRcnAzIAp-uSBssnEvx2dQxI/edit#gid=0

zeux:
In terms of decompression time, this tends to vary a bit between the modes and filters used. Using two models as a test, Buggy.gltf from glTF-Sample-Models (300K triangles, 245K vertices) and Thai Buddha (https://sketchfab.com/3d-models/thai-buddha-cba029e262bd4f22a7ee4fcf064e22ee, 6M triangles, 3M vertices), we get the following timings on Chrome Canary on i7 8700K:

Buggy.gltf:
meshopt: decodeVertexBuffer 4.5 ms, decodeIndexBuffer 2.7 ms, total ~7.5ms
Draco: total 186 ms (25x slower)

Thai Buddha:
meshopt: decodeVertexBuffer 78ms, decodeIndexBuffer 118ms, total 196ms
Draco: 6.19s DecodeBufferToMesh, 7.63s total (39x slower)

Based on the size and performance metrics, it looks like while Draco still is a viable alternative in cases when download size is critical and the content consists purely of elements that Draco supports well, this extension may provide a better balance and supports compression for all non-texture data that can be encoded in glTF.

zeux:
@donmccurdy @lexaknyazev JFYI I've resubmitted this with the new extension name and the enum changes we discussed on the previous PR. Also all numbers in this PR are now current (old PR had some numbers from earlier implementation that was less efficient). This PR should now be the central one, please let me know if you have further feedback! (also please let me know if you'd like me to add you to the contributor list, I don't know how that works)

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,476 @@
...
+This is useful to avoid confusion, and can also be used by loaders that support the extension to skip loading of these buffers.
+
+When a buffer is marked as a fallback buffer, the following must hold:
+
+- All references to the buffer must come from `bufferView`s that have a `EXT_meshopt_compression` extension specified
+- No references to the buffer may come from `EXT_meshopt_compression` extension JSON
+
+If a fallback buffer doesn't have a URI and doesn't refer to the GLB binary chunk, it follows that `EXT_meshopt_compression` must be a required extension.
+
+## Compressing geometry data
```
donmccurdy, line 143: Should this section be marked as non-normative, similar to https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#appendix-b-brdf-implementation?

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,476 @@
...
+
+Vertex data should be quantized using the appropriate representation; this extension cleanly interacts with KHR\_mesh\_quantization by compressing already quantized data.
+
+Morph targets can be treated identically to other vertex attributes, as long as vertex order optimization is performed on all target streams at the same time. It is recommended to use quantized storage for morph target deltas, possibly with a narrower type than that used for baseline values.
+
+When storing vertex data, mode 0 (attributes) should be used; for index data, mode 1 (triangles) or mode 2 (indices) should be used instead. Mode 1 only supports triangle list storage; indices of other topology types can be stored using mode 2. The use of triangle strip topology is not recommended since it's more efficient to store triangle lists using mode 1.
+
+Using filter 1 (octahedral) for normal/tangent data may improve compression ratio further.
+
+## Compressing animation data
```
donmccurdy, line 162: Should this section be marked as non-normative, similar to https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#appendix-b-brdf-implementation?

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,476 @@
...
+
+To reduce the number of keyframes, encoders can either selectively remove keyframes that don't contribute to the resulting movement, resulting in sparse input/output data, or resample the keyframes uniformly, resulting in uniformly dense data. Resampling can be beneficial since it means that all animation channels in the same animation can share the same input accessor, and provides a convenient quality vs size tradeoff, but it's up to the encoder to pick the optimal strategy.
+
+Additionally it's important to identify tracks with the same output value and use a single keyframe for these.
+
+To reduce the size of each keyframe, rotation data should be quantized using 16-bit normalized components; for additional compression, the use of filter 2 (quaternion) is recommended. Translation/scale data can be compressed using filter 3 (exponential) with the same exponent used for all three vector components.
+
+After pre-processing, both input and output data should be stored using mode 0 (attributes).
+
+# Appendix A: Bitstream
```
donmccurdy, line 174: Should this section be marked as non-normative, similar to https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#appendix-b-brdf-implementation?

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,476 @@
...
+When stored in a GLB file, the dummy buffer should have index 1 or above, to avoid conflicts with GLB binary buffer.
+
+This extension allows buffers to be tagged as fallback by using the `fallback` attribute as follows:
+
+```json
+{
+	"byteLength": 1432878,
+	"extensions": {
+		"EXT_meshopt_compression": {
+			"fallback": true
```
donmccurdy, line 128: The `fallback` property should be included in a schema file, as well.

zeux: Done!

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Done!

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Done!

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,476 @@
...
+
+While the extension JSON specifies a separate buffer to source compressed data from, the parent `bufferView` must also have a valid `buffer` reference as per glTF 2.0 spec requirement. To produce glTF files that *require* support for this extension and don't have uncompressed data, the referenced buffer can contain no URI as follows:
+
+```json
+{ "byteLength": 1432878 }
+```
+
+When stored in a GLB file, the dummy buffer should have index 1 or above, to avoid conflicts with GLB binary buffer.
+
+This extension allows buffers to be tagged as fallback by using the `fallback` attribute as follows:
```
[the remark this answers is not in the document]
zeux: I added "optionally" to the sentence, please let me know if this is sufficient or if I can word it more carefully - indeed, "fallback" annotations are purely optional.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: I'm not as sure about this one - this designates the format of the compressed data, and so it seems normative to me (assuming I understand the meaning of the word). It's essentially the equivalent of Draco or Basis bitstream specification.

donmccurdy: Oops, yes. On second thought I agree.

zeux:
@donmccurdy @lexaknyazev Any news on this? Looking at gltfpack releases and planning the next one in October, it would be good to get some early indication as to the fate of this extension / the modification to JSON structure if any, just to make it easier on the users and avoid multiple incompatible implementations in the wild. Please let me know if I can do anything to move this along.

donmccurdy:
In my opinion this can be merged whenever you feel it is ready. An EXT_ extension does not have formal approval requirements beyond the existence of multiple independent implementations. We are always happy to review for technical and compatibility concerns, but I don't see any likely issues here.

emackey:
@lilleyse You're using this too, right?  Any concern with merging this as-is?

zeux:
@donmccurdy Ah - I don't really know what the process is so was unclear on the next steps. FWIW the binary encoding has been stable for close to 6 months now (since end of March or thereabouts?), with no expected changes on that front; the JSON encoding has switched to using string-based enums as of June but again no changes since then and none anticipated. So I feel like this is ready.

I would propose that the next steps could be:

- We merge this as is (note that it has the "draft" status on the page), which would make it easier to merge implementations to other projects since there'd be a reference spec in the tree
- I can contribute implementations for this extension to Three.js/Babylon.js/cgltf to make sure we have reasonable coverage
- We can let this extension bake for a bit to make sure no issues emerge
- At some point we change the status to "Complete" (or, if any changes become necessary - I'm not anticipating any structural changes but of course the spec may need to be clarified based on feedback - we apply them first).

Does that sound reasonable?

donmccurdy:
This all sounds good to me. 👍

donmccurdy:
@bghgary @lilleyse @lexaknyazev any concerns?

bghgary:
Seems like a good plan. 👍

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+During the decoding process, decoder maintains four variables:
+
+- `next`: an integer referring to the expected next unique index (also known as high-watermark), starts at 0
+- `last`: an integer referring to the last encoded index, starts at 0
+- `edgefifo`: a 16-entry FIFO with two vertex indices in each entry; initial contents is undefined
+- `vertexfifo`: a 16-entry FIFO with a vertex index in each entry; initial contents is undefined
+
+To decode each triangle, the decoder needs to analyze the `code` byte, read additional bytes from `data` as necessary, and update the internal state correctly. The `code` byte encoding is optimized to reach a single byte per triangle in most common cases; the resulting data can often be compressed by a general purpose compressor running on the resulting .bin/.glb file.
+
+When extra data is necessary to decode a triangle and it represents an index value, the decoder uses varint-7 encoding, which encodes an integer as one or more bytes, with the byte with the 0 most significant bit terminating the sequence:
```
lexaknyazev, line 266: I think another name for this encoding is "Unsigned LEB128". Consider adding such a reference.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+Filters are functions that transform each encoded attribute. For each filter, this document specifies the transformation used for decoding the data; it's up to the encoder to pick the parameters of the encoding for each element to balance quality and precision.
+
+For performance reasons the results of the decoding process are specified to one unit in last place (ULP) in terms of the decoded data, e.g. if a filter results in a 16-bit signed normalized integer, decoding may produce results within 1/32767 of specifed value.
+
+## Filter 1: octahedral
+
+Octahedral filter allows to encode unit length 3D vectors (normals/tangents) using octahedral encoding, which results in a more optimal quality vs precision tradeoff compared to storing raw components.
+
+The input to the filter is four 8-bit or 16-bit components, where the first two specify the X and Y components in octahedral encoding encoded as signed normalized K-bit integers, the third component explicitly encodes 1.0 as a K-bit integer, and the last component may contain arbitrary data (which is useful for tangents).
```
lexaknyazev, line 398: To avoid any ambiguity, please
- mention whether K has any constraints;
- define "signed normalized K-bit integer" (e.g., two-complement with a sign in the MSB);
- add an example of "1.0 encoded as a K-bit integer" as only the first two components are specified as "signed normalized", mention that it's used to determine K, and what to do with invalid values in this component;
- define `INTN_MAX`.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+	float32_t one = input[3];
+
+	float32_t x = input[0] / one;
+	float32_t y = input[1] / one;
+	float32_t z = 1.0 - abs(x) - abs(y);
+
+	// octahedral fixup for negative hemisphere
+	float32_t t = min(z, 0.0);
+
+	x += copysign(t, x);
```
lexaknyazev, line 414: Consider either defining `copysign` in the spec or mentioning a normative reference for it (e.g. C99).

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+	y += copysign(t, y);
+
+	// renormalize (x, y, z)
+	float32_t l = sqrt(x * x + y * y + z * z);
```
lexaknyazev, line 418: For better readability, consider renaming `l` to `length`.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+	output[(maxcomp + 1) % 4] = round(x * 32767.0);
+	output[(maxcomp + 2) % 4] = round(y * 32767.0);
+	output[(maxcomp + 3) % 4] = round(z * 32767.0);
+	output[(maxcomp + 0) % 4] = round(w * 32767.0);
+}
+```
+
+## Filter 3: exponential
+
+Exponential filter allows to encode floating point values with a range close to the full range of a 32-bit floating point value, but allows more control over the exponent/mantissa to trade quality for precision, and has a bit structure that is more optimally aligned to the byte boundary.
```
lexaknyazev, line 466: Maybe mention here that byte boundary alignment for mantissa / exponent facilitates better compression.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+```
+float32_t decode(int32_t input) {
+	int32_t e = input >> 24;
+	int32_t m = (input << 8) >> 8;
+	return pow(2.0, e) * m;
+}
+```
+
+The valid range of `e` is [-100, +100], which facilitates performant implementations.
```
lexaknyazev, line 480: What should decoders do with values out of this range?

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+glTF files come with a variety of binary data - vertex attribute data, index data, morph target deltas, animation inputs/outputs - that can be a substantial fraction of the overall transmission size. To optimize for delivery size, general-purpose compression such as gzip can be used - however, it often doesn't capture some common types of redundancy in glTF binary data.
+
+To achieve optimal compression rates for mesh geometry data, glTF supports Draco mesh compression through KHR\_draco\_mesh\_compression extension. However, this comes at a cost of decoding performance (Draco decoding takes an appreciable time which can start offsetting the loading time for faster connections) and inability to directly upload buffer view data to GPU storage or, for most implementations, use quantized storage in the output. Additionally, animation and morph target data isn't supported.
```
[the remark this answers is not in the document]
zeux: Removed the paragraph and added another paragraph that's more neutral and describes applications of this extension instead.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+This extension provides a different option for compressing binary data. Similarly to supercompressed textures (see KHR\_image\_ktx2), this extension assumes that the buffer view data is optimized for GPU efficiency - using quantization and using optimal data order for GPU rendering - and provides a compression layer on top of bufferView data. Each bufferView is compressed in isolation which allows the loaders to maximally efficiently decompress the data directly into GPU storage.
```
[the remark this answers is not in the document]
zeux: Fixed

extensions/2.0/Vendor/EXT_meshopt_compression/schema/buffer.EXT_meshopt_compression.schema.json
```diff
@@ -0,0 +1,17 @@
...
+    "type": "object",
+    "description": "Compressed data for bufferView.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "fallback": {
+            "type": "boolean",
+            "description": "Set to true to indicate that the buffer is only referenced by bufferViews that have EXT_meshopt_compression extension and as such doesn't need to be loaded.",
+            "default": false
+        },
+        "extras": { }
```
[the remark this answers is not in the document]
zeux: Fixed

extensions/2.0/Vendor/EXT_meshopt_compression/schema/bufferView.EXT_meshopt_compression.schema.json
```diff
@@ -0,0 +1,50 @@
...
+            "type": "string",
+            "description": "The compression mode.",
+            "enum": [ "ATTRIBUTES", "TRIANGLES", "INDICES" ]
+        },
+        "filter": {
+            "type": "string",
+            "description": "The compression filter.",
+            "enum": [ "NONE", "OCTAHEDRAL", "QUATERNION", "EXPONENTIAL" ]
+        },
+        "extras": { }
```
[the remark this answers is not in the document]
zeux: Fixed

extensions/2.0/Vendor/EXT_meshopt_compression/schema/bufferView.EXT_meshopt_compression.schema.json
```diff
@@ -0,0 +1,50 @@
...
+            "type": "integer",
+            "description": "The length of the compressed data in bytes.",
+            "minimum": 1
+        },
+        "byteStride": {
+            "type": "integer",
+            "description": "The stride, in bytes.",
+            "minimum": 4,
+            "maximum": 252,
+            "multipleOf": 4,
```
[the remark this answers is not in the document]
zeux: Nice catch! I've removed the multipleOf specification since the actual behavior here depends on the mode and as such is too complex to represent here, and corrected a few other mistakes in the same schema.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+Filter specifies the algorithm used to transform the data after decompression, and can be one of:
+
+- Filter 0: none. Attribute data is used as is.
+- Filter 1: octahedral. Suitable for storing unit length vectors (normals/tangents) as 4-byte or 8-byte values with variable precision octahedral encoding.
+- Filter 2: quaternion. Suitable for storing rotation data for animations or instancing as 8-byte values with variable precision max-component encoding.
+- Filter 3: exponential. Suitable for storing floating point data as 4-byte values with variable mantissa precision.
+
+The filters are detailed further in Appendix B (Filters).
```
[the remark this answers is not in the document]
zeux: Done

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+Compression mode specifies the bitstream layout and the algorithm used to decompress the data, and can be one of:
+
+- Mode 0: attributes. Suitable for storing sequences of values of arbitrary size, relies on exploiting similarity between bytes of consecutive elements to reduce the size.
+- Mode 1: triangles. Suitable for storing indices that represent triangle lists, relies on exploiting topological redundancy of consecutive triangles.
+- Mode 2: indices. Suitable for storing indices that don't represent triangle lists, relies on exploiting similarity between consecutive elements.
+
+In all three modes, the resulting compressed byte sequence is typically noticably smaller than the buffer view length, *and* can be additionally compressed by using a general purpose compression algorithm such as Deflate for the resulting glTF file (.glb/.bin).
+
+The format of the bitstream is specified in Appendix A (Bitstream).
```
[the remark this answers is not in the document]
zeux: Done

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+The codecs used by this extension can represent geometry exactly, replicating both vertex and index data without changes in contents or order. However, to get optimal compression, it's necessary to pre-process the data.
+
+To get optimal compression, encoders should optimize vertex and index data for locality of reference. Specifically:
+
+- Triangle order should be optimized to maximize the recency of previously encountered vertices; this is similar to optimizing meshes for vertex reuse aka post-transform cache in GPU hardware.
+- Vertex order should be linearized in the order that vertices appear in the index stream to get optimal index compression
+
+When index data is not available (e.g. point data sets) or represents topology with a lot of seams (e.g. each triangle has unique vertex indices because it specifies flat-shaded normal), encoders could additionally optimize vertex data for spatial locality, so that vertices close together in the vertex stream are close together in space.
+
+Vertex data should be quantized using the appropriate representation; this extension cleanly interacts with KHR\_mesh\_quantization by compressing already quantized data.
```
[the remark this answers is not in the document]
zeux: Fixed

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+## Mode 0: attributes
+
+Attribute compression exploits similarity between consecutive elements of the buffer by encoding deltas. The deltas are stored for each separate byte which makes the codec more versatile since it can work with components of various sizes. Additionally, the elements are stored with bytes deinterleaved, which means that sequences of deltas are more easily compressible by some general purpose compressors that may run on the resulting data.
+
+To facilitate efficient decompression, deinterleaving and delta encoding are performed on attribute blocks instead of on the entire buffer; within each block, elements are processed in groups of 16.
+
+The encoded stream structure is as follows:
+
+- Header byte, which must be equal to `0xa0`
```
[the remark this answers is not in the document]
zeux: Thanks! I've changed all hexadecimal constants to use decorated lower-case hexadecimal and consistently used XYZW as upper-case placeholders for places where it's denoting a variable that is referred to later.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Thanks, added a reference here and in another section, both link to Wikipedia just in case.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Thanks, I've tightened this description a bit and added language around the 1.0 encoding (both what is encoded and how it's used).

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Defined copysign as C99 with an extra clarification.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Done

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Yup!

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: The behavior is unspecified (this range limit exists to make it possible for decoders to simply shift the exponent into the 32-bit floating point range without having to handle overflow conditions that require renormalizing the mantissa), left a comment to that effect + noted that encoders are expected to clamp.

zeux:
@lexaknyazev Thanks a lot for the feedback, I believe I've addressed it (mostly tried to use one commit per comment for easier diff review)

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,484 @@
...
+
+## Filter 1: octahedral
+
+Octahedral filter allows to encode unit length 3D vectors (normals/tangents) using octahedral encoding, which results in a more optimal quality vs precision tradeoff compared to storing raw components.
+
+The input to the filter is four 8-bit or 16-bit components, where the first two specify the X and Y components in octahedral encoding encoded as signed normalized K-bit integers (4 <= K <= 16, integers are stored in two's complement format), the third component explicitly encodes 1.0 as a signed normalized K-bit integer, and the last component may contain arbitrary data (which is useful for tangents).
+
+The encoding of the third component allows to compute K for each vector independently from the bit representation, and must encode 1.0 precisely which is equivalent to `1 << (K - 1)` as an integer; values of the third component that aren't equal to `1 << (K - 1)` for a valid `K` are invalid and the result of decoding such vectors is unspecified.
```
lexaknyazev, line 400: Maybe I'm missing something, but here's what it looks like for `K = 8`:
```
1 << (8 - 1) = 128 (10000000)
```
However, 9 bits are required to represent 128 as a signed integer.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,484 @@
...
+
+Filters are functions that transform each encoded attribute. For each filter, this document specifies the transformation used for decoding the data; it's up to the encoder to pick the parameters of the encoding for each element to balance quality and precision.
+
+For performance reasons the results of the decoding process are specified to one unit in last place (ULP) in terms of the decoded data, e.g. if a filter results in a 16-bit signed normalized integer, decoding may produce results within 1/32767 of specified value.
+
+## Filter 1: octahedral
+
+Octahedral filter allows to encode unit length 3D vectors (normals/tangents) using octahedral encoding, which results in a more optimal quality vs precision tradeoff compared to storing raw components.
+
+The input to the filter is four 8-bit or 16-bit components, where the first two specify the X and Y components in octahedral encoding encoded as signed normalized K-bit integers (4 <= K <= 16, integers are stored in two's complement format), the third component explicitly encodes 1.0 as a signed normalized K-bit integer, and the last component may contain arbitrary data (which is useful for tangents).
```
lexaknyazev, line 398: When set, should extra bits (e.g. top 6 for K = 10) in the first two components be ignored (masked out) or should the decoder fail?

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Sorry - late night edits. It's (1 << (8 - 1)) -1, will fix.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Will call this out explicitly, the top unused bits need to be equal to the sign bit of the K-bit integer (which is important for decoding performance so that decoder doesn't need to perform sign extension manually with variable K).

lexaknyazev: There's another edge case here: two-complement's asymmetry makes possible values like `-128/127 ≈ -1.008`. Should they be clamped?

zeux: Yup - good point - the expected range for K-bit integers is `-max .. max` with max = `(1 << (K-1)) - 1`. Similarly to the exponent range and 1.0 encoding here and some others, valid streams shouldn't contain values out of range, so that particular one should be invalid. I will add another paragraph here to specify the range and sign extension behavior more precisely.

zeux: Done

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Fixed

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -0,0 +1,488 @@
...
+	y += copysign(t, y);
+
+	// renormalize (x, y, z)
+	float32_t len = sqrt(x * x + y * y + z * z);
+
+	x /= len;
+	y /= len;
+	z /= len;
+
+	output[0] = round(x * INTN_MAX);
```
lexaknyazev, line 428: Last nit: while it's certainly possible to derive that `INTN_MAX` is either 127 or 32767 depending on `intN_t` type which depends on something else (`byteStride`?), it would be better to define that for clarity.

lexaknyazev (approved):

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Fixed

extensions/2.0/Vendor/EXT_meshopt_compression/schema/bufferView.EXT_meshopt_compression.schema.json
```diff
@@ -0,0 +1,50 @@
...
+        },
+        "byteLength": {
+            "type": "integer",
+            "description": "The length of the compressed data in bytes.",
+            "minimum": 1
+        },
+        "byteStride": {
+            "type": "integer",
+            "description": "The stride, in bytes.",
+            "minimum": 4,
```
lexaknyazev, line 26: Should this be `2`?

extensions/2.0/Vendor/EXT_meshopt_compression/schema/buffer.EXT_meshopt_compression.schema.json
```diff
@@ -0,0 +1,18 @@
...
+        "fallback": {
+            "type": "boolean",
+            "description": "Set to true to indicate that the buffer is only referenced by bufferViews that have EXT_meshopt_compression extension and as such doesn't need to be loaded.",
+            "default": false
+        },
+        "extensions": { },
+        "extras": { }
+    },
+    "required": [
+    ]
```
lexaknyazev, line 17: Please remove the empty array here.

zeux: Done

extensions/2.0/Vendor/EXT_meshopt_compression/schema/bufferView.EXT_meshopt_compression.schema.json
zeux: Yup - thanks! Fixed.

extensions/2.0/Vendor/EXT_meshopt_compression/schema/bufferView.EXT_meshopt_compression.schema.json
```diff
@@ -0,0 +1,50 @@
...
+        "mode": {
+            "type": "string",
+            "description": "The compression mode.",
+            "enum": [ "ATTRIBUTES", "TRIANGLES", "INDICES" ]
+        },
+        "filter": {
+            "type": "string",
+            "description": "The compression filter.",
+            "enum": [ "NONE", "OCTAHEDRAL", "QUATERNION", "EXPONENTIAL" ]
+        },
```
lexaknyazev, line 43: To be aligned with the readme, the schema should define `filter`'s default value.

lexaknyazev:
@zeux 
Sorry for all the delays with reviewing, this seems to be ready to go now. Feel free to open another PR if/when the status should be updated from "Draft" to "Complete".

zeux:
@lexaknyazev No worries, thanks a lot for the help!

PR #1835, opened 2020-07-01
KHR_materials_clearcoat - Mark BRDF implementation section as non-normative

rsahlin:
In order to be consistent with the current non normative BRDF implementation ([Appendix B in glTF spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#appendix-b-brdf-implementation))  
The BRDF implementation of clearcoat should be marked as non-normative.

This is also an important distinction between 3D Formats and 3D Commerce-Certification, we provide the data parameters needed to visualize PBR models, not normative BRDF implementation.  
The Certification group will standardise and specify how to become a certified viewer.

From a glTF perspective it is fine for implementers to use glTF as needed and any way they see fit- we do suggest how to implement and provide an opensource sample implementation.


rsahlin:
Closed since https://github.com/KhronosGroup/glTF/pull/1836 will do the same.

PR #1836, opened 2020-07-01
Clearcoat: Better separation of non-normative section, update trademarks.

emackey:
As pointed out by @rsahlin in https://github.com/KhronosGroup/glTF/pull/1688#discussion_r448262644, the parameter list is normative, and the implementation notes are non-normative.

Previously this extension had them mixed together without clear demarcation, so I've separated them here.

Also the trademark list at the bottom has been updated, copied from the new one in `KHR_xmp` that includes newer Khronos standards such as ANARI etc.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
@@ -64,7 +64,7 @@ The following parameters are contributed by the `KHR_materials_clearcoat` extens
 
 ## Implementation Notes
 
-__This section is non-normative.__
+*This section is non-normative.*
 
 All implementations should use the same calculations for the BRDF inputs. Implementations of the BRDF itself can vary based on device performance and resource constraints. See [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation) for more details on the BRDF calculations.
 
--- a/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md
@@ -52,7 +52,7 @@ The PBR clearcoat materials are defined by adding the `KHR_materials_clearcoat` 
 
 ### Clearcoat
 
-All implementations should use the same calculations for the BRDF inputs. Implementations of the BRDF itself can vary based on device performance and resource constraints. See [Appendix B](/specification/2.0/README.md#appendix-b-brdf-implementation) for more details on the BRDF calculations.
+The following parameters are contributed by the `KHR_materials_clearcoat` extension:
 
 |                                  | Type                                                                            | Description                            | Required             |
 |----------------------------------|---------------------------------------------------------------------------------|----------------------------------------|----------------------|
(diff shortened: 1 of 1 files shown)


rsahlin:
@emackey lol - we both created basically the same PR at the same time :-)
Only nitpick I can see is that as far as I can tell the glTF document has *This section is non-normative.*
I'm no MD ninja so I don't know if "__This section is non-normative.__" translates to the same thing.....

I'll close my PR :-)

emackey:
Thanks, I didn't intend for bold there, I switched it to italics.  Markdown mixup.

proog128:
There are a few pieces of information that should be normative, otherwise consistency will be hard to achieve:
* Roughness is squared GGX roughness to make it perceptually linear.
* The clearcoat IOR is 1.5 (F0 = 0.04).
* Clearcoat factor/texture are a directional-independent weight for the clearcoat layer. 0 disables it completely (for all incoming/outgoing directions).
* Roughness is fetched from the G channel.
* Clearcoat layer is above emission (clearcoat dims emission at grazing angles).
* Behavior of normal map.

The following should be kept in the non-normative section:
* How layering works.

The following could be added to the non-normative section to help with implementations:
* What BSDF is used (although the shape of the highlight/behavior of roughness should be similar to GGX).
* IBL precomputation etc.

rsahlin:
> * Roughness is squared GGX roughness to make it perceptually linear.
> * The clearcoat IOR is 1.5 (F0 = 0.04).

I'm not sure I agree that this should be normative - maybe F0, which is defined as %4 in the spec - but I think the squared roughess is related to the rest of the BRDF implementation (which is non-normative)
By not mandating the IOR we will allow viewers to dynamically selecting the value as they see fit.

proog128:
F0/IOR is a free parameter in the Fresnel term. Whatever Fresnel term you use in the end (full, Schlick), it has to be given to evaluate the material. With the 4% we make clear that the coat is glass-like/dielectric. Many more complex materials (Standard Surface, VRayMtl, ...) allow this parameter to be set by users. I cannot think of any reasons why a viewer wants to use a different value. In addition, the core spec also defines the reflectance for dielectric materials in the normative part, so it shouldn't be a problem for an extension.

About squared roughness, we somehow have to characterize the BRDF lobe or shape of the specular highlight. We can probably develop some complicated definition, but I think characterizing it as "reasonable similar to GGX with squared roughness" is simpler and more intuitive. It also would allow low-end viewers to use for example a Phong BRDF with some magic formulas to match the highlights. Details (what is "reasonable similar"?) can be defined by certification.

proog128:
I moved some parts of the non-normative implementation section to the normative section on this branch: [DassaultSystemes-Technology:clearcoat-normative](https://github.com/KhronosGroup/glTF/compare/clearcoat-normative...DassaultSystemes-Technology:clearcoat-normative), according to my previous comment. Please tell me what you think. If you like it, maybe it is possible to merge it somehow into this PR? Or should I create a new PR?

emackey:
Thanks @proog128.  I've added that commit to this branch, it looks good to me.

emackey:
Any further comments on this?  OK to merge?

rsahlin:
From my side I think it looks good :-)

A very reasonable balance between what is normative  and implementation notes - I think it will be ok for 3D commerce to certify this in the future by making sure the normative parts are followed. (together with the BRDF certification that is being worked on today)

Issue #1645, opened 2019-07-28
glTF file sample_techniques.gltf in KHR_techniques_webgl extension is invalid


zeux:
The file extensions/2.0/Khronos/KHR_techniques_webgl/examples/sample_techniques.gltf fails validation since it doesn't contain an array of bufferViews but it does have accessors that reference bufferView 0.

zeux:
I'm going to close this because KHR_techniques_webgl status is unclear anyway, and it's likely to be removed or delisted, see #1762.

PR #1852, opened 2020-08-15
Unlist KHR_techniques_webgl.

donmccurdy:
See https://github.com/KhronosGroup/glTF/pull/1762#issuecomment-585788164 — the `KHR_techniques_webgl` extension has not been ratified as a Khronos extension and does not appear likely to be soon. Removing it from the folder to avoid confusion, although we could certainly leave the draft on a branch if that's helpful to someone.


pjcozzi:
@donmccurdy I believe CesiumJS implements this.  Depending on if anyone is using it, we might want to keep the extension spec documentation around.

@lilleyse do you know if (1) any Cesium ion tools generate glTF with `KHR_techniques_webgl` either for 3D Tiles or COLLADA FX, or (2) any tools in the 3D Tiles ecosystem generate 3D Tiles with glTF  with `KHR_techniques_webgl`?  If the answer to (1) or (2) is yes, is `KHR_techniques_webgl` still the right approach?  I recall it being used for custom polyline shaders years ago, which still seems like an OK use case.

donmccurdy:
Yeah, happy to keep it somewhere, I just worry it shouldn't be listed as an unratified `KHR_` indefinitely. Could even rename it to an EXT_ if that makes things easier.

pjcozzi:
@donmccurdy sounds reasonable.  I just want to get a sense of what is being used in the wild today.

@lilleyse?

> @lilleyse do you know if (1) any Cesium ion tools generate glTF with KHR_techniques_webgl either for 3D Tiles or COLLADA FX, or (2) any tools in the 3D Tiles ecosystem generate 3D Tiles with glTF with KHR_techniques_webgl? If the answer to (1) or (2) is yes, is KHR_techniques_webgl still the right approach? I recall it being used for custom polyline shaders years ago, which still seems like an OK use case.

donmccurdy:
Following up from earlier discussion — the extension's presence has lead to some confusion (like https://github.com/KhronosGroup/glTF/issues/1645) so I do think we should be clearer about its status than the lack of a "Complete" status in its wording. Most extension PRs are closed when we decide not to ratify, not merged. But there have been suggestions other than removing it, including:

(a) putting it in a "drafts/" or "deprecated/" folder
(b) keeping it on a branch
(c) renaming to EXT_ and considering it complete.

As others have noted, there's valuable spec design on the extension that we don't want to lose completely, but that isn't really a risk with Git history either. The point was also raised that messaging for `KHR_materials_pbrSpecularGlossiness` may need to change with the upcoming PBR extensions. It may become deprecated (TBD?), but that would have a very different meaning from the state of `KHR_techniques_webgl`, which was never ratified at all.

@pjcozzi @lilleyse if you are using the extension, and plan to continue doing so, would (c) be an acceptable option for you?

/cc @bghgary @emackey

emackey:
For (a), we shouldn't use the term "drafts" because it sounds like new upcoming things, not old outgoing things, and may attract PRs.  "Deprecated" or "Archived" would be better.

And yes, "Deprecated" has a strong connotation when applied to `KHR_materials_pbrSpecularGlossiness`.  We can ask on the PBR Next call whether that's really the message we want to send once PBR Next is complete, but I suspect it is.

I'm somewhat opposed to renaming `KHR_techniques_webgl` to EXT.  The main reason for this particular extension is backwards compatibility to glTF 1.0, and renaming to EXT breaks all backwards compatibility to existing implementations of it.  Also, I think our goal should be to discourage use of it, not present it as an equal alternative to other material extensions.

So, my vote would be to move `KHR_techniques_webgl` to a new folder ("Archived" if deprecated is too strong) and then consider moving `KHR_materials_pbrSpecularGlossiness` to that folder down the road.

donmccurdy:
> I'm somewhat opposed to renaming KHR_techniques_webgl to EXT ... [which] breaks all backwards compatibility to existing implementations of it

That's a fair concern, and I do want to be sensitive to the fact that it's already in use. On the other hand, the extension has none of the IP protection that a KHR extension should have. If we are presenting it as a real option — even if "discouraged" — it should not be in a state that creates IP confusion. Perhaps another option would be to rename here, and just accept that older implementations will continue to support the KHR_ naming? 

Alternatively ... and a little outside anything we've discussed before ... we could just ratify the extension and immediately "archive" it. Not sure how I feel about that. 😕

pjcozzi:
> So, my vote would be to move KHR_techniques_webgl to a new folder ("Archived" if deprecated is too strong) and then consider moving KHR_materials_pbrSpecularGlossiness to that folder down the road.

This works for me.  The top of KHR_techniques_webgl could also clearly state that it was never ratified.

PR #1868, opened 2020-09-22
Added note about IOR and roughness for transmission

elalish:
I don't consider this a spec change so much as increased clarity, as this is a natural conclusion to draw from the physics. However, calling it out is important as it might not be obvious that IOR would affect the amount of roughness blur. 

I purposefully put "unless otherwise specified by another extension" because this will be handled differently when using the volume extension since at that point the transmission surface becomes single-sided (entering the volume).


abwood (approved):
This is great, thank you for this clarification!

emackey:
This PR addresses: https://github.com/KhronosGroup/glTF/pull/1718#issuecomment-693081802

emackey (approved):

MiiBond (approved):
Looks good to me! Thanks, @elalish

PR #1881, opened 2020-10-09
Align KHR_materials_variants schemas with the core spec

lexaknyazev:
This is a technical update to align `KHR_materials_variants` schemas with the core glTF design principles.

- All objects should inherit from `glTFProperty`.
  - `extras` and `extensions` need to be mentioned explicitly.
- Elements of a top-level array inherit from `glTFChildOfRootProperty`.
- Empty arrays are not allowed.

/cc @jercytryn @sleroux

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_variants/schema/glTF.KHR_materials_variants.schema.json
+++ b/extensions/2.0/Khronos/KHR_materials_variants/schema/glTF.KHR_materials_variants.schema.json
@@ -9,6 +9,7 @@
             "type": "array",
             "items": {
                 "type": "object",
+                "allOf": [ { "$ref": "glTFChildOfRootProperty.schema.json" } ],
                 "description": "An object defining a valid material variant",
                 "gltf_detailedDescription": "An object defining a valid material variant",
                 "properties": {
@@ -19,11 +20,15 @@
                     },
                     "extras": {
                         "type": "object"
-                    }
+                    },
+                    "extensions": { }
                 },
                 "required": [ "name" ]
-            }
-        }
+            },
+            "minItems": 1
+        },
+        "extensions": { },
+        "extras": { }
     },
     "required": [ "variants" ]
 }
--- a/extensions/2.0/Khronos/KHR_materials_variants/schema/mesh.primitive.KHR_materials_variants.schema.json
+++ b/extensions/2.0/Khronos/KHR_materials_variants/schema/mesh.primitive.KHR_materials_variants.schema.json
@@ -10,6 +10,7 @@
             "gltf_detailedDescription": "An array of object values that associate an indexed material to a set of variants.",
             "items": {
                 "type": "object",
+                "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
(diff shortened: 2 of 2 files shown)


emackey (approved):

emackey:
Follows #1681

sleroux (approved):
Thanks for the clarifying update!

sleroux (approved):

lexaknyazev:
@donmccurdy 
Any objections to merge this?

extensions/2.0/Khronos/KHR_materials_variants/schema/glTF.KHR_materials_variants.schema.json
```diff
@@ -9,6 +9,7 @@
             "type": "array",
             "items": {
                 "type": "object",
+                "allOf": [ { "$ref": "glTFChildOfRootProperty.schema.json" } ],
```
donmccurdy, line 12: For consistency: https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_lights_punctual/schema/glTF.KHR_lights_punctual.schema.json defines this differently:

```
"items": {
                "type": "object",
                "$ref": "light.schema.json"
},
```

... are these equivalent?

lexaknyazev: `$ref` would require adding another file like `variant.schema.json`. Given that we don't need any fields beyond those defined in `glTFChildOfRootProperty`, I think taking a shortcut here is fine.

donmccurdy: I didn't mean to suggest creating a new file — comparing these two lines:

```
"$ref": "glTFChildOfRootProperty.schema.json",
```

```
"allOf": [ { "$ref": "glTFChildOfRootProperty.schema.json" } ],
``` 

... is there any difference in meaning? Or why does the first require a new file when the second does not?

lexaknyazev: >  is there any difference in meaning?

The semantics are slightly different.

This case
```json
"items": {
  "type": "object",
  "$ref": "glTFChildOfRootProperty.schema.json"
},
```
means: "replace the value of `"items"` with the contents of `"glTFChildOfRootProperty.schema.json"`". In fact, `"type": "object"` is ignored there.

OTOH, this case
```json
"items": {
  "type": "object",
  "allOf": [ { "$ref": "glTFChildOfRootProperty.schema.json" } ],
```
means: "`items` is an object that must be valid against all the schemas defined in `allOf` array".

Given that `KHR_materials_variants` defines `name` as a required field, we cannot use the first approach here.

donmccurdy: I see, thanks!

donmccurdy (approved):

Issue #1537, opened 2019-01-18
Stride forces non-vertex attribute accessors to each have their own buffer view?


scurest:
Here is what the spec says about stride

> When a buffer view is used for vertex attribute data, it may have a byteStride property. This property defines the stride in bytes between each vertex.
>
> [...]
>
> When byteStride of referenced bufferView is not defined, it means that accessor elements are tightly packed, i.e., effective stride equals the size of the element. When byteStride is defined, it must be a multiple of the size of the accessor's component type. byteStride must be defined, when two or more accessors use the same bufferView.
>
> [...]
>
> **bufferView.byteStride**
>
> The stride, in bytes, between vertex attributes. When this is not defined, data is tightly packed. When two or more accessors use the same bufferView, this field must be defined.

In addition, before [#1501](https://github.com/KhronosGroup/glTF/pull/1501) the first quote read (note the more explicit last sentence)

> Buffer view could have byteStride property. It means byte-distance between consequential elements. This field is defined only for buffer views that contain vertex attributes.

From this conclude

1. A buffer view which is used by an accessor that is _not_ vertex attribute data must not have a stride property defined (even if that stride property is the implied value of the size of the element).
2. The requirement that a buffer view used by more than one accessor have a stride property defined is unconditional, ie. it applies regardless of whether the accessor hold vertex attribute data or not.
3. It follows that only vertex attribute accessors can share buffer views; other accessors must each have their own buffer view.

I was originally going to file an issue against the validator (cc @lexaknyazev) because (1) is not flagged ([example](https://gist.github.com/scurest/a5d136252fc98dce34756640926b46fc)) and (2) is only flagged when the two accessors are vertex attribute data. 

But I'm wondering if this is a spec issue instead with (2) instead, because

* there are example models that don't satisfy (2), eg. [AnimatedTriangle](https://github.com/KhronosGroup/glTF-Sample-Models/blob/2af6c42c2c183bf65b5f14414aebd6bb99407c33/2.0/AnimatedTriangle/glTF-Embedded/AnimatedTriangle.gltf) where accessors 2 and 3 both use buffer view 2 but it does not have a byte stride defined.
* the discussion in #827 suggests the stride requirement have no reason to be applied to non-vertex attribute data. In particular https://github.com/KhronosGroup/glTF/issues/827#issuecomment-326997197 is inconsistent with (3).
* (3) appears unnecessarily restrictive.

So assuming (2) is false, I think the spec should be updated to say something like

[...]

Issue #1885, opened 2020-10-18
EXT_mesh_gpu_instancing: Clarify bufferView usage


donmccurdy:
Similar to https://github.com/KhronosGroup/glTF/issues/1864#issuecomment-698579703, I suspect there is an implicit bufferView usage associated with `EXT_mesh_gpu_instancing`, which the spec currently doesn't mention. For example, you probably wouldn't want your instance attributes going into a mixed buffer view alongside animation data. Let's clarify this, so that authoring tools know what to do.

Issue #1198, opened 2018-01-03
Clarification on byteStride requirement


zellski:
[In the spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#data-alignment) we are told that "`byteStride` must be defined, when two or more accessors use the same `bufferView`."

How generally true is this? Is it a general statement for all accessors -- indices, attributes, animation?

I would've assumed so, but:
- Several of the Khronos sample models (generated via COLLADA2GLTF) place the indices of multiple mesh primitives into a single `bufferView`, shared by many accessors (using different `.byteOffset`), and yet there is no `.byteStride` to be seen.
- The Khronos validator seems to go out of its way to restrict testing for `.byteStride` to the accessors of mesh primitive attributes. No such test is done for the indices accessor, nor for animations.

Who's right? The spec, or the sample models & validator?

lexaknyazev:
The `byteStride` property is [valid only for vertex attributes](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#bufferviewbytestride). Using it with other data will result in validation error.

zellski:
Ahh, there it is. Thanks.

It might be nice to just s/two or more accessors/two or more vertex attribute accessors/ in the spec, but I see now that the information is there.

TrabacchinLuigi:
I still don't get it.
I'm working on a GLB/GLTF converter and i got this `MESH_PRIMITIVE_ACCESSOR_WITHOUT_BYTESTRIDE` so i went through the docs and read and read and i see no reason i should get this error, i keep all my buffers separated and tightly packed i have just one view on every one of them and many accessor with an offset

lexaknyazev:
The rule is: `bufferView.byteStride` must be defined when two or more vertex attribute accessors (e.g. `POSITION` and `NORMAL`) use it.

lexaknyazev:
@TrabacchinLuigi 
If there's a particular asset that wrongly triggers this error, please consider sharing it here (it's possible to drag-n-drop zip files to GitHub).

TrabacchinLuigi:
[server.zip](https://github.com/KhronosGroup/glTF/files/3838924/server.zip)
I belive this one triggers it wrongly, hope i'm not wasting your time... but this drome me crazy

lexaknyazev:
[...]

PR #1888, opened 2020-10-21
Tighten the rules around bufferView.byteStride

### LINKED ISSUE #1864: Interleaved animation data
The `bufferView.byteStride` was intended only for vertex attributes, its description says:
> The stride, in bytes, between vertex attributes.

It turns out that some tools may interleave animation data and some engines even support that, see
https://forum.babylonjs.com/t/gltf-animations-loading-sampler-data-from-interleaved-buffer/14091

We should clarify the spec whether it should be allowed.

/cc @donmccurdy @bghgary

zeux:
The intention of the spec has always been to only use byteStride for
buffer views with vertex data. However, the current language was lax in
a few places, which led to logical contradictions:

- When two accessors with non-vertex data used the same bufferView, it
could be interpreted as requiring byteStride on that view
- However, byteStride was only explicitly allowed for vertex attribute
data, making it seem as if it's not valid on non-vertex views.

glTF-Validator flags bufferViews that have a byteStride but store
non-vertex data as erroneous. Based on this and the above, this change
tightens the language to explicitly only allow byteStride on vertex data
views, and changes all references to it to say the same.

This would close #1864 by explicitly outlawing interleaved data that's
not vertex data, and also relates to #1537 (that is already closed, but this
change would provide an actual resolution there).

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -469,7 +469,7 @@ The following example defines two buffer views: the first is an ELEMENT_ARRAY_BU
 }
 ```
 
-When a buffer view is used for vertex attribute data, it may have a `byteStride` property. This property defines the stride in bytes between each vertex.
+When a buffer view is used for vertex attribute data, it may have a `byteStride` property. This property defines the stride in bytes between each vertex. Buffer views with other types of data should not define `byteStride`.
 
 Buffers and buffer views do not contain type information. They simply define the raw data for retrieval from the file. Objects within the glTF file (meshes, skins, animations) access buffers or buffer views via *accessors*.
 
@@ -649,7 +649,7 @@ When neither `sparse` nor `bufferView` is defined, `min` and `max` properties co
 
 The offset of an `accessor` into a `bufferView` (i.e., `accessor.byteOffset`) and the offset of an `accessor` into a `buffer` (i.e., `accessor.byteOffset + bufferView.byteOffset`) must be a multiple of the size of the accessor's component type.
 
-When `byteStride` of referenced `bufferView` is not defined, it means that accessor elements are tightly packed, i.e., effective stride equals the size of the element. When `byteStride` is defined, it must be a multiple of the size of the accessor's component type. `byteStride` must be defined, when two or more accessors use the same `bufferView`.
+When `byteStride` of referenced `bufferView` is not defined, it means that accessor elements are tightly packed, i.e., effective stride equals the size of the element. When `byteStride` is defined, it must be a multiple of the size of the accessor's component type. `byteStride` must be defined, when two or more vertex data accessors use the same `bufferView`.
 
 Each `accessor` must fit its `bufferView`, i.e., `accessor.byteOffset + STRIDE * (accessor.count - 1) + SIZE_OF_ELEMENT` must be less than or equal to `bufferView.length`.
 
@@ -2125,7 +2125,7 @@ The length of the bufferView in bytes.
 
 #### bufferView.byteStride
 
-The stride, in bytes, between vertex attributes.  When this is not defined, data is tightly packed. When two or more accessors use the same bufferView, this field must be defined.
+The stride, in bytes, between vertex attributes.  When this is not defined, data is tightly packed. When two or more vertex data accessors use the same bufferView, this field must be defined.
 
 * **Type**: `integer`
 * **Required**: No
--- a/specification/2.0/schema/bufferView.schema.json
+++ b/specification/2.0/schema/bufferView.schema.json
@@ -26,7 +26,7 @@
             "minimum": 4,
             "maximum": 252,
             "multipleOf": 4,
-            "gltf_detailedDescription": "The stride, in bytes, between vertex attributes.  When this is not defined, data is tightly packed. When two or more accessors use the same bufferView, this field must be defined.",
+            "gltf_detailedDescription": "The stride, in bytes, between vertex attributes.  When this is not defined, data is tightly packed. When two or more vertex data accessors use the same bufferView, this field must be defined.",
             "gltf_webgl": "`vertexAttribPointer()` stride parameter"
         },
         "target": {


zeux:
If we want to keep #1864 open I can remove the line about "other types of data"; the critical part here is to make the "two or more accessor" rule more explicit about only applying to vertex data, as multiple accessors that reference the same bufferView with tightly packed data by referring to individual sections of the buffer through byte offsets should be valid.

specification/2.0/README.md
```diff
@@ -469,7 +469,7 @@ The following example defines two buffer views: the first is an ELEMENT_ARRAY_BU
 }
 ```
 
-When a buffer view is used for vertex attribute data, it may have a `byteStride` property. This property defines the stride in bytes between each vertex.
+When a buffer view is used for vertex attribute data, it may have a `byteStride` property. This property defines the stride in bytes between each vertex. Buffer views with other types of data should not define `byteStride`.
```
donmccurdy, line 472: See https://github.com/KhronosGroup/glTF/issues/1885 for context here, but could EXT_mesh_gpu_instancing also use interleaved attributes? I think the answer is yes, and that `bufferView.target` would similarly be `ARRAY_BUFFER`, but I'm not sure about this.

donmccurdy: I'm not sure how to generalize and future-proof this language. Perhaps we should be describing "attribute data" to cover both vertex attributes and instance attributes?

zeux: > bufferView.target would similarly be ARRAY_BUFFER

I don't think ARRAY_BUFFER is right... It would be if the implementation is using vertex stream divisors to implement instancing, but using uniform or shader storage or texel buffers would be a more modern way of achieving this, for which ARRAY_BUFFER is incorrect. So I'd rather have a different type of target that we expect for the instance data stream...

zeux: Effectively I'd say the rule should be that the target defines the *usage* of the data, but not the implementation details of a given renderer. Ideally we wouldn't use GL names for this but I guess that ship has sailed.

donmccurdy: Fine with me to not use `ARRAY_BUFFER` for instance attributes. However, I expect we do want to allow instance attributes to be interleaved, right? Does gltfpack interleave these attributes? If so we would want to allow `.byteStride` in that use case.

zeux: gltfpack doesn't interleave attributes at the moment.

For whether interleaving is allowed in general on a given data type, I think we should look at the implementation complexity (where support for byte stride is easy when the data is mapped directly to a vertex stream on the GPU, and more difficult otherwise), possible gains in efficiency from interleaving due to increased cache coherency, and uniformity.

Instanced data is interesting because it has the widest possible set of implementation option out of all other types of data:

- It can be implemented by uploading data to the GPU and using vertex streams with instance frequency/divisor; in this case supporting interleaving is free.
- It can be implemented by uploading data to the GPU and using buffers (typed or untyped) with the shader reading data from the buffer based on gl_InstanceId or equivalent; in this case, interleaving can be difficult and I'd expect implementations to have to pre-process interleaved data on the CPU
- It can be implemented by reading the transforms on the CPU and dispatching multiple draw calls (which can still be more efficient vs non-instanced implementation for a few reasons); in this case, interleaving might require extra processing on the CPU side

I don't think there are efficiency gains that one should expect from interleaving the instance data however, since the model is generally one where you'd need to read all transforms to render the batch of instances. One could make the argument that for animations, interleaving is more valuable than it is for instances since cache utilization can be better when keyframes are ordered globally by time across multiple different tracks, so that sampling animation at time T requires fewer than 1 cache accesses per track.

If the spec *right now* only allows byteStride to be defined for vertex data (which I think was the intention), maybe it would make sense to phrase this as "Buffer views with other types of data should not define `byteStride`, unless interleaved data is explicitly supported by the extension that uses the buffer view as a data source"? That way we can have specific callouts to byteStride being supported.

Alternatively I can remove this specific note from the change, if it's not clear whether we should support interleaving for non-vertex data.

specification/2.0/README.md
```diff
@@ -2125,7 +2125,7 @@ The length of the bufferView in bytes.
 
 #### bufferView.byteStride
 
-The stride, in bytes, between vertex attributes.  When this is not defined, data is tightly packed. When two or more accessors use the same bufferView, this field must be defined.
+The stride, in bytes, between vertex attributes.  When this is not defined, data is tightly packed. When two or more vertex data accessors use the same bufferView, this field must be defined.
```
emackey, line 2128: This reference section was auto-generated from the schema files.  Please make the same change in both places.

(Soon we may adopt an automated process to generate this reference from the schema).

emackey: Specifically this line:

https://github.com/KhronosGroup/glTF/blob/6a3d023e63f3af893a3f66b7d8dc09f3c74ad21f/specification/2.0/schema/bufferView.schema.json#L29

zeux: Thanks! Fixed.

specification/2.0/README.md
donmccurdy: > If the spec right now only allows byteStride to be defined for vertex data (which I think was the intention)...

I guess this is what we need to determine, then... let me follow up in https://github.com/KhronosGroup/glTF/issues/1864.

hypnosnhendricks: fbx2gltf currently uses target=0 for all morph target data buffer views. I noticed gltfpack uses 34962 ( vertex data ). I ran into this exact contradiction around byteStride when packing morph target data without specifying a target for the buffer view. Perhaps this is an error on fbx2gltf's part and all morph target data should be changed to use 34962 target type.

We use GLTF/GLB to deliver assets to clients, who then process them based on the client they use (Unity, Unreal, etc). None of our clients can upload the GLB data directly into the GPU and have to process it for the particular engine. We would like to see GLTF/GLB be as flexible as possible when transmitting data to clients.

zeux: > fbx2gltf currently uses target=0 for all morph target data buffer views
Does it specify byteStride for these or not? If it doesn't then this change shouldn't impact that in any way, as there's no requirement in the existing spec that the vertex views must specify non-0 target.

rsahlin: I would like to include indices in the tightened specification, one reason for this is consistency - both type of bufferViews will be read by the GPU in the vertex/geometry processing step(s).
Another reason is with regards to component type:

`When a buffer view contain vertex indices or attributes, they must be its only content, i.e., it's invalid to have more than one kind of data in the same buffer view.`

Ie, it should not be valid to have 2 Accessors (indices) referencing the same BufferView, using different offsets and 0 byteStride to access the data using different component types (short/int)
At the moment it seems the validator does not flag for this.

rsahlin: > Effectively I'd say the rule should be that the target defines the semantics of the data, but not the implementation details of a given renderer.

I totally agree and for me it would help if target _must_ be specified if the BufferView is used for data that needs to be bound (in a buffer before being accessed by the GPU).
That way an implementations can look through the declared BufferViews and know which ones will require a GPU based target buffer.
Ideally this could be expanded to other targets so that implementations may decide what data to generate buffers for (without traversing the nodetree and check what will be rendered)

zeux: Since today index buffer views can't contain byteStride, I don't think it's a good idea to include that as that'd render many existing assets invalid. Ditto for the target - I agree it should have been a required attribute, but since it's not it's not an option to go back.

donmccurdy: > Ie, it should not be valid to have 2 Accessors (indices) referencing the same BufferView, using different offsets and 0 byteStride to access the data using different component types (short/int)

I don't think I understand this part — we wouldn't want to interleave indices, but what's the concern with multiple indices' accessors being end-to-end in a single buffer view?

hypnosnhendricks: There is also the rule about byteStride being required when two or more accessors use the same buffer view. 
(https://github.com/KhronosGroup/glTF/issues/1198). 

I ran into this issue because our morph targets use target=0 in the buffer views, so we ran into this valiadtion error when we pack the morph data into the same buffer view, even thought it is tightly packed and the stride equals FLOAT, VEC3, since the morph pos/nor/tan are are the same componentType and Type.

But, there is also the rule that the skinning matricies cannot have a byte stride, which also uses target=0 in the buffer view.

I had to add extra code to detect when I'm packing accessors into the same buffer view to specify a byte stride or not based on if it is vertex data, index data, or 'other' data, but not 'other' data that is bind matricies.

donmccurdy: I'm not sure what `target=0` means sorry, that's not a valid enum value here. Do you mean `target` is undefined?

Only vertex attributes may have a byte stride, and we can't loosen that safely within the 2.0 version (except perhaps in extensions that define new types of data that might benefit from a byteStride), so I think the purpose of this subthread is resolved. If there's something else I've missed in this thread, let's open a new PR or issue to continue its discussion.

donmccurdy (approved):
I believe this PR reflects the original intent of the specification, and clarifies it. There does not seem to be any other interpretation that would be compatible within the glTF 2.0 lifecycle, in any case. 

Looks good to me. 👍

lexaknyazev:
This clarification has been integrated into the new AsciiDoc version of the spec. Thanks!

PR #1890, opened 2020-10-26
Correct the formula of BRDF

06wj:


emackey:
@proog128 Ok to merge this?

proog128:
Yes, thank you!

emackey:
Note: There's an AsciiDoctor conversion of this README underway now.  We should double-check that this change makes it into the converted document, once that's available.

PR #1892, opened 2020-10-28
Testing wetzel automated reference output.  DO NOT MERGE

emackey:
## DO NOT MERGE

This is simply to show the diff of what would happen if we took automated [wetzel](https://github.com/CesiumGS/wetzel) output and blindly used it as a replacement for the reference section.

**NOTE:** If there are functional differences between this and what's in the reference section, it may indicate a disparity between the main reference section and the JSON schemas.

The command-line used to generate this was:

```
$ bin/wetzel.js -l2 -a=cqo ../glTF/specification/2.0/schema/glTF.schema.json > ../wetzel_output/ref_2020-10-28.md
```

PR #1897, opened 2020-11-03
Manual cleanup to bring schema and reference into better alignment.

emackey:
This PR manually cleans up some minor discrepancies that have separated the JSON schema from the generated reference section over the years.

I'd like to merge this first, in advance of re-running the automated conversion.  Running the automated conversion will break old URL hash references into the Markdown version, but those will break anyway when we migrate to AsciiDoctor, so we can save that for a later PR and get it all done in one shot.

### DIFF
--- a/specification/2.0/schema/extras.schema.json
+++ b/specification/2.0/schema/extras.schema.json
@@ -1,5 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-04/schema",
     "title": "Extras",
-    "description": "Application-specific data."
+    "description": "Application-specific data.",
+    "gltf_detailedDescription": "Application-specific data. **Implementation Note:** Although extras may have any type, it is common for applications to store and access custom data as key/value pairs. As best practice, extras should be an Object rather than a primitive value for best portability."
 }
--- a/specification/2.0/schema/extras.schema.json
+++ b/specification/2.0/schema/extras.schema.json
@@ -2,5 +2,5 @@
     "$schema": "http://json-schema.org/draft-04/schema",
     "title": "Extras",
     "description": "Application-specific data.",
-    "gltf_detailedDescription": "Application-specific data. **Implementation Note:** Although extras may have any type, it is common for applications to store and access custom data as key/value pairs. As best practice, extras should be an Object rather than a primitive value for best portability."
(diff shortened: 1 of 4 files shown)


emackey:
The fixes here are in response to findings in #1892.

donmccurdy (approved):

donmccurdy:
Nothing objectionable here, I'd be glad to merge this sooner than later. Thanks!

emackey:
Worth noting that none of the text in the diffs here is "new".  The schema changes are all copied from the corresponding parts of the README reference section, and vice-versa, the reference changes here all come from the schema.  The only thing I did was try to figure out which one should overwrite the other in which cases.

specification/2.0/schema/extras.schema.json
```diff
@@ -1,5 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-04/schema",
     "title": "Extras",
-    "description": "Application-specific data."
+    "description": "Application-specific data.",
+    "gltf_detailedDescription": "Application-specific data. **Implementation Note:** Although extras may have any type, it is common for applications to store and access custom data as key/value pairs. As best practice, extras should be an Object rather than a primitive value for best portability."
```
lexaknyazev, line 5: Does `**Implementation Note:**` decoration look as expected in the generated Markdown?

emackey: No, and worse, it sprinkles itself all over the reference section.

Would it be OK if I add a new type here called `gltf_sectionDescription` ?  The "detailedDescription" one is only ever used on properties, not at the top level of a reference section like this.  I have a local 2-line change to wetzel that will kick it out in the correct place.

emackey: Here's the proposed change to wetzel: https://github.com/CesiumGS/wetzel/pull/42

lexaknyazev: Nothing prevents us from adding custom properties, so let's do it.

emackey: Done.

emackey:
Seems like no further objections.  I'm going to merge this, and then re-run the comparison in #1892.

PR #1899, opened 2020-11-06
Test new automated reference section

emackey:
This is a big one.  We probably don't want to merge this until the AsciiDoctor transition is ready to go.

Things that changed:

- Schema titles have changed.  It was not feasible for wetzel to parse what "Material PBR Metallic Roughness" meant in terms of the specific type name.  So types like this now have a title of `material.pbrMetallicRoughess` to match the type name.

- The reference section in README.md is the raw output from wetzel (see commit descriptions below).

- A temporary `REFERENCE.adoc` has been added, to show what the AsciiDoctor version of the reference output looks like.

Ultimately I do want to commit the schema title changes.  But the README changes will become obsolete in the move to AsciiDoctor, and the REFERENCE section will be a temporary file during the AsciiDoctor build process, not preserved in git.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -1653,29 +1653,29 @@ This chunk must be padded with trailing zeros (`0x00`) to satisfy alignment requ
 
 ## Objects
 * [`accessor`](#reference-accessor)
-   * [`sparse`](#reference-sparse)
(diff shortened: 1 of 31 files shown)


emackey:
Closing in favor of #1901.

PR #1904, opened 2020-11-17
Remove comparison to GL coordinates in normalTexture schema.

### LINKED ISSUE #1902: "OpenGL conventions" aren't really "+X is right and +Y is up. +Z points toward the viewer"
This is the GLU convention but by default, OpenGL's clip space with default DEPTH_TEST is +Z *away* (far). (Yes it's left-handed!)
Here's an example with the default clip-space and DEPTH_TEST: https://jdashg.github.io/misc/webgl/webgl-clipspace.html

jdashg:
Resolves #1902.

### DIFF
--- a/specification/2.0/schema/material.schema.json
+++ b/specification/2.0/schema/material.schema.json
@@ -15,7 +15,7 @@
         "normalTexture": {
             "allOf": [ { "$ref": "material.normalTextureInfo.schema.json" } ],
             "description": "The normal map texture.",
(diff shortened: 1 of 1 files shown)


emackey:
Thanks @jdashg.  Can we actually just remove the part in parenthesis?  All we need to do here is specify what glTF's own convention is; we don't need to draw comparisons with OpenGL or GLU at all here.  Would that be OK?

emackey:
Thanks @jdashg.

PR #1907, opened 2020-11-24
Add JSON schemas to AsciiDoctor output.

emackey:
@oddhack This is a PR into your branch, to add the JSON schema file contents.

via https://github.com/CesiumGS/wetzel/pull/51

### DIFF
--- a/specification/2.0/Makefile
+++ b/specification/2.0/Makefile
@@ -8,19 +8,18 @@ all: $(TARGETS)
 
 # Generate the Properties Reference section of the spec from JSON schema.
 # This uses 'npx' to minimally install the 'wetzel' tool, if needed.
-# While the schema used by wetzel is the locally installed version, it
-# generates inline links to the schema under $(SCHEMALINK), typically
-# the canonical schema in the github default branch.
+# When using embedded schemas, SCHEMALINK should be a relative path.
 
 WETZEL = npx wetzel
-SCHEMALINK = https://github.com/KhronosGroup/glTF/tree/master/specification/2.0/schema
+SCHEMALINK = schema
+EMBEDSCHEMA = JsonSchemaReference.adoc
 
 # Base name of the generated properties reference file
 PROPREF = PropertiesReference
 # Schema files to leave out of $(PROPREF)
 IGNORESCHEMA = '["gltfchildofrootproperty.schema.json", "gltfid.schema.json", "gltfproperty.schema.json"]'
 $(PROPREF).adoc: $(wildcard schema/*.json)
-	$(WETZEL) -n -a=cqo -m=a -p "$(SCHEMALINK)" \
+	$(WETZEL) -n -a=cqo -m=a -p "$(SCHEMALINK)" -e "$(EMBEDSCHEMA)" \
 	    -i $(IGNORESCHEMA) \
 	    schema/glTF.schema.json > $@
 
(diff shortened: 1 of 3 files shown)


emackey:
for #1901

specification/2.0/Makefile
```diff
@@ -33,12 +32,12 @@ SPECDEPS = $(SPEC).adoc PropertiesReference.adoc
...
+# :allow-url-read: is necessary for the embedded render.githubusers.com
 # math images to be processed for the PDF target. See
 # https://github.com/asciidoctor/asciidoctor-pdf/issues/369
 $(SPEC).pdf: $(SPECDEPS)
 	$(ASCIIDOCTOR) -b pdf -a allow-uri-read -r asciidoctor-pdf $(SPEC).adoc -o $@
 
 GENERATED = $(PROPREF).adoc $(TARGETS)
 clean:
-	-rm -f $(GENERATED) $(TARGETS)
+	-rm -f $(GENERATED) $(TARGETS) $(EMBEDSCHEMA)
```
oddhack, line 43: Would add $(EMBEDSCHEMA) to the definition of GENERATED above.

specification/2.0/Makefile
```diff
@@ -8,19 +8,18 @@ all: $(TARGETS)
...
 WETZEL = npx wetzel
-SCHEMALINK = https://github.com/KhronosGroup/glTF/tree/master/specification/2.0/schema
+SCHEMALINK = schema
+EMBEDSCHEMA = JsonSchemaReference.adoc
 
 # Base name of the generated properties reference file
 PROPREF = PropertiesReference
 # Schema files to leave out of $(PROPREF)
 IGNORESCHEMA = '["gltfchildofrootproperty.schema.json", "gltfid.schema.json", "gltfproperty.schema.json"]'
 $(PROPREF).adoc: $(wildcard schema/*.json)
```
oddhack, line 21: Would add $(EMBEDSCHEMA) to targets here. Might want some sort of npx dependency too, but I'm not sure how that would work since the wetzel files are pulled from a remote.

specification/2.0/Makefile
```diff
@@ -8,19 +8,18 @@ all: $(TARGETS)
...
-	$(WETZEL) -n -a=cqo -m=a -p "$(SCHEMALINK)" \
+	$(WETZEL) -n -a=cqo -m=a -p "$(SCHEMALINK)" -e "$(EMBEDSCHEMA)" \
```
oddhack, line 22: Is this where $(EMBEDSCHEMA) is generated?

oddhack (changes requested):

oddhack:
I'll accept once you've made those changes, unless you want to hold off.

oddhack:
Going to go ahead and merge this now and make the changes I recommended above - something I wanted to tweak in the appendix section layout and it will be easier to check in my branch.

specification/2.0/Makefile
emackey: Yes. I suppose it would be more Makefile-style if we were to run wetzel twice for the two different targets.  In that case `-e` would be some kind of embed mode switch to indicate which of the two targets was being produced, and the output would always go through stdout.  Is it worth refactoring to get that behavior?

specification/2.0/Makefile
emackey: I think in the next round of changes I'm going to add the version number onto the end of wetzel.  Specifically, something like this:

```
WETZEL = npx wetzel@0.1.0-dev.1.2
```

That way, updates to new versions of wetzel must be manually done in the Makefile, not just pulled in automatically whenever someone publishes.  This would provide more stability of the history, for example pulling an old commit and re-running the old documentation would use the specific version of wetzel available at that point in history, and should exactly regenerate what the docs were back then.  Likewise, changes to the docs introduced by new wetzel versions could be found with bisect, etc.

specification/2.0/Makefile
oddhack: I don't think it is - the two targets should be generated at the same time to remain consistent. Not a big deal either way.

PR #1908, opened 2020-11-25
Clarify restrictions on the number of indices

### LINKED ISSUE #1674: How should implementations deal with index counts that don't match the primitive requirements
The current specification doesn't mention the (recommended) behavior of implementations when encountering data which is not fit for rendering.

We got a glTF file from a customer which had an index buffer with 5 indices for a TRIANGLES primitive. First this caused actually crash in a third party loading library (we're following up there independently), but we then came to the question what the recommended way to handle this is.
Should we drop the last two indices? Should we drop the whole primitive?

Also (probably because it isn't specified) the glTF validator didn't emit an error or warning and declared the asset valid. I suspect that this might trip up other implementations and it would be good to get some clarification in the specification.

zeux:
glTF specification is ambiguous as to what the expected behavior is when
a 2-index triangle primitive is specified.

As the spec never defined a valid treatment and this causes continuous
issues with various viewers and processors, this change updates the
spec, making the rules stricter. It is expected that existing assets
largely already conform to these rules as these just spell out the
"logical" requirements of a given mode: it must result in a non-empty
primitive and must not specify any "unused" indices.

Fixes #1674.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -784,6 +784,14 @@ All indices for indexed attribute semantics must start with 0 and be continuous 
 
 All attribute accessors for a given primitive must have the same `count`. When `indices` property is not defined, it indicates the number of vertices to render; when `indices` property is defined, it indicates the upper (exclusive) bound on the index values in the `indices` accessor.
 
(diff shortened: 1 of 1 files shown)


lexaknyazev:
This clarification has been integrated into the new AsciiDoc version of the spec. Thanks!

PR #1912, opened 2020-11-30
Update main glTF logo to have green "TM"

emackey:
Per request from Neil Trevett during PBR TSG, 2020-11-30.

The new image is a reencoding of an official logo from:  https://www.khronos.org/legal/trademarks/


emackey:
Note that browser file caching seems pretty strong here.  You may have to clear your cache to see the updated logo.  Only the "TM" part is different, it changed from black to green to match the rest of the logo.

PR #1914, opened 2020-12-01
Add VRM to "Formats Built on glTF"

donmccurdy:
Seems like a relevant addition here? Wouldn't generally list third-party extensions as formats, but I think this is sufficiently notable, and has its own ecosystem and file extension.


bghgary (approved):
Makes sense to me

PR #1917, opened 2020-12-02
Sectioned hanging text

DrX3D:
Sectioned all hanging text by creating an initial sub-section with the text from the parent. The end result is a document with all text in a specific section. No section contains sub-sections and text.

1. New subsection for main sections (_X_. _name_) are _X_.1 General
1. Other new subsections (_X.Y..._. _name_) are _X.Y..._.1 Overview
1. New anchor is created for each new sub-section by appending _-general_ or _-overview_ to the parent label

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -486,7 +486,7 @@ The next example defines the transformation for a node with attached camera usin
 === Buffers and Buffer Views
 
 
-[[buffers-overview]]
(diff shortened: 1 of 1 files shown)


DrX3D:
This is tied to Issue #1917  (https://github.com/KhronosGroup/glTF/pull/1917)

emackey (approved):
Awesome, thanks Leonard!

PR #1919, opened 2020-12-03
Flip credits & copyrights

emackey:
This moves the copyright up top, and puts all credits at the end.  Neil requested this on a recent call.


oddhack (approved):
LGTM. You can merge, or I will.

PR #1922, opened 2020-12-05
Material Variants: Modify "extras" to use same convention as elsewhere in glTF.

emackey:
I'm not a fan of this, as I'd rather see all "extras" everywhere get tightened to be an object, rather than the lone object get loosened to be "any".  But the former is too big a breaking change to be considered for glTF 2.0, and conformance will help the glTF Validator enforce uniform behavior, so here we are.

For https://github.com/KhronosGroup/glTF-Validator/issues/149#issuecomment-739353647

One question though, is this change itself too big a breaking change for a released extension?

/cc @lexaknyazev @bghgary @jercytryn

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_variants/schema/glTF.KHR_materials_variants.schema.json
+++ b/extensions/2.0/Khronos/KHR_materials_variants/schema/glTF.KHR_materials_variants.schema.json
@@ -18,9 +18,7 @@
                         "description": "The name of the material variant",
                         "gltf_detailedDescription": "The name of the material variant."
                     },
-                    "extras": {
-                        "type": "object"
-                    },
+                    "extras": { },
                     "extensions": { }
                 },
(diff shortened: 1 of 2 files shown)


donmccurdy:
> One question though, is this change itself too big a breaking change for a released extension?

To me this seems firmly in the "bug fix" category, it is very unlikely that anyone intended for the extension's `.extras` property to differ from other extensions.

sleroux (approved):
I mentioned on https://github.com/KhronosGroup/glTF-Validator/issues/149 that the intention was to mirror existing `extras` than making this an exception so I'm in favor of this. I agree with Don that people most likely are treating this the same as existing `extras`.

emackey:
@lexaknyazev Does this change fix the problem for the validator?

lexaknyazev:
I wouldn't call it a "problem" rather an inconsistency. So yes, this PR fixes it.

PR #1927, opened 2021-01-11
Rename KHR_draco_mesh_compression schema file to match the element it's attached to

kring:
The `KHR_draco_mesh_compression` extension is attached to a glTF `primitive`. But the schema file is incorrectly named `node.KHR_draco_mesh_compression.schema.json`. This PR simply renames it to `primitive.KHR_draco_mesh_compression.schema.json`.


emackey (approved):

emackey (changes requested):
Oops, approved too quickly.

@kring There's a broken link in the README that needs updating now.

https://github.com/CesiumGS/glTF/tree/draco-primitive-not-node/extensions/2.0/Khronos/KHR_draco_mesh_compression#json-schema

kring:
Oops, good catch! Fixed now.

emackey (approved):

kring:
Actually the element is called "mesh.primitive", not just "primitive". Updated again.

emackey:
Works. Thanks!

kring:
By the way, the context here is that I'm generating deserialization code from the schema, and the prefix on the schema files is the only semi-machine-readable indication of where an extension should be attached. I'll likely have a lot more changes like this.

emackey:
For what it's worth, I have a JSON mapping of them in another repo: https://github.com/AnalyticalGraphicsInc/gltf-vscode/blob/master/util/extensionMap2.0.json

emackey:
Also, some extensions, such as XMP (and its successor) will apply to multiple types of glTF objects, so cannot be guessed from a single filename, sadly.  That's part of the reason for the mapping in VSCode.

kring:
Oh nice, that's very helpful actually. Is it worth considering putting a mapping file like that directly into this repo?

kring:
Or it could go the other way, in order to keep the extensions more independent. A standard JSON file in each extension directory could describe the (potentially multiple) places where the extension is applied.

emackey:
Possibly.  When I wrote it, I was concerned that it could turn out too app-specific, or that I would be the only consumer.  But if multiple apps consume, it might be good to centralize.  I haven't thought about it much.

kring:
fwiw that file is exactly what I need. I'll copy it for now. Thanks for pointing me to it.

emackey:
Cool.  We should still accept this PR, for filename consistency.  But I'll wait for some other reviewers to chime in tomorrow.

bghgary (approved):

PR #1936, opened 2021-01-26
Add missing entry to Accessor component sizes tab

hazeycode:


lexaknyazev:
Thanks for the suggestion.

Signed int32 accessor type was omitted intentionally since not all platforms could support such vertex attribute type at a time of glTF 2.0 release. Unsigned int32 is supported only for index buffers.

If there's a specific use case for supporting int32 vertex attributes, it should be proposed as an extension.

hazeycode:
Thanks for the info @lexaknyazev, closing this

emackey:
Should the spec call this out explicitly, so it doesn't look like an omission?

PR #1943, opened 2021-02-15
Clarify R and RG textures handling

### LINKED ISSUE #1934: KHR_texture_basisu: should more channel types be supported?
I've been using the toktx tool to try to make compatible textures for use with the KHR_texture_basisu extension. I noticed sometimes toktx will produce a single channel RRR ktx2 texture. According to the [spec](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_texture_basisu#ktx-v2-images-with-basis-universal-supercompression), 

> RGB textures MUST use a single DFD channel with value KHR_DF_CHANNEL_ETC1S_RGB.
> RGBA textures MUST use two DFD channels with values KHR_DF_CHANNEL_ETC1S_RGB and KHR_DF_CHANNEL_ETC1S_AAA.

It seems like KHR_DF_CHANNEL_ETC1S_RRR could be a reasonable choice for a texture. Why isn't this allowed? Is there an argument passed to toktx that could resolve this?

lexaknyazev:
Fixes #1934.

This update future-proofs the extension for material textures that may intentionally use fewer than three color channels.

### DIFF
--- a/extensions/2.0/Khronos/KHR_texture_basisu/README.md
+++ b/extensions/2.0/Khronos/KHR_texture_basisu/README.md
@@ -136,17 +136,43 @@ To use KTX v2 image with Basis Universal supercompression without a fallback, de
 
 To cover a broad range of use cases, this extension allows two Basis Universal formats: ETC1S with BasisLZ and UASTC with optional Zstandard compression.
 
-- KTX header fields for ETC1S with BasisLZ payloads
+For the purposes of this extension, the following texture types are defined:
+
+- **RGB:** A texture that uses red, green, and blue channels. Alpha channel is either assumed to be always equal to `1.0` (`255`) or unused and not sampled at runtime. Such textures MAY be encoded with the sRGB transfer function.
+  > **Note:** RGB textures from the core glTF 2.0 specification include:
+  > - `baseColorTexture` (opaque case)
+  > - `normalTexture`
+  > - `emissiveTexture`
+  > - `metallicRoughnessTexture` (when be combined with `occlusionTexture`, otherwise red channel is ignored)
+
+- **RGBA:** A texture that uses all four channels. Such textures MAY be encoded with the sRGB transfer function (not applied to Alpha).
+  > **Note:** RGBA textures from the core glTF 2.0 specification include:
+  > - `baseColorTexture` (non-opaque case)
+
+- **Red:** A texture that uses only red channel. All other channels are unused and their values are not sampled at runtime. Such textures MUST NOT be encoded with the sRGB transfer function.
+  > **Note:** Red textures from the core glTF 2.0 specification include:
+  > - `occlusionTexture` (standalone)
+
+- **Red-Green:** A texture that uses only red and green channels. Blue and Alpha channels are unused and their values are not sampled at runtime. Such textures MUST NOT be encoded with sRGB transfer function.
+  > **Note:** The core glTF 2.0 specification has no examples of red-green textures.
+
+### KTX header fields for ETC1S with BasisLZ payloads
   - `supercompressionScheme` MUST be `1` (BasisLZ).
   - DFD `colorModel` MUST be `KHR_DF_MODEL_ETC1S`.
-  - RGB textures MUST use a single DFD channel with value `KHR_DF_CHANNEL_ETC1S_RGB`.
-  - RGBA textures MUST use two DFD channels with values `KHR_DF_CHANNEL_ETC1S_RGB` and `KHR_DF_CHANNEL_ETC1S_AAA`.
+  - RGB textures MUST use a single `KHR_DF_CHANNEL_ETC1S_RGB` DFD channel.
+  - RGBA textures MUST use two DFD channels: `KHR_DF_CHANNEL_ETC1S_RGB` and `KHR_DF_CHANNEL_ETC1S_AAA`.
+  - Red textures MUST use a single `KHR_DF_CHANNEL_ETC1S_RRR` DFD channel.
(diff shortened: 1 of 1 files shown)


extensions/2.0/Khronos/KHR_texture_basisu/README.md
```diff
@@ -141,12 +141,15 @@ To cover a broad range of use cases, this extension allows two Basis Universal f
   - DFD `colorModel` MUST be `KHR_DF_MODEL_ETC1S`.
   - RGB textures MUST use a single DFD channel with value `KHR_DF_CHANNEL_ETC1S_RGB`.
   - RGBA textures MUST use two DFD channels with values `KHR_DF_CHANNEL_ETC1S_RGB` and `KHR_DF_CHANNEL_ETC1S_AAA`.
+  - Red textures MUST use a single DFD channel with value `KHR_DF_CHANNEL_ETC1S_RRR`.
+  - Red-Green textures MUST use two DFD channels with values `KHR_DF_CHANNEL_ETC1S_RRR` and `KHR_DF_CHANNEL_ETC1S_GGG`.
```
donmccurdy, line 145: Is there a clearer way to identify "Red" and "Red-Green" textures? I think we are referring to textures in which the specification requires sampling only the "Red" channel, such as an `occlusionTexture` not reused by any other texture slot, and not a `baseColorTexture` that happens to be red.

lexaknyazev: When it comes to texture types, _red_ usually means exactly that: a single-channel texture containing values only in its first (i.e. R) channel and usually sampled as `r001` on the GPU. This is different from (deprecated) _luminance_ single-channel texture type that is sampled as `rrr1` or (also deprecated) _alpha_ single-channel texture type that is sampled as `000r`.

Similarly, _red-green_ means a two-channel texture that is sampled as `rg01`. This is different from (deprecated) _luminance-alpha_ texture type that is sampled as `rrrg`.

lexaknyazev: @donmccurdy 
Do you think we could come up with a better spec language, or should we merge it as-is? I could add my ⬆️ comment as a note if that helps.

donmccurdy: I'm thinking about how to implement this in glTF-Transform, and I don't know how, which makes me nervous about making it normative with "MUST". Given a user-provided PNG or JPEG texture, am I required to check every pixel's value for presence of green and blue channels?

It is much easier to detect occlusionTexture -> RRR, baseColorTexture -> RGB. Although from a client/loader perspective I suppose it would be nice to know what a `baseColorTexture` contains too...

lexaknyazev: Implementing it is very simple.

The target texture type is fully determined by the material slot. Specifically (for ETC1S):
- Base Color: RGB or RGB + AAA
- Emissive: RGB
- Normal: RGB (ETC1S not recommended)
- ORM or RM: RGB (ETC1S not recommended)
- Occlusion-only: Red
- Transmission-only: Red
- Clearcoat-Roughness: Red-Green

donmccurdy: I see — so far so good then. But I find the "red texture" language ambiguous because it is not clear it refers to the texture's use (material slots) and not the texture's data. Perhaps we can say something at the end of this section, like:

> Types of textures (RGB, Red, Red-Green, etc.) are defined according to their usage by material slots, and not by the pixel data the texture may happen to contain.

***

Suppose a model had an ORM texture with a fully non-metallic <kbd>B</kbd> channel. Would we prohibit the author from using two DFD channels with RRR and GGG here? Based on earlier language, "RGB textures MUST use a single DFD channel with value `KHR_DF_CHANNEL_ETC1S_RGB`", I assume so?

lexaknyazev: For practical purposes, "Red texture" means something like "only red channel is sampled and all others are ignored".

Content-aware optimization like conditionally dropping zero-filled blue channel sounds interesting. I didn't think about that. Looking at the existing material slots, the only case where it's applicable is the combined ORM map.

Anyway, this extension describes how to store different texture types with Basis, not how to match them with material slots. The list above is just informative.

lexaknyazev: We should explicitly call out material mappings in an implementation note to this spec including edge cases and definitions of "red" and "red-green" texture types in this context.

lexaknyazev: Note, that such optimization would affect transcode target selection as per the DevGuide. Still, I'm in favor of allowing the O-R case as it would noticeably improve visual quality.

donmccurdy: Ok, it'd be good to get a short summary of that into this spec when we can, but no problem with merging it now.

donmccurdy (approved):

extensions/2.0/Khronos/KHR_texture_basisu/README.md
lexaknyazev: @donmccurdy 
To properly spec the ORM usage for the red-green case, we have to require that implementations exactly follow the dev guide wrt choosing transcode targets.

Specifically, can we guarantee that all implementations would sample blue channel as zero at runtime when the source texture is Red-Green?

I'm sure that as of today, no engine would support that.

Issue #1937, opened 2021-01-26
KHR_materials_variants: Why is mappings not an array of material ids?


zeux:
I suspect this is too late to change, but I thought I'd at least ask what the rationale is.

When looking at the variants extension, it seems to me that the core idea is very simple:

- On a scene level, an array of named variants is declared
- On a primitive level, for each variant there's an override that replaces the original primitive.material reference with a variant-specific one

The simplest way to model this would be to specify an array of material indices that matches the length of the variants array. This is consistent with e.g. how morph targets are defined.

So instead of:

```
        "extensions": {
          "KHR_materials_variants" : {
            "mappings": [
              {
                "material": 2,
                "variants": [0, 3],
              },
              {
                "material": 4,
                "variants": [1],
              },
              {
                "material": 5,
                "variants": [2],
              },
            ],
          }
        }
```

You'd simply have:

```
        "extensions": {
          "KHR_materials_variants" : {
            "mappings": [2, 4, 5, 3]
          }
        }
```

This is much more concise, and also simpler to represent in-memory as instead of an array-of-structs-that-contains-an-array you just have an array of primitives, which is quicker to parse and more efficient to store. It's also the natural runtime representation for engines that want to support switching between different variants.

Any holes in the mappings array can just be plugged with the primitive material.

Why was the current representation chosen?

donmccurdy:
I can speculate that one advantage of the more verbose layout might be the ability to associate application data (extras, or other extensions?) with a primitive:variant mapping. For example, a model of a car with `[basic, standard, deluxe]` variants might contain different captions for the seats in each variant, describing features specific to that variant. There would be no obvious place to store that information if the mappings were simply integers, although something like another parallel array could still be contrived.

But, I have no idea if that was the original motivation. @jercytryn, @sleroux, or @pushmatrix might be able to say. I suspect it is too late to change in any case, at least for the current extension within glTF 2.0. We can always use feedback for future specification versions.

sleroux:
> I can speculate that one advantage of the more verbose layout might be the ability to associate application data (extras, or other extensions?)

[...]

PR #1947, opened 2021-03-01
KHR_materials_variants: Relax index uniqueness restrictions

### LINKED ISSUE #1941: KHR_materials_variants: Uniqueness of material indices is overly strict
KHR_materials_variants contains the following language:

> Across the entire mappings array, each variant index and each material must be used no more than one time.

It makes sense that the variant index must be used no more than one time - duplicate indices would create ambiguity.
However, I don't think the specification should restrict the duplicate occurrence of materials.

Per #1937, mappings are represented as objects to allow for application extensibility - to specify data with `extras`.
The restriction above maintains that an application must use the same extras data for any two variants that happen to have an identical material (or duplicate the material).

This is overly restrictive; if the extras data contains variant/mesh-specific annotations, this data doesn't have to match between different variants even if they use the same material.

I propose to change the language to:

> Across the entire mappings array, each variant index must be used no more than one time.

Since this is a relaxation of the current requirements, it shouldn't present compatibility issues wrt assets, assuming that existing implementations don't have the same restriction internally.

zeux:
KHR_materials_variants currently restricts both variant and material index to be unique within each mapping

It makes sense that the variant index must be used no more than one time - duplicate indices would create ambiguity.
However, the specification should restrict the duplicate occurrence of materials.

Per #1937, mappings are represented as objects to allow for application extensibility - to specify data with extras.
The restriction above maintains that an application must use the same extras data for any two variants that happen to have an identical material (or duplicate the material).

This is overly restrictive; if the extras data contains variant/mesh-specific annotations, this data doesn't have to match between different variants even if they use the same material.

This change removes the uniqueness requirement for material index while maintaining it for variant indices.

Since this is a relaxation of the current requirements, it shouldn't present compatibility issues wrt assets, assuming that existing implementations don't have the same restriction internally.

Fixes #1941.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_variants/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_variants/README.md
@@ -64,7 +64,7 @@ Viewer-specific logic defines, for a given instance of a glTF asset, up to one s
 
 For a given primitive, each mapping item represents a material that should be applied to the primitive when one of its _variants_ is active. Available _mappings_ are defined on the primitive node as an array of objects. Each object specifies a material by its index in the root level `materials` array and an array of variants each by their respective indices in the root level `variants` array, defined above.
 
-Across the entire _mappings_ array, each variant index and each material must be used no more than one time.
+Across the entire _mappings_ array, each variant index must be used no more than one time.
 
 When the active variant is referenced in a mapping, a compliant viewer will apply its material to the primitive.
 


sleroux:
Hey Zeus,

Sorry for the late response on this. The reasoning behind the use of the extras object is sound and it would be more clear if we were to remove the restriction of requiring materials to also be unique here. I'm in favor of relaxing this restriction as it would make it easier to implement extras and mesh/primitive specific annotations without losing the structure the extension provides.

sleroux (approved):

lexaknyazev:
@donmccurdy 
Do we need more approvals or could this PR be merged as-is?

donmccurdy (approved):
Looks good to me, I'm inclined to consider this a "bug fix" given the intended use. 👍

PR #1948, opened 2021-03-01
KHR_lights_area

bhouston:
here is a quick proposal for a rect area light (expanded to include sphere and circle per requests below.)  This specification does not determine its implementation, just its specification and intended behavior.

Real-time rect area lights are though generally implementing using the 2016 SIGGRAPH LTC method described here: https://eheitzresearch.wordpress.com/415-2/

Three.js area light:
* Docs: https://threejs.org/examples/webgl_lights_rectarealight.html
* Example: https://threejs.org/docs/#api/en/lights/RectAreaLight

One extra thought: Maybe this should have been build upon the existing ```KHR_lights_punctual``` extension and just add another light type to it called "rect"?

### DIFF
--- a/extensions/2.0/Khronos/KHR_lights_rect_area/README.md
+++ b/extensions/2.0/Khronos/KHR_lights_rect_area/README.md
@@ -1,0 +1,142 @@
+# KHR\_lights\_rect\_area
+
+## Contributors
+
+* Ben Houston, Threekit, <https://twitter.com/BenHouston3D>
+
+Copyright (C) 2021 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Proposed
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension defines a rectangular area light for use with glTF 2.0.
+
+Many 3D tools and engines support built-in implementations of rect area types. Using this extension, tools can export and engines can import these lights. 
+
+This extension defines a single "rect" area light type: `rect`.
+
+These lights are referenced by nodes and inherit the transform of that node.
+
+A conforming implementation of this extension must be able to load light data defined in the asset and has to render the asset using those lights. 
+
+## Defining Rect Area Lights
+
+Lights are defined within a dictionary property in the glTF scene description file, by adding an `extensions` property to the top-level glTF 2.0 object and defining a `KHR_lights_rect_area` property with a `lights` array inside it.
+
+The area light is assumed to point down the -Z axis.
+
+`width` is along the X-axis, and assumed to be centered.
+
+`height` is defined along the y-axis, and assumed to be centered.
+
+The following example defines a white-colored rect area light.
+
+```javascript
+"extensions": {
+    "KHR_lights_rect_area" : {
+        "lights": [
+            {
+                "width": 1.0,
+                "height": 1.0,
+                "intensity": 1.0,
+                "color": [
+                    1.0,
+                    1.0,
+                    1.0
+                ],
+            }
+        ]
+    }
+}
+```
+
+## Adding Light Instances to Nodes
+
+Lights must be attached to a node by defining the `extensions.KHR_lights_rect_area` property and, within that, an index into the `lights` array using the `light` property.
+
+```javascript
+"nodes" : [
+    {
+        "extensions" : {
+            "KHR_lights_rect_area" : {
+                "light" : 0
+            }
+        }
+    }            
+]
+```
+
+The light will inherit the transform of the node.
+
+## Rect Area Properties
+
+| Property | Description | Required |
+|:-----------------------|:------------------------------------------| :--------------------------|
+| `name` | Name of the light. | No, Default: `""` |
+| `color` | RGB value for light's color in linear space. | No, Default: `[1.0, 1.0, 1.0]` |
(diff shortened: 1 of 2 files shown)


extensions/2.0/Khronos/KHR_lights_rect_area/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+| Property | Description | Required |
+|:-----------------------|:------------------------------------------| :--------------------------|
+| `name` | Name of the light. | No, Default: `""` |
+| `color` | RGB value for light's color in linear space. | No, Default: `[1.0, 1.0, 1.0]` |
+| `intensity` | Brightness of light in. The units that this is defined in depend on the type of light in terms of luminous flux per unit area (1 lux = 1 lumen/m^2 = 1 candela * sr / m^2) | No, Default: `1.0` |
+| `width` | The centered width of the light along the X-axis. | :white_check_mark: Yes |
+| `height` | The centered height of the light along the Y-axis. | :white_check_mark: Yes |
+
+QUESTION: If an area light is scaled down by parent transforms, does it emit less light because it has less world space area?  Currently this would be the behavior given the current definition.  This seems intuitively like it makes sense.
```
lexaknyazev, line 89: What do you think about the simplifying the object a bit, specifically:
- define an area light source as 1x1 square, set intensity relative to 1m<sup>2</sup>;
- remove `width` and `height` properties;
- derive real-world size from node transforms?

mvaligursky:
In Playcanvas we decided to go with shape property, which can be specified for current Directional, Spot or Omni lights, and these are the shapes we support:
```
 * @description Infinitesimally small point light source shape.
export const LIGHTSHAPE_PUNCTUAL = 0;

 * @description Rectangle shape of light source.
export const LIGHTSHAPE_RECT = 1;

 * @description Disk shape of light source.
export const LIGHTSHAPE_DISK = 2;

 * @description Sphere shape of light source.
export const LIGHTSHAPE_SPHERE = 3;
```

So it'd be great to make this extension more generic than only supporting rect area lights?

example of some types:
http://playcanvas.github.io/#graphics/area-lights.html

bhouston:
> In Playcanvas we decided to go with shape property, which can be specified for current Directional, Spot or Omni lights, and these are the shapes we support:

It increases the complexity a fair bit but disk and sphere are useful light types.  Capsule is the other light area type that I know UE4 supports.

extensions/2.0/Khronos/KHR_lights_rect_area/README.md
bhouston: You are current that one could do this, but in my experience every implementation I've see has width and height as explicit parameters:

https://docs.unrealengine.com/en-US/BuildingWorlds/LightingAndShadows/LightTypes/RectLights/index.html

http://playcanvas.github.io/#graphics/area-lights.html

https://docs.chaosgroup.com/display/VRAYNUKE/Rectangle+or+Area+Light+%7C+VRayLightRect

lexaknyazev: Okay, I see. My concern here is that this design makes it possible to represent the same light in infinite amount of ways by combining light's own dimensions with node transforms. Such flexibility should be thoroughly evaluated for a portable runtime transmission format.

On the other hand, if we can precisely specify how all these properties interact with each other, it might be better to keep them.

bhouston:
@mvaligursky based on your feedback I have converted this from just a "rect" light extension to one that covers "sphere", "disk" and "rects."

While this generalization makes sense, I do worry that it increases the implementation burden on 3D engines for supporting this extension, and thus may make it harder to get buy-in. 

But I think it is important to be responsive to feedback and to go with the decision-making flow of this group.

kowsheek:
We have a similar custom extension that we use in [BRIO](http://brioxr.com/) for functionalities in ray tracing. This spec would be really good to have a standardized approach.

Only change I'd recommend here is to change the `type` field to `shape`.

bhouston:
@kowsheek thanks for the support -- I've made that change!  It is a good one.

bhouston:
Started a survey here on whether to expand to sphere and disk based on @mvaligursky's suggestion: https://twitter.com/BenHouston3D/status/1368966602205630464

hypnosnhendricks:
https://alextardif.com/arealights.html

extensions/2.0/Khronos/KHR_lights_rect_area/README.md
siliconvoodoo: I implemented it using a naming convention and I went for 1x1 + node transform.

siliconvoodoo:
I don't think a sphere is an area light. It's a sphere light. Or a gaussian sphere light or whatever. Area must be flat. Or you're in for capsule too. I'd limit this to a few basic flat shapes. Polygonal or spheres and capsules and whatnot, could be obtained using a different extension lights_mesh and use a node that points to a mesh so the shape is free.

fawadtariq:
Are there any updates on this ? @bhouston @donmccurdy

KiaArmani:
Are there any news on this feature? Is there anything that prevents this from being pushed further? I'd love to help!

brandon-lb:
Yes this would be really nice to have!

BruceCherniak:
What is the community's "2024" thoughts on an area light extension?

With @julienduroure  fantastic glTF Blender exporter additions, glTF is a first-class citizen. 
 Emissive geometries can somewhat fulfill the need for area lights, yet there are many reasons not to rely solely on these in a pathtracer, as well as the orthogonality in handling "materials as lights".  True area light definitions would be a real benefit to the glTF spec.

PR #1950, opened 2021-03-03
Change IBL units from lux to nits

bhouston:
There appears to be a mistake in the EXT_lights_image_based extension.  It specifies the IBL to be in units of lux.  It should instead be nits.

Motivated by this comparison of light units for various 3D engines and formats:
https://docs.google.com/spreadsheets/d/1Ce9XCC2Ub9eVjAQdYbYdLvlkDBJ_0GtAyRbSF0KJWzE/edit?ts=603fccbd#gid=0

Which prompted this twitter discussion:
https://twitter.com/BenHouston3D/status/1367187488536723462


bsdorra:
I’ve some troubles to understand the physical meaning of luminance(nits) for IBL emission. An (infinite) IBL is a inherently non-physical light source which doesn’t have a physical dimension and its emission changes globally with evaluation angle. So both integration domains for luminance (area and solid angle) are non-existant. In fact, an IBL can be seen as an infinite amount of directional lights, and as it's being done for directional lights, its unit can be interpreted as the directional contribution to a scene as Lux. So how much light does a receiver receives from a sample direction per unit-area in lm/sr2.

bhouston:
I misunderstood the purpose of this extension.  It appears that it is prefiltered cube maps.  I figured it wasn't pre-filtered.  Thus I believe the units are correct.

PR #1951, opened 2021-03-08
KHR_lights_dome

bhouston:
This is a polished version of EXT_lights_image_based.

### DIFF
--- a/extensions/2.0/Khronos/KHR_lights_dome/README.md
+++ b/extensions/2.0/Khronos/KHR_lights_dome/README.md
@@ -16,28 +16,52 @@ Written against the glTF 2.0 spec.
 
 This extension provides the ability to define a single image-based dome light in a glTF scene.  A dome light illuminate the scene as if the light was coming from a sphere outside of the scene extents.  Only one dome light per scene is supported to avoid confusing this dome light extension with being a specular or diffuse light probe solution.
 
-Many 3D tools and engines support image-based global illumination but the exact technique and data formats employed vary.  This extension differs from the EXT_lights_image_based extension in that it does not specify the exact implementation of global illumination from the dome light but rather specifies the precise meaning of the dome light, leaving the implementation as an exercise for the engine.  It also makes interchange of the data simplier because this extension does not make assumptions of implementation.
+Many 3D tools and engines support image-based global illumination but the exact technique and data formats employed vary.
 
-The extension supports only a single type of dome light, an image-based dome light, enough though other simplifer dome light specifications exist (spherical harmonic, solid color, etc.).  The lack of choice is motivated by the fact that all other types of dome lights can be viewed as simplification of a sufficiently high resolution image-based dome light.
+### Differences from EXT_lights_image_based
 
-The extension also specifies that all cube maps images must be encoded in RGBE if they are low dynamic range.  If the image format is high dynamic range it is assumed to be in linearly encoded.  While RGBE is unlikely to be the run-time format, it is the format that has the largest dynamic range and is also the most well established format for 8-bit per channel HDR images.  It is expected that the run-time engine will re-encode this input image format to whatever is appropriate for the engine itself.
+This extension is similar to the EXT_lights_image_based extension but differs in a bunch of important ways.  These differences are motivated by trying to not presuppose the implementation of the rendering method, rather than letting that leak into the data transfer format specification.  It also simplifies what is stored to the minimum, any additional data can be derived as necessary by the rendering engine.
 
-This format is purposely highly opinionated in order to simplify implementation.
+The differences are:
+
+- The cube map is stored in non-prefiltered form (in nits units), rather than prefiltered (in lux units.)
+- Only the top level full resolution cube map is stored, rather than the mip chain.
+- The cube map is assumed to be encoded in RGBE if low dynamic range, rather than RGBD.
+- No spherical harmonic is stored, if needed it can be generated at run-time.
+- No rotation quaternion on the image cube map is specified.
+
+The aim of these differences is to simplify the adoption of this extension by rendering engines.
 
 ## Defining Dome Lights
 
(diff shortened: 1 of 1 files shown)


extensions/2.0/Khronos/KHR_lights_dome/README.md
```diff
@@ -0,0 +1,129 @@
...
+### Differences from EXT_lights_image_based
+
+This extension is similar to the EXT_lights_image_based extension but differs in a bunch of important ways.  These differences are motivated by trying to not presuppose the implementation of the rendering method, rather than letting that leak into the data transfer format specification.  It also simplifies what is stored to the minimum, any additional data can be derived as necessary by the rendering engine.
+
+The differences are:
+
+- The cube map is stored in non-prefiltered form (in nits units), rather than prefiltered (in lux units.)
+- Only the top level full resolution cube map is stored, rather than the mip chain.
+- The cube map data, if specified via a PNG, is assumed to be encoded in R11G11B10F, rather than RGBD.
+- No spherical harmonic is stored, if needed it can be generated at run-time.
```
rsahlin, line 31: I would suggest adding the (diffuse) spherical harmonics, even though it can be generated at run-time we do want to keep runtime processing down and the data to store it is very small.

extensions/2.0/Khronos/KHR_lights_dome/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+Many 3D tools and engines support image-based global illumination but the exact technique and data formats employed vary.
+
+### Differences from EXT_lights_image_based
+
+This extension is similar to the EXT_lights_image_based extension but differs in a bunch of important ways.  These differences are motivated by trying to not presuppose the implementation of the rendering method, rather than letting that leak into the data transfer format specification.  It also simplifies what is stored to the minimum, any additional data can be derived as necessary by the rendering engine.
+
+The differences are:
+
+- The cube map is stored in non-prefiltered form (in nits units), rather than prefiltered (in lux units.)
```
rsahlin, line 28: I think this is a good idea and it has been suggested before, if I remember correctly there is also a suggestion as to what algorithm to use for filtering.
It would be great if we could propose an algorithm to use for filtering as well.

extensions/2.0/Khronos/KHR_lights_dome/README.md
```diff
@@ -0,0 +1,129 @@
...
+- Only the top level full resolution cube map is stored, rather than the mip chain.
+- The cube map data, if specified via a PNG, is assumed to be encoded in R11G11B10F, rather than RGBD.
```
rsahlin, line 30: I think we should consider using KTX for textures, I also think this is one of the major concerns regarding the existing EXT_lights_image_based (that it uses PNG and not KTX)

bhouston: Good idea.  I asked that on twitter here: https://twitter.com/BenHouston3D/status/1368956654776164364. I didn't get a response as to whether it is guar rented to be lossless.  Remember if one encodes R11G11B10F into an image it better be lossless or you'll get garbage out as the lossy encoder won't be error tracking properly.  Also KTX/Basis in glTF doesn't yet support a true HDR texture right?

lexaknyazev: KTX v2 container as it natively supports cubemaps. For the payload format I'd strongly suggest BC6H as it has much better precision and lower memory footprint than packed floats.

Platforms that don't support BC6H natively, can decode it to one of three uncompressed floating-point formats with these highly-optimized wasm decoders: https://github.com/KhronosGroup/BC6H-Decoder-WASM

extensions/2.0/Khronos/KHR_lights_dome/README.md
bhouston: I think if you do run-time pre-filtering you do not need the SH representation at all.  If you only wanted a low resolution light, which an SH offers, you can just give it small 128x128 cube map.  I think forcing implementations to adopt SH is not required for good results, and it increases complexity.

extensions/2.0/Khronos/KHR_lights_dome/README.md
bhouston: I could make this extension dependent upon KTX2 + BC6H support.  Is BC6H always supported if KTX2 is supported by glTF?  Do the run-time glTF engines already support BC6H?

lexaknyazev: BC6H support is a hardware feature.

bhouston: There appears to be no support for BC6H on mobile at this time: https://docs.unity3d.com/Manual/class-TextureImporterOverride.html.   This is likely to be the main users of glTF.  The BC6H encoder to R11G11B10F seems to be about 10 KB in size.  That is small enough to be acceptable.  This means that we could support a true native HDR format.

lexaknyazev: To further reduce transmission size, BC6H data could be RDO-encoded (TBD) and losslessly compressed with Zstandard (already included in KTX v2, currently used only with UASTC codec as a part of `KHR_texture_basisu`).

Another work is in progress to improve mobile support.

bhouston: @lexaknyazev do you have a slide deck that covers the benefits and downsides of these formats?  I heard there is one.

lexaknyazev: It was an internal presentation I did over a year ago and it's quite outdated now.

Here are some slides from it (updated a bit):
image
image
image
image
image
image
image

bhouston:
Closing in favour of https://github.com/KhronosGroup/glTF/pull/1956

PR #1954, opened 2021-03-17
KHT_texture_float_rgb9e5: Float format for material textures

rsahlin:
The purpose of this extension is to enable use of HDR material textures by means of texture float format VK_FORMAT_E5B9G9R9_UFLOAT_PACK32.  
KTX v2 shall be used as a container, the extension is defined in a way that follows the paradigm of KHR_texture_basisu.  

The intended usecase for this extension is any usecase where HDR image resources needs to be specified, for instance in an environment map (IBL) but it could be relevant for _any_ material texture where increased range is wanted.  

The format VK_FORMAT_E5B9G9R9_UFLOAT_PACK32 is chosen based on previous discussions with input from, Alexey, Don, Norbert and others.

### DIFF
--- a/extensions/2.0/Khronos/KHR_texture_float_rgb9e5/README.md
+++ b/extensions/2.0/Khronos/KHR_texture_float_rgb9e5/README.md
@@ -1,0 +1,219 @@
+﻿# KHR_texture_float_rgb9e5
+
+## Contributors
+
+* Rickard Sahlin, mailto:rickard.sahlin@ikea.com  
+* Alexey Knyazev [@lexaknyazev](https://github.com/lexaknyazev)  
+* Don McCurdy, Google, [@donrmccurdy](https://twitter.com/donrmccurdy)  
+* Norbert Nopper, mailto:nopper@ux3d.io  
+
+Copyright (C) 2021 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds the ability to specify textures using KTX v2 images with float format as defined by VK_FORMAT_E5B9G9R9_UFLOAT_PACK32.
+
+When the extension is used, it's allowed to use value `image/ktx2` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_float_rgb9e5` texture extension object.
+
+At runtime, engines are expected to check hardware support for VK_FORMAT_E5B9G9R9_UFLOAT_PACK32, if not present the texture should be converted to a suitable supported format.
+[TBD - what should be done when there is no hw support for VK_FORMAT_E5B9G9R9_UFLOAT_PACK32?]
+
+## glTF Schema Updates
+
+The `KHR_texture_float_rgb9e5` extension is added to the `textures` object and specifies a `source` property that points to the index of the `image` which defines a reference to the KTX v2 image in VK_FORMAT_E5B9G9R9_UFLOAT_PACK32.
+
+The following glTF will load `image.ktx2` in clients that support this extension, and otherwise fall back to `image.png`.
+
+```json
+{
+    "asset": {
+        "version": "2.0"
+    },
+    "extensionsUsed": [
+        "KHR_texture_float_rgb9e5"
+    ],
+    "textures": [
+        {
+            "source": 0,
+            "extensions": {
+                "KHR_texture_float_rgb9e5": {
+                    "source": 1
+                }
+            }
+        }
+    ],
+    "images": [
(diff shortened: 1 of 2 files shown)


lexaknyazev:
As I [mentioned](https://github.com/KhronosGroup/glTF/pull/1951#discussion_r589637391) earlier, BC6H is a superior choice. For platforms without hardware support, [these decoders](https://github.com/KhronosGroup/BC6H-Decoder-WASM) provide nearly real-time processing speed.

rsahlin:
Sure, I understand that @lexaknyazev - but how do we solve that when we want to combine it with non pre-filtered cubemaps?

lexaknyazev:
Pre-filtering is unrelated here. BC6H is a pixel format, just like RGB9E5 but with different transmission / bandwidth characteristics.

rsahlin:
Well - one specific usecase for a float (HDR) texture format is as a cubmap in an IBL.  
As such it is interesting to choose a format that can be accepted by us as a group,  including possible need for specifying a non-prefiltered cubemap.  
I'm fine with either and will create a PR using BC6H.

rsahlin:
> As I [mentioned](https://github.com/KhronosGroup/glTF/pull/1951#discussion_r589637391) earlier, BC6H is a superior choice.

Well that depends on your usecase - BC6H is a compressed format and will suffer from (more) artefacts than for instance RGB9E5.  
I want us to be able to provide both a (near) lossless float format and one that is more effective when it comes to transmission/ram.

lexaknyazev:
RGB9E5 is hardly lossless. Due to its shared exponent encoding, a channel that has the maximum value dictates the precision of two other channels. The only lossless option is F16 that would take twice as much GPU memory.

Block-compression of BC6H is optimized for F16 workflow and with a high-quality RDO encoder (TBD) would naturally provide a range of quality / size options without complicating the spec and client implementations.

rsahlin:
Sure - both formats are lossy.
My point is to provide the option of having one format that provides a higher quality at a (memory) cost.

rsahlin:
Maybe we should consider VK_FORMAT_R16G16B16_SFLOAT - which is a 48 bit format instead of the 32 bits that RGB9E5 occupies - ie 50% more than RGB9E5 and not double.  
Or make this into a half float extension that supports both the 3 and 4 component version of F16.

lexaknyazev:
3-component half-float format does not exist in hardware in most cases. When used, it's expanded to RGBA_16F in the driver thus wasting even more memory and processing cycles.

rsahlin:
Could you explain what hardware and 3D graphics API are you referring to?
To the best of my knowledge GPU hardware has supported the 3 component half float texture format (natively) for many years.

lexaknyazev:
3-component half-float has **no** support in:
- Apple Metal
- Microsoft Direct3D

3-component half-float has **limited** support in Vulkan:
- No support on NVIDIA or AMD on desktop.
- No support on Qualcomm (Adreno) on mobile.
- With OpenGL (ES), it's usually emulated on these paltforms.

rsahlin:
Ok, I did not know that about Metal/Direct3D....

So what would you suggest as a high(er) quality uncompressed float format?

lexaknyazev:
> So what would you suggest as a high(er) quality uncompressed float format?

The harsh reality is that there are very few choices, each with its own drawbacks:
- Two packed 32 bpp formats for ES 3.0+ platforms: RGB9E5 (with cross-channel influence) and RG11FB10F (with grayscale discoloration).
- One true uncompressed 64 bpp format with very broad support: RGBA16F that uses lots of transmission bandwidth / GPU memory.

BC6H outperforms all of these even if the image is 2x upscaled (to prevent any possible artifacts) before applying compression. Here's a brief example:

| Format             | Texture Size | Byte Length |
|--------------------|--------------|-------------|
| RGBA16F            | 2048x2048    | 33554432    |
| RGB9E5 / RG11FB10F | 2048x2048    | 16777216    |
| BC6H               | 4096x4096    | 16777216    |

rsahlin:
Ok thanks @lexaknyazev 
I have discussed this with my colleagues and in addition to the compressed BC6H format we would like to propose a F16 format - to be used in cases where the increased range is needed (at the expense of transfer size and memory consumption)
This would also enable F16 format supported by EXR (which has been requested before) - but by using KTX instead.

lexaknyazev:
> in cases where the increased range is needed

Could you elaborate? BC6H supports the same `[0.0 ... 65504.0]` range as F16, except `NaN` and `Infinity` values.

rsahlin:
Well, it's still increased range compared to what is currently available.
Maybe I should add _increased range and lossless image quality is needed_

extensions/2.0/Khronos/KHR_texture_float_r16g16b16/README.md
```diff
@@ -0,0 +1,242 @@
+# KHR_texture_float_r16g16b16
```
aaronfranke, line 1: The extension name does not match the name in the PR title.

PR #1955, opened 2021-03-17
KHT_texture_float_bc6h: Float format for material textures

rsahlin:
The purpose of this extension is to enable use of HDR material textures by means of texture float format VK_FORMAT_BC6H_UFLOAT_BLOCK.
KTX v2 shall be used as a container, the extension is defined in a way that follows the paradigm of KHR_texture_basisu.  
This extension requires KHR_texture_basisu to be supported (for transcoding purposes)

The intended usecase for this extension is any usecase where HDR image resources needs to be specified, for instance in an environment map (IBL) but it could be relevant for any material texture where increased range is wanted.

The format VK_FORMAT_BC6H_UFLOAT_BLOCK is chosen based on previous discussions with input from, Alexey, Don, Norbert and others.

### DIFF
--- a/extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
+++ b/extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
@@ -1,0 +1,219 @@
+﻿# KHR_texture_float_bc6h
+
+## Contributors
+
+* Rickard Sahlin, mailto:rickard.sahlin@ikea.com  
+* Alexey Knyazev [@lexaknyazev](https://github.com/lexaknyazev)  
+* Don McCurdy, Google, [@donrmccurdy](https://twitter.com/donrmccurdy)  
+* Norbert Nopper, mailto:nopper@ux3d.io  
+
+Copyright (C) 2021 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.  
+This extension requires the KHR_texture_basisu extension to be supported.  
+
+## Overview
+
+This extension adds the ability to specify textures using KTX v2 images with float format as defined by VK_FORMAT_BC6H_UFLOAT_BLOCK.
+
+When the extension is used, it's allowed to use value `image/ktx2` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_float_bc6h` texture extension object.
+
+At runtime, engines are expected to check hardware support for VK_FORMAT_BC6H_UFLOAT_BLOCK, if not present the texture should be transcoded converted to a suitable supported format.
+
+## glTF Schema Updates
+
+The `KHR_texture_float_bc6h` extension is added to the `textures` object and specifies a `source` property that points to the index of the `image` which defines a reference to the KTX v2 image in VK_FORMAT_BC6H_UFLOAT_BLOCK.
+
+The following glTF will load `image.ktx2` in clients that support this extension, and otherwise fall back to `image.png`.
+
+```json
+{
+    "asset": {
+        "version": "2.0"
+    },
+    "extensionsUsed": [
+        "KHR_texture_float_bc6h"
+    ],
+    "textures": [
+        {
+            "source": 0,
+            "extensions": {
+                "KHR_texture_float_bc6h": {
+                    "source": 1
+                }
+            }
+        }
+    ],
+    "images": [
+        {
+            "uri": "image.png"
+        },
+        {
+            "uri": "image.ktx2"
+        }
+    ]
+}
+```
+
+When used in the glTF Binary (GLB) format the `image` that points to the KTX v2 resource uses the `mimeType` value of `image/ktx2`.
+
+```json
+{
+    "asset": {
+        "version": "2.0"
+    },
+    "extensionsUsed": [
+        "KHR_texture_float_bc6h"
+    ],
+    "textures": [
+        {
+            "source": 0,
+            "extensions": {
+                "KHR_texture_float_bc6h": {
+                    "source": 1
+                }
+            }
+        }
+    ],
+    "images": [
+        {
+            "mimeType": "image/png",
+            "bufferView": 1
+        },
+        {
+            "mimeType": "image/ktx2",
+            "bufferView": 2
+        }
+    ]
+}
+```
+
+### Using Without a Fallback
+
+To use KTX v2 image with Basis Universal supercompression without a fallback, define `KHR_texture_float_bc6h` in both `extensionsUsed` and `extensionsRequired`. The `texture` object will then have its `source` property omitted as shown below.
+
+```json
+{
+    "asset": {
+        "version": "2.0"
+    },
+    "extensionsUsed": [
+        "KHR_texture_float_bc6h"
+    ],
+    "extensionsRequired": [
+        "KHR_texture_float_bc6h"
+    ],
+    "textures": [
+        {
+            "extensions": {
+                "KHR_texture_float_bc6h": {
+                    "source": 0
+                }
+            }
+        }
+    ],
+    "images": [
+        {
+            "uri": "image.ktx2"
+        }
+    ]
+}
+```
+
+### JSON Schema
+
+[texture.KHR_texture_float_bc6h.schema.json](schema/texture.KHR_texture_float_bc6h.schema.json)
+
+## KTX v2 Images with VK_FORMAT_BC6H_UFLOAT_BLOCK
+
+To cover usecases where a texture source shall have increased dynamic range
+
+- Swizzling metadata (`KTXswizzle`) MUST be `rgba` or omitted.
+- Orientation metadata (`KTXorientation`) MUST be `rd` or omitted.
+- Color space information in the DFD MUST match the expected usage, namely:
+  - For textures with **color data** (e.g., base color maps),
+    - `colorPrimaries` MUST be `KHR_DF_PRIMARIES_BT709`;
+    - `transferFunction` MUST be `KHR_DF_TRANSFER_SRGB`.
+  - For textures with **non-color data** (e.g., normal maps),
+    - `colorPrimaries` MUST be `KHR_DF_PRIMARIES_UNSPECIFIED`;
+    - `transferFunction` MUST be `KHR_DF_TRANSFER_LINEAR`.
+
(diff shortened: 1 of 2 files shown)


extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
```diff
@@ -0,0 +1,219 @@
...
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.  
+This extension requires the KHR_texture_basisu extension to be supported.
```
lexaknyazev, line 20: `KHR_texture_basisu` is unrelated to floating-point texture support.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
```diff
@@ -0,0 +1,219 @@
...
+## KTX v2 Images with VK_FORMAT_BC6H_UFLOAT_BLOCK
+
+To cover usecases where a texture source shall have increased dynamic range
+
+- Swizzling metadata (`KTXswizzle`) MUST be `rgba` or omitted.
+- Orientation metadata (`KTXorientation`) MUST be `rd` or omitted.
+- Color space information in the DFD MUST match the expected usage, namely:
+  - For textures with **color data** (e.g., base color maps),
+    - `colorPrimaries` MUST be `KHR_DF_PRIMARIES_BT709`;
+    - `transferFunction` MUST be `KHR_DF_TRANSFER_SRGB`.
```
lexaknyazev, line 143: sRGB transfer function is not applicable to floating-point formats. This field must be set to `KHR_DF_TRANSFER_LINEAR`.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+## KTX v2 Images with VK_FORMAT_BC6H_UFLOAT_BLOCK
+
+To cover usecases where a texture source shall have increased dynamic range
+
+- Swizzling metadata (`KTXswizzle`) MUST be `rgba` or omitted.
+- Orientation metadata (`KTXorientation`) MUST be `rd` or omitted.
+- Color space information in the DFD MUST match the expected usage, namely:
+  - For textures with **color data** (e.g., base color maps),
+    - `colorPrimaries` MUST be `KHR_DF_PRIMARIES_BT709`;
```
lexaknyazev, line 142: For material textures, there's virtually no benefit in using high-precision textures with the standard (aka bt.709) gamut. Bringing WCG material texture support to glTF would require much more infrastructure updates than adding a new pixel format.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
```diff
@@ -0,0 +1,219 @@
...
+}
+```
+
+### JSON Schema
+
+[texture.KHR_texture_float_bc6h.schema.json](schema/texture.KHR_texture_float_bc6h.schema.json)
+
+## KTX v2 Images with VK_FORMAT_BC6H_UFLOAT_BLOCK
+
+To cover usecases where a texture source shall have increased dynamic range
```
lexaknyazev, line 136: Higher-fidelity material textures and non-prefiltered IBLs have different input data requirements and unlikely to be served by a single extension.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+When the extension is used, it's allowed to use value `image/ktx2` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_float_bc6h` texture extension object.
+
+At runtime, engines are expected to check hardware support for VK_FORMAT_BC6H_UFLOAT_BLOCK, if not present the texture should be transcoded converted to a suitable supported format.
+
+## glTF Schema Updates
+
+The `KHR_texture_float_bc6h` extension is added to the `textures` object and specifies a `source` property that points to the index of the `image` which defines a reference to the KTX v2 image in VK_FORMAT_BC6H_UFLOAT_BLOCK.
+
+The following glTF will load `image.ktx2` in clients that support this extension, and otherwise fall back to `image.png`.
```
lexaknyazev, line 34: There is no fallback from a floating-point format to a regular PNG unless the range is restricted to `0.0 ... 1.0`.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
rsahlin: Yes it is, I have a hunch that transcoding may be needed to get support on targets that does not support BC6H.

lexaknyazev: `KHR_texture_basisu` provides a fixed functionality of supporting ETC1S and UASTC codecs. Both of them are (and will remain) completely unrelated to BC6H.

Decoders for BC6H were developed separately and have no dependencies on Basis Universal.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
rsahlin: AFAIK the BT709 standard supports a bitdepth of 8 or 10 - could you please explain why this would not be beneficial in our usecase?

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
rsahlin: So what is your suggestion?

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
rsahlin: So you are saying that we do not want a possible LDR fallback?

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
rsahlin: Yes, I know that but I thought that BC6H support was coming to basisu (meaning it could also be used as a transcoder should BC6H not be supported on target)?

lexaknyazev: The basisu glTF extension is complete and therefore its functionality is frozen forever (except minor spec clarifications and guidelines).

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
lexaknyazev: That entirely depends on the goals and usage. For example:
- No real fallback for non-prefiltered IBLs.
- Questionable (to say the least) fallback for WCG.
- Limited fallback for standard-range high-precision content.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
lexaknyazev: There were couple attempts at defining an IBL extension within the current ecosystem. My suggestion is to converge there. In parallel, we could work on WCG support in glTF. This is a big topic with major ecosystem implications, so it should start with a high-level discussion.

rsahlin: Sure - I agree with that.
However, in the meantime I think we could make progress with either BC6H or RGB9E5 which are needed in order for us to define enviroment light extension.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
rsahlin: Removed fallback

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
lexaknyazev: If the immediate goal of these extensions is to provide storage support for IBL, that's fine. Still, I wouldn't expect the finalized IBL extension to support more than one storage option.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
lexaknyazev: Because to get that benefit in rendering context, all these specific conditions should be met:
* Original texture source data is authored with precision greater than _8-bit sRGB_, so there is actual data in those extra bits.
  * Note that _8-bit sRGB_ has roughly 12 bits of linear precision for values near zero and about 7 bits of linear precision for values near one.
* That extra precision is amplified enough during shading so that it makes a difference on a standard screen after the tone mapping.
  * A good example would be a simple linear gradient that is stretched over a very long object so that a single `1/255` step takes lots of screen space thus making that gradient look quantized.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
rsahlin: > If the immediate goal of these extensions is to provide storage support for IBL, that's fine. Still, I wouldn't expect the finalized IBL extension to support more than one storage option.

Well, the goal is to provide support for float source images (textures) - one immediate usage of this would be the IBL.  
Written the way it is it shall also be possible to use as a texture source for materials, ie the same way that KHR_texture_basisu is used.

extensions/2.0/Khronos/KHR_texture_float_bc6h/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+This extension adds the ability to specify textures using KTX v2 images with compressed float format as defined by VK_FORMAT_BC6H_UFLOAT_BLOCK.  
+It allows specification of a source texture with increased dynamic range to be used.  
+This source texture shall be used in shader (BRDF) calculations in a way that retains the increased range.  
+However, this extension does not define the way pixel data is written to framebuffer or how swapchain is presented using the increased range.  
+
+When the extension is used, it's allowed to use value `image/ktx2` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_float_bc6h` texture extension object.
+
+At runtime, engines are expected to check hardware support for VK_FORMAT_BC6H_UFLOAT_BLOCK, if not present the extension is not supported.  
+This means that it is not allowed to transcode the texture data to some other format.
```
lexaknyazev, line 32: glTF does not prescribe runtime behaviour so this statement cannot be included. If there's an application-specific use-case, this should be a vendor extension.

rsahlin: Removed wording and added conformance section.

fire:
Is there are corresponding uncompressed float format? Something like tinyexr compiled to wasm? Feel free to direct me to another location.

PR #1956, opened 2021-03-17
KHR_environment_map

rsahlin:
This is a continuation of #1850 which seems to have gone stale and is based on a fork of glTF in the UX3D repo.  

The purpose of this extension is to add support for defining an environment map (cubemap or spherical harmonics) to a glTF scene.  
I have done my best to update according to comments in #1850  

This extension is based on EXT_lights_image_based, with the major difference being that KTX2 is used as a container format for cubemap textures. 
Cubemaps may choose to include prefiltered mip-levels , if not included this shall be done by implementations.
This extension specifies a set of texture formats that are allowed for the cubemaps.


extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,143 @@
...
+* Mike Bond, Adobe, <mailto:mbond@adobe.com>
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+This extension may use KHR_texture_basisu, KHR_texture_float_bc6h or KHR_texture_float_rgb9e5
```
lexaknyazev, line 18: `KHR_texture_basisu` is strictly limited to 8-bit data. How is it used with HDR IBL values?

rsahlin: This extension shall not mandate that an HDR IBL _must_ be used.  
If wanted an LDR image may be used.  
I'm also thinking that we may need some other mechanism to provide uncompressed 8 bit data for environment maps that does not need HDR but want to provide a lossless image (like PNG).

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,143 @@
...
+or use a float texture format as defined by KHR_texture_float_bc6h or KHR_texture_float_rgb95e.   
+
+
+```json
+"extensions": {
+    "KHR_lights_environment" : {
+        "lights": [
+            {
+                "intensity": 1.0,
+                "irradianceCoefficients": [...3 x 9 array of floats...],
```
bhouston, line 58: This is in my opinion derived data from the specular cube map.  Thus I question whether we need to have it here.  It seems like pre-mature optimization and also increases implementation work because now all engines need to support Spherical Harmonics.

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,143 @@
...
+            }
+        ]
+    }
+}
+```
+
+## Specular Radiance Cubemaps
+
+The cubemap used for specular radiance is defined as a cubemap containing separate images for each cube face. 
+The mip levels shall evenly map to roughness values from 0 to 1 in the PBR material and should be generated with a principled multi-scatter GGX normal distribution. The data in the maps represents illuminance in candela per square meter (nit).
```
bhouston, line 69: Do we need to specify the cube map mipmaps as opposed to just the top level one?  This is again an optimization because they are derived data.  Only the top most cube map is required.  Also by providing them we are deciding that they can only implement roughness via using a series of cube-maps, when in reality there are alternative implementations.  For example, one could use a method that retain more resolution than a mip chain even at high roughness levels.  Or one could have create a tree where one specifies LOD and roughness.  For example, at a 0 roughness level one could still have a mip chain for that cube map so that it avoids jaggies when zoomed out.

I would much prefer we focus on non-duplicated non-derived data in the specification of this extension.  If someone prefers to send duplicate pre-computed data, then should specify a vendor extension for that payload IMO.

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,143 @@
...
+        "extensions": {
+            "KHR_lights_environment": {
+                "environment": 0
+            }
+        }
+    }
+  ]
+```
+
+## Adding Light Instances to Scenes
```
bhouston, line 103: The following section is duplicated with the above "Using the Environment Light" section.

bhouston (commented):
Thank you @rsahlin for this.  You have done the heavily lifting dealing with the image formats, especially the new extensions to support varying image formats.

bhouston:
With regards to names, I prefer not "environment", rather Sky Dome, Dome Light or Sky Light as I think that is more descriptive.  Here is a survey of what other renderers and engines use:

- V-Ray: Dome Light
- Arnold: Sky Dome Light
- UE4: Sky Light
- Unity: Sky Box

extensions/2.0/Khronos/KHR_lights_environment/README.md
bhouston: I think that using LDR is not recommended.  Do we need this flexibility?  I feel it makes the extension have larger surface area? Could we be prescriptive to best practices only?

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,143 @@
...
+Secondly, it ensures that rendering of the image-based lighting is consistent across runtimes.
+
+A conforming implementation of this extension must be able to load the image-based environment data and render the PBR materials using this lighting.  
+
+This extension is based on EXT_lights_image_based, with added support for KTX2 as a carrier for [cubemap and] HDR texture formats.  
+
+The environment light is defined by a cubemap.  
+[TBD - shall a separate KHR_texture_cubemap extension be created?]
+Cubemaps can be supplied pre-filtered, by including the pre-filtered mip-levels, otherwise filtering will be performed at load time.  
+[TBD - point to relevant filtering algorithm]
```
bhouston, line 41: Here is a reference to the theoretical prefiltering method: https://developer.nvidia.com/gpugems/gpugems3/part-iii-rendering/chapter-20-gpu-based-importance-sampling. Faster methods are used in production than this which are more approximative.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: From a technical perspective I understand - but I'm all for a specification that allows for flexibility in use.
If wanted that should include use of LDR images, this could for instance be to save transmission size or GPU memory.

My hope is to have a specification that sort of seamlessly accepts a cubemap containing _any_ supported image format in viewer/engine. 
Hopefully the inclusion of LDR (existing JPG/PNG images) textures will not make the extension bigger or harder to implement.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: > * Unity: Sky Box

I can see that the first three could be potential naming kandidates - however, I feel that the skybox is not what we are defining.  
It's more like a leftover from the 90's when you wrapped your game with a view of the sky.
According to Unity manual there is no light contribution from skybox:
https://docs.unity3d.com/Manual/skyboxes-using.html#:~:text=%20To%20make%20Unity%20do%20this%3A%20%201,and%2C%20from%20the%20list%2C%20click%20Skybox.%20More%20

It seems to me that `environment` is used more in academia, for instance one of the early papers:
https://graphics.stanford.edu/papers/ravir_thesis/chapter4.pdf
As such I like it just because it's not attached to an engine implementation - it's more of a description of the methodology.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: I would like to have both the Lambertian contribution from the spherical harmonics and the specular from the texture.
As a runtime format I think it makes sense to bake this data offline since it is so small - otherwise implementations needs to do the convolution in rumtime (which I think is more cumbersome than implementing the spherical harmonics)

For the implementation, I think that is quite straightforward and I'll make sure to include links to examples.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: > Do we need to specify the cube map mipmaps as opposed to just the top level one?

I need to work on the wording - my intention is to allow specification of the roughness mipmap in the file - for instance if you use a compressed format you would generally not want to filter then compress at runtime :-)

For uncompressed formats I think it makes sense to allow exclusion of the (roughness) mipmaps - generating this on the fly is both quite simple and quick.

extensions/2.0/Khronos/KHR_lights_environment/README.md
bhouston: > As a runtime format I think it makes sense to bake this data offline since it is so small

Three.js does not use the SH do add the Lambertian contribution in the default workflow.  It just derived it from the convolution of the specular maps.

There are significant limitations to SH representations, even though I used to be a supporter -- and I did write the initial implementation that turned into Three.js SH support: https://github.com/mrdoob/three.js/pull/6405/files

For example UE4 no longer supports SH for sly lights (our specific use case) and instead just uses cube maps.  They say: "Prior to Unreal Engine 4.18, Static Sky Lights used to be represented by Lightmass with a 3rd Order Spherical Harmonic which didn't capture the detail that could be present in a sunrise or sunset."  https://docs.unrealengine.com/en-US/BuildingWorlds/LightingAndShadows/LightTypes/SkyLight/index.html

The changing of the state-of-the-art implementation leads me to want to avoid presupposing unnecessarily the implementation, rather than a description of the physical inputs (which are much less likely to become obsolete.)

extensions/2.0/Khronos/KHR_lights_environment/README.md
bhouston: > According to Unity manual there is no light contribution from skybox:

Unity does take lighting contributions from a Skybox: https://learn.unity.com/tutorial/creating-a-high-dynamic-range-image-skybox#5f358fb3edbc2a1563767802

> I need to work on the wording - my intention is to allow specification of the roughness mipmap in the file - for instance if you use a compressed format you would generally not want to filter then compress at runtime :-)

I believe you are presupposing implementation details which are not necessary in this mipmap chain, rather than specifying the physical inputs which are more universal.

The last time this extension was attempted, it was moved to EXT in part because it contained pre-filtered MIPS.  Here is another Three.js contributor making that case: https://twitter.com/EmmettLalish/status/1367196117801181184. And here is the author of Filmant making that case: https://twitter.com/romainguy/status/1367236298218426369.  (And remember I am also making this case.)

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,174 @@
...
+* Ben Houston, ThreeKit, <mailto:bhouston@threekit.com>
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+This extension may use KHR_texture_basisu, KHR_texture_float_bc6h or KHR_texture_float_f16
```
bhouston, line 19: I would suggest we only pick a single texture format for HDR data that is going to be standard, such as float_bc6h (which is specifically designed for the HDR use case) and only support that one.  It will make implementation easier and also conformance testing.  People can just transcode to their preferred format (of which there are many.). There is an 11KB JS decoder for BC6H which we can recommend.  For example Filament loves R11G11B10F: https://twitter.com/romainguy/status/1367926957267021824

lexaknyazev: > There is an 11KB JS decoder for BC6H which we can recommend.

Did you mean [7kB WASM decoders](https://github.com/KhronosGroup/BC6H-Decoder-WASM/tree/main/build), or is there another implementation?

bhouston: Sorry, that was the one I was referring to -- you had referred me to it earlier: https://github.com/KhronosGroup/BC6H-Decoder-WASM/tree/main/build

rsahlin: Saying now that a format is going to be standard is like taking up that magic 8-ball and asking it a question :-)

Jokes aside, I do really believe that we shall create a pragmatic yet flexible solution.
We should look outside the WebGL (and browser) usage and aim for the future.
The wasm libraries, while useful as a reference, will only do so much if your viewer is not JS based - it's not a silver bullet.
For our viewer in Java it can only serve as a reference - I'd rather start with uncompressed formats and add compressed formats using extensions such as the KHR_texture_basisu

lexaknyazev: Assuming that the Java viewer runs on a desktop, BC6H support is already there, in hardware. Besides, the compiled WASM modules could be auto-translated to C or compiled directly with LLVM. JavaScript is irrelevant here.

Uncompressed and non-universal formats are anything but pragmatic (although still fine as a vendor-specific use case).

rsahlin: Since BC6H is a texture format that needs support on runtime platform implementations need to have a fallback.
Implementing this fallback will be an effort - if your target is a browser and your language is javascript, calling the wasm module containing the transcoder will be trivial. 
Hence, I see js as relevant for the implementation effort.

For other languages the threshold will vary depending on your architecture and design goals.

>Uncompressed and non-universal formats are anything but pragmatic (although still fine as a vendor-specific use case).

What do you mean by non-universal formats?

lexaknyazev: As a standard, WebAssembly goes far beyond the web (despite its name). Standalone binary modules could be further compiled to native code or embedded as-is in almost any runtime environment via [Wasmer](https://wasmer.io/).

Calling exported Wasm functions from Java, for example, is almost as trivial as doing that from JS:
https://github.com/wasmerio/wasmer-java#example

Formats like RGB16F or RGB32F are:
- uncompressed by design and remain uncompressed on the GPU (sub-optimal bandwidth usage);
- don't compress well with zlib or zstd without domain-specific prepass (sub-optimal transmission);
- require per-pixel work (converting to RGBA16F) on a fair amount of platforms (sub-optimal loading, non-universal).

As such, I'd rather prefer standardizing on a dedicated floating-point lossless compression scheme that may decode to RGB16F/RGB32F if we can confirm that some environment maps absolutely need that level of precision.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: This is why I would like to have the option of Spherical Harmonics - choose solution based on your usecase.

If diffuse light does not encode well in the SH - skip it and let the implementation revert to fragment shader lookup.
If your diffuse light does encode well in the SH - then you free up a uniform sampler slot and get the diffuse light contribution in the vertex stage.

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -14,9 +14,10 @@
 Draft
 
 Open issues:
-Should Spherical Harmonics be supported or should clients be forced to evaluate the specular cubemap for diffuse contribution.  
-The question comes down to if resolution supplied by 9 coefficients SH is enough for most usecases?
-Since the SH are optional, and I believe the quality using 9 coefficients is reasonable,  I am leaning towards keeping them.   
+Should Spherical Harmonics be supported or should clients be forced to evaluate the specular cubemap for diffuse contribution?  
+The question comes down to if resolution supplied by 9 coefficients SH is enough for most usecases?  
+It could also be the case that a scene only wants diffuse contribution.  
+A link to an implementation that uses the cubemap for specular, diffuse and reflection would be beneficial.
```
bhouston, line 20: By default three.js viewer experiences use a singular IBL cube map and generate the roughness mip chain and then uses that for both specular and diffuse environmental IBL lighting.

https://threejs.org/examples/webgl_materials_envmaps_exr.html

Click on the "debug" option to see how these are all stored in a single image.

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -34,24 +43,21 @@ Secondly, it ensures that rendering of the image-based lighting is consistent ac
 A conforming implementation of this extension must be able to load the image-based environment data and render the PBR materials using this lighting.  
 
 The environment light is defined by a cubemap, this is to simplify realtime implementations as these are likely to use cubemap texture format.  
-Cubemaps can be supplied pre-filtered, by including the pre-filtered mip-levels, where each mip-level corresponds to increasing roughness values.
-If cubemap does not include mip-levels then filtering will be performed at load time.
-If a compressed texture format is used then pre-filtered mip-levels must be included.  
+Cubemaps shall be supplied without pre-filtered mip-maps for roughness values > 0, pre-filtering shall be done by the client.  
+[See Specular radiance cubemaps](#specular-radiance-cubemaps)  
+
+If a compressed texture format is used then pre-filtered mip-levels for roughness values > 0 shall be specified.
```
bhouston, line 50: > "If a compressed texture format is used then pre-filtered mip-levels for roughness values > 0 shall be specified."

This is undesired in my opinion.  I would prefer that no pre-filtered mip-levels are provided for both compressed and non-compressed texture formats.

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -13,14 +13,23 @@
 
 Draft
 
+Open issues:
+Should Spherical Harmonics be supported or should clients be forced to evaluate the specular cubemap for diffuse contribution.  
+The question comes down to if resolution supplied by 9 coefficients SH is enough for most usecases?
+Since the SH are optional, and I believe the quality using 9 coefficients is reasonable,  I am leaning towards keeping them.   
+
+If compressed texture format is used how should pre-filtering for roughness reflection be done?
```
bhouston, line 22: The compressed texture can be ready by the filtering code directly to create the mip chain of roughness, etc.  Remember that the three.js PMREM reads from one texture to create the pre-filtered results.  If the compressed format is supported, it will just work.  If it isn't supported one can transcode to a supported format.  There is no cost to using a compressed format when creating the filtered mips.

extensions/2.0/Khronos/KHR_lights_environment/README.md
bhouston: Notice the mip levels for Three.js stop getting smaller even at higher roughness levels:

Screen Shot 2021-04-09 at 3 55 38 PM

This is why I do not want to link roughness levels to a cube map mip chain -- because clients can do better.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: Extension is now using KTX 2 and supports formats with higher range (HDR)
I think this conversation is resolved.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: Removed this section

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: > As a standard, WebAssembly goes far beyond the web (despite its name). Standalone binary modules could be further compiled to native code or embedded as-is in almost any runtime environment via [Wasmer](https://wasmer.io/).

I think that's taking it a bit far - tbh Wasm was created with web and javascript focus in mind. 
Sure it's a solution that _can_ be used in a multitude of usecases and enviroments - but that does not mean that we as a group shall see a wasm module as 'effort free' library to implement.  

Using wasmer to call make a JNI call to wasm _seams easy enough - however,  it may not suit the architecture of the project. 
You would be introducing another 3rd party library and possibly breaking design goals by introducing JNI calls (which wont be debuggable in the JRE)

All I'm saying is that it's not as simple as:
- "Look, implementations will get this functionallity for free since it's available in a wasm lib."

rsahlin: This extension now requires KHR_texture_ktx - I see the question of what formats and how as resolved in this extension.
Please, let's continue discussion of relevant formats in KHR_texture_ktx.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: I see your point @bhouston - I just think there is a risk of confusion.
Sure, it's perfectly possible to sample the compressed textures to create filtered mips.
However, since there is (most likely) no compression capability on the target - the filtered mips will be in a non compressed format.
Implementations will end up with more uniform samplers and the filtered mips will be in a non compressed format, requiring more GPU memory.
Sure, the mips are smaller in size but it would still add some 30% or the original uncompressed size - likely to be larger than the compressed mip-level 0.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: As I stated above - the cost is in inconsistency with regards to GPU memory usage.
As a content creator you have a hard time knowing what the GPU budget will end up as.

rsahlin: I propose to make it optional to include the pre-filtered mips for compressed texture formats (bc6h).
This is solely for usecases where content creators want to have a better control and knowledge of target GPU memory.

@bhouston Is this a solution that is ok with you?

extensions/2.0/Khronos/KHR_lights_environment/README.md
lexaknyazev: > I think that's taking it a bit far - tbh Wasm was created with web and javascript focus in mind.

The scope has grown since then.

> Look, implementations will get this functionality for free since it's available in a wasm lib.

Porting its source (~1k LoC of simple math) to plain C or Java is so easy that I could do that over a weekend (not the current one, obviously). Additionally, I've got a GLSL implementation of it ready (not public yet) that performs the decoding in a single fragment shader. So, yes, implementations will get this for free eventually. Wasmer is just another approach as it could use the modules as-is.

rsahlin: I guess we just have to agree to disagree :-)

netpro2k:
Starting to look into how best to represent environment map data for scenes in Mozilla Hubs. This definitely looks pretty close to our needs. 

Its likely for our usecase we will want to be able to define multiple environment maps per scene, either on a per material level, or perhaps in more of a "light probe" way where environment maps will be effective in some spatial region, and possibly be mixed. The later certainly sounds out of scope for this extension, but allowing references on materials may make sense (the currently specified one on the scene being the default for materials not overriding it). In any case we can certainly at least share the actual `lights` definitions even if we are just pointing at it from an other extension or app specific data.

We also have a `MOZ_lightmap` extension which allows attaching baked light maps to materials (right now not very formally defined, but very simply adds a single texture to materials directly mapping to three.js's lightmap implementation) so we also have a desire to be able to use environment cubuemaps only for specular reflections and not diffuse lighting (as that is already baked into the lightmaps). Again, its possible here we can just use the `specularCubemaps` data as specified in this extension and ignore/omit the `irradianceCoefficients` for these cases, but just bringing it up as a point of discussion in case its generally applicable.

extensions/2.0/Khronos/KHR_lights_environment/README.md
netpro2k: Not an expert by any means so I don't have very strong opinions on if SH are a reasonable enough approximation, but if they are it does seem in line with the goals of GLTF to have these pre computed, as the goal is to have as direct a path from disk to gpu memory as possible.

extensions/2.0/Khronos/KHR_lights_environment/README.md
netpro2k: While I agree having to provide the full mip chain complicates the extension and in turn any tooling wishing to create assets using the extension, I don't see a good way around that for compressed formats without the runtime decompress the textures (and keep them decompressed when uploading to the GPU). 

Making it optional for uncompressed formats seems reasonable enough, though its trading tooling implementation complexity for runtime implementation complexity. I can see arguments either way there.

Can you elaborate on what you meant re:
> I believe you are presupposing implementation details which are not necessary in this mipmap chain, rather than specifying the physical inputs which are more universal.

extensions/2.0/Khronos/KHR_lights_environment/README.md
netpro2k: The PMREM textures three generates are quite interesting. I wonder the what the tradeoffs of actually using a single image like this for this, created offline, for this extension instead of cubemaps with mip chain might be. It would certainly simplify some things, and make the input format requirements pretty flexible (we wouldn't really need to require KTX), but would certainly complicate both tooling and runtime requirements quite a bit since its being quite prescriptive...

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,196 @@
...
+* Ben Houston, ThreeKit, <mailto:bhouston@threekit.com>
+
+## Status
+
+Draft
+
+Open issues:
+Should Spherical Harmonics be supported or should clients be forced to evaluate the specular cubemap for diffuse contribution?  
+The question comes down to if resolution supplied by 9 coefficients SH is enough for most usecases?  
+It could also be the case that a scene only wants diffuse contribution.
```
netpro2k, line 19: Also the inverse may be true, the scene may only want specular contribution, when it already has lighting data, be it through another lighting extension like `KHR_lights_punctual`, or something external in the runtime.

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,196 @@
...
+Implementations may calculate the irradiance from the specular radiance cubemap and use instead.  
+One possible benefit with spherical harmonics is that it is generally enough evaluate the harmonics on a per vertex basis, instead of sampling a cubemap on a per fragment basis.  
+
+[Realtime Image Based Lighting using Spherical Harmonics](https://metashapes.com/blog/realtime-image-based-lighting-using-spherical-harmonics/)  
+[An Efficient Representation for Irradiance Environment Maps](http://graphics.stanford.edu/papers/envmap/)  
+
+### Using the environment light
+
+The environment light is utilized by a scene.  
+Each scene can have a single environment light attached to it by defining the `extensions.KHR_lights_environment` property and, within that, an index into the `lights` array using the `light` property.
```
netpro2k, line 147: It may be desirable to also be able to add environment light references to materials as well, with the one on the scene here being a default fallback.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: I agree @netpro2k, SH are approximation - they will not be able to realistically capture all environments.
For low frequency type of environments they do a good job, for high frequency or HDR type of environments they are not ideal.
I will keep them based on the glTF goal of providing data that reduces load time processing.

rsahlin:
Hi netpro2k and thanks for your comments!

> Its likely for our usecase we will want to be able to define multiple environment maps per scene, either on a per material level, or perhaps in more of a "light probe" way where environment maps will be effective in some spatial region, and possibly be mixed. The later certainly sounds out of scope for this extension, but allowing references on materials may make sense (the currently specified one on the scene being the default for materials not overriding it). In any case we can certainly at least share the actual `lights` definitions even if we are just pointing at it from an other extension or app specific data.

A good idea to be able to specify multiple environment maps!
Not sure it should be on material level though (since it's not really a property of the material)
Think it makes sense on a node (spatial) level - but not sure on how to actually achieve it.

Do you have a suggestion on how to spatially override the scene environment map?

rsahlin:
> We also have a `MOZ_lightmap` extension which allows attaching baked light maps to materials (right now not very formally defined, but very simply adds a single texture to materials directly mapping to three.js's lightmap implementation) so we also have a desire to be able to use environment cubuemaps only for specular reflections and not diffuse lighting (as that is already baked into the lightmaps). Again, its possible here we can just use the `specularCubemaps` data as specified in this extension and ignore/omit the `irradianceCoefficients` for these cases, but just bringing it up as a point of discussion in case its generally applicable.

Good point - maybe I could add a value indicating the contribution/use of the `specularCubemaps`
LIke a bitflag with values:
SPECULAR - cubemap contributes to direct (incoming) light.
DIFFUSE - cubemap contributes to diffuse light - if used together with `irradianceCoefficients` implementations may choose which one they prefer.
REFLECTION - cubemap used for environment reflections - this means that a filtered cubemap needs to be created if not already present.

Do you think this could work?

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: Not sure as how to derive the use of another environment light based on a material property....
Could you please elaborate a bit more on the usecase for this?

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: Good point - please see my comment below.

bhouston:
> maybe I could add a value indicating the contribution/use of the specularCubemaps

This is non-physical, advanced and I would suggest we do not do that.  This is similar to the concept to advanced light linking, where only certain objects are affected by certain lights.  This is not supported in the punctual light extension either because it is really advanced.

bhouston:
I think that if we removed SH's from the current definition and had the mipmaps optional (I prefer removed, but it isn't that important), this could get approved by the committee.  I think because we haven't done that this extension is not getting support from the standards committee and we are moving around laterally as a result.

I think there is a clear way forward to getting this approved.  I would suggest we just go ahead and get it done?

extensions/2.0/Khronos/KHR_lights_environment/README.md
netpro2k: What I meant is objects using a given material would also be lit by a given environment map. This is how environment maps are currently specified in threejs.

The primary use case I am looking at for this is large environments where a single environment map will not be sufficient to cover the entire model. You can imagine an interior with multiple rooms, each of which might need their own environment map. There are definitely many ways to slice that and I dont think it has to be at the material level.

netpro2k:
> I think that if we removed SH's from the current definition and had the mipmaps optional (I prefer removed, but it isn't that important), this could get approved by the committee.

While I definitely appreciate the tooling simplicity this would provide,I think my main concern with this extension essentially being just "a way to associate cubemaps with a gltf" would be that it is then a bit under specified what that model should look like at runtime.

I am not particularly tied to including SHs but I do think this extension needs to at least mostly specify how whatever data it does include is to be used.

Re mips, I am still unclear on what the suggested solution is for compressed textures if we do not require including the full mip chain other than decompressing it, generating the mip chain, and then uploading the entire thing uncompressed. I suppose this is not completely unacceptable but certainly seems undesirable from a "I know what the runtime costs of this asset I am creating are going to be" perspective.

> This is non-physical, advanced and I would suggest we do not do that. This is similar to the concept to advanced light linking, where only certain objects are affected by certain lights.

I would definitely agree on the point of light linking being out of scope for this, not sure about including something to indicate how the cubemap should be used. Especially if dropping SHs (since previously it was discussed that you could omit the cubemap or SH if you only wanted specular or diffuse components) it does seem like it may be useful to be able to specify multiple ways to use the cubemap data. Don't feel particularly strong about this point either though, I can see a good argument for this being only a "purely physical" lighting representation to keep things simple. 

> Not sure it should be on material level though (since it's not really a property of the material)
>Think it makes sense on a node (spatial) level - but not sure on how to actually achieve it.

Hmm yeah its a bit conceptually weird to associate it with a material, though this is what three.js does. Associating with nodes could also make sense, though I do think any sort of spatial mapping (like specifying a bounding area of effect, falloff, etc) is probably getting out of scope for this and should be done in another extension. 

Specifying just 1 environment map for the entire gltf seems a bit restricted to the sort of "model viewer" usecase, where I am thinking more along the lines of a "level" asset which will likely need multiple.

I guess I am trying to come up with something in between single environment map per scene and a full "light probe" extension, but that might end up being too awkward.

rsahlin:
> This is non-physical, advanced and I would suggest we do not do that. This is similar to the concept to advanced light linking, where only certain objects are affected by certain lights. This is not supported in the punctual light extension either because it is really advanced.

Hi  @bhouston and thanks for your comments.
I see your point - however I'm not sure I agree on the similarity to light linking.
This would be a declaration of what is baked into the cubemap (by the exporter) and a way for engines to properly interpret this data.

rsahlin:
> I think that if we removed SH's from the current definition and had the mipmaps optional (I prefer removed, but it isn't that important), this could get approved by the committee. I think because we haven't done that this extension is not getting support from the standards committee and we are moving around laterally as a result.

I haven't got the indication that this extension is 'held back' by the inclusion of SH.
Re the previous comments, for instance by @netpro2k , I really do believe that including the SH's are consistent with the goals for glTF.
Whenever possible, reduce client side need for pre-precessing - and as stated before, I see the need for the usecase where only the diffuse (environment) light is used as contributing factor.

For the mipmaps - these should _only_ be included if compressed textures are used - otherwise mipmaps _shall_ be generated on the client side.

netpro2k:
> as it is now the IBL's are defined as an array in the asset, ie per asset there may be multiple IBL's...
>  Do you think this is enough, since a viewer could programatically change the scene IBL?

That seems reasonable enough as a scope for this extension. Then another extension (or just app specific data) can reference the lights and use them for something more complicated like a light probe setup with areas of influence.

bhouston:
Interesting work by the WebXR group on lighting estimation: https://immersive-web.github.io/lighting-estimation/

They support LightProbe.  The light probe has the following information:
- 1 9 coefficient spherical harmonic.
- 1 directional light with unbounded RGB intensity.
- one reflection map specified in either RGBA_F16 (HDR) or RGBA8 (LDR.)

One or more of these I guess can be empty.

```
interface XRLightEstimate {
  readonly attribute Float32Array sphericalHarmonicsCoefficients;
  readonly attribute DOMPointReadOnly primaryLightDirection;
  readonly attribute DOMPointReadOnly primaryLightIntensity;
};
```

```
dictionary XRLightProbeInit {
  XRReflectionFormat reflectionFormat = "srgba8";
};

partial interface XRSession {
  Promise<XRLightProbe> requestLightProbe(optional XRLightProbeInit options = {});
  readonly attribute XRReflectionFormat preferredReflectionFormat;
};
```

Some differences are obvious, such as estimating light direction and intensity.  glTF can already support this via the lights extension.  But interestingly, they can break out the directional light, then they can get away with a 9 coefficient for the spherical harmonic (well assuming there is only one major bright light, otherwise 9 coefficient SH fails again.). Lastly, they use a env map for the reflections, only the top level with no pre-convolutions.  It is up to the consuming engine to convolve it for their purposes.

It is an interesting choice - but everything is familiar to this group.

rsahlin:
Hi @johguenther 

> 1. Allow rotation or not? It's useful to tweak the look and illumination, and it would allow for a dynamic (animated, rotating)
 environment. The alternative to apply the inverse transform to the whole scene (and camera) seems an ugly workaround to me.

I don't really understand the usecase here - could you please elaborate?
Conceptually I see the environment as fixed and non rotatable, just like it is in the real world.

> 3.If the latter, is it a problem if there are multiple environments at the same time? 

I would say yes, since we do want the implementation to be accessible. 
I know that most game engines support multiple light probes / SH's - but this is extension is not meant to provide this functionallity.
The goal is to be able to specify one environment light (direct and diffuse light) - once this is in place we can look into an extension that does a similar thing but on a Node level (light probes)
I see the two as different in what they do - one environment vs multiple lights/probes. 

Hope you understand what I mean?

Best regards
/Richard

johguenther:
> I don't really understand the usecase here - could you please elaborate? Conceptually I see the environment as fixed and non rotatable, just like it is in the real world.

It probably depends on the point of view: I see it mainly as a light source (after all it's a `KHR_lights` extension) and secondarily as environment, whereas you place the emphasis the other way around. A usecase example is a studio environment (also existing in the real world) where the light boxes / diffusors can be rotated around the subject.

> The goal is to be able to specify one environment light (direct and diffuse light) - once this is in place we can look into an extension that does a similar thing but on a Node level (light probes) I see the two as different in what they do - one environment vs multiple lights/probes.

Fair enough. Is there already such a "light probes on nodes" extension in the making (a PR)?

johguenther:
I was thinking just of the HDRI environment texture, to be placed as a light at nodes. An implementation / renderer could derive the SH coefficients or other pre-filtering from that in a preprocessing step, if needed.

rsahlin:
> I was thinking just of the HDRI environment texture, to be placed as a light at nodes. An implementation / renderer could derive the SH coefficients or other pre-filtering from that in a preprocessing step, if needed.

This is what this extension aims to do for the scene - as opposed to node level.
That is, if no irradiance map (SH) is specified then the cubemap _shall_ be used to calculate irradiance.

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,261 @@
...
+                "layer": 0
+            }
+        ],
+        "images": [
+            {
+                "name": "environment cubemap 0",
+                "uri": "cubemap0.ktx2",
+                "mimeType": "image/ktx2"
+            }
+        ],
```
donmccurdy, line 89: I'm curious, why create a new `images` array under the extension rather than using the root `images` array? Similarly, it might be more idiomatic to use the root `textures` array rather than adding `cubemaps` here. For example:

```jsonc
// root
{
  "textures": [{
    "source": 0, 
    "extensions": {"KHR_lights_environment":{"layer": 0}} // optional
  }],
  "images": [{
    "uri": "cubemap.ktx2",
    "mimeType": "image/ktx2"
  }],
  "extensions": [{
    "lights": [{
      "name": "SceneEnvironment",
      "cubemap": 0,
      "irradianceCoefficients": [...],
      ...
    }]
  }],
  ...
}
```

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,261 @@
...
+            }
+        ]
+    }
+}
+```
+
+## Specular Radiance Cubemaps
+
+The cubemap used for specular radiance is defined as a cubemap containing separate images for each cube face.  
+The data in the maps represents illuminance in candela per square meter.
```
donmccurdy, line 106: I'm not sure how to interpret `cd / m²` —  wouldn't we need to know the physical area represented by a pixel to use that? Perhaps `cd / sr` (sr = steradians)? I notice Unreal Engine uses `cd / m²`, but it also requires a reference size (in meters) used for projection onto the scene. It would be helpful to know if other formats (`.hdr` or `.exr`) have chosen any existing precedent here.

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,261 @@
...
+}
+```
+
+## Specular Radiance Cubemaps
+
+The cubemap used for specular radiance is defined as a cubemap containing separate images for each cube face.  
+The data in the maps represents illuminance in candela per square meter.  
+Using this information it is possible to calculate the non-direct light contribution to the scene if irradiance coefficients are not included.  
+
+Cube faces are defined in the KTX2 format specification, implementations of this extension must follow the KTX2 cubemap specification.
```
donmccurdy, line 109: It's probably necessary to say something about the color space. For material textures, the embedded color space can (and typically should) be ignored. For environment cubemaps I'm not sure – given the specified units of `cd / m²` is this always Linear-sRGB? Or do we need to allow other options?

shannon:
I just wanted to provide some feedback on this extension as I recently implemented the specular cubemap and irradiance coefficient portion (I haven't done the localized bit yet).

### Schema layout

I do think this extension would be simpler if it reused the root image and texture collections for the cubemap. As it stands now I had to put in extra extension specific loading in place to load the cubemap, which really just includes fetching and parsing the KTX. Normally though, I can just loop through the images / textures and load from there since there is already logic in place for loading ktx2 image types. Now I have to go check if there are khr_lights_environment.lights to load as well. It's not the end of the world but it just seemed a bit unnecessary. I think @donmccurdy's [suggestion](https://github.com/KhronosGroup/glTF/pull/1956#discussion_r779140239) in the review comments makes sense there.

### Irradiance Coefficients

So let me preface by saying my mathematics is not the strongest but I have read through the referenced papers on spherical harmonics and looked through the various code samples included on deriving SH. What I found was that there are slight variations that led to some confusion about what the "correct" approach would be. For example, there was one example that used the texel solid angle for applying weight to the coefficients and another that simply divided by the number of samples rather than keeping a weighted sum. So because of this discrepancy I went looking for other papers and came across [Stupid SH Tricks](https://www.ppsloan.org/publications/StupidSH36.pdf) which did a good job of breaking it down but it also used another slight variation in that it used an approximation for the solid angle.

Another major thing to consider is that if the irradianceCoefficients are supplied in the gltf file, they must match what the runtime derivation would be if they weren't. I know this seems obvious but with the above I think this may vary more than expected. In my opinion, because the gltf file should be universal, it kind of forces us to pick one rather than leaving it up to the runtime implementation. 

In addition, while looking for implementations I came across bablyonjs's implementation of spherical harmonics. Superficially [here](https://github.com/BabylonJS/Babylon.js/blob/5e6321d887637877d8b28b417410abbbeb651c6e/src/Shaders/ShadersInclude/harmonicsFunctions.fx#L10). You can see that their coefficients have integrated the reconstructions coefficients directly and have divided by pi to simulate Lambertian (more on this bit below). Should this spec take a similar approach? At least the reconstruction portion might make sense. If so the included irradianceCoefficients need to specify this or they definitely won't match runtime derivation or the shader code.

### How irradiance coefficients fit into the GGX/Lambertian BRDF model

I spent an embarrassingly long time trying to figure out why my diffuse portion was so washed out. I am using the same GGX/Lambertian BRDF model from the glTF-Sample-Viewer. I had assumed that I could swap out the Lambertian portion with the reconstructed irradiance. It did not look good. 

So I came across another implementation which did the weight a little bit differently [here](https://github.com/nicknikolov/cubemap-sh/blob/master/index.js#L100). And I thought, that can't be right. They are summing each color channel, rather than each texel. So I thought it must be off by PI somewhere as that is pretty close to 3 (the number of channels). So I divided the weight by PI and voila it looked perfect. But I could not understand why I needed to do this. Perhaps I missed it but I couldn't find it anywhere in the referenced papers. It wasn't until later that I came across the bablyonjs implementation and I realized that this is required in the ggx/lambertian model.

In any case, my point is that it would be helpful if this extension called this out somewhere or provided reference to a paper which described the relationship between SH and GGX/Lambertian.

### Cubemap Texture Formats

I think it's important to note that WebGPU does not support 3 channel texture formats (with the exception of packed formats). It also does not provide a way to write 3 channel buffers to 4 channel formats. This means that you have to pad the buffer before offloading it to the GPU resulting in extra CPU side preprocessing. I don't know if that is a problem for this spec so much as a problem for WebGPU. However, the list of supported texture formats here seemed somewhat arbitrary. It seems like we should at least also support the 4 channel variations. Alpha can be ignored in irradiance coefficient derivation anyways.

rsahlin:
Hi @shannon  and thank you so much for your great effort and comments - greatly appreciated!
Just back from christmas vacation hence the late reply :-)

### Schema Layout   
I totally see your point and have been thinking about where to include the cubemaps.   
Yes, the current solution is somewhat deviates from the normal loading procedure.  
Let's continue this discussion in the above comments.

### Irradiance Coefficients

Yes, there are slight differences 'out there' as to how the coefficients are calculated and used.  
One reason I have not (yet) detailed this process further is because glTF is a fileformat and I want to get a good balance in providing the means to include data and mandating exactly how this data is used.
That said, I do believe we should agree on one algorithm to use for calculating the coefficients (because this is the data stored in the file)  
I have not seen `Stupid SH Tricks` at a quick glance it look brilliant!
Maybe there is something there we can point to - or do you have a prefered solution for calculating the coefficients?

I would assume the divide by PI depends on if a solid angle is used when calculating the coefficients or not.  
Meaning that the need for division by PI in the shader is coupled to that (algorithm used to calculate the coefficients).     


### Cubemap Texture Formats  

This texture format limitation is nothing strange to me at all.  
Desktop discreet GPUs normally don't support 3 channel texture formats (that are not word aligned) - however I do not see the need to include RGBA texture formats - the 4'th channel will simply be unused or combined with some other one channel texture of choice :-)

Or do you see a need to actually provide the alpha channel in the cubemap?

Again - thank you so much for your comments and effort! :-)

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: Hi @donmccurdy  and thanks for your comment

I agree that it is not the obvious choice to put the cubemaps inside the light extension.  
The reason for this is backwards compatibility and separation.  

By backwards compatibility I mean that since we introduce a new fileformat (ktx) and new texture formats this may break current implementations if they go through the main `images` and/or `textures` arrays and treat them like they are jpg/png.  
This is a behavior which I think must be allowed with the glTF 2.0 spec - there simply is no other formats for images/textures.  

If the cubemaps are included in the root images/textures arrays - how are implementations that does not support this extension supposed to handle those resources?  

By separation I mean that from a data packaging perspective it may make sense to put the cubemaps inside the extension to make it abundantly clear that these resources are only ever used if the extension is supported (and used)

I understand that the behavior to load the cubemaps will be slightly different than to _normal_ resources - but I think it's worth it due to the above reasons.   

Please let me know what you think!

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: The way I think about it is that the output uses the same units meaning that the area represented is only resolved in the final step, usually in the display.
Modern displays knows how many candelas they can output and the same units are used for instance in HDR content.

Does this answer your question?

On a side note, this is one reason why I think it is important to proceed with https://github.com/KhronosGroup/glTF/pull/2083

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: My assumption was that the colorspace is given by the vkFormat.
According to the KTX V2 spec the dataformat colorspace information is only needed when the format is VK_FORMAT_UNDEFINED

This would give:  
VK_FORMAT_R16G16B16_SFLOAT being a linear format in the default colorspace  
VK_FORMAT_R8G8B8_SRGB being nonlinear in the default colorspace  
The default colorspace being define by the ITU BT 709 color primaries.

Do you think this makes sense?

extensions/2.0/Khronos/KHR_lights_environment/README.md
shannon: IMO, I don't think this would break any existing implementation in the same way that the KHR_texture_basisu extension didn't by adding image/ktx2 mime type in the images. Otherwise an implementation that didn't support KHR_texture_basisu should run into the same issue you described. 

In practice I think most implementation would load what they support and either ignore the rest or load them generically.

For example, in my image loading implementation (WebGPU/WebGL based), for jpg/png or any web loadable image type I load it using an Image element and use createImageBitmap to send it to the GPU. For everything else I simply load it as an array buffer, working under the assumption that any spec compliant object that references that image will know what to do it with it from there. As is the case with KHR_texture_basisu, and now this KHR_lights_environment would as well. In my case with the current spec of this extension, I reused the same Image loading logic I just had to call it again from the extension specific logic rather during the root load logic.

shannon:
@rsahlin more than happy to contribute where I can.

I don't really have a preference for which algorithm to use. I ended up using the Stupid SH Tricks algorithm because it was the simplest to implement and since it used an approximation I assumed it must be slightly faster.
```
float f[],s[];
float fWtSum=0;
Foreach(cube map face)
    Foreach(texel)
        float fTmp = 1 + u^2+v^2;
        float fWt = 4/(sqrt(fTmp)*fTmp);
        EvalSHBasis(texel,s);
        f += t(texel)*fWt*s; // vector
        fWtSum += fWt;
f *= 4*Pi/fWtSum; // area of sphere
```

My understanding is that this is just the same thing but with an approximation of the solid angle. I did replace it with an actual solid angle calculation and it made no visible difference. In both cases I needed to keep the 1/PI or it was very blown out. After more reading on it, I do think it really just needs to be divided by pi to fit into the Lambertian BRDF formula. 
See:
https://google.github.io/filament/Filament.md.html#toc5.3.4.6 
https://google.github.io/filament/Filament.md.html#toc4.5

In either case, if this spec defined that the full solid angle should be calculated I would just drop the approximation. It only affects the preprocess time and not render time anyways.

I've included the two relevant JS and WGSL shader snippets here for reference:

```js
async deriveIrradianceCoefficients(commandEncoder, cubemap) {
        const start = performance.now();
        //Downsample to SPHERICAL_HARMONICS_SAMPLE_SIZE (128x128 seems to be sufficient)
        //Downsampling also ensures that the data is in rgba32float format regardless of the input ktx format.
        const data  = await this.downsampleCubemap(commandEncoder, cubemap);

        const size = SPHERICAL_HARMONICS_SAMPLE_SIZE;

        const coefficients = [...new Array(9)].map(() => new Float32Array(3));

        const texel = (x, y, f) => {
            const i = ((f * size * size) + ((size * y) + x)) * 4;
            return new Float32Array(data.buffer, data.byteOffset + (i * 4), 3);
        }

        let weightSum = 0;
        for(let f = 0; f < 6; f++) {
            for (let y = 0; y < size; y++) {
                for (let x = 0; x < size; x++) {
                    const u = ((x + 0.5) / size) * 2.0 - 1.0;
                    const v = ((y + 0.5) / size) * 2.0 - 1.0;

                    const temp   = 1.0 + u * u + v * v;
                    const weight = 4.0 / (Math.sqrt(temp) * temp);
                    // const weight = texelSolidAngle(x, y, size, size);
    
                    const [dx, dy, dz] = cubeCoord(u, v, f);
                    const color = texel(x, y, f);

                    for(let c = 0; c < 3; c++) { //this is faster than vec3 methods
                        const value = color[c] * weight;

                        //band 0
                        coefficients[0][c] += value * 0.282095;

                        //band 1
                        coefficients[1][c] += value * 0.488603 * dy;
                        coefficients[2][c] += value * 0.488603 * dz;
                        coefficients[3][c] += value * 0.488603 * dx;

                        //band 2
                        coefficients[4][c] += value * 1.092548 * dx * dy;
                        coefficients[5][c] += value * 1.092548 * dy * dz;
                        coefficients[6][c] += value * 0.315392 * (3.0 * dz * dz - 1.0);
                        coefficients[7][c] += value * 1.092548 * dx * dz;
                        coefficients[8][c] += value * 0.546274 * (dx * dx - dy * dy);  
                    }
                    weightSum += weight;
                }
            }
        }

        for(let c = 0; c < 9; c++) {
            vec3.scale(coefficients[c], coefficients[c], 4 * Math.PI / weightSum);
        }
        
        console.log(`Derived in ${performance.now() - start}ms`, coefficients);
        return coefficients;
}
```
```wgsl
fn getIrradiance(n: vec3<f32>) -> vec3<f32>{
            let c1 = 0.429043;
            let c2 = 0.511664;
            let c3 = 0.743125;
            let c4 = 0.886227;
            let c5 = 0.247708;

            var L00  = lighting.irradianceCoefficients[0];
            var L1_1 = lighting.irradianceCoefficients[1];
            var L10  = lighting.irradianceCoefficients[2];
            var L11  = lighting.irradianceCoefficients[3];
            var L2_2 = lighting.irradianceCoefficients[4];
            var L2_1 = lighting.irradianceCoefficients[5];
            var L20  = lighting.irradianceCoefficients[6];
            var L21  = lighting.irradianceCoefficients[7];
            var L22  = lighting.irradianceCoefficients[8];

            return (
                c1 * L22 * (n.x * n.x - n.y * n.y) +
                c3 * L20 * n.z * n.z +
                c4 * L00 -
                c5 * L20 +
                2.0 * c1 * (L2_2 * n.x * n.y + L21 * n.x * n.z + L2_1 * n.y * n.z) +
                2.0 * c2 * (L11 * n.x + L1_1 * n.y + L10 * n.z)
            ) / vec3<f32>(${M_PI});
}
```

For the cubemap texture formats. I don't really feel strongly about this, I just noticed that while trying to get an early estimation of the time it would take to prefilter and derive the coefficients, I found that padding the buffer from 3 channels to 4 channels before sending it to the GPU was a large portion of the time spent. About 200ms of blocking CPU time on a decent desktop. This leads to a lot of jank on startup. I'm still optimizing this so it's possible I can get rid of this in some other way.

extensions/2.0/Khronos/KHR_lights_environment/README.md
rsahlin: Aaah, now I see what you mean.

That would mean that the expected behavior at load time is to check what textures are used and then proceed to load image resources based on that usage.

I will move the images from being included _in_ the extension to be at root level images array instead.

rsahlin:
Great work @shannon - thanks :-)
I will look into this as soon as I have bandwidth - probably within 2-3 weeks.

rsahlin:
I am still spending my time completing the KHR_displaymapping_pq extension and will hopefully get around to updating this extension within a couple of weeks.

shannon:
I just wanted to share a running implementation of this extension in case anyone is interested.
 
[demo](https://revelryengine.github.io) | [source code](https://github.com/revelryengine/renderer)

This is not a finished engine so a couple of things to note:

- It will use WebGPU on Chrome Desktop and fall back to WebGL2 if not supported. WebGPU is still very much in development so it may break on a browser update. You can [force WebGL2](https://revelryengine.github.io?forceWebGL2=true) mode if that's the case. 
- You can change the environment by clicking the Lightbulb Icon ->Environment Lighting.
- You can also change the model by clicking the Cube Icon.
- One very useful thing for this particular extension is to click the Gear Icon -> Debug -> Light Info Irradiance. This was crucial for my debugging as mentioned in the conversation above.

A good portion of the shader code was ported or referenced from the [glTF-Sample-Viewer project](https://github.com/KhronosGroup/glTF-Sample-Viewer).

I wrote a [deno script](https://github.com/revelryengine/sample-environments) to generate the sample environments from polyhaven HDRs.  (Yay for WebGPU in deno!).

I did implement the localized portion of this extension but I don't have any good sample models to really test it fully. I will try to make a better one at some point as I was really just adding the bounding boxes to existing environments. Which is not ideal since the scene doesn't really line up with anything like the scene from the [reference paper](https://developer.arm.com/documentation/102179/0100/Implement-reflections-with-a-local-cubemap).

aidinabedi:
Thank you @rsahlin and everyone that's working on this. I would love to see this in glTF and am curious what the status is.

I'm asking because we're looking at proposing something that would fit very nicely with this. It's based on an Unreal feature called [HDRI Backdrop](https://docs.unrealengine.com/4.26/en-US/BuildingWorlds/LightingAndShadows/HDRIBackdrop/), which is a more generalized version of a skybox that is more suited for product visualization.


Regarding `KHR_lights_environment`, I also have some detailed questions:
- What is the motivation behind the `layer` property? Is it common to have multiple cubemaps in the same KTX V2 file? Is it in the case of mipmaps?
- Have you considered also supporting equirectangular/panorama images as an alternative to KTX cubemaps? This would make the extension less reliant on special tooling and keep the door open for a future HDR format extension (that we're also looking into).

aidinabedi:
That's excellent to hear. Will you be at [Khronos AWE event](https://www.khronos.org/events/awe-2023) next week? Would be a great opportunity to discuss some of these details there as well.


> * I do not know if it is common to have multiple cubemaps in a KTX V2 file - but the option is there so I feel we must adress it.
>   The property defaults to 0 , meaning that you can just ignore it if not used :-)

In my limited experience, I haven't encountered multiple cubemaps in a single KTX file (and can't really imagine much use case for it). Additionally, since [KHR_texture_basisu](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_texture_basisu/README.md) itself doesn't provide any `layer` property I think it would be more consistent to go without it. And this ties into one of the reasons I'm bringing it up. I see a lot of benefit from re-using the existing `textures` and `images` arrays defined in the core glTF spec (rather than defining a new distinct set of `images` and `cubemaps` that are bound to a specific extension). This would make the extension more compatible with other extensions that also need cubemaps while also avoiding the precedent of each extension defining their own set of cubemaps and images. [KHR_texture_basisu](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_texture_basisu/README.md#using-ktx-v2-images-with-basis-universal-supercompression-for-material-textures) allows defining non-2D textures as long as they are not used by any material maps.

> * Since glTF is a transmission (delivery) format we usually go for solutions that minimizes the need for target processing.
>   Meaning that, processing that can be done prior to deployment (of the glTF) should not be done on the target

I would like to argue that panoramic images are the de facto transmission/delivery format for environment maps. You'll find endless panoramic HDRIs available online, but few (if any) cubemaps. Additionally, I believe one of the goals of glTF (at least in my view) is to standardize features already implemented - and I have yet to encounter a 3D engine (whether web or native) that doesn't support panoramic images out-of-the-box. For some it's also the only way to input an environment (like Unreal, Sketchfab, and even [the glTF sample viewer](https://i.imgur.com/xDWlhA3.png)). On top of that it would additionally removes the extension's reliance on a single image format, which makes it significantly more attractive for engines and platforms that don't prioritize supporting KTX.


Sincerely,
Aidin

shannon:
I have been trying to optimize my implementation of this a bit to get rid of the jank on startup. Something I am noticing is that the prefilter shader is not cheap and it needs to run faces*roughnessLevel*(ggx+Charlie), which leads to a total of 120 draw calls. On some environments this is pretty fast but on others it can be pretty slow and in webgl2 this leads to blocking the main UI thread (compositor).

Does it make sense to store all of this in the same KTX file ahead of time? With roughness stored as mip levels along with a separate LUT. Before I implemented this extension, I stored the ggx and Charlie as two separate cubemaps already prefiltered and loaded them directly. Perhaps this could be a use case for the layers in KTX.

I'm really just thinking out loud as I am not an expert in this area. My understanding is this would require defining a specific BRDF for this extension. For now I am just going to break up the prefiltering to something like running one roughness level per frame. Instead of all in one frame.

rsahlin:
I have created a couple of test models - one with the intention of checking roughness values and how the reflection changes with roughness going from 0 to 1

The spheres are 1m in diameter, there are two rows - one close to the viewer (at Z=0) and one row at a distance (Z = 15m)
This is also to check that mip-level selection is consistent across depth.
The spheres are pure white metal (ie 100% reflective) - just to make it easier to spot problems.
I am using a custom studio cubemap
image

rsahlin:
And here is a test model to show what the reflection will be with different IOR values - going from IOR = 1 to the left up to IOR = 3 on the right.
This is done using the KHR_materials_ior extension
I have removed all lights from the scene so that it's only the reflection that is visible.

image

aaronfranke:
See also [OMI_environment_sky](https://github.com/omigroup/gltf-extensions/tree/main/extensions/2.0/OMI_environment_sky) for an extension that specifies the skybox and environment properties, which affect the sun. Here is the pull request with screenshots: https://github.com/omigroup/gltf-extensions/pull/244

Igor-dvr:
There was a couple of PRs already and a lot of feadback:
- https://github.com/KhronosGroup/glTF/pull/1951
- https://github.com/KhronosGroup/glTF/pull/1850
- https://github.com/omigroup/gltf-extensions/pull/244
- https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_lights_image_based

What is needed to move forward? Implementation in programs like Blender?

Should this be called dome/environment light and compatible with OpenUSD dome light https://openusd.org/dev/user_guides/schemas/usdLux/DomeLight.html ?
Also Blender PR Environment/Dome light https://projects.blender.org/blender/blender/pulls/146193

extensions/2.0/Khronos/KHR_lights_environment/README.md
```diff
@@ -0,0 +1,265 @@
...
+
+The texture type of the KTX v2 file shall be 'Cubemap'  
+The texture format must be one of the following:  
+
+VK_FORMAT_R8G8B8_SRGB  
+VK_FORMAT_R8G8B8_UNORM  
+VK_FORMAT_E5B9G9R9_UFLOAT_PACK32  
+VK_FORMAT_B10G11R11_UFLOAT_PACK32  
+VK_FORMAT_R16G16B16_SFLOAT  
+VK_FORMAT_R16G16B16_UNORM
```
siliconvoodoo, line 241: VK_FORMAT_R32G32B32A32_SFLOAT
VK_FORMAT_BC6H_UFLOAT_BLOCK16
VK_FORMAT_ASTC_4x4_SFLOAT_BLOCK
other interesting compressed formats?

extensions/2.0/Khronos/KHR_lights_environment/README.md
siliconvoodoo: in my opinion using Photometric units (`cd / m²` is Nits) is a bad habit that glTF started with with the KHR_lights_punctual extension using `lm/sr` instead of `W/sr`. A proper modern data format should store Radiometric units, not these unmanageable photometric horrors. Do you know how to convert from Lumen to something workable in a shader (that being Watts)? Here it is:

image

Lagarde et al. 2014.

good luck with that.
Blender gave up and used `683×` as-is without looking at the spectrum integral at all. This bad choice is forcing people into conventional approximations, it's dirty IMHO.

Now considering we are sort of stuck with these (lest we revisit KHR_lights_punctual), then I'd be careful about the range this forces people with, you can't store meaningful lumens in a R8G8B8 format. You'd need metadata to explain the exponent, multiplier and biases.

PR #1959, opened 2021-03-27
Remove EXT_mesh_gpu_instancing 4x3 sample

donmccurdy:
The `MAT4x3` is not allowed by the final version of the `EXT_mesh_gpu_instancing` spec, so we should remove the sample to avoid confusion.


donmccurdy:
Perhaps we should move the remaining samples to the glTF-Sample-Models repository, too?

PR #1964, opened 2021-04-09
KHR_texture_ktx

rsahlin:
The purpose of this PR is to add support for textures using KTX v2 images, the intention is not to provide support for already supported formats such as the typical 8 bpp formats suppoted by JPEG or PNG, but rather to add support for formats that have increased precision and/or range.

### DIFF
--- a/extensions/2.0/Khronos/KHR_texture_ktx/README.md
+++ b/extensions/2.0/Khronos/KHR_texture_ktx/README.md
@@ -1,0 +1,244 @@
+﻿# KHR_texture_ktx
+
+## Contributors
+
+* Rickard Sahlin, Ikea 
+* Ben Houston, ThreeKit
+
+
+Copyright (C) 2021 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds the ability to specify textures using KTX v2 images.  
+An implementation of this extension can use such images as an alternative to the PNG or JPEG images available in glTF 2.0 when higher precision or range is needed.  
+
+When the extension is used, it's allowed to use value `image/ktx2` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_ktx` texture extension object.
+
+
+## glTF Schema Updates
+
+The `KHR_texture_ktx` extension is added to the `textures` object and specifies a `source` property that points to the index of the `image` which defines a reference to the KTX v2 image.
+
+The following glTF will load `image.ktx2`
+
+```json
+{
+    "asset": {
+        "version": "2.0"
+    },
+    "extensionsUsed": [
+        "KHR_texture_ktx"
+    ],
+    "textures": [
+        {
+            "source": 0,
+            "extensions": {
+                "KHR_texture_ktx": {
+                    "source": 1
+                }
+            }
+        }
+    ],
+    "images": [
+        {
+            "uri": "image.png"
+        },
+        {
+            "uri": "image.ktx2"
+        }
+    ]
+}
+```
+
+When used in the glTF Binary (GLB) format the `image` that points to the KTX v2 resource uses the `mimeType` value of `image/ktx2`.
+
+```json
+{
+    "asset": {
+        "version": "2.0"
+    },
+    "extensionsUsed": [
+        "KHR_texture_ktx"
+    ],
+    "textures": [
+        {
+            "source": 0,
+            "extensions": {
+                "KHR_texture_ktx": {
+                    "source": 1
+                }
+            }
+        }
+    ],
+    "images": [
+        {
+            "mimeType": "image/png",
+            "bufferView": 1
+        },
+        {
+            "mimeType": "image/ktx2",
+            "bufferView": 2
+        }
+    ]
+}
+```
+
+### Using Without a Fallback
+
+To use KTX v2 image without a fallback, define `KHR_texture_ktx` in both `extensionsUsed` and `extensionsRequired`. The `texture` object will then have its `source` property omitted as shown below.
+
+```json
+{
+    "asset": {
+        "version": "2.0"
+    },
+    "extensionsUsed": [
+        "KHR_texture_ktx"
+    ],
+    "extensionsRequired": [
+        "KHR_texture_ktx"
+    ],
+    "textures": [
+        {
+            "extensions": {
+                "KHR_texture_ktx": {
+                    "source": 0
+                }
+            }
+        }
+    ],
+    "images": [
+        {
+            "uri": "image.ktx2"
+        }
+    ]
+}
+```
+
+### JSON Schema
+
+[texture.KHR_texture_ktx.schema.json](schema/texture.KHR_texture_ktx.schema.json)
+
+## KTX v2 Images
+
+
+
+For the purposes of this extension, the following texture types and formats are defined:
+
+- **RGBA:** A texture that uses all four channels.  
+
+VK_FORMAT_R16G16B16A16_SFLOAT  
+VK_FORMAT_R16G16B16A16_UNORM  
+VK_FORMAT_R32G32B32A32_SFLOAT  
+
+
+- **RGB:** A texture that uses red, green, and blue channels. Alpha channel is unused and not sampled at runtime.  
+
+VK_FORMAT_R16G16B16_SFLOAT  
+VK_FORMAT_R16G16B16_UNORM  
+VK_FORMAT_R32G32B32_SFLOAT  
+
+- **Red:** A texture that uses only red channel. All other channels are unused and their values are not sampled at runtime.  
+
+VK_FORMAT_R16_SFLOAT  
+VK_FORMAT_R16_UNORM  
+VK_FORMAT_R32_SFLOAT  
+
+## KTX header fields
+
+- vkFormat must be one of:  
+VK_FORMAT_R16G16B16A16_SFLOAT  
+VK_FORMAT_R16G16B16A16_UNORM  
+VK_FORMAT_R32G32B32A32_SFLOAT  
+VK_FORMAT_R16G16B16_SFLOAT  
+VK_FORMAT_R16G16B16_UNORM  
+VK_FORMAT_R32G32B32_SFLOAT  
+VK_FORMAT_R16_SFLOAT  
+VK_FORMAT_R16_UNORM  
+VK_FORMAT_R32_SFLOAT  
+
+
+
+### Additional requirements
+
+Regardless of the format used, these additional restrictions apply for compatibility reasons:
+
+- Swizzling metadata (`KTXswizzle`) MUST be `rgba` or omitted.
+- Orientation metadata (`KTXorientation`) MUST be `rd` or omitted.
+
+
+## Known Implementations
+
+Authoring:
+
+
+Viewing:
+
+
+## Resources
+
+[KTX File Format Specification, version 2](https://github.khronos.org/KTX-Specification/)
+
+[KTX Reference Software](https://github.com/KhronosGroup/KTX-Software/)
+
+## Appendix: Full Khronos Copyright Statement
+
+Copyright 202§ The Khronos Group Inc.
+
+Some parts of this Specification are purely informative and do not define requirements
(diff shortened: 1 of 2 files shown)


lexaknyazev:
Given how glTF handles existing PNG/JPEG/Basis textures, this extension could only be used with _non-color_ texture slots (e.g. normals or ORM) for the following reason.

None of the proposed pixel formats support sRGB transfer because they already have high enough precision. Yet, all texture slots that use _color_ data are currently locked to 8-bit sRGB inputs. Down-sampling data at runtime would subvert the whole reason of having it stored in high precision. If the goal of this extension is to allow any existing texture slot to support high-precision inputs, the extension spec should redefine that explicitly by stating that in those cases a linear transfer function is used.

However the extension still cannot go beyond `0.0 ... 1.0` range as values outside of it are undefined for all existing color and non-color texture slots (ignoring IBL for now).

rsahlin:
Right - we have a spec that declares sRGB transfer function for `baseColorTexture` and `emissiveTexture`, this is a property of the material object.
We also have extensions that may provide additional color texture sources.

What if:
We add wording to this extension that changes the `baseColorTexture` and `emissiveTexture` transfer function so that it is sRGB only when the image source is image/png or image/png.
This would be similar to what KHR_texture_basisu does to image.mimetype (it 'adds' the use of image/ktx2).

>When this extension is used it is allowed to provide color textures that does not use sRGB transfer function for texture sources refered to by material source objects.  
In the core spec this affects `baseColorTexture`and `emissiveTexture`, in extensions it will affect any color texture source.  
This means that when this extension is used a texture that stores color values shall not be sRGB encoded, if it contains higher range than 8 bits per color component.

extensions/2.0/Khronos/KHR_texture_ktx/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+### JSON Schema
+
+[texture.KHR_texture_ktx.schema.json](schema/texture.KHR_texture_ktx.schema.json)
+
+## Color Texture Sources
+
+When this extension is used it is allowed to provide color textures that does not use sRGB transfer function for texture sources refered to by material source objects.
+In the core spec this affects baseColorTextureand emissiveTexture, in extensions it will affect any color texture source.  
+This means that when this extension is used a texture that stores color values shall not be sRGB encoded, if it contains higher range than 8 bits per color component.
```
lexaknyazev, line 136: Let's rephrase it in a more normative way (also shorter):
> Texture data provided by this extension is always encoded using linear transfer function. When this extension is used with material texture slots that expect sRGB-encoded values, only the fallback 8-bit texture (when present) uses sRGB encoding.
> > **Implementation Note**: For the core glTF 2.0 specification, these texture slots include `emissiveTexture` and `baseColorTexture`; for material extensions, it's usually texture slots named `*ColorTexture`.

extensions/2.0/Khronos/KHR_texture_ktx/README.md
```diff
@@ -0,0 +1,262 @@
...
+VK_FORMAT_R16G16B16_SFLOAT  
+VK_FORMAT_R16G16B16_UNORM  
+VK_FORMAT_R32G32B32_SFLOAT  
+
+- **Red:** A texture that uses only red channel. All other channels are unused and their values are not sampled at runtime.  
+
+VK_FORMAT_R16_SFLOAT  
+VK_FORMAT_R16_UNORM  
+VK_FORMAT_R32_SFLOAT  
+
```
lexaknyazev, line 162: For completeness, add Red-Green 16-bit formats.

extensions/2.0/Khronos/KHR_texture_ktx/README.md
rsahlin: Great, thanks @lexaknyazev Updated - resolving this conversation.

extensions/2.0/Khronos/KHR_texture_ktx/README.md
rsahlin: Added, thanks @lexaknyazev

extensions/2.0/Khronos/KHR_texture_ktx/README.md
```diff
@@ -0,0 +1,293 @@
...
+This extension defines a number of formats that are supported for images used as texture sources.  
+It does not define how these formats are stored internally, it is up to implementations to use a suitable texture format.  
+The source texture shall be used in shader (BRDF) calculations in a way that retains the increased range.  
+However, this extension does not define the way pixel data is written to framebuffer or how swapchain is presented using the increased range.  
+
+If ktx image 
+
+## KTX header fields
+
+`supercompressionScheme` MUST be 3 (ZLIB).
```
lexaknyazev, line 183: Why this restriction? Zstandard is faster and more efficient.

extensions/2.0/Khronos/KHR_texture_ktx/README.md
```diff
@@ -0,0 +1,293 @@
...
+  - For textures with **non-color data** (e.g., normal maps),
+    - `colorPrimaries` MUST be `KHR_DF_PRIMARIES_UNSPECIFIED`;
+    - `transferFunction` MUST be `KHR_DF_TRANSFER_LINEAR`.
+
+  > **Implementation Note**: Engines may need to generate missing mipmaps at runtime when a full mip pyramid is not available.
+
+
+### Using KTX v2 Images with Basis Universal Supercompression for Material Textures
+
+When a texture referencing a KTX v2 image with Basis Universal supercompression is used for glTF 2.0 material maps (both color and non-color), the KTX v2 image MUST be of **2D** type as defined in the KTX v2 Specification, Section 4.1.
```
lexaknyazev, line 223: This statement contradicts the supercompression restriction above.

rsahlin: This is a cut & paste error - it should read:
Using KTX v2 Images for Material Textures

extensions/2.0/Khronos/KHR_texture_ktx/README.md
rsahlin: Just because I'm implementing using ZLIB - I'll add Zstandard as well.

extensions/2.0/Khronos/KHR_texture_ktx/README.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+### JSON Schema
+
+[texture.KHR_texture_ktx.schema.json](schema/texture.KHR_texture_ktx.schema.json)
+
+## Color Texture Sources
+
+Texture data provided by this extension is encoded depending on the target format and material teture slot usage.    
+When this extension is used with material texture slots that expect sRGB-encoded values (RGB or RGBA format) and the texture format is 8 bits per pixel (VK_FORMAT_R8G8B8 or VK_FORMAT_R8G8B8A8) the texture uses sRGB encoding.
```
lexaknyazev, line 143: To align with the KTX spec, sRGB-encoded textures must use `VK_FORMAT_R8G8B8(A8)_SRGB` format and the appropriate DFD transfer function value. The subsequent sections mention only linear `_UNORM` variants.

rsahlin: Thanks @lexaknyazev I have added SRGB formats for RGB and RGBA and updated the wording.
Also added 16 bit linear encoded formats for color texture slots.

extensions/2.0/Khronos/KHR_texture_ktx/README.md
```diff
@@ -0,0 +1,329 @@
...
+```
+
+### JSON Schema
+
+[texture.KHR_texture_ktx.schema.json](schema/texture.KHR_texture_ktx.schema.json)
+
+## Color Texture Sources
+
+Texture data provided by this extension is encoded depending on the target format and material texture slot usage.    
+When this extension is used with material texture slots that expect sRGB-encoded values (RGB or RGBA format) and the texture format is 8 bits per pixel, the texture format must be one of VK_FORMAT_B8G8R8_SRGB or VK_FORMAT_B8G8R8A8_SRGB and contain an sRGB encoded image.
```
lexaknyazev, line 143: Looks like a typo: these should be `R8G8B8(A8)` not `B8G8R8(A8)`. Also below.

rsahlin: Haha - thanks @lexaknyazev - you are of course correct :-)
Will change.

atteneder:
Thanks for this draft,

The list of supported `vkFormat` values is a specific subset (meaning KTX2 supports more than that). I assume this was chosen, because those are uncompressed and universally supported across platforms, right?

Imagine upcoming demand to broaden the KTX2 support even further (e.g. to compressed formats like ASTC):

- Would it make sense to include the in this PR? (or even lift the limit to a subset)
- Should each set of formats have a dedicated extension (e.g. KHR_texture_astc) ? If so, should the naming of this extension (KHR_texture_ktx) be made more specific?
- Does that even make sense (supporting non-universally supported formats)?

Regards,
Andi

rsahlin:
Thanks for your comment @atteneder 

Yes, the formats are choses as either:
* A substitute for the (uncompressed) formats used to store PNG/JPEG source images.
* Uncompressed formats that deliver increased range or precision compared to PNG/JPEG source images.

I am not sure how to handle (directly) compressed formats such as ASTC - instead of using these the intention is to use KHR_texture_basisu - which can then be transcoded to ASTC on the target device.
Is there a reason why you do not think that KHR_texture_basisu can deliver the same result as ASTC?

So, in conclusion - I see this extension as _just_ providing uncompressed texture formats that are universally available (or super simple to convert to/from)

Hope this answers your question?

lexaknyazev:
> Is there a reason why you do not think that KHR_texture_basisu can deliver the same result as ASTC?

The only reason is that ASTC supports many more configurations and block sizes than the ASTC subset used for Basis Universal.

Still, ASTC should not be included in this extension because emulating it on devices without hardware support is far from simple - the format is very complex.

atteneder:
> Hope this answers your question?

Indeed, thanks a lot!

Admitted, the KHR_texture_astc example does not make much sense in the presence of BasisU. But even if I cannot come up with some sane example/use-case right now, I guess one extension per KTX2 format subset is the direction this is taking.

rsahlin:
Ah, ok I see thanks for the explanation @lexaknyazev
I agree - to provide support for a specific compressed texture format is not part of this extension :-)

rsahlin:
@atteneder Please feel free to chime in on what VK_ formats that should be included in this extension.

aaronfranke:
This PR has been superseded by [KHR_texture_basisu](https://github.com/KhronosGroup/glTF/pull/1751), which adds support for KTX v2 `.ktx2` images that can be used as textures. This PR should be closed.

I do see that this PR has `"layer"`, while KHR_texture_basisu does not have this. What is the use case for this? If it's still needed, perhaps this can be reworked into another extension that builds upon KHR_texture_basisu.

donmccurdy:
@aaronfranke this PR doesn't overlap with the features of `KHR_texture_basisu`, which allows only “Basis Universal” UASTC and ETC1S compressed textures. KTX2 is also capable of supporting dozens of other compressed and uncompressed formats, including floating-point textures used in image base lighting or lightmaps. `KHR_texture_ktx` was proposed to allow (some of) these other texture formats that `KHR_texture_basisu` does not.

But I think the OP is accurate that it makes sense to continue with this extension only in conjunction with adding features to glTF that require some of these formats, like IBL. And fully specifying an IBL extension is non-trivial.

mosra:
A use case I have for this extension is large-ish texture atlases using texture arrays, i.e., using a 2D array KTX2 file and the `"layer"` in this extension. And not using it for anything fancy like IBL or any fancy format besides (block-compressed) 8-bit sRGB, just in places where regular 2D images would be used for regular textures.

What's nice about how the extension is made is that it still allows a fallback to 2D images if the importer doesn't support KTX2 files or texture arrays, and how there isn't anything else to be specified to make interaction with e.g. `KHR_texture_transform` work. For me, and together with offering an alternative to `MSFT_texture_dds` with a better format, that's the main selling point of this extension. 

I feel like expanding the pixel format or texture format support should be a responsibility of some other extension, not this one, and I'm fine with it restricting the allowed pixel formats to a subset compatible with PNG/JPEG by default. Same for disallowing e.g. cube map KTX files. I think just even allowing non-Basis (array) KTX2 files to be used warrants existence of this extension, and the extras could be tangential `KHR_texture_cubemap`, `KHR_texture_half_float` etc. extensions, without tying a particular feature to a concrete file format.

Issue #1640, opened 2019-07-11
Axis convention language in spec


dubois:
From the spec:
> glTF uses a right-handed coordinate system, that is, the cross product of +X and +Y yields +Z. glTF defines +Y as up. The front of a glTF asset faces +Z.

This doesn't define right-handed because (as also mentioned in #566), x cross y = z is a mathematical identity that holds even in left-handed engines like Unity. To do otherwise would be to require left-handed math libraries in left-handed engines and then the world would end.

This would be more accurate:

> glTF uses a right-handed axis convention; that is, the cross product of right and forward yields up. glTF defines +Y as up, +Z as forward, and -X as right.

The language "+Z as forward, and -X as right" might be a little controversial since there was a lot of discussion in #1043 over whether "forward" even has a meaning (it does!); but nailing this down is critical for tool authors. eg, if two different gltf importers give different results when importing an asset into _insert engine here_, one should be able to authoritatively say which one(s) are buggy. (see also https://github.com/KhronosGroup/UnityGLTF/issues/319)

One could even argue that "front of an asset faces Z" is a bit of an overreach; it goes beyond specifying correct gltf tool behavior (importers, format converters) and starts dictating correct behavior for users creating assets.

Anyway, I'm very happy that glTF has taken a stand against ambiguous axis conventions like "Y up, right handed"! In a past life I worked with robots, and when there is no simple "flip an axis" trick to convert from one convention to another (eg [ROS](https://www.ros.org/reps/rep-0103.html) to Unity) it really brings these problems into sharp relief.

donmccurdy:
Based on our own samples, `"+Z as forward, and -X as right"` appears to be correct:

![Screen Shot 2021-04-09 at 10 37 11 AM](https://user-images.githubusercontent.com/1848368/114219520-9b17db80-991f-11eb-90f4-d7c6d55057c8.png)

I think we should make the requested clarification; any concerns with this proposal?

MikelSalazar:
The problem is that the standard we are using right now is LPT-ENU and, as the second part of the name implies, it defines the axes as East(X), North(Y) and Up (Z). Plus, in almost every geolocation system, the third axis is defined as altitude/height-over-ellipsoid, because in many applications (i.e. anything related with 2D maps) there is no third dimension.

As a 3D modeler/developer, I prefer the system  defined in GLTF but I can also understand why the situation is as it is. In any case, this is something we have to clarify very well in the documentation.

[...]

PR #1969, opened 2021-04-15
Clarify axis convention language

dubois:
Apply the change suggested in #1640

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -251,7 +251,7 @@ For property names, glTF uses [camel case](http://en.wikipedia.org/wiki/CamelCas
 
 ## Coordinate System and Units
 
-glTF uses a right-handed coordinate system, that is, the cross product of +X and +Y yields +Z. glTF defines +Y as up. The front of a glTF asset faces +Z.
+glTF uses a right-handed coordinate system. glTF defines +Y as up, +Z as forward, and -X as right. The front of a glTF asset faces +Z.
 
 ![](figures/coordinate-system.png)
 


specification/2.0/README.md
```diff
@@ -251,7 +251,7 @@ For property names, glTF uses [camel case](http://en.wikipedia.org/wiki/CamelCas
 
 ## Coordinate System and Units
 
-glTF uses a right-handed coordinate system, that is, the cross product of +X and +Y yields +Z. glTF defines +Y as up. The front of a glTF asset faces +Z.
+glTF uses a right-handed coordinate system; that is, the cross product of right and forward yields up. glTF defines +Y as up, +Z as forward, and -X as right.
```
donmccurdy, line 254: The line, "The front of a glTF asset faces +Z." is a useful clarification of "+Z as forward", as opposed to forward in some scene navigation or default camera orientation sense. I think I would keep that part. Otherwise looks good, thank you!

dubois: Done.

donmccurdy: Had to read this through a couple times:

> ... the cross product of right and forward yields up ...

I think you mean `left ⨉ forward`? It may be better to stick with the original phrasing of this clause, "cross product of +X and +Y yields +Z". The right-hand rule does not really dictate the right/forward/up convention to my understanding (see various combinations here: https://www.evl.uic.edu/ralph/508S98/coordinates.html).

And thanks for correcting the comma. :)

donmccurdy: On second reading, my link above contradicts us:

> The default coordinate system in OpenGL(TM) is right-handed: the positive x and y axes point right and up, and the negative z axis points forward. 

This might need a closer review, @lexaknyazev would be interested in your thoughts.

lexaknyazev: +X right and -Z forward is the right-handed system same as +X left and +Z forward, just rotated by 180 degrees.

specification/2.0/README.md
```diff
@@ -251,7 +251,7 @@ For property names, glTF uses [camel case](http://en.wikipedia.org/wiki/CamelCas
 
 ## Coordinate System and Units
 
-glTF uses a right-handed coordinate system, that is, the cross product of +X and +Y yields +Z. glTF defines +Y as up. The front of a glTF asset faces +Z.
+glTF uses a right-handed coordinate system; that is, the cross product of right and forward yields up. glTF defines +Y as up, +Z as forward, and -X as right. The front of a glTF asset faces +Z.
```
donmccurdy, line 254:
```suggestion
glTF uses a right-handed coordinate system; that is, the cross product of +X and +Y yields +Z. glTF defines +Y as up, +Z as forward, and -X as right. The front of a glTF asset faces +Z.
```

lexaknyazev: Arguable, the whole "cross product" part is redundant as it's true for any right-handed system.

specification/2.0/README.md
donmccurdy: Ah thanks! I'm misreading things here. 

Sounds like this can be resolved.

specification/2.0/README.md
lexaknyazev, line 254: While this statement is technically correct, it's hard to parse because the "right" direction is defined after it's used.

lexaknyazev: Removing the "the cross product ..." would eliminate the confusion without affecting normative definitions.

donmccurdy:
GitHub is showing that I deleted a comment from @lexaknyazev — I'm not sure how (I only deleted a "suggested edit" of my own?) but let me know if something has gone missing here.

> Screen Shot 2021-04-14 at 9 56 41 PM

lexaknyazev:
> GitHub is showing that I deleted a comment from @lexaknyazev — I'm not sure how (I only deleted a "suggested edit" of my own?) but let me know if something has gone missing here.

Removing a suggestion apparently removes all comments on it. I [repeated](https://github.com/KhronosGroup/glTF/pull/1969#discussion_r613754019) my comment above.

specification/2.0/README.md
dubois: Missed this comment! Done.

emackey:
Looks like this was manually merged into the jon-adoc branch [here](https://github.com/KhronosGroup/glTF/blob/7748663963f65dba3cd6727a2fb2037f8ca58413/specification/2.0/Specification.adoc#coordinate-system-and-units).

Thanks @dubois, I'm going to close this PR as the old Markdown version is going away soon, but this change is already incorporated in the upcoming version.

Issue #1970, opened 2021-04-17
Showing status of extension PRs


donmccurdy:
A small suggestion for communicating status of extensions — From the outside it is likely difficult to understand the status and outlook of any particular extension PR on this github repository. PRs are sometimes opened speculatively with a KHR_ prefix and never move forward, while other KHR_-prefixed extensions may be very close to completion.

We have markdown status table (https://github.com/KhronosGroup/glTF/blob/master/extensions/README.md#in-progress-khronos-and-multi-vendor-extensions) but it requires additional PRs to keep that up to date, and most users understandably don't.  There's a nicer pattern in the DefinitelyTyped repository of using GitHub Projects for this purpose:

https://github.com/DefinitelyTyped/DefinitelyTyped/projects/5

I think we should consider creating a similar board for extensions. This has advantages: (1) maintainers can move extensions around without PRs, or even with automation (e.g. based on GitHub Reviews). (2) GitHub automatically shows this status in the sidebar of the PR, so it will be much more obvious to external users what stage the extension is at. A mockup to show how this might appear:

![Screen Shot 2021-04-17 at 11 18 43 AM](https://user-images.githubusercontent.com/1848368/115122847-ffb1e680-9f6e-11eb-8fad-49f1f61581d3.png)

donmccurdy:
Writing up a general summary of review process for various extension types, since I've wanted to reference this before. Is this reasonable? And would folks be OK with my creating an "extension status board" as described above?

***

- **Vendor** (`???_`): Vendors may merge vendor-prefixed extensions into this repository without formal requirements, but we would ask that the vendor confirm it is (1) ready to be merged and (2) in use or previously in use.
- **Multi-vendor** (`EXT_`): Where _multiple_ vendors are agreed and implementing an extension, we can merge it with an EXT_* prefix. Again, there are no formal approval requirements for such extensions, but reviewers will attempt to provide feedback for ecosystem coherence.
- **Khronos** (`KHR_`): Khronos extensions require formal interest, review, and approval by the 3D Formats working group, and have strict requirements when referencing existing IP. In general, `KHR_` extensions require significant buy-in from many implementors. Most pull requests for KHR_ extensions do not meet this threshold, and should be considered drafts until formally merged and ratified.

[...]

PR #1991, opened 2021-06-10
EXT_animation_quantization extension

PredatorCZ:
Purpose of this extension is to reduce memory and transmission impacts when storing animation data.
I wanted to make it as simple and intuitive as possible by using already present solutions within standard.
Something similar to mesh quantization and texture transform.
It also uses a few unmarked citations, if you're okay with that.
I have a few concerns, however they're not that important now.

### DIFF
--- a/extensions/2.0/Vendor/EXT_animation_quantization/README.md
+++ b/extensions/2.0/Vendor/EXT_animation_quantization/README.md
@@ -1,0 +1,170 @@
+# EXT_animation_quantization
+
+## Contributors
+
+* Lukas Cone
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+In a current standard implementation, animation tracks must be stored as `FLOAT` component type with exception of rotation. Translation and scale must be stored as `VEC3` and rotation as `VEC4`.
+Such implementation can be heavy on memory and transmition sizes.
+
+This extension expands the set of allowed component types for animation sampler storage to provide a memory/precision tradeoff - depending on the application needs, 16-bit or 8-bit storage can be sufficient. It also adds ability to use sampler per axis for translations as well as uniform scale sampling.
+
+In order to keep simplicity, this extension doesn't introduce new component types, but new schema.
+
+Translation tracks can use sampler per axis, this allows more freedom of reducing frames.
+
+Scale tracks as well can use sampler per axis, but also a single `SCALAR` accesor can be used for uniform scaling.
+
+Rotation tracks can be now processed in multiple ways. Quaternion doesn't need to store W component, such component can be recomputed from X, Y, Z values. If rotation is applied only for single axis, there is no need to store other axes, since they should be 0 values. In this case `SCALAR` or `VEC2` can be used istead of `VEC4`.
+
+This extension also reintroduces schema from `KHR_texture_transform` to increase quantized precision.
+
+Time values can now be quantized as well.
+
+This extension is NOT optional, files that use it must specify it in `extensionsRequired` array.
+
+## glTF Schema Updates
+
+```json
+ "channels": [
+    {
+        "sampler": 0,
+        "target": {
+            "node": 1,
+            "path": "translation.x"
+        },
+        "extensions": {
+            "EXT_animation_quantization": {
+                "offset": [-0.5],
+                "scale": [1],
+                "timeScale": 2.3333,
+            }
+        }
+    },
+]
+```
+
+### Properties
+
+| |Type|Description|Required|
+|-|-|-|-|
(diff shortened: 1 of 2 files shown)


PredatorCZ:
Can anyone please look into this? Any feedback would be helpful so we can move this further :)
Thanks.

donmccurdy:
I'm curious if you've considered the existing [`KHR_meshopt_compression`](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/EXT_meshopt_compression) extension for your use cases? It would provide smaller transmission size, but presumably reverts to float32 in application memory (/cc @zeux to confirm?). 

For the engines I work with (three.js, A-Frame) I don't think we can justify changing the animation API to support quantization for the memory benefits alone — we'd be de-quantizing at loading time, and so there's no advantage over MeshOpt. That's not an objection to this proposal, just an observation for these two engines as to why we likely wouldn't implement.

See https://github.com/KhronosGroup/glTF/issues/1970#issuecomment-885142415 for a short description of the extension process and requirements. If you need this extension for your own application, you are welcome to submit it under a vendor prefix. To merge an EXT_ extension will require multiple vendors expressing interest.

zeux:
@donmccurdy Yeah that's correct. The exponential filter in particular was built for cases like animation translation (and of course quaternion filter is for rotations), although it's useful for any other float streams - it allows both to reduce the floating point precision in a way that can be compressed well, and to also align exponents between consecutive floats which results in better compression still.

This coupled with the fact that translation/scale is relatively rare and I'd expect optimizing compressors to "fold" the constant tracks into a single value, meant that in my experiments the need for integer quantization for animations wasn't very critical for transmission size.

It would be interesting to compare transmission sizes on some real use cases between this proposal +- meshopt and baseline.

PredatorCZ:
Thanks for feedback. I looked carefully when I was making this proposal, I know about meshopt_compression and I know there is proposal for use of draco on animations as well. What I wanted, was exactly the same thing that mesh_quantization does.
Keeping everything simple as in mesh_quantization by using already defined data types and formats.
Unfortunately I haven't seen much love for meshopt, no support for Blender plugin or gltf_validator among a few that I use.

It might not be beneficial to every engine (those who dequantize at load time), for other low level languages like C/C++ and even Python's numpy might be. For example C/C++ doesn't need to dequantize at load time, but at sample time, etc...

I could post some real use examples, however my guess would be between baseline and meshopt.

About the vendor prefix, I don't feel comfortable registering my own vendor (a small github user among large corporates? cmon), nor to use other vendor's prefix for no reason. So choosing EXT was best choice where I hoped others would like to contribute as well, that unfortunately did not happened. From what I read in https://github.com/KhronosGroup/glTF/issues/1970#issuecomment-885142415 I feel like chances of proposing something from individuals seems very slim next to non-existant and I am not so sure if I would propose this extension if this comment was there much earlier.

zeux:
FWIW my personal opinion (although I don't think it's a universally held view) is that supporting various compression/quantization extensions in Blender export is an anti-pattern, and really Blender exporter should export a faithful representation of the original scene with high fidelity, and separate tools like gltfpack/gltf-transform should be in charge of reducing the memory/transmission footprint. But that's a separate conversation :)

(gltf-transform doesn't currently support meshopt compression but I'm working on exposing a cleaner JS encoder interface which should hopefully make it much easier to integrate there)

PredatorCZ:
I 100% agree with you, I personally don't use any form of compressions for exporters, I can optimize that myself.
I was talking about importer, should've said that :)

donmccurdy:
> ...for other low level languages like C/C++ and even Python's numpy might be. For example C/C++ doesn't need to dequantize at load time, but at sample time...

Similarly, JavaScript can make use of Uint8Array — the challenge is also that an existing engine's animation system is rarely designed around any particular input file format and its quantization strategy. If the existing keyframe interpolation cannot be dequantized at sample time, the engine will have to do so at loading time instead. Useful for tightly controlled workflows, but unfortunately less likely to get wide adoption.

> I don't feel comfortable registering my own vendor (a small github user among large corporates? ...

Personally, I think that's completely reasonable. For example, if someone were doing hobby game development and wanted to use glTF with some custom extensions for their game, that's great! Whether the vendor prefix and the extensions would be listed in the official registry I'm not sure about, I guess no one has asked for this yet. 🙂

> gltf-transform doesn't currently support meshopt compression but I'm working on exposing a cleaner JS encoder interface which should hopefully make it much easier to integrate there

I'd be glad to support that! Would it be worthwhile for gltf-transform to support meshopt _decompression_ sooner, or do you think the decoder API is likely to change with that work?

zeux:
> I'd be glad to support that! Would it be worthwhile for gltf-transform to support meshopt decompression sooner, or do you think the decoder API is likely to change with that work?

No, decoder API is stable and shouldn't change.

PR #1971, opened 2021-04-19
Khr materials emissive strength

bhouston:
First commit.

PR #1994, opened 2021-06-11
KHR_materials_emissive_strength

### LINKED ISSUE #1083: HDR values for emissiveFactor?

emackey:
View the [README](https://github.com/KhronosGroup/glTF/blob/5768b3ce0ef32bc39cdf1bef10b948586635ead3/extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md).

This supersedes #1971.

Fixes #1083.


extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
```diff
@@ -0,0 +1,166 @@
...
+## Overview
+
+The core glTF 2.0 material model includes an `emissiveFactor` and an `emissiveTexture` to control the color and
+intensity of the light being emitted by the material, clamped to the range [0.0, 1.0]. However, in PBR environments
+with high-dynamic range reflections and lighting, stronger emission effects may be desirable.
+
+In this extension, a new `emissiveStrength` scalar factor is supplied, that governs the upper limit of emissive
+strength per material. This strength can be colored and tempered using the core `emissiveFactor`
+and `emissiveTexture` controls, permitting the strength to vary across the surface of the material.
+Supplying values above 1.0 for `emissiveStrength` can have an influence on reflections, tonemapping, blooming, and more.
```
lexaknyazev, line 48: Should this statement refer to `100.0` as the threshold value?

extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
```diff
@@ -0,0 +1,166 @@
...
+The `emissiveStrength` has been set to 500.0, making the texture five times brighter than it otherwise
+would have been.
+
+### Parameters
+
+The following parameters are contributed by the `KHR_materials_emissive_strength` extension:
+
+| Name                   | Type       | Description                                    | Required             |
+|------------------------|------------|------------------------------------------------|----------------------|
+| **emissiveStrength**   | `number`   | The maximum strength of the emissive texture.  | No, default: `100.0` |
```
lexaknyazev, line 99: Values less than `100.0` not only lower the maximum strength of the emitted light but also increase effective precision of the supplied texture since the same 8 bits may represent finer details than by default.

This effect could be exploited further by adding another filed like `emissiveOffset` thus allowing authoring tools to find the most optimal representation for a given source asset.

extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
emackey: Great catch, thanks!

extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
emackey: True, but I don't think there's much demand to increase the precision of 8-bit images with this kind of offset.  If high precision is needed, it's more likely folks would want more bits.  But that topic touches many more textures beyond emission, so I think we should keep this extension focused on solving the more pressing emissive clamp problem.

lexaknyazev: Scaling strength beyond 100% inevitably stretches quantized 8-bit values to a wider range thus already reducing the original precision. Having a simple offset is usually enough to preserve it for textures that do not need real HDR storage.

lexaknyazev: OTOH, a non-zero offset would make the texture always emissive which may be quite restrictive for this approach to be useful.

Anyway, we should add a note to the spec that very high strength values may lead to visible banding.

lexaknyazev: One more nit here: a value below `100.0` could always be baked into `emissiveFactor`. Since this makes  an asset more portable, it may be reasonable to raise the minimum allowed strength to `100.0`.

lyuma:
Why is this extension incompatible with `KHR_materials_pbrSpecularGlossiness`?
I don't see any reason why it can't be used equivalently in a spec/gloss workflow. I would advocate for removing that exclusion.

donmccurdy:
@lyuma — on its own, no, there's no reason this extension is necessarily incompatible with `KHR_materials_pbrSpecularGlossiness`. But a number of other `KHR_materials_*` extensions are coming to glTF or available already (see https://github.com/KhronosGroup/glTF/blob/master/extensions), with some general implications:

1. Asking engines to support complex new features (clearcoat, transmission, volume, sheen, ...) in both PBR workflows is increasingly untenable. 
1. The `KHR_materials_pbrSpecularGlossiness` extension is less important: with `KHR_materials_ior` and `KHR_materials_specular`, the metal/rough workflow can replicate all of the features of the spec/gloss workflow, and an exporter can losslessly convert metal/rough -> spec/gloss.

For those reasons we are focusing future development on the metal/rough workflow, and encouraging implementations to do the same. IMHO this also aligns well with external directions like Autodesk Standard Surface, and the metal/rough workflow in glTF has been much more widely adopted.

*** 

I'm not necessarily opposed to relaxing the restriction for this particular extension, because it's somewhat of a "bug fix" to the existing emission value, but the overall trend away from `KHR_materials_pbrSpecularGlossiness` is something to be aware of.

emackey:
@lyuma Thanks, I pulled the exclusion out.  We do want to encourage everyone using `KHR_materials_pbrSpecularGlossiness` to switch to `KHR_materials_specular` at their earliest convenience, but there's nothing about this emission multiplier that interacts with either of those extensions.  Core material emission was already expected to be compatible with both.

extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
```diff
@@ -0,0 +1,173 @@
...
+### Physical Units
+
+For implementations where a physical light unit is needed, the units for the multiplicative product of
+`emissiveStrength`, `emissiveFactor`, and `emissiveTexture` are candela per square
+meter (**cd / m<sup>2</sup>**), sometimes called *nits*.
+
+**Implementation Note**:
+Because the value is specified per square meter, it indicates the brightness of any given point along
+the surface.  However, the exact conversion from physical light units to the brightness of rendered
+pixels requires knowledge of the camera's exposure settings, which are left as an implementation
```
rsahlin, line 62: I am not in favor of talking about camera exposure setting since there is no such thing in glTF.
I would like a mention that going from internal high range float values to those of the display is up to the implementation. 
This would usually involve some sort of tonemapping where the higher internal (float) range is converted to that of the output device.
This is needed regardless of camera exposure settings.

extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
```diff
@@ -0,0 +1,173 @@
...
+strength per material.
+
+**Implementation Note**: This strength can be colored and tempered using the core material's `emissiveFactor`
+and `emissiveTexture` controls, permitting the strength to vary across the surface of the material.
+Supplying values above 1.0 for `emissiveStrength` can have an influence on
+reflections, tonemapping, blooming, and more.
+
+### Physical Units
+
+For implementations where a physical light unit is needed, the units for the multiplicative product of
```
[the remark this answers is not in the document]
emackey: Emissive units are defined in [section 3.9.3](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#additional-textures) of the core spec, under the emissive bullet.  They were added following long-running discussions in the Khronos PBR TSG which came to a conclusion at the 2021-07-19 meeting, documented in the minutes for that meeting.

Khronos later published [a blog post](https://www.khronos.org/blog/khronos-prepares-gltf-2.0-to-be-submitted-as-an-international-standard) that listed this correction as a headline feature under the "glTF Technical Clarifications and Bug Fixes" section.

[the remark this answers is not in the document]
abwood: Just to clarify, this extension defines emissiveStrength, which is a unitless multiplier. With the core spec defining physical units for emission in section 3.9.3, we can remove this section on Physical Units here. Instead, we should state that this property is unitless and retains the original units of emission. With this more relaxed definition, this extension would remain compatible with any future changes to physical units (e.g., via a new extension).

donmccurdy:
Implemented in glTF-Transform:
- https://github.com/donmccurdy/glTF-Transform/pull/422

donmccurdy:
Not a particularly realistic sample, but here's a voxel desk with emissiveStrength=5.

[desk.glb.zip](https://github.com/KhronosGroup/glTF/files/7518199/desk.glb.zip)

https://user-images.githubusercontent.com/1848368/141247915-cf37c96c-b1e8-4446-b78f-a852df962583.mov

atteneder:
I haven't wrapped my head around this yet, so maybe someone can explain this to me:

Won't a unit-less multiplier lead to very different results in different render pipelines?

Simpler render pipelines will just output the material's emission value into the frame buffer and it is displayed as is (after optional tone-mapping). In those cases a multiplier like 2x or 5x makes sense.

Other render pipelines work with physical units and cameras with physically plausible exposure models. glTF's emission unit is nits and many real-world examples go into the hundreds of nits (e.g. modern mobile phones pride themselves with displays of 1000 nits brightness).

How are users supposed to create models that look correct in both cases?

Let's take the LED display of our beloved BoomBox as example. I guess a realistic, physical value for it would be 200 nits, but that would totally blow out in a simple render pipeline. Otherwise if the `emissiveIntensity` is low (in the single digits), it probably won't show in a render pipeline with correct exposure in a daylight scenario.

As a reference, Unity's HDRP offers artists two modes:

- exposureWeight 0 (zero) means the camera's exposure does not affect the final brightness (much like the implementation note in the glTF spec suggests)
- exposureWeight 1 means the camera's exposure scales the overall brightness physically correct.

Artists can even blend between those modes.

Curious of your thoughts

emackey:
The physical units are in the core glTF specification, under [additional textures, emissive section](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#additional-textures).

> For implementations where a physical light unit is needed, the units for the multiplicative product of the emissive texture and factor are candela per square meter (<strong>cd / m<sup>2</sup></strong>), sometimes called <em>nits</em>.

> **Implementation Note**  Because the value is specified per square meter, it indicates the  brightness of any given point along the surface.  However, the exact  conversion from physical light units to the brightness of rendered  pixels requires knowledge of the camera’s exposure settings, which are  left as an implementation detail unless otherwise defined by a glTF  extension.   Many rendering engines simplify this calculation by assuming that an emissive factor of 1.0 results in a fully exposed pixel.

This extension merely supplies an additional multiplier.  The multiplier cannot change the units.  For example if the multiplier itself were specified in nits, the final emissive value from glTF would end up being _square nits_ which is a nonsensical unit.

> How are users supposed to create models that look correct in both cases?

That's a question for glTF in general, not affected by this extension.  I believe that for a single model in isolation, it is essentially impossible to get the correct physical units for an emissive material without prior knowledge of external settings such as camera exposure controls, scene tone-mapping settings, and the like.  So in many cases, we expect models to use the "simple" path where an exposure value of 1.0 corresponds to a fully-exposed pixel, for portability across environments using different exposure settings.  But for models needed in situations where exposure is controlled and physical units are known, those physical units are specified in glTF, and are unchanged by this extension.

atteneder:
An idea how to make models look consistent across viewers: Recommend a fallback behavior to viewers that don't calculate physically correct camera exposure. That would be a scene specific multiplier.

To stay with the example, let's give the BoomBox display an `emissiveStrength` of 200 (nits). Viewers with physical camera exposures will calculate correct results, since this is a physical, measurable value.

A simpler viewer may define a certain maximum brightness value that leads to full (white) brightness. The value would depend on the scene. For a bright daylight scene this might be 1500 nits and just around 100 nits for moonlit night. The final emissive value would then be `emissiveFactor * emissiveTexture * emissiveStrength / sceneReferenceValue`.
This would apply only if the emissive strength extension is present.

Is that how this is intended to be used?

emackey:
> Is that how this is intended to be used?

No.

The original glTF 2.0 schema accidentally clamped `emissiveFactor` to a maximum value of `[1.0, 1.0, 1.0]`.

This extension undoes the accidental clamp, by multiplying a new value into `emissiveFactor`.

atteneder:
> > Is that how this is intended to be used?
> 
> No.
> 
> The original glTF 2.0 schema accidentally clamped `emissiveFactor` to a maximum value of `[1.0, 1.0, 1.0]`.
> 
> This extension undoes the accidental clamp, by multiplying a new value into `emissiveFactor`.

Thank you very much for explaining and clarifying!

I understand and agree that it's outside of a glTF's scope to be aware of the external exposure and tone-mapping settings of the viewer.

Let me ask another way: Wouldn't it lead to much more consistent results to to recommend it this way?

I mean according to the spec, the product `emissiveFactor * emissiveTexture` has the unit nits. Shouldn't we recommend to artists to enter physically plausible values for `emissiveStrength` instead of arbitrary, viewer specific multipliers?

I'm just afraid that if this is under-specified and left to viewers/artists, no emissive glTF material will look consistent across viewers.

Thanks for considering

emackey:
Units don't work that way.  For example:

x meters * y meters = xy square meters (length * length = area)

x meters * y scalar = xy meters (length * scalar = length)

So following that logic,

x nits (factor) * y nits (multiplier) = xy square nits (nonsensical units) final output

x nits (factor) * y scalar (unitless) = xy nits, final output.

The multiplier isn't arbitrary.  It _must_ be unitless, to preserve the physical units on the emissive system in glTF.

emackey:
And yes, that does mean that if your `emissiveFactor` is 1.0, and `emissiveStrength` is 200, the result is:

1 nit * 200 (scalar) = 200 nits

atteneder:
> And yes, that does mean that if your `emissiveFactor` is 1.0, and `emissiveStrength` is 200, the result is:
> 
> 1 nit * 200 (scalar) = 200 nits

Thanks, agreed! I don't want to suggest giving `emissiveStrength` a unit (sorry if I explained that poorly).

What I'm suggesting is the final value of `emissiveFactor.rgb * sRGB_to_Linear(emissiveTexture.rgb) * emissiveStrength` (in nits) *should* always be a physically plausible value. For that to be true on many real world examples (lamps or screens), the `emissiveStrength` value has to be in a range well above 100.

@donmccurdy 's desk.glb example above shows a voxely screen with an `emissiveStrength` of 5. Presuming there's no emissive texture and the emissiveFactor is pure white, this resulting emission is 5 nits, right? 5 nits does not look that bright in real world, does it?

A justification could be that the 2.0 spec does not require viewers to work with physical units for emission at all and so in Don's example `emissiveStrength` is just an arbitrary multiplier value.

I'd argue that if consistency across viewers is a priority, this extension should require nits as a unit (for the `emissiveFactor * emissiveTexture` term).

Hope that made it clearer and I'm curious what your thoughts are.

P.S.:
I'm currently prototyping this in Unity and indeed struggle to get it consistent across platforms/render pipelines.

emackey:
> 5 nits does not look that bright in real world, does it?

It depends on camera exposure.  Almost any number of nits from 1 to a million can be underexposed or overexposed simply by adjusting real-world camera exposure settings.  In the case of this one model of a glowing monitor, we can read the specs on a variety of real-world monitor screens, and find that typical values range from 80 to 500 nits depending on the screen, and so 5 nits seems far too dim for this model.  But what then?  For a while we had 1000 nits as the default, telling people to expect 1000 to exactly equal fully exposed white, and then at a different stage we had 100 nits as the default, saying 100 was white.  In both cases, there are far more real-world situations where those numbers are wrong vs. right.  We can't tell people what full exposure is here in a material definition, as it depends on the environment, which could be anything from a starry moonless night, where 5 nits is a bright glow, to a harsh summer day outdoors, where hundreds of nits are still too dim to read.

Compounding the problem, most renderers don't consider physical units.  Instead, they treat 1.0 as fully exposed, and anything above 1.0 as overexposed or in need of clamping, bloom, tone-mapping, and/or other such treatments.  Most of the glTF models we've encountered in the wild so far use this metric instead of the physical one, where "5" means 5x overexposed, which is not necessarily 5 nits (but could be on a dark night).  I don't know of an easy fix for this problem, but we've found that simply picking an arbitrary number of nits based on the brightness of a typical computer monitor is _not_ the correct solution.

emackey:
One solution could be a completely separate extension, where the glTF `scene` and/or `camera` object is given exposure controls, indicating how many nits are needed to exactly reach a full white exposure.

rsahlin:
> What I'm suggesting is the final value of `emissiveFactor.rgb * sRGB_to_Linear(emissiveTexture.rgb) * emissiveStrength` (in nits) _should_ always be a physically plausible value. For that to be true on many real world examples (lamps or screens), the `emissiveStrength` value has to be in a range well above 100.

Great suggestion @atteneder 
This is precisely what I am looking and something we really need in our QA process for instance.
One part to get there is the by adding an extension for mapping of values to display:

PR displaymapping https://github.com/KhronosGroup/glTF/pull/2083

This will give a plausible range of min/max values and a theoretical range of 0 - 10000 cd/m2 (when displays catch up)
To this extension I am considering adding `automatic aperture`, or perhaps called `auto-average-luminance` , which will control overall scene brightness so that a bright outdoor scene and a moonlit scene can  be rendered with expected results.
I believe this would give the `sceneReferenceValue` you describe earlier.

Do you think this would work for your usecase Andreas?

atteneder:
Thanks for your thoughts, @emackey  and  @rsahlin 

I'm not able to draw a conclusion yet...I need to do more research.

I agree, specifying an arbitrary value (in nits) that's supposed to give white (1.0) emission makes no sense as it depends on the scene/camera exposure.

About bringing exposure controls into glTF scene or camera: That's a good idea, but it only works when the scene/camera settings are not separate to the actual object/model. For lots of use-cases this is not true or desired. As an artist (or manufacturer) I want my glTF to look consistent across viewers.

@rsahlin I agree we should keep display-mapping in mind for this one, thanks for pointing it out. Ultimately I think those extensions should work independently, but I probably need to educate myself more about display-mapping first.

TerenceRussell:
Holy cats, still waiting for this bug fix! What's with this discussion of "real lighting units"?  Its a scalar workaround for a clamping issue with the original emissiveFactor color scalar.

extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
```diff
@@ -0,0 +1,167 @@
...
+The `emissiveStrength` has been set to 5.0, making the texture five times brighter than it otherwise
+would have been.
+
+### Parameters
+
+The following parameters are contributed by the `KHR_materials_emissive_strength` extension:
+
+| Name                   | Type       | Description                                    | Required           |
+|------------------------|------------|------------------------------------------------|--------------------|
+| **emissiveStrength**   | `number`   | The maximum strength of the emissive texture.  | No, default: `1.0` |
```
lexaknyazev, line 101: This description should be synced with the schema.

extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
```diff
@@ -0,0 +1,167 @@
...
+* Bruce Cherniak, Intel
+* Gary Hsu, Microsoft [@bghgary](https://twitter.com/bghgary)
+* Nicholas Barlow, Microsoft
+* Adam Morris, Target [@weegeekps](https://github.com/weegeekps)
+* Sandra Voelker, Target
+* Ben Houston, Threekit, [@bhouston](https://twitter.com/BenHouston3D)
+* Norbert Nopper, UX3D, [@UX3DGpuSoftware](https://twitter.com/UX3DGpuSoftware)
+* Eric Chadwick, Wayfair, [echadwick-wayfair](https://github.com/echadwick-wayfair)
+
+Copyright (C) 2021 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
```
lexaknyazev, line 26: 2022?

extensions/2.0/Khronos/KHR_materials_emissive_strength/schema/glTF.KHR_materials_emissive_strength.schema.json
```diff
@@ -0,0 +1,17 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_materials_emissive_strength glTF extension",
+    "type": "object",
+    "description": "glTF extension that adjusts the strength of emissive material properties.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "emissiveStrength": {
+            "type": "number",
+            "description": "The strength adjustment to be multiplied with the material's emissiveFactor and emissiveTexture.",
```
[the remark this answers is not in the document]
emackey: Fixed.

extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
emackey: Fixed.

extensions/2.0/Khronos/KHR_materials_emissive_strength/README.md
emackey: Fixed.

abwood (approved):
This extension has been ratified by Khronos on May 20, 2022.

PR #1997, opened 2021-06-28
Update normative language

lexaknyazev:
### General updates
* Use BCP 14 normative language.
* Properly list all external references.
* Expand the Samplers section.
* Expand material texture descriptions.
* Reduce emphasis on OpenGL.
* Minor language edits.

### Specific updates
* Restrict certain PNG and JPEG variations.
* Full rewrite of JSON encoding section, relaxing some of previously-unenforced restrictions.
* Explicitly allow IRIs.
* Explicitly disallow `bufferView.byteStride` for buffer views that do not contain vertex attributes.
* Elaborate on image media types handling.
* Restore accidentally removed vertex color usage.
* Discourage use of zero-length GLB BIN chunks.

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -100,9 +100,9 @@ glTF is an API-neutral runtime asset delivery format.  glTF bridges the gap betw
 [[introduction-conventions]]
 == Document Conventions
 
-The glTF Specification is intended for use by both implementors of the asset exporters or converters (e.g. digital content creation tools) and application developers seeking to import or load glTF assets, forming a contract between these parties.
+The glTF Specification is intended for use by both implementors of the asset exporters or converters (e.g. digital content creation tools) and application developers seeking to import or load glTF assets, forming a basis for interoperability between these parties.
 
-Specification text may address either party; typically the intended audience can be inferred from context, though some sections are defined to address only one of these parties.
+Specification text can address either party; typically the intended audience can be inferred from context, though some sections are defined to address only one of these parties.
 
 Any requirements, prohibitions, recommendations or options defined by <<introduction-normative-terminology, normative terminology>> are imposed only on the audience of that text.
 
@@ -321,15 +321,13 @@ Future extensions may include more compression methods including those for anima
 
 
 [[versioning]]
-== Versioning (Informative)
+== Versioning
 
-This section is marked as informative as there is no binding responsibility on implementations of the glTF — these guarantees are however a contract between the 3D Formats Working Group and developers using this Specification.
+Any updates made to the glTF Specification in a minor version **MUST** be backwards and forwards compatible. Backwards compatibility means that any client implementation that supports loading a glTF 2.x asset will also be able to load a glTF 2.0 asset. Forwards compatibility means that a client implementation that only supports glTF 2.0 is able to load glTF 2.x assets while gracefully ignoring any new features it does not understand.
 
-Any updates made to glTF in a minor version must be backwards and forwards compatible. Backwards compatibility means that any client implementation that supports loading a glTF 2.x asset will also be able to load a glTF 2.0 asset. Forwards compatibility means that a client implementation that only supports glTF 2.0 is able to load glTF 2.x assets while gracefully ignoring any new features it does not understand.
+A minor version update **MAY** introduce new features but **MUST NOT** change any previously existing behavior. Existing functionality **MAY** be deprecated in a minor version update, but it **MUST NOT** be removed.
 
-A minor version update may introduce new features but must not change any previously existing behavior. Existing functionality may be deprecated in a minor version update, but it must not be removed.
-
-Major version updates may be incompatible with previous versions.
+Major version updates **MAY** be incompatible with previous versions.
 
 
 [[file-extensions-and-media-types]]
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -1889,6 +1889,16 @@ Normal vectors **MUST** be normalized before being used in lighting equations. W
 The texture binding for occlusion maps **MAY** optionally contain a scalar `strength` value that is used to reduce the occlusion effect. When present, it affects the occlusion value as `1.0 + strength * (occlusionTexture - 1.0)`.
 
 - *emissive* : The emissive texture and factor control the color and intensity of the light being emitted by the material. The texture **MUST** contain 8-bit values encoded with the <<srgb, sRGB opto-electronic transfer function>> so RGB values **MUST** be decoded to real linear values before they are used for any computations. To achieve correct filtering, the transfer function **SHOULD** be decoded before performing linear interpolation.
++
+For implementations where a physical light unit is needed, the units for the multiplicative product of the emissive texture and factor are candela per square meter (**cd / m^2^**), sometimes called *nits*.
++
+[NOTE]
+.Implementation Note
+====
+Because the value is specified per square meter, it indicates the brightness of any given point along the surface.  However, the exact conversion from physical light units to the brightness of rendered pixels requires knowledge of the camera's exposure settings, which are left as an implementation detail unless otherwise defined by a glTF extension.
+
+Many rendering engines simplify this calculation by assuming that an emissive factor of `1.0` results in a fully exposed pixel.
+====
 
 The following example shows a material that is defined using `pbrMetallicRoughness` parameters as well as additional textures:
 
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -91,7 +91,7 @@ for identification and belong to their respective owners.
 [[introduction-general]]
 == General
 
-This document, referred to as the “glTF Specification” or just the “Specification” hereafter, describes the glTF file format.
+This document, referred to as the "`glTF Specification`" or just the "`Specification`" hereafter, describes the glTF file format.
 
 glTF is an API-neutral runtime asset delivery format.  glTF bridges the gap between 3D content creation tools and modern graphics applications by providing an efficient, extensible, interoperable format for the transmission and loading of 3D content.
 
@@ -99,7 +99,7 @@ glTF is an API-neutral runtime asset delivery format.  glTF bridges the gap betw
 [[introduction-conventions]]
 == Document Conventions
 
-The glTF Specification is intended for use by both implementors of the asset exporters or converters (e.g. DCC tools) and application developers seeking to import or load glTF assets, forming a contract between these parties.
+The glTF Specification is intended for use by both implementors of the asset exporters or converters (e.g. digital content creation tools) and application developers seeking to import or load glTF assets, forming a contract between these parties.
 
 Specification text may address either party; typically the intended audience can be inferred from context, though some sections are defined to address only one of these parties.
 
@@ -129,7 +129,7 @@ Unless otherwise noted in the section heading, all sections and appendices in th
 [[introduction-technical-terminology]]
 === Technical Terminology
 
-The glTF Specification makes use of common engineering and graphics terms such as *Vertex*, *Normal*, *Texture*, etc to identify and describe certain GPU constructs and their attributes, states, and behaviors.
+The glTF Specification makes use of common engineering and graphics terms such as *Vertex*, *Normal*, *Texture*, etc. to identify and describe certain GPU constructs and their attributes, states, and behaviors.
 
 [[introduction-normative-references]]
 === Normative References
@@ -140,84 +140,119 @@ The following documents are referenced by normative sections of the specificatio
 
 ==== External Specifications
 
-[[bcp14]]
+* [[bcp14]]
 Bradner, S., _Key words for use in RFCs to Indicate Requirement Levels_, BCP 14, RFC 2119, March 1997. Leiba, B., _Ambiguity of Uppercase vs Lowercase in RFC 2119 Key Words_, BCP 14, RFC 8174, May 2017.
 <https://www.rfc-editor.org/info/bcp14>
 
-[[utf8]
+* [[utf8]]
 The Unicode Consortium, _The Unicode Standard_
 <https://www.unicode.org/versions/latest/>
 
-[[json]]
+* [[json]]
 Bray, T., Ed., _The JavaScript Object Notation (JSON) Data Interchange Format_, STD 90, RFC 8259, DOI 10.17487/RFC8259, December 2017, <https://www.rfc-editor.org/info/rfc8259>
 
-[[ieee-754]]
+* [[ieee-754]]
 ISO/IEC 60559:2020
 _Floating-point arithmetic_
 <https://www.iso.org/standard/80985.html>
 
-[[png]]
+* [[png]]
 ISO/IEC 15948:2004
 _Portable Network Graphics (PNG): Functional specification_
 <https://www.iso.org/standard/29581.html>
++
+[NOTE]
+.Note
+====
+A free version of this standard is available from W3C: https://www.w3.org/TR/PNG/
+====
 
-[[jpeg]]
+
+
+* [[jpeg]]
 ISO/IEC 10918-1:1994
 _Digital compression and coding of continuous-tone still images: Requirements and guidelines_
 <https://www.iso.org/standard/18902.html>
++
+[NOTE]
+.Note
+====
+An earlier edition of this standard called ITU Recommendation T.81 is available from W3C: https://www.w3.org/Graphics/JPEG/itu-t81.pdf
+====
 
-[[jfif]]
+* [[jfif]]
 ISO/IEC 10918-5:2013
 _Digital compression and coding of continuous-tone still images: JPEG File Interchange Format (JFIF)_
 <https://www.iso.org/standard/54989.html>
++
+[NOTE]
+.Note
+====
+An earlier edition of this standard is available from W3C: https://www.w3.org/Graphics/JPEG/jfif3.pdf
+====
 
-[[data-uri]]
+* [[exif]]
+CIPA DC-008-Translation-2019
+_Exchangeable image file format for digital still cameras_
+<https://www.cipa.jp/std/documents/download_e.html?DC-008-Translation-2019-E>
+
+* [[data-uri]]
 Masinter, L., _The "data" URL scheme_, RFC 2397, DOI 10.17487/RFC2397, August 1998, <https://www.rfc-editor.org/info/rfc2397>
 
-[[uri]]
+* [[uri]]
 Berners-Lee, T., Fielding, R., and L. Masinter, _Uniform Resource Identifier (URI): Generic Syntax_, STD 66, RFC 3986, DOI 10.17487/RFC3986, January 2005, <https://www.rfc-editor.org/info/rfc3986>
 
-[[iri]]
+* [[iri]]
 Duerst, M. and M. Suignard, _Internationalized Resource Identifiers (IRIs)_, RFC 3987, DOI 10.17487/RFC3987, January 2005, <https://www.rfc-editor.org/info/rfc3987>
 
-[[http]]
+* [[http]]
 Fielding, R., Ed., and J. Reschke, Ed., _Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing_, RFC 7230, DOI 10.17487/RFC7230, June 2014, <https://www.rfc-editor.org/info/rfc7230>
 
-[[srgb]]
-IEC 61966-2-1:1999.
+* [[srgb]]
+IEC 61966-2-1:1999
 _Default RGB colour space - sRGB_
 https://webstore.iec.ch/publication/6169
++
+[NOTE]
+.Note
+====
+The encoding characteristics of sRGB are freely available from ICC: https://www.color.org/chardata/rgb/srgb.xalter
+====
 
-[[mikktspace]]
+* [[bt709]]
+Recommendation ITU-R BT.709-6 _Parameter values for the HDTV standards for production and international programme exchange_
+https://www.itu.int/rec/R-REC-BT.709-6-201506-I
+
+* [[mikktspace]]
 _MikkTSpace_
 https://github.com/mmikk/MikkTSpace
 
 ==== Media Type Registrations
 
-[[gltf-json]]
+* [[gltf-json]]
 IANA.
 _model/gltf+json Media Type_.
 https://www.iana.org/assignments/media-types/model/gltf+json
 
-[[gltf-binary]]
+* [[gltf-binary]]
 IANA.
 _model/gltf-binary Media Type_.
 https://www.iana.org/assignments/media-types/model/gltf-binary
 
-[[gltf-buffer]]
+* [[gltf-buffer]]
 IANA.
 _application/gltf-buffer Media Type_.
 https://www.iana.org/assignments/media-types/application/gltf-buffer
 
-[[octet-stream]]
+* [[octet-stream]]
 IANA.
 _application/octet-stream Media Type_.
 https://www.iana.org/assignments/media-types/application/octet-stream
 
-[[image-jpeg]]
+* [[image-jpeg]]
 Freed, N. and N. Borenstein, _Multipurpose Internet Mail Extensions (MIME) Part Two: Media Types_, RFC 2046, DOI 10.17487/RFC2046, November 1996, <https://www.rfc-editor.org/info/rfc2046>
 
-[[image-png]]
+* [[image-png]]
 IANA.
 _image/png Media Type_.
 https://www.iana.org/assignments/media-types/image/png
@@ -244,9 +279,9 @@ A glTF asset is represented by:
 * Binary files (`.bin`) containing geometry, animation, and other buffer-based data.
 * Image files (`.jpg`, `.png`) containing texture images.
 
-Binary and image resources **MAY** also be embedded directly in JSON using <<data-uri,Data URI>> or stored side-by-side with JSON in GLB container.
+Binary and image resources **MAY** also be embedded directly in JSON using <<data-uri,Data URI>> or stored side-by-side with JSON in <<glb-file-format-specification,GLB>> container.
 
-Valid glTF asset **MUST** specify its version.
+A valid glTF asset **MUST** specify its version.
 
 image:figures/files.png[pdfwidth=4in,align=left]
 
@@ -270,6 +305,7 @@ The following are outside the scope of the initial design of glTF:
 * *glTF is not intended to be human-readable,* though by virtue of being represented in JSON, it is developer-friendly.
 
 While version 2.0 of glTF does not define compression for geometry and other rich data, there are extensions providing such options.
+
 * https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md[KHR_draco_mesh_compression] for geometry.
 * https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_texture_basisu/README.md[KHR_texture_basisu] for GPU-compatible block-compressed textures.
 
@@ -291,8 +327,8 @@ Major version updates **MAY** be incompatible with previous versions.
 [[file-extensions-and-media-types]]
 == File Extensions and Media Types
 
-* JSON glTF files **SHOULD** use `.gltf` extension and <<gltf-json,`model/gltf+json`>> Media Type.
-* glTF files stored in GLB container **SHOULD** use `.glb` extension and <<gltf-binary,`model/gltf-binary`>> Media Type.
+* <<json,JSON>> glTF files **SHOULD** use `.gltf` extension and <<gltf-json,`model/gltf+json`>> Media Type.
+* glTF files stored in <<glb-file-format-specification,GLB>> container **SHOULD** use `.glb` extension and <<gltf-binary,`model/gltf-binary`>> Media Type.
 * Files representing binary buffers **SHOULD** use either:
 ** `.bin` file extension with <<octet-stream,`application/octet-stream`>> Media Type;
 ** `.bin`, `.glbin`, or `.glbuf` file extensions with <<gltf-buffer,`application/gltf-buffer`>> Media Type.
@@ -309,7 +345,7 @@ Major version updates **MAY** be incompatible with previous versions.
 
 Although glTF Specification does not define any subset of the <<json,JSON>> format, implementations **SHOULD** be aware of its peculiar properties that could affect asset interoperability.
 
-1. glTF JSON data **SHOULD** be written with UTF-8 encoding without BOM. This requirement is not applied when a glTF implementation does not control string encoding. glTF implementations **SHOULD** adhere to the <<json,RFC 8259>>, Section 8.1. with regards to treating BOM presence.
+1. glTF JSON data **SHOULD** be written with UTF-8 encoding without BOM. This requirement is not applied when a glTF implementation does not control string encoding. glTF implementations **SHOULD** adhere to <<json,RFC 8259>>, Section 8.1. with regards to treating BOM presence.
 
 2. ASCII characters stored in glTF JSON **SHOULD** be written unescaped.
 +
@@ -426,7 +462,7 @@ The image above shows relations between top-level arrays in a glTF asset.  See t
 [[asset]]
 == Asset
 
-Each glTF asset **MUST** have an `asset` property. The `asset` object **MUST** contain `version` property which specifies the target glTF version of the asset. Additionally, an optional `minVersion` property **MAY** be used to specify the minimum glTF version support required to load the asset. The `minVersion` property allows asset creators to specify a minimum version that a client implementation **MUST** support in order to load the asset. This is very similar to the `extensionsRequired` concept described in <<specifying-extensions>>, where an asset **SHOULD NOT** be loaded if the client does not support the specified extension. Additional metadata **MAY** be stored in optional properties such as `generator` or `copyright`.  For example,
+Each glTF asset **MUST** have an `asset` property. The `asset` object **MUST** contain a `version` property which specifies the target glTF version of the asset. Additionally, an optional `minVersion` property **MAY** be used to specify the minimum glTF version support required to load the asset. The `minVersion` property allows asset creators to specify a minimum version that a client implementation **MUST** support in order to load the asset. This is very similar to the `extensionsRequired` concept described in <<specifying-extensions>>, where an asset **SHOULD NOT** be loaded if the client does not support the specified extension. Additional metadata **MAY** be stored in optional properties such as `generator` or `copyright`.  For example,
 
 [source,json]
 ----
@@ -498,7 +534,7 @@ RGB color values **MUST** use <<srgb,sRGB>> color primaries.
 [NOTE]
 .Implementation Note
 ====
-Color primaries define the interpretation of each color channel of the color model, particularly with respect to the RGB color model. In the context of a typical display, color primaries describe the color of the red, green and blue phosphors or filters. The same primaries are also defined in Recommendation ITU-R BT.709. Since the overwhelming majority of currently used consumer displays are using the same primaries as default, client implementations usually do not need to convert color values. Future specification versions or extensions may allow other color primaries (such as P3).
+Color primaries define the interpretation of each color channel of the color model, particularly with respect to the RGB color model. In the context of a typical display, color primaries describe the color of the red, green and blue phosphors or filters. The same primaries are also defined in <<bt709,Recommendation ITU-R BT.709>>. Since the overwhelming majority of currently used consumer displays are using the same primaries as default, client implementations usually do not need to convert color values. Future specification versions or extensions may allow other color primaries (such as P3).
 ====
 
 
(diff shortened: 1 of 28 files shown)


javagl:
I started a review of the changes, but it's pretty large. I locally took some notes (a bit of nitpicking, and some things that I wondered about, but where, at some point, I have to assume that they have been sorted out in certain issues, and the changes here are just the _result_ of a larger discussion). Two things for the process:

- Is there any timeline for the review?
- Do you prefer inlined comments in the review, or a list with entries like
  > - https://github.com/KhronosGroup/glTF/pull/1997/files#diff-ca628c11679a8c4c0c4031440c7c1cda27c8896ebc87c3bd1e8640efc8942d99R147 Broken formatting of "[[utf8]" (missing "]", I guess)

  (The first may be more convenient (and avoid duplication), but the second one may be more handy and less scattered, by having all comments by reviewers in one place. I'm locally keeping the notes in such a list, but would transfer them into inline comments eventually, if you prefer that)

lexaknyazev:
Since inline comments naturally support threading, I'd prefer to use them since the diff is indeed large (although most of it is trivial). Feel free to ask about any confusing changes.

specification/2.0/Specification.adoc
```diff
@@ -91,45 +91,168 @@ for identification and belong to their respective owners.
 [[introduction-general]]
 == General
 
-The GL Transmission Format (glTF) is an API-neutral runtime asset delivery format.  glTF bridges the gap between 3D content creation tools and modern graphics applications by providing an efficient, extensible, interoperable format for the transmission and loading of 3D content.
+This document, referred to as the “glTF Specification” or just the “Specification” hereafter, describes the glTF file format.
```
oddhack, line 94: Use "`word`" for proper curly quotes.

specification/2.0/Specification.adoc
```diff
@@ -91,45 +91,168 @@ for identification and belong to their respective owners.
...
-* Image files (`.jpg`, `.png`) for textures
+* A JSON-formatted file (`.gltf`) containing a full scene description: node hierarchy, materials, cameras, as well as descriptor information for meshes, animations, and other constructs.
+* Binary files (`.bin`) containing geometry, animation, and other buffer-based data.
+* Image files (`.jpg`, `.png`) containing texture images.
 
-Assets defined in other formats, such as images, may be stored in external files referenced via URI, stored side-by-side in GLB container, or embedded directly into the JSON using https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs[data URIs].
+Binary and image resources **MAY** also be embedded directly in JSON using <<data-uri,Data URI>> or stored side-by-side with JSON in GLB container.
 
-Valid glTF asset must specify its version.
+Valid glTF asset **MUST** specify its version.
```
oddhack, line 249: "*A* valid..."

specification/2.0/Specification.adoc
```diff
@@ -146,91 +269,146 @@ The following are outside the scope of the initial design of glTF:
...
+** Client implementations **MAY** support JPEG images that use <<exif,Exchangeable image file format (Exif)>>.
 
 [[json-encoding]]
 == JSON Encoding
 
-To simplify client-side implementation, glTF has additional restrictions on JSON format and encoding.
+Although glTF Specification does not define any subset of the <<json,JSON>> format, implementations **SHOULD** be aware of its peculiar properties that could affect asset interoperability.
 
-1. JSON must use UTF-8 encoding without BOM.
+1. glTF JSON data **SHOULD** be written with UTF-8 encoding without BOM. This requirement is not applied when a glTF implementation does not control string encoding. glTF implementations **SHOULD** adhere to the <<json,RFC 8259>>, Section 8.1. with regards to treating BOM presence.
```
oddhack, line 312: "should adhere to RFC 8259" (remove "the").

specification/2.0/Specification.adoc
```diff
@@ -242,13 +420,13 @@ While the spec does not explicitly disallow non-normalized URIs, their use may b
...
+Each glTF asset **MUST** have an `asset` property. The `asset` object **MUST** contain `version` property which specifies the target glTF version of the asset. Additionally, an optional `minVersion` property **MAY** be used to specify the minimum glTF version support required to load the asset. The `minVersion` property allows asset creators to specify a minimum version that a client implementation **MUST** support in order to load the asset. This is very similar to the `extensionsRequired` concept described in <<specifying-extensions>>, where an asset **SHOULD NOT** be loaded if the client does not support the specified extension. Additional metadata **MAY** be stored in optional properties such as `generator` or `copyright`.  For example,
```
oddhack, line 429: "must contain *a* version property..."

oddhack:
I started running through this and soon noticed a lot of omission of definite & indefinite articles (such as "the keyword", "a buffer", "an object") where they're grammatically appropriate. On the grand scale of things this is more jarring than a spec bug, but it is still worth addressing.

lexaknyazev:
> I started running through this and soon noticed a lot of omission of definite & indefinite articles (such as "the keyword", "a buffer", "an object") where they're grammatically appropriate.

You're right. This is a long-standing issue since the very first glTF drafts.

neiltrevett:
> When is the the deadline for this update?

We have two deadlines:
1. Any IP-impacting changes must be implemented and agreed by the meeting next week (July 7th) so we can vote to submit the specification for ratification. No IP-impacting changes would then be permitted without restarting the ratification review
2. Any non-IP-impacting changes (formatting, clarifications informative additions, adding missing articles that don't change meaning) can be made at any time during the 6-week ratification review and beyond, up to when we formally release the specification, or submit it to JTC 1, which ever is the sooner - latest early September.

Neil

specification/2.0/Specification.adoc
```diff
@@ -91,45 +91,168 @@ for identification and belong to their respective owners.
...
+
+The following documents are referenced by normative sections of the specification:
+
+==== External Specifications
+
+[[bcp14]]
+Bradner, S., _Key words for use in RFCs to Indicate Requirement Levels_, BCP 14, RFC 2119, March 1997. Leiba, B., _Ambiguity of Uppercase vs Lowercase in RFC 2119 Key Words_, BCP 14, RFC 8174, May 2017.
+<https://www.rfc-editor.org/info/bcp14>
+
+[[utf8]
```
donmccurdy, line 147: minor: missing closing bracket here

specification/2.0/Specification.adoc
```diff
@@ -91,45 +91,168 @@ for identification and belong to their respective owners.
...
+The glTF Specification makes use of common engineering and graphics terms such as *Vertex*, *Normal*, *Texture*, etc to identify and describe certain GPU constructs and their attributes, states, and behaviors.
+
+[[introduction-normative-references]]
+=== Normative References
+
+References to external documents are considered normative references if the Specification uses any of the normative terms defined in <<introduction-normative-terminology>> to refer to them or their requirements, either as a whole or in part.
+
+The following documents are referenced by normative sections of the specification:
+
+==== External Specifications
```
donmccurdy, line 141: In rendered form it's difficult to find something quickly among these references, because the keywords are somewhere in the middle of the line. Would it be appropriate to put the title first, rather than the source, or prefix with a keyword like "_JSON_"?

specification/2.0/Specification.adoc
```diff
@@ -146,91 +269,146 @@ The following are outside the scope of the initial design of glTF:
...
 * *glTF is not intended to be human-readable,* though by virtue of being represented in JSON, it is developer-friendly.
 
-While version 2.0 of glTF does not define compression for geometry and other rich data, the https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md[KHR_draco_mesh_compression extension] provides that option. Future extensions may include compression methods for textures and animation data.
+While version 2.0 of glTF does not define compression for geometry and other rich data, there are extensions providing such options.
+* https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md[KHR_draco_mesh_compression] for geometry.
+* https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_texture_basisu/README.md[KHR_texture_basisu] for GPU-compatible block-compressed textures.
```
donmccurdy, line 274: These two bullets are collapsed into a single paragraph, may need an extra blank line above them.

specification/2.0/Specification.adoc
```diff
@@ -146,91 +269,146 @@ The following are outside the scope of the initial design of glTF:
...
-  - When using MIME type `application/gltf-buffer`, the binary file extension may be `*.bin`, `*.glbin`, or `*.glbuf`.  See the https://www.iana.org/assignments/media-types/application/gltf-buffer[gltf-buffer registration].
-* Texture files use the official `image/*` type based on the specific image format. For compatibility with modern web browsers, the following image formats are supported: `image/jpeg`, `image/png`.
-+
-[NOTE]
-.Implementation Note
-====
-Implementations should use the image type pattern matching algorithm from the https://mimesniff.spec.whatwg.org/#matching-an-image-type-pattern[MIME Sniffing Standard] to detect PNG and JPEG images as file extensions may be unavailable in some contexts.
-====
+[[file-extensions-and-media-types]]
+== File Extensions and Media Types
```
donmccurdy, line 292: I understand why the use of a `.gltf` extension for JSON glTF files is not required — we don't want to prohibit use cases like REST APIs, or perhaps like `.vrm`.

But, is it worth including some language that normatively prohibits a JSON glTF file using a `.glb` prefix, or a binary GLB file using `.gltf`? Such mixups would be unfortunate.

specification/2.0/Specification.adoc
```diff
@@ -146,91 +269,146 @@ The following are outside the scope of the initial design of glTF:
...
 [[json-encoding]]
 == JSON Encoding
 
-To simplify client-side implementation, glTF has additional restrictions on JSON format and encoding.
+Although glTF Specification does not define any subset of the <<json,JSON>> format, implementations **SHOULD** be aware of its peculiar properties that could affect asset interoperability.
 
-1. JSON must use UTF-8 encoding without BOM.
+1. glTF JSON data **SHOULD** be written with UTF-8 encoding without BOM. This requirement is not applied when a glTF implementation does not control string encoding. glTF implementations **SHOULD** adhere to the <<json,RFC 8259>>, Section 8.1. with regards to treating BOM presence.
+
+2. ASCII characters stored in glTF JSON **SHOULD** be written unescaped.
```
donmccurdy, line 314: The earlier language seemed to say this referred to property names and enums but not values (e.g. URIs). Since some ASCII characters should be escaped in URIs, we may need to be more specific. "... written unescaped, with the exception of defined URI values."?

specification/2.0/Specification.adoc
```diff
@@ -313,18 +493,12 @@ All angles are in radians.
...
 Positive rotation is counterclockwise.
 
-The <<transformations,node transformations>> and <<animations,animation channel paths>> are 3D vectors or quaternions with the following data types and semantics:
-
-* translation: A 3D vector containing the translation along the x, y and z axes
-* rotation: A quaternion (x, y, z, w), where w is the scalar
-* scale: A 3D vector containing the scaling factors along the x, y and z axes
-
-RGB color values use sRGB color primaries.
+RGB color values **MUST** use <<srgb,sRGB>> color primaries.
```
donmccurdy, line 496: Which RGB color values — outside of textures — use sRGB color primaries? baseColorFactor and emissiveFactor are linear, I'm not sure if those are considered "color values" even though they are factors, but I think this line may be ambiguous.

specification/2.0/Specification.adoc
```diff
@@ -572,20 +735,26 @@ The following example defines two buffer views: the first is an ELEMENT_ARRAY_BU
 }
 ----
 
-When a buffer view is used for vertex attribute data, it may have a `byteStride` property. This property defines the stride in bytes between each vertex.
+When a buffer view is used for vertex attribute data, it **MAY** have a `byteStride` property. This property defines the stride in bytes between each vertex. Buffer views with other types of data **MUST NOT** not define `byteStride`.
```
donmccurdy, line 738: Would it be valid for an extension to define a new "kind" of data that supports interleaving? Trying to accommodate `EXT_mesh_gpu_instancing` instance attributes and perhaps future extensions here.

donmccurdy (commented):
Reviewed as far as the beginning of the _Geometry_ section – will try to review further soon!

specification/2.0/Specification.adoc
lexaknyazev: Color primaries (aka chromaticities) have nothing to do with linear / non-linear encoding. They are essentially "units" of color values.

Factors are unitless multipliers for texture values; when the texture is missing, it's defined as having all values as `1.0`.

specification/2.0/Specification.adoc
lexaknyazev: The earlier language was causing more confusion than good and was not enforced anyway. IRIs are almost universally supported now, so only reserved characters have to be always percent-encoded (not the same as JSON escaping -- these are unrelated).

A simple example. Let's say a texture is called `🚀.png`. There are three valid options to encode it:

IRI within UTF-8 JSON (no escaping at all):
```json
{
  "uri": "🚀.png"
}
```

IRI within ASCII JSON (decoded by JSON parsers to the form above):
```json
{
  "uri": "\uD83D\uDE80.png"
}
```

URI within ASCII JSON (percent-encoding handled by URI parsers):
```json
{
  "uri": "%F0%9F%9A%80.png"
}
```

However when a filename contains _reserved_ characters (e.g. `#.png`), they always have to be percent-encoded.

donmccurdy: How about "... SHOULD be written without JSON escaping." to avoid ambiguity w/ URI escaping?

specification/2.0/Specification.adoc
donmccurdy: What is an example of an RGB color value in the glTF specification? I would guess most readers of this line will not know the distinction between color primaries and encoding. Would this be a fair clarification?

> RGB color values **MUST** use <<srgb,sRGB>> color primaries and linear encoding, excluding textures  as described in Section XX.XX.XX.

... is it correct to say that `baseColorFactor` and `COLOR_0`, and `KHR_lights_punctual`'s `light.color` are all sRGB primaries with linear encoding?

specification/2.0/Specification.adoc
```diff
@@ -91,45 +91,168 @@ for identification and belong to their respective owners.
...
+Fielding, R., Ed., and J. Reschke, Ed., _Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing_, RFC 7230, DOI 10.17487/RFC7230, June 2014, <https://www.rfc-editor.org/info/rfc7230>
+
+[[srgb]]
+IEC 61966-2-1:1999.
+_Default RGB colour space - sRGB_
+https://webstore.iec.ch/publication/6169
+
+[[mikktspace]]
+_MikkTSpace_
+https://github.com/mmikk/MikkTSpace
```
javagl, line 193: Should the MikkTSpace link point to GitHub, or to http://www.mikktspace.com/ ?

specification/2.0/Specification.adoc
```diff
@@ -776,24 +929,28 @@ The following fragment shows an example of `sparse` accessor with 10 elements de
...
+----
 
-Each `accessor` must fit its `bufferView`, i.e., `accessor.byteOffset + STRIDE * (accessor.count - 1) + SIZE_OF_ELEMENT` must be less than or equal to `bufferView.length`.
+**MUST** be less than or equal to `bufferView.length`.
 
-For performance and compatibility reasons, each element of a vertex attribute must be aligned to 4-byte boundaries inside `bufferView` (i.e., `accessor.byteOffset` and `bufferView.byteStride` must be multiples of 4).
+For performance and compatibility reasons, each element of a vertex attribute **MUST** be aligned to 4-byte boundaries inside a `bufferView` (i.e., `accessor.byteOffset` and `bufferView.byteStride` **MUST** be multiples of 4).
 
-Accessors of matrix type have data stored in column-major order; start of each column must be aligned to 4-byte boundaries. To achieve this, three `type`/`componentType` combinations require special layout:
+Accessors of matrix type have data stored in column-major order; start of each column **MUST** be aligned to 4-byte boundaries. To achieve this, three `type`/`componentType` combinations require special layout:
```
javagl, line 953: Maybe the examples could be descibed formally as something like

> When `rows * SIZE_OF_COMPONENT` is not a multiple of 4, then `(rows * SIZE_OF_COMPONENT) % 4` padding bytes have to be inserted at the end of each column, except for the last column in the accessor

(where `rows` is the number of rows of the matrix)

specification/2.0/Specification.adoc
```diff
@@ -924,77 +1101,106 @@ The following example defines a mesh containing one triangle set primitive:
 
 Each attribute is defined as a property of the `attributes` object. The name of the property corresponds to an enumerated value identifying the vertex attribute, such as `POSITION`. The value of the property is the index of an accessor that contains the data.
 
-Valid attribute semantic property names include `POSITION`, `NORMAL`, `TANGENT`, `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_0`, `JOINTS_0`, and `WEIGHTS_0`.  Application-specific semantics must start with an underscore, e.g., `_TEMPERATURE`.
+Valid attribute semantic property names include `POSITION`, `NORMAL`, `TANGENT`, `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_0`, `JOINTS_0`, and `WEIGHTS_0`.
```
javagl, line 1104: The wording for "valid...names include" seems a bit vague here. This might not be important, because formal details are given a bit later, but at this point, it made me wonder whether this means that "Clients MUST support the attribute semantic property names [...]..."

specification/2.0/Specification.adoc
```diff
@@ -924,77 +1101,106 @@ The following example defines a mesh containing one triangle set primitive:
...
+
+Application-specific attribute semantics **MUST** start with an underscore, e.g., `_TEMPERATURE`. Application-specific attribute semantics **MUST NOT** use _unsigned int__ component type.
```
javagl, line 1106: The formatting is a bit broken here, probably related to the second `__` after "unsigned int"...

specification/2.0/Specification.adoc
```diff
@@ -924,77 +1101,106 @@ The following example defines a mesh containing one triangle set primitive:
...
-                                  `5123` (UNSIGNED_SHORT)            | See <<skinned-mesh-attributes,Skinned Mesh Attributes>>
-| `WEIGHTS_0` | `"VEC4"`        | `5126` (FLOAT) +
-                                  `5121` (UNSIGNED_BYTE) normalized +
-                                  `5123` (UNSIGNED_SHORT) normalized | See <<skinned-mesh-attributes,Skinned Mesh Attributes>>
+                `"VEC4"`        | _float_ +
+                                  _unsigned byte_ normalized +
+                                  _unsigned short_ normalized | RGB or RGBA vertex color linear multiplier
+| `JOINTS_0`  | `"VEC4"`        | _unsigned byte_
+                                  _unsigned short_            | See <<skinned-mesh-attributes,Skinned Mesh Attributes>>
+| `WEIGHTS_0` | `"VEC4"`        | _float_) +
```
javagl, line 1128: Closing ) after float

specification/2.0/Specification.adoc
```diff
@@ -924,77 +1101,106 @@ The following example defines a mesh containing one triangle set primitive:
...
+                                  _unsigned byte_ normalized +
+                                  _unsigned short_ normalized | See <<skinned-mesh-attributes,Skinned Mesh Attributes>>
 |====
 
-`POSITION` accessor *must* have `min` and `max` properties defined.
+`POSITION` accessor **MUST** have its `min` and `max` properties defined.
 
-`TEXCOORD`, `COLOR`, `JOINTS`, and `WEIGHTS` attribute semantic property names must be of the form `[semantic]_[set_index]`, e.g., `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_0`. Client implementations must support at least two UV texture coordinate sets, one vertex color, and one joints/weights set. Extensions can add additional property names, accessor types, and/or accessor component types.
+`TEXCOORD`, `COLOR`, `JOINTS`, and `WEIGHTS` attribute semantic property names **MUST** be of the form `[semantic]_[set_index]`, e.g., `TEXCOORD_0`, `TEXCOORD_1`, `COLOR_0`. All indices for indexed attribute semantics **MUST** start with 0 and be continuous positive integers: `TEXCOORD_0`, `TEXCOORD_1`, etc. Indices **MUST NOT** use leading zeroes to pad the number of digits (e.g. `TEXCOORD_01`).
```
javagl, line 1135: I think this should be "consecutive" integers (not "continuous") - to be confirmed by some English guy!

specification/2.0/Specification.adoc
```diff
@@ -1013,23 +1219,27 @@ primitives[i].attributes.POSITION +
...
 |====
 
-`POSITION` accessor *must* have `min` and `max` properties defined.
+`POSITION` accessor **MUST** have `min` and `max` properties defined.
 
-All Morph Target's accessors *must* have the same `count` as the accessors of the original primitive.
+All Morph Target's accessors **MUST** have the same `count` as the accessors of the original primitive.
 
-A Morph Target may also define an optional `mesh.weights` property that stores the default targets weights. In the absence of a `node.weights` property, the primitives attributes are resolved using these weights. When this property is missing, the default targets weights are assumed to be zero.
+A mesh with Morph Targets **MAY** also define an optional `mesh.weights` property that stores the default targets weights. In the absence of a `node.weights` property, the primitives attributes are resolved using these weights. When this property is missing, the default targets weights are zeros.
```
javagl, line 1242: It was not clear (for me) what "resolved" should mean here, exactly...

specification/2.0/Specification.adoc
```diff
@@ -1064,13 +1274,9 @@ The following example extends the Mesh defined in the previous example to a morp
 }
 ----
 
-After applying morph targets to vertex positions and normals, tangent space may need to be recalculated. See <<appendix-a-tangent-space-recalculation,Appendix A>> for details.
+The number of morph targets is not limited. Client implementations **SHOULD** support at least eight morphed attributes. This means that they **SHOULD** support at least eight morph targets that contain a `POSITION` attribute, or four morph targets that contain a `POSITION` and a `NORMAL` attribute, or two morph targets that contain `POSITION`, `NORMAL` and `TANGENT` attributes.
```
javagl, line 1277: I see that this previously was an implementation note. But now, it made me (honestly) wonder whether an implementation SHOULD also support eight morph targets that (only) contain a NORMAL attribute. 

I think this could be avoided by more generically (and elaborately) saying that the implementation

> ...SHOULD support eight morph targets when each morph target has one attribute, four morph targets where each morph target has two attributes, or two morph targets where each morph target has three or four attributes

specification/2.0/Specification.adoc
```diff
@@ -1171,14 +1384,18 @@ The mesh for a skin is defined with vertex attributes that are used in skinning
 }
 ----
 
-The number of joints that influence one vertex is limited to 4 per set, so referenced accessors must have `VEC4` type and following component types:
+The number of joints that influence one vertex is limited to 4 per set, so referenced accessors **MUST** have `VEC4` type and following component types:
+
+* *`JOINTS_0`*: _unsigned byte_ or _unsigned short_
+* *`WEIGHTS_0`*: _float_, or _normalized unsigned byte_, or _normalized unsigned short_`
```
javagl, line 1390: Here and below, an unnecessary backtick after "unsigned short"

specification/2.0/Specification.adoc
```diff
@@ -1367,31 +1593,94 @@ The following example shows an image pointing to an external PNG image file and
...
 
-Samplers are stored in the `samplers` array of the asset. Each sampler specifies filter and wrapping options corresponding to the GL types. The following example defines a sampler with linear mag filtering, linear mipmap min filtering, and repeat wrapping in S (U) and T (V).
+==== Overview
+Samplers are stored in the `samplers` array of the asset. Each sampler specifies filtering and wrapping modes.
+
+Client implementations **SHOULD** follow specified filtering modes. When the latter are undefined, client implementations **MAY** set their own default texture filtering settings.
+
+Client implementations **MUST** follow specified wrapping modes.
+
+==== Filtering
```
javagl, line 1632: There are some strong and difficult requirements mentioned here. Although it (only) says "**SHOULD** follow specified filtering mode", it ends with "client implementations **MUST** be able to generate mipmaps at runtime". 

It's quite a burden to put on clients in view of the broader context of glTF. The point is: Clients will just pass the magnification info to OpenGL/WebGL, and have no influence whatsoever on whether these constraints are met. They will be, usually, but that's essentially in the responsibility of the graphics card / driver vendors. 

tl;dr: Should (or can) the glTF spec dictate MipMap-lookup-rules?

(Related: A bit later, in the context of "non power-of-two textures", this question is somewhat offloaded by talking about the PLATFORMS that the clients are running on...)

specification/2.0/Specification.adoc
```diff
@@ -1407,32 +1696,26 @@ Samplers are stored in the `samplers` array of the asset. Each sampler specifies
...
+==== Non-power-of-two Textures
+
+Client implementations **SHOULD** resize non power-of-two textures when running on certain platforms that have limited support for such texture dimensions.
 
 [NOTE]
-.Non-Power-Of-Two Texture Implementation Note
+.Implementation Note
 ====
-glTF does not guarantee that a texture's dimensions are a power-of-two.  At runtime, if a texture's width or height is not a power-of-two, the texture needs to be resized so its dimensions are powers-of-two if the `sampler` the texture references
+Specifically if the `sampler` the texture references
```
javagl, line 1707: This is a bit context-free. It could repeat the point: 

> Specifically, client implementations **SHOULD** resize non-poper-of-two textures (so that their horizontal and vertical sizes are powers of two) if the `sampler` the texture references ...

specification/2.0/Specification.adoc
```diff
@@ -1525,31 +1852,26 @@ The following examples shows a material that is defined using `pbrMetallicRoughn
...
 
-The `alphaMode` property defines how the alpha value of the main factor and texture should be interpreted. The alpha value is defined in the `baseColor` for metallic-roughness material model.
+The `alphaMode` property defines how the alpha value is interpreted. The alpha value is taken from the fourth component of the _base color_ for metallic-roughness material model.
 
 `alphaMode` can be one of the following values:
 
 * `OPAQUE` - The rendered output is fully opaque and any alpha value is ignored.
-* `MASK` - The rendered output is either fully opaque or fully transparent depending on the alpha value and the specified alpha cutoff value. This mode is used to simulate geometry such as tree leaves or wire fences.
+* `MASK` - The rendered output is either fully opaque or fully transparent depending on the alpha value and the specified _alpha cutoff_ value. This mode is used to simulate geometry such as tree leaves or wire fences.
 * `BLEND` - The rendered output is combined with the background using the normal painting operation (i.e. the Porter and Duff over operator). This mode is used to simulate geometry such as guaze cloth or animal fur.
```
javagl, line 1876: I wonder whether the Porter and Duff Over-Operator has to be described/specificed. (This is similar to the point about the treatment of the MipMap lookups, in that it is usually beyond the control of the client implementation). However, here, it is not imposing any requirements like MUST/SHOULD, but just said "to be this way"...

specification/2.0/Specification.adoc
```diff
@@ -1707,18 +2037,19 @@ where
...
 ====
-glTF 2.0 only supports animating node transforms and Morph Targets weights. A future version of the specification may support animating arbitrary properties, such as material colors and texture transform matrices.
+glTF 2.0 only supports animating node transforms and morph target weights. Extensions or a future version of the specification may support animating arbitrary properties, such as material colors and texture transform matrices.
 ====
 
 [NOTE]
 .Note
 ====
-glTF 2.0 defines only animation storage, so this specification doesn't define any particular runtime behavior, such as: order of playing, auto-start, loops, mapping of timelines, etc...
+glTF 2.0 defines only storage of animation keyframes, so this specification doesn't define any particular runtime behavior, such as: order of playing, auto-start, loops, mapping of timelines, etc... When loading a glTF 2.0 asset, client implementations may select an animation entry and pause it on the first frame, play it automatically, or ignore all animations until further user requests. When a playing animation is stopped, client implementations may reset the scene to the initial state or to freeze it at the current frame.
```
javagl, line 2052: The last "to" does not seem to belong there.

javagl:
I went through the remaining document. (Sorry, some comments are "nitpicking" about formatting, but I hope that they are not too distracting). 

One (inlined) point that might relevant for the "normativeness" is: In how far can the Filtering behavior be prescribed (e.g. the necessity to be able to generate MipMaps)?

Other, general points:

- Sometimes, the wording is "instanced", and sometimes "instantiated". Do we make a difference here?
- The connection between the component types (and the "normalized" flag) and the corresponding values (like 5126 for float) is not really clear, now that the references to the constants like `GL_FLOAT` have been omitted...

---

This may just be a lack of understanding on my side:

- I never deeply dug through the maths of PBR, but... when talking about "metal" and "non-metal" [here](https://github.com/KhronosGroup/glTF/pull/1997/files#diff-ca628c11679a8c4c0c4031440c7c1cda27c8896ebc87c3bd1e8640efc8942d99R1753) with the "metal" value being in [0,1], then I wonder: Is 0.0001 already "metal", or is 0.9999 still a "non-metal"? (Maybe this becomes clear in the appendix, but I didn't know what to do with this information, at this point, from a specification perspective...)
- It's nice that the sRGB computations are illustrated with an [example here](https://github.com/KhronosGroup/glTF/pull/1997/files#diff-ca628c11679a8c4c0c4031440c7c1cda27c8896ebc87c3bd1e8640efc8942d99R1788), but ... someone who's reading that will most likely have NO idea where these numbers came from. There's a link to some IEC spec, for ~175 bucks, so that may be safe from a "normative-ness" standpoint. But someone who tries to understand and implement that may have to do some websearches (and eventually stumble over http://davengrace.com/cgi-bin/cspace.pl ...)

---


Formatting/linking issues:

- "PropertiesReference.adoc" and "JsonSchemaReference.adoc" are not linked yet, but I assume that they will be in the future
- https://github.com/KhronosGroup/glTF/pull/1997/files#diff-ca628c11679a8c4c0c4031440c7c1cda27c8896ebc87c3bd1e8640efc8942d99R1976 The matrices are not properly formatted
- Formulas at https://github.com/KhronosGroup/glTF/blob/15f3ea26d5e539a67afe9dcfe29410d736097002/specification/2.0/Specification.adoc#b32-specular-brdf are not properly formatted

javagl (commented):
To summarize: I did not find any obvious issues. Some details are discussed in the inlined and PR  comments.

specification/2.0/Specification.adoc
```diff
@@ -409,29 +577,23 @@ The node hierarchy is defined using a node's `children` property, as in the foll
...
-TRS properties are converted to matrices and postmultiplied in the `T * R * S` order to compose the transformation matrix; first the scale is applied to the vertices, then the rotation, and then the translation.
-
-When a node is targeted for animation (referenced by an `animation.channel.target`), only TRS properties may be present; `matrix` will not be present.
+Any node **MAY** define a local space transformation either by supplying a `matrix` property, or any of `translation`, `rotation`, and `scale`  properties (also known as _TRS properties_). `translation` and `scale` are 3D vectors in the local coordinate system. `rotation` is a unit quaternion value, XYZW, in the local coordinate system, where W is the scalar.
 
+When `matrix` is defined, it **MUST** be decomposable to TRS properties.
 [NOTE]
 .Implementation Note
 ====
-If the determinant of the transform is a negative value, the winding order of the mesh triangle faces should be reversed. This supports negative scales for mirroring geometry.
```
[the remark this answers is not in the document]
lexaknyazev: This part has been relocated to the geometry section with more normative language. Looks like all major implementations (except Cesium) follow it already.

specification/2.0/Specification.adoc
```diff
@@ -539,17 +696,23 @@ The following example defines a buffer. The `byteLength` property specifies the
 }
 ----
 
-A *bufferView* represents a subset of data in a buffer, defined by a byte offset into the buffer specified in the `byteOffset` property and a total byte length specified by the `byteLength` property of the buffer view.
+The byte length of the referenced resource **MUST** be greater than or equal to the `buffer.byteLength` property.
+
+Buffer data **MAY** alternatively be embedded in the glTF file via `data:` URI with base64 encoding. When `data:` URI is used for buffer storage, its `mediatype` field **MUST** be set to `application/octet-stream` or `application/gltf-buffer`. The byte length of the base64-encoded data **MUST** be equal to the `buffer.byteLength` property.
```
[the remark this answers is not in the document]
lexaknyazev: Requiring an external file containing the binary buffer to have an exact size seems too restrictive as applications may want to store something (e.g. a checksum) at the end of it. This use case does not affect glTF operations. I don't think it's practical to do same with `data:`-URIs, but we should probably allow it anyway.

Thoughts?

specification/2.0/Specification.adoc
```diff
@@ -539,17 +696,23 @@ The following example defines a buffer. The `byteLength` property specifies the
...
 
-When a buffer view contain vertex indices or attributes, they must be its only content, i.e., it's invalid to have more than one kind of data in the same buffer view.
+A _buffer view_ represents a contiguous segment of data in a buffer, defined by a byte offset into the buffer specified in the `byteOffset` property and a total byte length specified by the `byteLength` property of the buffer view.
+
+Buffer views used for images, vertex indices, vertex attributes, or inverse bind matrices **MUST** contain only one kind of data, i.e. the same buffer view **MUST NOT** be used both for vertex indices and vertex attributes.
+
+When a buffer view is used by vertex indices or attribute accessors it **SHOULD** specify `bufferView.target` with a value of `34963` or `34962` respectively.
```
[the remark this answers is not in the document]
lexaknyazev: Yes, the intention is to have these magic numbers only in the schemas. We should add a brief explanation to the spec about that. There are only four affected places: accessors (incl. sparse), `bufferView.target`, samplers, and `primitive.mode`.

specification/2.0/Specification.adoc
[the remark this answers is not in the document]
javagl: Yes, I remember seeing it later, but did not remove this comment, sorry.

specification/2.0/Specification.adoc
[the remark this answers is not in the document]
javagl: I don't know exactly whether there may be a case that somebody wants to add, for example, some padding bytes to the data for whatever reason. The main point of the question was 1. I didn't see a reason for the difference between `>=` for buffer and `==` for data URI, and 2. I didn't see a strong reason to not also allow `>=` for the latter. But I don't have strong opinions here. 

(In doubt, one can argue to start with the "more strict" version, because _relaxing_ a requirement later is usually far easier than _strengthening_ a requirement - and unless there's a use case for something like "padded data URI data", we can leave it as it is)

specification/2.0/Specification.adoc
```diff
@@ -293,17 +471,19 @@ Entities of a glTF asset are referenced by their indices in corresponding arrays
...
+In this example, `buffers` and `bufferViews` arrays have only one element each. The bufferView refers to the buffer using the buffer's index: `"buffer": 0`.
+
+Indices **MUST** be non-negative integer numbers. Indices **MUST** always point to existing elements.
 
-Whereas indices are used for internal glTF references, _names_ are used for application-specific uses such as display. Any top-level glTF object can have a `name` string property for this purpose. These property values are not guaranteed to be unique as they are intended to contain values created when the asset was authored.
+Whereas indices are used for internal glTF references, optional _names_ are used for application-specific uses such as display. Any top-level glTF object **MAY** have a `name` string property for this purpose. These property values are not guaranteed to be unique as they are intended to contain values created when the asset was authored.
 
-For property names, glTF uses https://en.wikipedia.org/wiki/CamelCase[camel case] `likeThis`. Camel case is a common naming convention in JSON and WebGL.
+For property names, glTF usually uses camel case, `likeThis`.
```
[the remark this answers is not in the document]
lexaknyazev: Attribute semantics since they use uppercase.

specification/2.0/Specification.adoc
lexaknyazev: Completely reworded that part to avoid confusing readers with sRGB.

specification/2.0/Specification.adoc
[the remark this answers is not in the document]
lexaknyazev: Removed the equality requirement since it prevents buffer storage type conversion.

specification/2.0/Specification.adoc
lexaknyazev: Done.

specification/2.0/Specification.adoc
lexaknyazev: Done.

specification/2.0/Specification.adoc
lexaknyazev: Please review the updated language in that section below.

javagl (commented):
I did another quick pass, focussing on the changes, and think that they indeed add clarity. I cannot say much about the fallback behavior for the MipMapping on a technical level, but reducing the requirement from MUST to SHOULD IMHO makes sense, because it's usually beyond the control of the client. 

No further comments for now, except for a detail regarding the `ROWS * SIZE_OF_COMPONENT` part (inlined).

lexaknyazev:
Let's continue the refinements in #1901.

Issue #2008, opened 2021-07-30
Need to improve `cubicspline` interpolation for rotations


jbherdman:
In the current spec, Appendix C indicates that `cubicspline` interpolation of quaternions is performed as if the quaternions were simply 4D vectors, followed by a renormalize of the resulting 4D-vector so that it becomes a unit quaternion.  This interpretation is further clarified via the closed issue #1386.

This decision on interpolation logic is a problem, because it does not match the de facto industry standard of `sqlerp()` as introduced by Shoemake in 1985.  (Note that `squad()`, another de facto standard introduced by Shoemake in 1987, is basically just `sqlerp()` with specific rules about how to auto-compute "smooth" interior-tangents between quaternion keyframes.  We can side-step that added complexity by continuing to rely on explicit in-/out-tangents written to the glTF file.)

Looking at how quaternion interpolation is currently specified for glTF, the closest example that I can find is Blender's system.  This has been termed "RQBez" ("Renormalized quaternion Bezier curve") in literature.  However, it should be noted that the math from Appendix C does not fully replicate Blender's runtime system.  There are corner-cases that exist in the Blender code which are not shared with the glTF interpolation (flipping to "shortest arc" while interpolating, for example).  That is entirely reasonable, but it means that even "Blender quaternion animation" cannot be directly exported to the glTF interpolation model.

The better solution would be to adopt the de facto standard of `sqlerp()`.  This would allow most existing 'upstream' programs to output their native animation directly (other than upstream programs that rely on Euler angles for rotation animation).  This leads to smaller file sizes, as rotation-animation will generally not need to be 'resampled' for export.  It also avoids the 'RQBez' error-case when the interpolated 4D-vector passes too close to (0,0,0,0).

The main problem, of course, is that this ship has already sailed.  In theory, the current `cubicspline` behaviour for rotation is likely fixed in stone.  In practice, I would point out that very few public projects appear to properly support the existing `cubicspline` interpolation.  Presumably, though, the most prudent course of action would be to add a new interpolation-type, rather than changing the existing `cubicspline` logic.

lexaknyazev:
Technically, schemas are more fixed than the appendix; to add a new enum, we'd need to define an extension. However, given that
> very few public projects appear to properly support the existing `cubicspline` interpolation

it may be possible to somehow expand the existing language around splines.

@bghgary could you please look at #2009?

bghgary:
[...]

PR #2009, opened 2021-07-30
Adds 'CUBICSLERP' interpolation type for rotations

jbherdman:
Attempt at resolving #2008 by adding a new interpolation-type for rotations, which matches the de facto industry standard `sqlerp()`.


jbherdman:
Replaced by #2018

PR #2012, opened 2021-08-10
More language updates

lexaknyazev:
@emackey 
Could you please update pre-defined strings in `wetzel` (and publish a new version to npm) so that they use `**MUST**` instead of `must`?

### DIFF
--- a/specification/2.0/schema/material.schema.json
+++ b/specification/2.0/schema/material.schema.json
@@ -10,7 +10,7 @@
         "extras": { },
         "pbrMetallicRoughness": {
             "allOf": [ { "$ref": "material.pbrMetallicRoughness.schema.json" } ],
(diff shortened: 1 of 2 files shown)


specification/2.0/schema/material.schema.json
```diff
@@ -10,7 +10,7 @@
         "extras": { },
         "pbrMetallicRoughness": {
             "allOf": [ { "$ref": "material.pbrMetallicRoughness.schema.json" } ],
-            "description": "A set of parameter values that are used to define the metallic-roughness material model from Physically-Based Rendering (PBR) methodology. When undefined, all the default values of `pbrMetallicRoughness` **MUST* apply."
```
emackey, line 13: This wasn't introduced in this PR, but there's a `**MUST*` here, missing an asterisk.

PR #2014, opened 2021-08-16
Add the **MUST** keyword to wetzel's output

emackey:
Requested in #2012.


emackey:
@lexaknyazev I downloaded the output from this.  It looks like it's working correctly.

emackey:
Nope, I missed one.  Update coming...

emackey:
The missed case didn't actually appear in the glTF docs, but it's fixed now anyway.  This is ready.

PR #2016, opened 2021-08-25
Update EXT_meshopt_compression specification to fix and clarify bitstream format

zeux:
To make sure that the spec for EXT_meshopt_compression is actually implementable, I've worked with @magcius to create a reference JS decoder by following the spec, and validate it using decoding tests from the original implementation.

While not a perfect clean-room implementation (the original implementation had to be consulted a couple of times because the spec was difficult to understand), this seems sufficiently comprehensive to make sure that the spec actually describes the intended bitstream correctly.

During this process, two bugs in the spec were found, both of them in the octahedral filter:

- fixed point encoding of 1.0 was stored in the .Z element ([2]); the comment correctly pointed that out, but pseudo-code used index 3
- the octahedral transform used the wrong sign (+= copysign instead of -= copysign).

These two bugs have been corrected to match the actual implementation. Other than that, no bugs were found so the spec seems correct, but this change contains many clarifications and wording tweaks made during the implementation process to hopefully make the spec more precise and easier to understand at the same time.

The reference decoder still needs a little bit of cleanup and it will be part of meshoptimizer repository, whereas this change contains the spec updates.

### DIFF
--- a/extensions/2.0/Vendor/EXT_meshopt_compression/README.md
+++ b/extensions/2.0/Vendor/EXT_meshopt_compression/README.md
@@ -224,13 +224,13 @@ The header bits establish the delta encoding mode (0-3) for each group of 16 ele
 - bits 2: Deltas are stored in 4-bit sentinel encoding; the size of the encoded block is [8..24] bytes
 - bits 3: All 16 byte delta are stored as bytes; the size of the encoded block is 16 bytes
 
-When using the sentinel encoding, each delta is stored as a 2-bit or 4-bit value in a single 4-byte or 8-byte block, with deltas stored from most significant to least significant bit inside the byte. That is, the 2-bit encoding is packed as:
+When using the sentinel encoding, each delta is stored as a 2-bit or 4-bit value in a single 4-byte or 8-byte block, with deltas stored from most significant to least significant bit inside the byte. That is, the 2-bit encoding is packed as follows with 4 deltas per byte:
 
 ```
 (delta3 << 0) | (delta2 << 2) | (delta1 << 4) | (delta0 << 6)
 ```
 
-And the 4-bit encoding is packed as:
+And the 4-bit encoding is packed as follows with 2 deltas per byte:
 
 ```
 (delta1 << 0) | (delta1 << 4)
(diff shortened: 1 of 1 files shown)


extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -191,31 +193,52 @@ The encoded stream structure is as follows:
...
 
 The header bits establish the delta encoding mode (0-3) for each group of 16 elements that follows:
 
 - bits 0: All 16 byte deltas are 0; the size of the encoded block is 0 bytes
-- bits 1: Deltas are using 2-bit sentinel encoding; the size of the encoded block is [4..20] bytes
-- bits 2: Deltas are using 4-bit sentinel encoding; the size of the encoded block is [8..24] bytes
-- bits 3: All 16 byte deltas are stored verbatim; the size of the encoded block is 16 bytes
+- bits 1: Deltas are stored in 2-bit sentinel encoding; the size of the encoded block is [4..20] bytes
+- bits 2: Deltas are stored in 4-bit sentinel encoding; the size of the encoded block is [8..24] bytes
+- bits 3: All 16 byte delta are stored as bytes; the size of the encoded block is 16 bytes
```
magcius, line 225: s/delta/deltas/

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -191,31 +193,52 @@ The encoded stream structure is as follows:
...
+maxBlockElements = min((8192 / byteStride) & ~15, 256)
 ```
 
-The attribute block structure consists of `byteStride` blocks (one for each byte of the element) with the following structure:
+Each attribute block consists of `byteStride` "data blocks" (one for each byte of the element), and each "data block" contains deltas stored for groups of elements. Each group always contains 16 elements; when the number of elements that needs to be encoded isn't divisible by 16, it gets rounded up and the remaining elements are ignored after decoding. In other terms:
 
-- Header bits, 2 bits for each group of 16 elements (`blockSize`/16 2-bit values), padded to a byte
-- Delta blocks, with variable number of bytes stored for each group of 16 elements
+```
+groupCount = ceil(blockElements / 16)
```
magcius, line 207: Clarify where blockElements comes from. I did it as follows in my implementation, but I would prefer it to be spelled out, especially if this is not correct.

> ```
> blockElements = min(remainingBlockElements, maxBlockElements)
> ```
> 
> where `remainingBlockElements` is the number of elements that have yet to be written. A stream with a byteStride of 64 containing 200 elements would be broken up into two attribute blocks: one containing 128 elements, and the other containing 72 elements. And these blocks would have 8 and 5 groups, respectively.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -307,15 +334,15 @@ The third index, `c`, is equal to `last-1` for `0xXd` and `last+1` for `0xXe`.
 `last` is set to `c` (effectively decrementing or incrementing it accordingly).
 
 Edges (c, b) and (a, c) are pushed to edge FIFO (in this order).
```
magcius, line 336: Missing a "the"

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
zeux: Yeah this is correct. I've removed "up to" from the description of the element count and changed `remainingBlockElements` to simply `remainingElements` to be more clear, but otherwise incorporated this into the spec edits.

zeux:
FWIW this should be ready for review/merge cc @donmccurdy @lexaknyazev

lexaknyazev (approved):
Apologies for the delay. LGTM.

PR #2018, opened 2021-08-30
Add 'EXT_animation_sqlerp' extension

### LINKED ISSUE #2008: Need to improve `cubicspline` interpolation for rotations

jbherdman:
Attempt to resolve #2008 via an extension, rather than direct spec change.

### DIFF
--- a/extensions/2.0/Vendor/EXT_animation_sqlerp/README.md
+++ b/extensions/2.0/Vendor/EXT_animation_sqlerp/README.md
@@ -1,0 +1,157 @@
+# EXT_animation_sqlerp
+
+## Contributors
+
+* Joe Herdman, [@jbherdman](https://github.com/jbherdman)
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds a new animation interpolation mode to glTF, based on the de facto industry standard of using spherical cubic interpolation (sqlerp) for quaternion animation.
+
+
+## glTF Schema Updates
+
+This extension allows for an additional interpolation-type in the `animation.sampler.interpolation` enumeration:
+   * `"CUBICSLERP"` The animation's interpolation is computed using spherical cubic interpolation (sqlerp) with specified tangents. This interpolation mode is only valid when targeting a rotation. The number of output elements must equal three times the number of input elements. For each input element, the output stores three elements, an in-tangent quaternion, a spline quaternion, and an out-tangent quaternion. There must be at least two keyframes when using this interpolation.
+
+See [CUBICSLERP Interpolation](#cubicslerp-interpolation) for additional information about spherical cubic interpolation (sqlerp).
+
+The `EXT_animation_sqlerp` extension is added to the `animation.channel` object and specifies a `sampler` property.  The `sampler` property references an `animation.sampler` (within the same `animation`) which defines animation data using the new `CUBICSLERP` interpolation type.
+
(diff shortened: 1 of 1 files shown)


UX3D-nopper:
> > Currently, when LINEAR, we implicitly guide to slerp.
> 
> No, the spec explicitly mentions slerp (Section C.4).
> 
> > But what about quaternion cupic spline interpolation?
> 
> Quaternions are interpolated as regular vectors and normalized afterwards (Section C.5).

Actually, I want to fix C.5, as in C.4 a LINEAR interpolation "becomes" a slerp. Why not in C.5 a CUBICSPLINE becomes sqlerp?
In one case another interpolation method is used, in the other case the approach is different.
For me, this is inconsistent, when there is potential for a better approach.

jbherdman:
Closing this PR as the extension/behaviour has been found to not meet my company's specific goals.  The technology/math is sound, but it turns out there are other incompatible restrictions on our end.  See https://github.com/KhronosGroup/glTF/issues/2008#issuecomment-956506135

@UX3D-nopper Feel free to continue including this in #2033 if it suits your purposes.

Also, after fully implementing this, I'm not sure the extra `animation.channel.extensions.EXT_animation_sqlerp.sampler` parameter is the right way to go.  The existing `animation.sampler.interpolation` is still going to have the new `cubicslerp` value in it, which means that loaders that are not aware of this extension may encounter parsing problems?  It might be a better approach to force this extension as "required" when it is used, and just set the extended `cubicslerp` sampler via the usual `animation.channel.sampler` property.

UX3D-nopper:
Okay, good to know. It simplifies the animation2 extension discussion.

Thank you for contributing.

PR #2019, opened 2021-08-30
EXT_feature_metadata feedback

lilleyse:
See https://github.com/CesiumGS/glTF/pull/4 for @pjcozzi's feedback and see https://github.com/CesiumGS/glTF/pull/3 for the original PR.


lilleyse:
My mistake, I targeted the wrong origin when opening this PR. This is for an in-progress extension that Cesium is working on and the discussion is here https://github.com/CesiumGS/glTF/pull/9.

PR #2022, opened 2021-09-03
removed duplicate (1 - f0) in implementation notes

elalish:
@WestLangley noticed that for our suggested implementation, if you plug in a pure dielectric material at normal incidence, our equations reduce to `f_diffuse = (1 / pi) * (1 - f0)^2 * baseColor`, which seems to be an error, as the energy-conservation term `(1 - f0)` is double-counted. I believe this is the right fix, but would appreciate more eyes on it.

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -4071,7 +4071,6 @@ material = mix(dielectric_brdf, metal_brdf, metallic)
 Taking advantage of the fact that `roughness` is shared between metal and dielectric and that the Schlick Fresnel is used, we can simplify the mix and arrive at the final BRDF for the material:
 
 ```
-const dielectricSpecular = 0.04
 const black = 0
 
 c_diff = lerp(baseColor.rgb, black, metallic)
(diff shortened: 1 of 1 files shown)


specification/2.0/README.md
```diff
@@ -4074,7 +4074,7 @@ Taking advantage of the fact that `roughness` is shared between metal and dielec
 const dielectricSpecular = 0.04
```
proog128, line 4074: The constant `dielectricSpecular` could be removed, it's not used anymore.

proog128 (approved):

specification/2.0/README.md
elalish: Done.

bghgary:
This PR addresses this issue: https://github.com/KhronosGroup/glTF/issues/1945

emackey:
@lexaknyazev What is the cutoff for getting fixes like this copied into `jon-adoc`?

elalish:
Ping for merge, now that this has been discussed on the call?

PR #2027, opened 2021-09-20
Adding the KHR_materials_iridescence effect

UX3D-nopper:
Replaces https://github.com/KhronosGroup/glTF/pull/1742


bhouston:
The weirdest thing we found recently is that you can incorporate roughness into the iridescence effect and it allows you to simulate iridescence when it is created via metal tempering, otherwise it isn't accurate for metal tempering cases: https://twitter.com/BenHouston3D/status/1437767108423737347

Basically most iridescence we are doing is based on an oily film, which usually has a roughness close to 0.  As you increase the roughness of the iridescence layer, such as when it is on a rough metallic surface and maintains that roughness because it is created via metal tempering, the resulting effect becomes more averaged and doesn't shift around as much based on viewing angle.  Maybe the math for this isn't yet found in an academic paper, but it does seem straight forward, such integrate across the solid angle implied by the roughness.

That said, implementing ideas into the formal glTF standard without a corresponding academic paper to cite is probably too forward looking.

UX3D-nopper:
> The weirdest thing we found recently is that you can incorporate roughness into the iridescence effect and it allows you to simulate iridescence when it is created via metal tempering, otherwise it isn't accurate for metal tempering cases: https://twitter.com/BenHouston3D/status/1437767108423737347
> 
> Basically most iridescence we are doing is based on an oily film, which usually has a roughness close to 0. As you increase the roughness of the iridescence layer, such as when it is on a rough metallic surface and maintains that roughness because it is created via metal tempering, the resulting effect becomes more averaged and doesn't shift around as much based on viewing angle. Maybe the math for this isn't yet found in an academic paper, but it does seem straight forward, such integrate across the solid angle implied by the roughness.
> 
> That said, implementing ideas into the formal glTF standard without a corresponding academic paper to cite is probably too forward looking.
> 
> BTW credit for extending iridescence with roughness to simulate metallic tempering goes to @DanielSturk.

Today I had a call with Pascal. He pointed me to this publication:
https://cg.cs.uni-bonn.de/en/publications/paper-details/kneiphof-2019-iridescence/
Think this is enough academic.

bhouston:
Nice paper!  Do you have the PDF somewhere?  Couldn't find a place to download it.

So should we incorporate either the base PBR roughness into the iridescence calculation or introduce a new iridescence-specific roughness into this extension as we did with the clear coat layer?  I feel that we may be able to get away with coupling it with the base PBR roughness.

bhouston:
Found the slides here: https://cg.cs.uni-bonn.de/aigaion2root/attachments/Real-time%20Image-based%20Lighting%20of%20Microfacet%20BRDFs%20with%20Varying%20Iridescence%20Slides.pdf. That is probably enough.

bhouston:
I think we have too many parameters in this extension and we should figure out how to trim them.

I notice in the original Belcour/Belcour paper they write: "We control variations of the thin- film (η2 = 2.0) using a texture that remaps Dinc to the [0..415]nm range, mimicking a condensation effect that is more or less revealed depending on the slab orientation. "

This suggests we do not need the "thinFilmThicknessMinimum" parameter.

Given that "baseExtinctionCoefficient" is for the underlying metal layer and we have already chosen not to model this, I do not feel we should put it into the iridescence extension.  It is the wrong place anyhow - if we want to model it, we should always model it in the base metal layer where the PBR metalness parameter is defined, rather than requiring one to use the iridescence layer.

This gets rid of two of the current parameters.

UX3D-nopper:
> I think we have too many parameters in this extension and we should figure out how to trim them.
> 
> I notice in the original Belcour/Belcour paper they write: "We control variations of the thin- film (η2 = 2.0) using a texture that remaps Dinc to the [0..415]nm range, mimicking a condensation effect that is more or less revealed depending on the slab orientation. "
> 
> This suggests we do not need the "thinFilmThicknessMinimum" parameter.
> 
> Given that "baseExtinctionCoefficient" is for the underlying metal layer and we have already chosen not to model this, I do not feel we should put it into the iridescence extension. It is the wrong place anyhow - if we want to model it, we should always model it in the base metal layer where the PBR metalness parameter is defined, rather than requiring one to use the iridescence layer.
> 
> This gets rid of two of the current parameters.

Hmm, just one section below, in Fig. 15, the range is from ∈ [0..1470] nm. These are just examples. The minimum is a suggestion, that one can better adjust the values in a range.

baseExtinctionCoefficient is added by Pascal, I will look into it as well.

UX3D-nopper:
@PascalSchoen I removed the base extinction, as in this case Ben is right. I also changed the parameter names to be more glTF style.
@lexaknyazev I have granted you access to the fork. Please feel free to make any formal changes regarding the parameters.

@PascalSchoen Can we somehow approximate just using iridesenceIOR and the IOR? Furthermore, having IOR in this extension, adds an dependency to the IOR extension.

bhouston:
> Hmm, just one section below, in Fig. 15, the range is from ∈ [0..1470] nm. These are just examples. The minimum is a suggestion, that one can better adjust the values in a range.

What I am trying to say is that all of their ranges start at 0, not an arbitrary minimum.  Given that we could have minimums on nearly every texture modulator the system, I would argue that we have already decided not to do that, thus we could leave out the minimum here.

UX3D-nopper:
The problem is, that you loose the prcision, thta's the reason why we should keep it.
Let's say, you use values between 1200 and 1400 for your asset, then the texture is mapped to this 200 range.
Otherwise the steps are too large.

bhouston:
> Let's say, you use values between 1200 and 1400 for your asset, then the texture is mapped to this 200 range.
Otherwise the steps are too large.

Are you sure the steps are too large?  I believe that a minimum is never used in any of the papers we are referencing.

We have the same issue for emissive textures, etc, which can have arbitrary large ranges.

I believe there are existing solutions, such as use a float texture.  Also this gets solved by a graph-based material system like MaterialX.

I would advocate just a scale factor + texture as we do with every other slot, otherwise I feel we are adding complexity for reasons we are not completely sure of yet.

UX3D-nopper:
Please have a look at this presentation at page 71 and 72
https://blog.selfshadow.com/publications/s2017-shading-course/drobot/s2017_pbs_multilayered.pdf
There are cases between e.g. 280 and 282nm.
Also, the art control (see on page 70), is min / max thickness range as well.
If a e.g.[0;1200] is mapped to a [0:255], one unit step would be around 4.7 nm.

bhouston:
> Also, the art control (see on page 70), is min / max thickness range as well.

You are correct, this does suggest it is useful.  I just worry about the fact that you can easily run out of uniforms or whatnot on a lot of platforms.  But maybe it is less of an issue than it was on the past.  It is too bad webglstats.com is dead.

PascalSchoen:
@UX3D-nopper @bhouston, I adjusted the model to work without having the `baseExtinctionCoefficient` for metallic materials. The way I modelled it is to use the `baseColor` as the input for the `artisticIOR(reflectivity, edgeColor)` function (with `baseColor` used for both `reflectivity` and `edgeColor`) defined in [this paper](http://jcgt.org/published/0003/04/03/paper.pdf).

I validated the shading model with the implementation in [Pixar RenderMan 24](https://rmanwiki.pixar.com/display/REN24/LamaConductor). Here a comparison:

glTF Sample Viewer:
Screenshot 2021-09-23 at 17 07 18

RenderMan 24:
Screenshot 2021-09-23 at 17 07 25

I have to merge the current `master` branch in and then I'll also push the implementation in the glTF Sample Viewer.

UX3D-nopper:
> @UX3D-nopper @bhouston, I adjusted the model to work without having the `baseExtinctionCoefficient` for metallic materials. The way I modelled it is to use the `baseColor` as the input for the `artisticIOR(reflectivity, edgeColor)` function (with `baseColor` used for both `reflectivity` and `edgeColor`) defined in [this paper](http://jcgt.org/published/0003/04/03/paper.pdf).
> 
> I validated the shading model with the implementation in [Pixar RenderMan 24](https://rmanwiki.pixar.com/display/REN24/LamaConductor). Here a comparison:
> 
> glTF Sample Viewer:
> Screenshot 2021-09-23 at 17 07 18
> 
> RenderMan 24:
> Screenshot 2021-09-23 at 17 07 25
> 
> I have to merge the current `master` branch in and then I'll also push the implementation in the glTF Sample Viewer.

Looks great, looking forward to see it live.

bhouston:
I did some experiments with complex fresnel that includes both the n and k terms and I found it to be underwhelming in terms of the resulting impact:

https://www.shadertoy.com/view/ttySzt

Thus I question wether it is important to have extinction in this model at all?

bhouston:
BTW I have also done the artistic fresnel thing before and it is neat, but because it is attempting to simulate the n/k terms of complex fresnel and I found that the n/k terms have almost no visual impact, I question whether we need to realize the n/k terms via artistic fresnel.  I think we just do not need it: https://www.shadertoy.com/view/wlySRt

Is there a visual impact to what you are doing with artistic fresnel derivation on physical style materials?  I couldn't find any real-world examples where this made a significant impact in my testing.  Can you show with and without?

UX3D-nopper:
> BTW I have also done the artistic fresnel thing before and it is neat, but because it is attempting to simulate the n/k terms of complex fresnel and I found that the n/k terms have almost no visual impact, I question whether we need to realize the n/k terms via artistic fresnel. I think we just do not need it: https://www.shadertoy.com/view/wlySRt
> 
> Is there a visual impact to what you are doing with artistic fresnel derivation on physical style materials? I couldn't find any real-world examples where this made a significant impact in my testing. Can you show with and without?

@PascalSchoen Maybe you can better answer, if this is significant to you

UX3D-nopper:
One can try out the current implementation here: https://github.com/KhronosGroup/glTF-Sample-Viewer/pull/352

extensions/2.0/Khronos/KHR_materials_iridescence/schema/glTF.KHR_materials_iridescence.schema.json
```diff
@@ -0,0 +1,50 @@
...
+    "description": "glTF extension that defines an iridescence effect.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "iridescenceFactor": {
+            "type": "number",
+            "description": "The iridescence intensity factor.",
+            "default": 0.0,
+            "minimum": 0.0,
+            "maximum": 1.0,
+            "gltf_detailedDescription": ""
```
emackey, line 14: Three parameters have a blank value for `gltf_detailedDescription`.  These should be either removed or populated.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+If `iridescenceFactor` is zero (default), the iridescence extension has no effect on the material. 
+All textures in this extension use a single channel in linear space. 
+The thickness of the thin-film is set to `iridescenceThicknessMaximum` if `iridescenceThicknessTexture` is not given.
+If `iridescenceThicknessTexture` is set, the thickness of the thin-film varies between  `iridescenceThicknessMinimum` and `iridescenceThicknessMaximum` as follows: 
+```
+thickness = mix(iridescenceThicknessMinimum, iridescenceThicknessMaximum, iridescenceThicknessTexture.g)
+```
+The thin-film layer can have a different IOR than the underlying material. With `iridescenceIOR` one can set an IOR value for the thin-film layer independently.
```
proog128, line 77: Where does the thin film layer belong in the layering stack? How does it interact with other extensions like clearcoat and sheen?

UX3D-nopper: @UX3D-kanzler @UX3D-becher @UX3D-eckerlein @PascalSchoen 
Can you please add the missing details depneding on the current implementation:
https://github.com/KhronosGroup/glTF-Sample-Viewer/pull/352

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+The iridescence effect is modeled via a microfacet BRDF with a modified Fresnel reflectance term that accounts for inter-reflections as shown in [Laurent Belcour and Pascal Barla (2017)](https://belcour.github.io/blog/research/publication/2017/05/01/brdf-thin-film.html).
+
```
proog128, line 80: In the sheen extension there is a [long, non-normative section "Implementation Notes"](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_sheen/README.md#implementation-notes) that gives a sample implementation of the sheen effect. I like it a lot, because I think it is a good starting point for a custom implementation. Do you plan to add such a section to this extension?

UX3D-nopper: Good point. @UX3D-becher

UX3D-nopper: Tobias, feel free to add things and add your name to the creators of this extension.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,148 @@
...
+|------------------------------|---------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|-----------------------|
+|**iridescenceFactor**         | `number`                                                            | The iridescence intensity.                                                                       | No, default: `0.0`    |
+|**iridescenceTexture**        | [`textureInfo`](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | The iridescence intensity stored in the red channel of the texture.                                                               | No                    |
+|**iridescenceIOR**               | `number`                                                            | The index of refraction of the dielectric thin-film layer.                                       | No, default: `1.8`    |
```
abwood, line 59: Question that came up recently from @elalish, what effect does IOR have on thin film?

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,148 @@
...
+}
+```
+
+### Iridescence
+
+All implementations should use the same calculations for the BRDF inputs. Implementations of the BRDF itself can vary based on device performance and resource constraints. See [appendix](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#appendix-b-brdf-implementation) for more details on the BRDF calculations.
+
+|                              | Type                                                                | Description                                                                                      | Required              |
+|------------------------------|---------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|-----------------------|
+|**iridescenceFactor**         | `number`                                                            | The iridescence intensity.                                                                       | No, default: `0.0`    |
```
abwood, line 57: We discussed whether or not we could remove this factor and instead use thickness to fade out the effect. Is that possible with this model?

emackey: Specifically, what's the difference between "intensity" dropping to zero vs. "thickness" dropping to zero?  Is there a visible difference, making it worthwhile to use two channels instead of a single combined channel?  (It would imply minimum locked to zero, of course).

emackey: I just noticed the conversation of this on the mailing list... apologies for the duplication!

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
UX3D-nopper: Here it is nicely described:
https://docs.arnoldrenderer.com/display/A5AFMUG/Thin+Film

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
UX3D-nopper: Fot the non-mailing list users: The intensity is just an artistic control, so thickness is sufficient.

UX3D-nopper:
If the extensions is present in a material and only the default values are used, the effect is off. This is because the iridescenceFactor is 0.0 and this behaviour we do have in other extensions as well.
By setting the iridescenceFactor to 1.0, the effect is fully enabled. Any value in between is like a blending between the effect off and fully enabled.
If no texture is defined, the value of the texture is treated like it is 1.0. So iridescence = iridescenceFactor * iridescenceTexture.r becomes iridescence = 0.0 * 1.0.
If the iridescenceTexture is defined, it is behaving like a mask. Providing an empty texcture is an invalid glTF as in other extensions.

A similar definition is e.g. in the clear coat extension:
https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_clearcoat

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+To calculate `F_iridescence` we need to calculate the spectral differences. 
+At first, optical path differences (`diff`) between rays reflected at the thin-film and at the base material have to be calculated:
+```
+diff = 2.0 * IOR_iridescence * thin_film_thickness * cos(viewAngle)
+```
+With the path difference, the phase shift can be calculated in Fourier space.
+
+<img src="https://render.githubusercontent.com/render/math?math=\displaystyle%20\Delta%20\phi%20=%202%20\pi%20\lambda^{-1}\text{diff}">
```
abwood, line 125: It is unclear to me how the phase shift defined in this equation ties into the `F_iridescence` calculation?

emackey:
@echadwick-wayfair Another way to look at the answer to your question is:

- The `Factor` defaults to `0.0`, meaning if `Factor` isn't there, the feature is turned off.
- The `Texture` defaults to `1.0`, because 1 is the identity for multiplication.  If the texture isn't there, its effects are ignored (multiplied by 1) and only the Factor is relevant.

echadwick-wayfair:
Where can I see the glTF syntax used by the IridescentDance.glb as shown here? https://gltf.ux3d.io/

I can't figure out how to get strong iridescent coloration in my own model; the effect for me is barely visible.

UX3D-nopper:
> Where can I see the glTF syntax used by the IridescentDance.glb as shown here? https://gltf.ux3d.io/
> 
> I can't figure out how to get strong iridescent coloration in my own model; the effect for me is barely visible.

@UX3D-kanzler Cab you please create a basic glTF with the effect, ZIP it and attach to this thread?

UX3D-kanzler:
This is the model that I have used to compare the effect of different thickness ranges. There one can see the values that are used to create materials with various saturation levels.
[iridescence.zip](https://github.com/KhronosGroup/glTF/files/7594610/iridescence.zip)

PascalSchoen:
I added two more examples to the samples repository: https://github.com/KhronosGroup/glTF-Sample-Models/pull/331

There is one example for each metallic and dielectric base materials, which contain a set of spheres each with varying thin-film thicknesses, thin-film IOR and material IOR.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,207 @@
...
+|------------------------------|---------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|-----------------------|
+|**iridescenceFactor**         | `number`                                                            | The iridescence intensity.                                                                       | No, default: `0.0`    |
+|**iridescenceTexture**        | [`textureInfo`](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | The iridescence intensity stored in the red channel of the texture.                                                               | No                    |
+|**iridescenceIOR**               | `number`                                                            | The index of refraction of the dielectric thin-film layer.                                       | No, default: `1.8`    |
```
proog128, line 59: Why is `iridescenceIOR` not texturable? In contrast to KHR_materials_ior which defines the IOR of the 3D volume, the thin film IOR is a 2D surface effect, so texturing is easily possible. I understand that due to several technical reasons less textures is better than more textures, but it feels somewhat arbitrary. What if a texture is used as an atlas to represent different materials on a single shape? Maybe the reason for this decision should be stated somewhere in the spec.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,207 @@
...
+<figure>
+<img src="./figures/thickness-comparison.jpg"/>
+<figcaption><em>Comparison of different iridescence thickness ranges for a constant iridescence IOR value of 1.8. </em></figcaption>
+</figure>
+
+The thin-film layer can have a different IOR than the underlying material. With `iridescenceIOR` one can set an IOR value for the thin-film layer independently. 
+This also has an effect on the optical path difference:
+
+<figure>
+<img src="./figures/ior-comparison.jpg"/>
```
proog128, line 100: Are IOR values in the image (1.7, 2.0, 2.3) representative for thin-film materials? I would have expected lower values, maybe in the range 1.3 to 1.6 (water, oil, glass), but I have no idea what materials are used in fashion or car paints. That said, a table with IORs of a couple of typical materials used for thin-film would be nice, similar to [the table in KHR_materials_ior](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_ior#extending-materials). That would give artists a realistic, physical starting point for adjustments.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+
+The iridescence effect is modeled via a microfacet BRDF with a modified Fresnel reflectance term that accounts for inter-reflections as shown in [Laurent Belcour and Pascal Barla (2017)](https://belcour.github.io/blog/research/publication/2017/05/01/brdf-thin-film.html).
+
+
+## Implementation Notes
+
+*This section is non-normative.*
+
+For the calculation of `f_specular`, the original Fresnel term `F_schlick` is exchanged with a newly introduced `F_iridescence` term.
```
proog128, line 112: The term `F_schlick` does not exist in the base spec.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,207 @@
...
+At first, optical path differences (`diff`) between rays reflected at the thin-film and at the base material have to be calculated:
+```
+diff = 2.0 * IOR_iridescence * thin_film_thickness * cos(viewAngle)
+```
+With the path difference, the phase shift can be calculated in Fourier space.
+
+<img src="https://render.githubusercontent.com/render/math?math=\displaystyle%20\Delta%20\phi%20=%202%20\pi%20\lambda^{-1}\text{diff}">
+
+A complete implementation for evaluating `F_iridescence` can be found in the shader file of the Khronos Sample Viewer:
+[iridescence.glsl](https://github.com/PascalSchoen/glTF-Sample-Viewer/blob/feature/KHR_materials_iridescence/source/Renderer/shaders/iridescence.glsl)
```
proog128, line 128: Some notes on the `fresnelConductorExact` function which is being used in the sample code:

```
vec3 eta_1 = vec3(eta1);

// Force eta_2 -> 1.0 when Dinc -> 0.0
vec3 eta_2 = mix(eta_1, vec3(eta2), smoothstep(0.0, 0.03, Dinc));

vec3 eta_3, kappa_3;
artisticIor(baseF0, mix(vec3(0.0), baseF0, metallic), eta_3, kappa_3);

// ...

fresnelConductorExact(cosTheta1, eta_2[i] / eta_1[i], 0.0, R12p[i], R12s[i]);
fresnelConductorExact(cosTheta2[i], eta_3[i] / eta_2[i], kappa_3[i] / eta_2[i], R23p[i], R23s[i]);
```

First call is first interface, second call is second interface (if I am not mistaken). The name implies that this is a Fresnel for conductors (I found a *very* similar looking function in [Mitsuba](https://github.com/mitsuba-renderer/mitsuba/blob/master/src/libcore/util.cpp#L739), in case you took it from there, note that this is GPL license). However, for the first interface, I would expect a dielectric Fresnel. You even mention this in the property table: "iridescenceIOR - The index of refraction of the dielectric thin-film layer." For the second interface, I would expect it to be either dielectric or conductor, depending on the metalness (`metallic` parameter) of the base material.

`eta_3` and `kappa_3`, i.e., the complex IOR of the base, is computed in `artisticIor`. There are several issues:
* `artisticIor` implements Gulbrandsen's Fresnel term. I am wondering why you chose that term. It's not used in the glTF spec, and its use is actually discouraged because it has several problems. See (Naty Hoffman's talk "Fresnel Equations Considered Harmful")[https://renderwonk.com/publications/mam2019/] or the discussions in KHR_materials_specular for the reasons.
* The `mix` in the call to `artisticIor` sets `edgeColor` to `baseF0` if `metallic = 1`. This is not how the glTF base material behaves. According to the base spec, F90 = 1 for metals.

In summary, I think the Fresnel at the first interface should be a regular dielectric Fresnel (maybe Schlick, maybe the full Fresnel equations) and the Fresnel at the second interface should be `F` from the base spec, KHR_materials_ior and KHR_materials_specular (i.e., it correctly accounts for metallic, baseColor, specular, specular color, and volume IOR).

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Clearcoat
+Considering KHR_materials_clearcoat extension, the clearcoat layer is evaluated before the thin-film layer. 
+Light that is passing through the clearcoat is then processed as described taking into account differences in IOR.
```
proog128, line 132: > Light that is passing through the clearcoat is then processed as described taking into account differences in IOR. 

This should stay non-normative. But note that the non-normative section in KHR_materials_clearcoat actually defines the opposite. This was discussed several times if I recall correctly. It's a simplification. We do not compute the ratio of clearcoat IOR and base IOR when crossing the clearcoat/base interface.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Theory, Documentation and Implementations
+
+[A Practical Extension to Microfacet Theory for the Modeling of Varying Iridescence](https://belcour.github.io/blog/research/publication/2017/05/01/brdf-thin-film.html)  
+[Arnold for Maya user guide - Thin Film](https://docs.arnoldrenderer.com/display/A5AFMUG/Thin+Film)  
+[Artist Friendly Metallic Fresnel](http://jcgt.org/published/0003/04/03/)  
+[Practical Multilayered Materials in Call of Duty: Infinite Warfare](https://blog.selfshadow.com/publications/s2017-shading-course/drobot/s2017_pbs_multilayered.pdf)  
+[Real-time Image-based Lighting of Microfacet BRDFs with Varying Iridescence](https://cg.cs.uni-bonn.de/en/publications/paper-details/kneiphof-2019-iridescence/)  
+[Real-Time Rendering, Fourth Edition; page 361ff](https://www.realtimerendering.com/)  
+[Rendering Iridescent Objects in Real-time](http://dspace.library.uu.nl/handle/1874/287110)
```
proog128, line 149: Would be nice to add the names of the authors and the year when the article was published. See [here](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#references) for examples.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
abwood: As discussed on Monday's call, we decided that IOR does not need to be texturable. Our opinion is that due to the physical process of how the film is applied via spray, the IOR is unlikely to vary across a material and does not warrant a dedicated texture.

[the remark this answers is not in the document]
abwood: From the [Arnold Renderer Documentation](https://docs.arnoldrenderer.com/display/A5AFMUG/Thin+Film), they choose to use a default IOR of 1.5.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
VanCantus: @proog128 clearcoat is defined in the documentation. For sheen: it should have the same behaviour as with the default specular highlight as it's just a replacement for or a blend between it.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,207 @@
...
+The thin-film layer can have a different IOR than the underlying material. With `iridescenceIOR` one can set an IOR value for the thin-film layer independently. 
+This also has an effect on the optical path difference:
+
+<figure>
+<img src="./figures/ior-comparison.jpg"/>
+<figcaption><em>Comparison of different iridescence IOR values for a thin-film thickness range between 200 nm (bottom) and 800 nm (top).</em></figcaption>
+</figure>
+
+
+The iridescence effect is modeled via a microfacet BRDF with a modified Fresnel reflectance term that accounts for inter-reflections as shown in [Laurent Belcour and Pascal Barla (2017)](https://belcour.github.io/blog/research/publication/2017/05/01/brdf-thin-film.html).
```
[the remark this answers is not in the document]
VanCantus: I added the required implementation details to the "Implementation Notes" section.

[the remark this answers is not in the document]
PascalSchoen: I added the required implementation details to the "Implementation Notes" section.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
PascalSchoen: @proog128 clearcoat is defined in the documentation. For sheen: it should have the same behaviour as with the default specular highlight as it's just a replacement for or a blend between it.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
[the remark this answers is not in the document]
proog128: Thanks!

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
proog128: The default IOR (1.3) is now in a sensible range, so I hope it's not too misleading for readers anymore.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+
+## Implementation Notes
+
+*This section is non-normative.*
+
+For the calculation of `f_specular`, the original Fresnel term `F_schlick` is exchanged with a newly introduced `F_iridescence` term. 
+To artistically control the strength of the iridescence effect, the two Fresnel terms are mixed by using the `iridescenceFactor`:
+```
+F = mix(F_schlick, F_iridescence, iridescenceFactor)
```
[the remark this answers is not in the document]
proog128: This is clear now, thanks!

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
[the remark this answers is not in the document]
proog128: Thanks very much for the detailed implementation section! It's a very nice (practical) reference for implementations.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,207 @@
...
+diff = 2.0 * IOR_iridescence * thin_film_thickness * cos(viewAngle)
+```
+With the path difference, the phase shift can be calculated in Fourier space.
+
+<img src="https://render.githubusercontent.com/render/math?math=\displaystyle%20\Delta%20\phi%20=%202%20\pi%20\lambda^{-1}\text{diff}">
+
+A complete implementation for evaluating `F_iridescence` can be found in the shader file of the Khronos Sample Viewer:
+[iridescence.glsl](https://github.com/PascalSchoen/glTF-Sample-Viewer/blob/feature/KHR_materials_iridescence/source/Renderer/shaders/iridescence.glsl)
+
+### Clearcoat
```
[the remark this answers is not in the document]
proog128: Seems to be done, thanks!

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
proog128: Thanks!

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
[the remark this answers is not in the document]
proog128: Fine for me.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,336 @@
...
+float R21 = R12;
+float T121 = 1.0 - R12;
+
+// Second interface
+vec3 baseIOR = Fresnel0ToIor(baseF0 + 0.0001); // guard against 1.0
+vec3 R1 = IorToFresnel0(baseIOR, iridescenceIOR);
+vec3 R23 = F_Schlick(R1, cosTheta2);
+```
+
+`iridescenceIOR` is the index of refraction of the thin-film layer, `outsideIOR` the IOR outside the thin-film (e.g. air or the clearcoat material) and `baseIOR` the IOR of the base material. The latter is calculated from its F0 value using the inverse of the special normal incidence case of Fresnel's equations:
```
proog128, line 148: > `outsideIOR` the IOR outside the thin-film (e.g. air or the clearcoat material) and `baseIOR` the IOR of the base material

This could be more specific. In which case is the clearcoat IOR used exactly? When `KHR_materials_clearcoat` is present? When `clearcoatFactor > 0`? Always, even if clearcoat is disabled (`clearcoat == 0`)? Maybe the answer should go into the normative part.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,336 @@
...
+*This section is non-normative.*
+
+For the calculation of `f_specular`, the original Fresnel term `F_schlick` (see [B.3.4. Fresnel](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#fresnel) in the glTF 2.0 specification) is exchanged with a newly introduced `F_iridescence` term.
+To artistically control the strength of the iridescence effect, the two Fresnel terms are mixed by using the `iridescenceFactor`:
+
+```glsl
+F = mix(F_schlick, F_iridescence, iridescenceFactor)
+```
+
+The calculation of `F_iridescence` is described in the following sections. For glTF an approximation of the original model (defined in the Mitsuba code example from [Belcour/Barla](https://belcour.github.io/blog/research/publication/2017/05/01/brdf-thin-film.html) is used.
```
proog128, line 129: * There is a closing bracket missing in this paragraph.
* The citation style differs from the one before. Before it's "Laurent Belcour and Pascal Barla (2017)", now it's "Belcour/Barla".
* It seems like at least in the core spec links in the document target the anchors in the reference section. The reference section then includes the external link. I think that makes sense, as the external link is not duplicated many times across the document and can easily be replaced if required.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Material Interfaces
+
+The iridescence model defined by Barla and Belour models two material interfaces - one from the outside to the thin-film layer and another one from the thin-film to the base material. These two interfaces are defined as follows:
```
proog128, line 133: * Typo: Belour -> Belcour
* And another nitpick: Order of Barla and Belcour is reversed here compared to before.

proog128 (changes requested):
@PascalSchoen This is a huge improvement over the last version. Thank you very much!

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
PascalSchoen: Fixed

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
PascalSchoen: Fixed

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
PascalSchoen: @proog128 I updated the documentation quite a bit and added some improved implementation details

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
PascalSchoen: I think that's still up to debate. The current implementation in the sample viewer is, that we blend between an IOR of `1.0` (air) and `1.5` (clearcoat IOR) using the `clearcoatFactor`. That's just a workaround and not physically plausible, but also weighting the clearcoat effect using a factor is not physically based. Another solution for clearcoat would be to ignore it for this extension. Artistically this might be a good approach as it wouldn't mess with an existing iridescence effect when adding clearcoat but it would again contradict the iridescence model.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
proog128: Thanks!

This made me notice a tiny issue here:

```
iridescence_fresnel_lum = luminance(iridescence_fresnel)
return (1 - iridescene_fresnel_lum) * base + iridescence_fresnel * specular_brdf
```

In many cases this is not energy conserving (per RGB channel). To make it energy conserving, you have to use the maximum:

```
iridescence_fresnel_max = max(iridescence_fresnel.r, iridescence_fresnel.g, iridescence_fresnel.b)
return (1 - iridescence_fresnel_max) * base + iridescence_fresnel * specular_brdf
```

PascalSchoen: I also know, that this might not be energy conserving but might/should preserve the perceived luminance. We can also go with the maximum though, which would make the diffuse darker (especially in cases where one RGB component of the iridescence Fresnel is very high).

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -188,7 +300,7 @@ and depends on the first-order optical path difference <img src="https://render.
 OPD = 2.0 * iridescenceIOR * iridescenceThickness * cosTheta1;
 ```
 
-`phi12` and `phi23` define the base phases per interface and are approximated with `0.0` if the IOR of the hit material (`iridescenceIOR` or `baseIOR`) is higher than the IOR of the previous material (`outsideIOR` or `iridescenceIOR`) and `M_PI` otherwise. Also here, polarization is ignored.
+`phi12` and `phi23` define the base phases per interface and are approximated with `0.0` if the IOR of the hit material (`iridescenceIOR` or `baseIOR`) is higher than the IOR of the previous material (`outsideIOR` or `iridescenceIOR`) and <img src="https://render.githubusercontent.com/render/math?math=\color{gray}\displaystyle%20\pi"> otherwise. Also here, polarization is ignored.
```
abwood, line 303: Could we use the UTF-8 character, π, in place of the image.

PascalSchoen: Fixed

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
PascalSchoen: The clearcoat dependency has been removed as discussed in the PBR TSG meeting.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
PascalSchoen: The maximum RGB component of the iridescence Fresnel is used for weighting as discussed in the PBR TSG meeting.

proog128 (approved):

UX3D-nopper:
@proog128 I suggest you add your name to this extension, as you helped a lot in the end. Maybe someone else wants to join as well.

echadwick-wayfair:
Should the Contributors list everyone who comtributed to discussions in the working group calls, like it does for Specular?
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md

UX3D-nopper:
> Should the Contributors list everyone who contributed to discussions in the working group calls, like it does for Specular? https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md

For instance, yes. But we need to ask these folks.

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,429 @@
...
+```
+metal_brdf =
+  iridescent_conductor_layer(
+    iridescence_strength = iridescence,
+    iridescence_thickness = iridescenceThickness,
+    iridescence_ior = iridescenceIOR,
+    outside_ior = 1.0,
+    base_f0 = baseColor,
+    specular_brdf = specular_brdf(
+      α = roughness ^ 2)))
```
abwood, line 122: You caught this during the meetings and I just wanted to add a reminder. We've got an extra parenthesis here 😃.

PascalSchoen: Good catch 😄 Totally forgot about that. It's fixed now. 😉

PascalSchoen: Good catch 😄 Totally forgot about that. It's fixed now. 😉

PascalSchoen (approved):

extensions/2.0/Khronos/KHR_materials_iridescence/README.md
```diff
@@ -0,0 +1,429 @@
...
+* Ben Houston, threekit
+* Mathias Kanzler, UX3D
+* Norbert Nopper, UX3D [@UX3DGpuSoftware](https://twitter.com/UX3DGpuSoftware)
+
+Copyright (C) 2018-2021 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
```
emackey, line 15: The PBR TSG held a poll on yesterday's call, and decided to move this extension from "Draft" to "Release Candidate" status.  This is a pre-ratification step that indicates to early adopters that the group believes the design has stabilized enough for the first set of implementations to be developed.

```suggestion
Release Candidate
```

donmccurdy:
One implementation of this extension (for import, processing, export) has been added to glTF Transform and will be released in v2.1. 

- https://github.com/donmccurdy/glTF-Transform/pull/518

sebavan:
Hello,

I am currently implementing the extension for Babylon and noticed a slight variation from the original 2017 doc about energy conservation:

image

The fact that we would use different fresnel mix is preventing us to simply bake the new values in f0 as Unity does in its implementation for instance.

Also I am not able to find any doc about the energy conservation technique we are implementing here ? I wonder how much we would deviate from the ground truth with it in place ?

sebavan:
I am currently implementing the Babylon extension and noticed a variation from the original 2017 implementation mainly About the energy conservation:
```
function iridescent_fresnel_mix(iridescence_fresnel, base, specular_brdf) {
    // Get maximum component value of iridescence fresnel color
    iridescence_fresnel_max = max(iridescence_fresnel.r, iridescence_fresnel.g, iridescence_fresnel.b)

    return (1 - iridescene_fresnel_max) * base + iridescence_fresnel * specular_brdf
}
```

Is this not preventing to "bake" the irridescent fresnel as a slight modification of the base f0 ? which would introduce quite a different version of the overall shader ?

Also is there any reference about the conservation technique detailed here ? and how much impact they would have on the overall render ?

PascalSchoen:
@sebavan this is defined in the specular extension and adopted by this extension (see [KHR_materials_specular #Implementation](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md#implementation)). 

The issue is, that if we simply use `1 - iridescence_fresnel` for the weight of the `base`, you get inverse colors there. If you don't weight `base` at all, you gain energy. I updated the PR with the same description and a reference to `KHR_materials_specular`.

emackey:
On the call today it was suggested we rename `"iridescenceIOR"` to `"iridescenceIor"` to fix the camelCase.  @lexaknyazev has some additional nonbreaking corrections to some of the description fields, to align them with language used in the core spec.

sebavan:
All good for me I changed it in Babylon and it should be up in the next couple hours :-)

UX3D-nopper:
We can also make the changes in Gestaltor.

UX3D-nopper:
However, did we think about the camel case rule for OpenGL. or OpenXR?
IOR is Index Of Refraction and IOR in short. So, in this case, it should stay `iridescenceIOR`, correct?
Otherwise it should be OpenGl or OpenXr, right?

lexaknyazev:
@UX3D-nopper 
All existing glTF JSON properties follow `lowerCamelCase` convention. The `KHR_materials_ior` extension did not make an exception for its `ior` property despite the latter being an acronym. So for consistency, we should call iridescence IOR as `iridescenceIor`.

UX3D-nopper:
So the rule is like theon one here: https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined
Or do we have a better document for this?

UX3D-kanzler:
I have opened a pull request for the sample models (iridescenceIOR->iridescenceIor): https://github.com/KhronosGroup/glTF-Sample-Models/pull/349

donmccurdy:
PR for gltf-transform: https://github.com/donmccurdy/glTF-Transform/pull/570

echadwick-wayfair:
It would be helpful if the code block in the [Extending Materials section](https://github.com/KhronosGroup/glTF/blob/1ad76e411dbb4892cda68806b038b4b7c9ecfcd7/extensions/2.0/Khronos/KHR_materials_iridescence/README.md#extending-materials) included each of the Properties, to illustrate all their formatting, e.g.:

```
{
    "materials": [
        {
            "extensions": {
                "KHR_materials_iridescence": {
                    "iridescenceFactor": 1.0,
                    "iridescenceTexture": {
                      "index": 0
                    },
                    "iridescenceIor": 1.3,
                    "iridescenceThicknessMinimum": 200.0,
                    "iridescenceThicknessMaximum": 400.0,
                    "iridescenceThicknessTexture": {
                      "index": 0
                    }
                }
            }
        }
    ]
}
```

emackey:
This was ratified by the Khronos Board on July 22, 2022.

PR #2028, opened 2021-09-20
Integrate latest AsciiDoc updates

lexaknyazev:
- Clarified Points and Lines rendering.
- More detailed explanation of sampler wrapping modes.
- Reordered and expanded skinning section.
- Added examples of non-ASCII paths.
- Removed the reference to the external extension document (it will be replaced anyway).
- Added references to GL-originated integer enum values. 
- Minor tweaks, language and typography fixes.

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -2181,7 +2181,7 @@ glTF 2.0 defines only storage of animation keyframes, so this specification does
 [NOTE]
 .Implementation Note
 ====
-glTF 2.0 does not specifically define how an animation will be used when imported but, as a best practice, it is recommended that each animation is self-contained as an action. For example, "Walk" and "Run" animations might each contain multiple channels targeting a model's various bones. The client implementation may choose when to play any of the available animations.
+glTF 2.0 does not specifically define how an animation will be used when imported but, as a best practice, it is recommended that each animation is self-contained as an action. For example, "`Walk`" and "`Run`" animations might each contain multiple channels targeting a model's various bones. The client implementation may choose when to play any of the available animations.
 ====
 
 All animations are stored in the `animations` array of the asset. An animation is defined as a set of channels (the `channels` property) and a set of samplers that specify accessors with key frame data and interpolation method (the `samplers` property).
@@ -2666,7 +2666,7 @@ This chapter presents the bidirectional scattering distribution function (BRDF) 
 [NOTE]
 .Note
 ====
-See <<Pharr2016,Pharr et al. (2016), Chapter 5.6 "Surface Reflection">>, for an introduction to radiometry and the BRDF.
+See <<Pharr2016,Pharr et al. (2016), Chapter 5.6 "`Surface Reflection`">>, for an introduction to radiometry and the BRDF.
 ====
 
 The BRDF of the metallic-roughness material is a linear interpolation of a metallic BRDF and a dielectric BRDF. The BRDFs share the parameters for roughness and base color. The blending factor `metallic` describes the metalness of the material.
@@ -2700,7 +2700,7 @@ Metallic surfaces reflect most illumination, only a small portion of the light i
 [NOTE]
 .Note
 ====
-See <<Pharr2016,Pharr et al. (2016), Chapter 8.2 "Specular Reflection and Transmission">>.
+See <<Pharr2016,Pharr et al. (2016), Chapter 8.2 "`Specular Reflection and Transmission`">>.
 ====
 
 This effect is described by the Fresnel term `conductor_fresnel` with the wavelength-dependent refractive index and extinction coefficient. To make parameterization simple, the metallic-roughness material combines the two quantities into a single, user-defined color value `baseColor` that defines the reflection color at normal incidence, also referred to as `f0`. The reflection color at grazing incidence is called `f90`. It is set to `1.0` because the grazing angle reflectance for any material approaches pure white in the limit. The conductor Fresnel term modulates the contribution of a specular BRDF parameterized by the `roughness` parameter.
@@ -2723,7 +2723,7 @@ Unlike metals, dielectric materials transmit most of the incident illumination i
 [NOTE]
 .Note
 ====
-See <<Pharr2016,Pharr et al. (2016), Chapter 8.2 "Specular Reflection and Transmission">>.
+See <<Pharr2016,Pharr et al. (2016), Chapter 8.2 "`Specular Reflection and Transmission`">>.
 ====
 
 This makes dielectrics like glass, oil, water or air transparent. Other dielectrics, like most plastic materials, are filled with particles that absorb or scatter most or all of the transmitted light, reducing the transparency and giving the surface its colorful appearance.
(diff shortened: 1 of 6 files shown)


specification/2.0/Specification.adoc
```diff
@@ -1340,25 +1361,48 @@ A significant number of authoring and client implementations associate names wit
...
+
+The joint hierarchy used for controlling skinned mesh pose is simply the node hierarchy, with each node designated as a _joint_ by a reference from the `skin.joints` array. Each skin's joints **MUST** have a common parent node (direct or indirect) called _common root_, which may or may not be a joint node itself. When a skin is referenced by a node within a scene, the common root **MUST** belong to the same scene.
+
+[NOTE]
+.Implementation Note
+====
+A node object does not specify whether it is a joint. Client implementations may need to traverse the `skins` array first, marking each joint node.
+====
+
+A joint node **MAY** have regular nodes attached to it, even a complete node sub graph with meshes.
```
bghgary, line 1395: nit: What does "regular" mean?
```suggestion
A joint node **MAY** have other nodes attached to it, even a complete node sub graph with meshes.
```

specification/2.0/Specification.adoc
```diff
@@ -1340,25 +1361,48 @@ A significant number of authoring and client implementations associate names wit
...
+
+[NOTE]
+.Implementation Note
+====
+It's often used to have an entire geometry attached to a joint node without having it being skinned (e.g., a sword attached to a hand). Note that the node transform is the local transform of the node relative to the joint, like any other node in the glTF node hierarchy as described in the <<transformations,Transformations>> section.
```
bghgary, line 1400: nit:
```suggestion
It's common to have an entire geometry attached to a joint node without having it being skinned (e.g., a sword attached to a hand). Note that the node transform is the local transform of the node relative to the joint, like any other node in the glTF node hierarchy as described in the <<transformations,Transformations>> section.
```

specification/2.0/Specification.adoc
```diff
@@ -1340,25 +1361,48 @@ A significant number of authoring and client implementations associate names wit
...
+====
+
+To apply skinning, a regular transformation matrix is computed for each joint and then used to transform individual vertices as described in the next section.
```
bghgary, line 1403: I'm not sure I understand the purpose of this statement. What is the ideal behind adding this?

specification/2.0/Specification.adoc
```diff
@@ -1405,7 +1448,11 @@ In the example below, the translation of `node_0` and the scale of `node_1` are
 [[skinned-mesh-attributes]]
 ==== Skinned Mesh Attributes
 
-The mesh for a skin is defined with vertex attributes that are used in skinning calculations in the vertex shader. The `JOINTS_0` attribute data contains the indices of the joints from the corresponding `skin.joints` array that affect the vertex. The `WEIGHTS_0` attribute data defines the weights indicating how strongly the joint influences the vertex. The following mesh skin defines `JOINTS_0` and `WEIGHTS_0` vertex attributes for a triangle mesh primitive:
+The skinned mesh **MUST** have vertex attributes that are used in skinning calculations. The `JOINTS_0` attribute data contains the indices of the joints from the corresponding `skin.joints` array that affect the vertex. The `WEIGHTS_0` attribute data defines the weights indicating how strongly the joint influences the vertex.
+
+The per-vertex transformation matrix is computed as a weighted linear sum of the joint transformation matrices. Note that per-joint inverse bind matrices (when present) **MUST** be applied before the regular node transforms.
```
bghgary, line 1453: I'm not sure about the usage of "regular". Would "base" be better?

bghgary (commented):
Reviewed the skinning section.

specification/2.0/Specification.adoc
lexaknyazev: The older version of the spec didn't mention what to do with the joint matrices at all; it just had a link to the overview image that is outdated now.

specification/2.0/Specification.adoc
lexaknyazev: Yeah, I certainly overused the word "regular" in this section.

specification/2.0/Specification.adoc
bghgary: Ok, does this line need to be here if it's used only in the next section?

specification/2.0/schema/animation.channel.target.schema.json
```diff
@@ -10,7 +10,7 @@
             "description": "The index of the node to target."
         },
         "path": {
-            "description": "The name of the node's TRS property to modify, or the \"weights\" of the Morph Targets it instantiates. For the \"translation\" property, the values that are provided by the sampler are the translation along the x, y, and z axes. For the \"rotation\" property, the values are a quaternion in the order (x, y, z, w), where w is the scalar. For the \"scale\" property, the values are the scaling factors along the x, y, and z axes.",
+            "description": "The name of the node's TRS property to modify, or the `\"weights\"` of the Morph Targets it instantiates. For the `\"translation\"` property, the values that are provided by the sampler are the translation along the X, Y, and Z axes. For the `\"rotation\"` property, the values are a quaternion in the order (x, y, z, w), where w is the scalar. For the `\"scale\"` property, the values are the scaling factors along the X, Y, and Z axes.",
```
emackey, line 13: I don't recall if we are still using backtics for code-quoting in any other JSON schema descriptions.  When I was first working on the VSCode extension, these descriptions were processed as plain text, not markdown.  I don't think that's changed, but I haven't checked in some years since then.  If it works in the wetzel output though, that might be all that's needed.

lexaknyazev: That is needed for the wetzel output to be consistent. We can look into VSCode UI later.

specification/2.0/Specification.adoc
```diff
@@ -568,7 +568,7 @@ Chromaticity coordinates define the interpretation of each primary color channel
 [[scenes-overview]]
 === Overview
 
-glTF 2.0 assets **MAY** contain zero or more _scenes_, the set of visual objects to render. Scenes are defined in a `scenes` array. All nodes listed in `scene.nodes` array **MUST** be root nodes (see the next section for details).
+glTF 2.0 assets **MAY** contain zero or more _scenes_, the set of visual objects to render. Scenes are defined in a `scenes` array. All nodes listed in `scene.nodes` array **MUST** be root nodes, i.e., they **MUST NOT** be listed in a `node.children` array of any node.
```
emackey, line 571: Can the same root node appear in more than one scene?

specification/2.0/Specification.adoc
```diff
@@ -1995,9 +2040,9 @@ This material does not emit light and will be black unless some lighting is pres
 This specification does not define size and style of non-triangular primitives (such as _points_ or _lines_), and applications **MAY** use various techniques to render these primitives as appropriate. However, the following conventions are **RECOMMENDED** for consistency:
```
emackey, line 2040: > This specification does not define size and style...

I think "and" here should be "or".

emackey:
Yes, everything in this PR looks good to me.

lexaknyazev:
> Can `TEXCOORD_0` be used as a morph target attribute? 

The spec says:
> Client implementations **SHOULD** support at least three attributes -- `POSITION`, `NORMAL`, and `TANGENT` -- for morphing.

I think it's safe to say that it does not forbid such usage given that later the spec mentions morphing custom attributes. So let's mention `TEXCOORD_n` and `COLOR_n` explicitly.

WDYT?

emackey:
> I think it's safe to say that it does not forbid such usage given that later the spec mentions morphing custom attributes. So let's mention `TEXCOORD_n` and `COLOR_n` explicitly.

@bghgary or @donmccurdy Do you know if this would cause heartache for Babylon or ThreeJS?

lexaknyazev:
The missing spec part is the entries in the table with allowed accessor types. A similar fix is also needed for the `KHR_mesh_quantization` extension that extends the table.

Despite all these updates, I still see this as a bug fix because of the "at least" language.

donmccurdy:
three.js will currently ignore morph attributes other than POSITION and NORMAL, so adding more (like `TEXCOORD_n` or `COLOR_n`) shouldn't be any problem. I agree that the spec has been clear that this was a possibility. It would certainly be interesting to have workflows for morphing other attributes.

Issue #2013, opened 2021-08-11
In the 2.0 Specification Are Morph Targets going to contain the same Morphs OR can they contain different amounts within a Primitive??


genusistimelord:
Like for example if the first target is 

```
{
    "NORMAL": 33,
    "POSITION": 32,
    "TANGENT": 34
},
```
then will all other targets in the array be exactly the same Or can they contain anything within the Primitive?

lexaknyazev:
Since a morph target contains attribute deltas, any omitted attributes should keep their base values provided by the `primitives[i].attributes` object.

Although the spec does not explicitly disallow per-target differences, it would good to ensure that implementations are fine with that.

@donmccurdy @bghgary @emackey

bghgary:
Babylon.js implement should allow this. Will need a test model to be sure.

genusistimelord:
Thank you for your Quick Reply. This has been very helpful.

I hope you will be able to Add this to the Read me Specification so others can know this =]!

donmccurdy:
three.js:

- morph POSITION is required
- morph NORMAL is optional, and may be disabled when limited vertex attributes are available
- morph TANGENT is ignored entirely

I'm not sure what would happen if a morph target provided normals but not positions, this is probably not supported. morphing other attributes (e.g. color or UVs) is technically possible, but I don't believe any implementation exists.

Is there a real use case for morphing normals while not also morphing positions? or do you have in mind to provide normals for some targets but not others? I'm not sure what the clarification should be in this case.

emackey:
Re-reading the original post in this issue, perhaps a concrete example would shed some light:

```
"primitives" : [
                {
                    "attributes" : {
                        "POSITION" : 0,
                        "NORMAL" : 1
                    },
                    "indices" : 3,
                    "targets" : [
                        {
                            "POSITION" : 4,
                            "NORMAL" : 5
                        },
                        {
                            "POSITION" : 6,
                        },
                        {
                            "POSITION" : 7,
                            "NORMAL" : 8
                        },
```

I did some quick testing with the `MorphStressTest` sample model, removing `NORMAL` from some of the morph targets in the second primitive.  The results are disastrous in several engines.

I recommend we disallow this, by saying that each of the morph targets in a given primitive MUST have the same attributes as the first.

lexaknyazev:
[...]

PR #2030, opened 2021-09-22
Clarify morphed attribute formats & cleanup indexed semantics

lexaknyazev:


lexaknyazev:
While we're here, is there anything to clarify wrt #2013?

specification/2.0/Specification.adoc
```diff
@@ -1164,31 +1164,28 @@ The following example defines a mesh containing one indexed triangle set primiti
...
-                                  _unsigned short_ normalized | UV texture coordinates for the second set
-| `COLOR_0`   | `"VEC3"` +
-                `"VEC4"`        | _float_ +
+                                  _unsigned short_ normalized | UV texture coordinates
+| `COLOR_n`   | VEC3 +
+                VEC4            | _float_ +
                                   _unsigned byte_ normalized +
                                   _unsigned short_ normalized | RGB or RGBA vertex color linear multiplier
-| `JOINTS_0`  | `"VEC4"`        | _unsigned byte_
+| `JOINTS_n`  | VEC4            | _unsigned byte_
```
emackey, line 1186: `JOINTS_n` are integers, IDs of particular bones in the rig.  You can't morph from bone 3 to bone 5; who knows if bone 4 is anything related to 3 or 5.

specification/2.0/Specification.adoc
```diff
@@ -1164,31 +1164,28 @@ The following example defines a mesh containing one indexed triangle set primiti
...
                                   _unsigned short_            | See <<skinned-mesh-attributes,Skinned Mesh Attributes>>
-| `WEIGHTS_0` | `"VEC4"`        | _float_ +
+| `WEIGHTS_n` | VEC4            | _float_ +
```
emackey, line 1188: Morphing `WEIGHTS_n` is somewhat more doable than morphing IDs.  But the weights are _required_ to sum to 1.0, even during intermediate, interpolated states.  I can't quite see how this could work in practice.  Also given that the IDs need to remain fixed, morphing their weights is of very limited usefulness.

I think we should disallow both `JOINTS` and `WEIGHTS` in morphing.

emackey (changes requested):
I'm pretty sure we cannot morph a skinning rig.

specification/2.0/Specification.adoc
lexaknyazev: This table describes the base mesh accessors, not morph targets.

specification/2.0/Specification.adoc
lexaknyazev: Ditto.

emackey (approved):
Thanks @lexaknyazev.  I was looking at the diff view and lacking that context.

Issue #1806, opened 2020-05-03
Camera perspective yfov upper limit


lexaknyazev:
_Originally posted by @vpenades [here](https://github.com/KhronosGroup/glTF-Validator/issues/140)._
> The official specification states that the value must be over 0, and although it does not explicitly state the upper limit, it states that the value is defined as radians, which sets an implicit upper limit.
>
> Also, the spec says the value must be used in formulas using the form `Tan(0.5*yfov)` which implicitly sets the upper limit below π.
>
> I guess the spec does not define the upper limit because it's a non rational number? but math libraries certainly define it...

I think it's a spec overlook. Given that camera properties go straight into pre-defined equations, there's no reason to allow values greater than or equal to π. We have similar restriction for punctual spot lights.

@donmccurdy @bghgary 
Do you think it's possible to change the schema? At the very least, we need to update the description.

bghgary:
> Do you think it's possible to change the schema?

Not normally. What happens when the values are greater than π?

lexaknyazev:
Reportedly, some math libraries do not like such values. Given that `yfov` is used as `tan(yfov / 2)`, the following happens:
- `[0]` - undefined (already forbidden)
- `(0, π)` - normal behavior
- `[π]` - undefined
- `(π, 2π)` - flipped frustrum
- `[2π]` - undefined
- `(2π, 3π)` - normal behavior
- (continue forever)

![image](https://user-images.githubusercontent.com/3479527/80995253-083d0a00-8e4f-11ea-899b-7eeda82f55b3.png)

bghgary:
I think we can update the description and add checks for validator. Schema changes seem too dangerous at this point (though I don't see anything specific will break).

lexaknyazev:
@donmccurdy @pjcozzi @emackey 
Any opinions here?

vpenades:
given that most math libraries work in 32bit floating point, and fov ranges very close to π don't make much sense, what about setting the upper limit to (float32)π - (float32)epsilon ?

lexaknyazev:
@vpenades 
That would be an implementation detail - some platforms (such as ECMAScript) may still use float64. I wouldn't expect consistent rendering results with any `yfov` values close to π as the final results would be unstable anyway. Values such as `1.5 * π` are likely a bigger issue.

The spec question here is whether we're able to update the schema, or just the description.

donmccurdy:
No preference on this.

[...]

PR #2031, opened 2021-09-22
Use draft/2020-12 schemas

lexaknyazev:
The new JSON schemas are used only in the Appendix A for now. Changes:
- Switched `draft-04` to `draft/2020-12`
- Added `$id`
- `"enum": [ <value> ]` -> `"const": <value>`
- URIs use `iri-reference` format.
- `minimum: number` + `exclusiveMinimum: bool` -> `exclusiveMinimum: number`

Updated the description of `camera.perspective.yfov` (in both places), see #1806.


javagl:
<sup>Sorry for not reviewing and commenting quickly enough on some of the recent PRs. It's sometimes hard to allocate the time for a "proper" review on a short term notice, although I know that the pace is currently very high...</sup>

It's hard to stay up to date with the draft versions of the JSON schema (I hope that they manage to settle for a "JSON Schema 1.0.0-final" at some point...). But from looking at http://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values , I think that `enum` should still be valid, and I wonder why the change of `"enum": [ <value> ]` -> `"const": <value>` was done. 

The way how enums are specified in the current glTF schema underwent some iterations (also because of the JSON Schema Draft changes). It's hard to make this consistent for arbitrary types of enums. And by "types", I mean the difference between `integer` and `string`, but also whether there is a "closed" set of valid values (like `perspective`/`orthographic` for camera), or an "open" set of valid values (like the image MIME type). 

(Is there even a sensible way to explicitly differentiate between the two? I think the semantics of `anyOf` also changed between the draft versions, and our current use is rather a "workaround" for some of the aforementioned difficulties of modeling enums cleanly...).

---

The change of

- `minimum: number` + `exclusiveMinimum: bool` -> `exclusiveMinimum: number`

is also a bit surprising. I think that the original structure using a `number` and a `bool` was a nice way of modeling the different aspects independently:

- The minimum value
- Whether the minimum value is exclusive or not

When there is *only* a `minimum` **or** a `exclusiveMinimum`, then this has some implications that I consider a bit undesirable:

- It changes the types. This will affect downstream pipelines (code generators), likely causing all sorts of incompatibilities...
- It raises the question of how to handle the case that `minimum` **AND** `exclusiveMinimum` are given...
- From quickly (!) looking over the diff, there seems to be a small inconsistency, with using [`exclusiveMinimum` as a `number` here](https://github.com/KhronosGroup/glTF/pull/2031/files#diff-7c288190b7450c9c62c46d1fb3698a5d784e0e9b6ef4bc894babf813edbe7349R20) and (still) using [`exclusiveMinimum` as a `bool` here](https://github.com/KhronosGroup/glTF/pull/2031/files#diff-8f5b436eac4d94a29216bacbd95e442106fa532727964c70de96180a3c2a73b8R19)

Of course, the latter could trivially be fixed. But I'd be in favor to *not* change the type of this field, but to keep the separate `number`+`bool` fields instead. I might be convinced otherwise, but ... only with some effort, because I think that there are relatively  _strong_ reasons to _not_ change this, unless it is strictly necessary...

lexaknyazev:
- Our `anyOf` workaround was implemented to add descriptions to integer enum values. `const` is basically a syntactic sugar for `enum` with only one element.
- The `exclusiveMinimum` update, which happened in `draft-06`, is described as:
  > changed from a boolean to a number to be consistent with the principle of keyword independence
  Of course, both properties cannot be present at the same time anmore.
- There is no inconsistency between the mentioned links as they point to different schema versions (old `draft-04` files are still in place).

javagl:
As usual, when I'm "surprised" by something that you do, it's an error or lack of understanding on my side. Thanks for the clarification 👍

PR #2032, opened 2021-09-23
Use generated version info

lexaknyazev:


oddhack:
It looks fine from an asciidoc point of view - is it generating the results you need?

Also, can we sync up on the release plan to the registry? We should get #1901 merged to master, the default branch changed, and the PDF / HTML artifacts committed to glTF-Registry before 6 AM Pacific time. I'm not sure if you are going to take care of all of that yourself, or would like some assistance from me.

oddhack (approved):

PR #2033, opened 2021-09-23
Adding a KHR_animation2 extension

UX3D-nopper:


extensions/2.0/Khronos/KHR_animation2/schema/animation.channel.schema.json
```diff
@@ -0,0 +1,21 @@
...
+    "properties": {
+        "sampler": {
+            "allOf": [ { "$ref": "glTFid.schema.json" } ],
+            "description": "The index of a sampler in this animation used to compute the value for the target.",
+            "gltf_detailedDescription": "The index of a sampler in this animation used to compute the value for the target, e.g., a node's translation, rotation, or scale (TRS)."
+        },
+        "target": {
+            "allOf": [ { "$ref": "animation.channel.target.schema.json" } ],
+            "description": "The property as a JSON pointer to target."
+        },
```
lexaknyazev, line 16: The `target` object could be replaced with the `pointer` property - there's nothing else there anyway.

extensions/2.0/Khronos/KHR_animation2/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Compatibility
+
+Tools e.g. like [Blender](https://www.blender.org/) do follow the same approach:  
+For each channel/property, one can have an individual animation curve.  
+In this apprach, the cases are reduced, where a complete or partial sampling is required.  
+
+### Minimum data set
+
+For e.g. a translation, all three values have to be stored. If the translation is only changed e.g. for the x value, the y and z values always have to be stored as well.
```
lexaknyazev, line 31: When all three components are affected by the animation, there will be three independent channels. The extension should allow interleaving output accessors for better memory management.

extensions/2.0/Khronos/KHR_animation2/schema/animation.channel.schema.json
UX3D-nopper: Yes, can then kick out the other schema.

extensions/2.0/Khronos/KHR_animation2/README.md
UX3D-nopper: Hmm, I have to think about this, but I do not see, why this should not be allowed? Isn't the bufferview with the byteStride solving this?

lexaknyazev: The core spec allows buffer views with `byteStride` only for vertex attributes. The extension should simply say that its accessors are allowed to be interleaved as well.

lexaknyazev:
@UX3D-nopper 
Logistical suggestion. Could you please rebase (and then force-push in your fork) the `ux3d:extensions/KHR_animation2` branch so that this PR does not contain 30 unrelated commits? Thanks.

UX3D-nopper:
> @UX3D-nopper
> Logistical suggestion. Could you please rebase (and then force-push in your fork) the `ux3d:extensions/KHR_animation2` branch so that this PR does not contain 30 unrelated commits? Thanks.

Will cleanup everything before the merge, that it is technically one commit. But let's see where the discussion goes first.

bhouston:
When I implemented the three.js animation system as well as the Clara.io system I thought about this design.  In the end I came up with a few more track types: https://github.com/mrdoob/three.js/tree/dev/src/animation/tracks

* Quaternion - because quaternions need to be interpolated wholistically.
* String - this is because they are not linearly interpolated, rather step-based.
* Boolean - for visibility and similar, again because it isn't linearly, rather step-based.

That said, I did you the idea of tracks that has string-based regex-like system that point to their targets, which I called a PropertyBinding: https://github.com/mrdoob/three.js/blob/dev/src/animation/PropertyBinding.js

UX3D-nopper:
> When I implemented the [three.js animation system](https://github.com/mrdoob/three.js/issues/6881) as well as the Clara.io system I thought about this design. In the end I came up with a few more track types: https://github.com/mrdoob/three.js/tree/dev/src/animation/tracks
> 
> * Quaternion - because quaternions need to be interpolated wholistically.
> * String - this is because they are not linearly interpolated, rather step-based.
> * Boolean - for visibility and similar, again because it isn't linearly, rather step-based.
> 
> But in the current situations, this is likely not needed for the current purposes.
> 
> That said, I did, like your proposal, have the idea of tracks use JSON pointers, which is similar to the string-based regex-like system that Three.js settled on: https://threejs.org/docs/#api/en/animation/PropertyBinding.parseTrackName

Regarding quaternion, isn't it you first calculate the "left" and "right" keyframe quaternion and do then the slerp?
I think it is like this e.g. in Blender.

scurest:
Blender cannot do slerp for exactly this reason: the curve for each component is evaluated independently. See eg. [this thread](https://devtalk.blender.org/t/quaternion-interpolation/15883).

UX3D-nopper:
> Blender cannot do slerp for exactly this reason: the curve for each component is evaluated independently. See eg. [this thread](https://devtalk.blender.org/t/quaternion-interpolation/15883).

Okay, thank you. Hmmm, then I have to rethink my proposal.

UX3D-nopper:
Seems my suggestion would be a step back regardign the one channel thingy. Will revisit it.

UX3D-nopper:
This extensions could be integrated as well:
https://github.com/KhronosGroup/glTF/pull/2018

UX3D-nopper:
See also here:
https://github.com/KhronosGroup/glTF/pull/1301/files

UX3D-nopper:
Updated the spec and tried to combine https://github.com/KhronosGroup/glTF/pull/1301 and https://github.com/KhronosGroup/glTF/pull/2018

jbherdman:
I have some concerns about the repurposing of the existing `animation.channel.target.path` property.  Is this going to work properly for older parsers that are unaware of this extension's existence?

My own basic use case would be to write out transform-animation as per the main spec, and then use this extension only for animation of additional properties (such as camera-fov).  This would ideally allow `KHR_animation2` to be a "used" extension rather than a "required" extension, and maximize backwards compatibility with older loaders.

I sense that this is maybe not an issue, in relation to the official JSON schema, but that doesn't quite line up with the "Allowed values" language used in the main spec.  Hopefully someone more knowledgeable can chime in with a quick answer one way or the other.

UX3D-nopper:
Yes

> I have some concerns about the repurposing of the existing `animation.channel.target.path` property. Is this going to work properly for older parsers that are unaware of this extension's existence?
> 
> My own basic use case would be to write out transform-animation as per the main spec, and then use this extension only for animation of additional properties (such as camera-fov). This would ideally allow `KHR_animation2` to be a "used" extension rather than a "required" extension, and maximize backwards compatibility with older loaders.
> 
> I sense that this is maybe not an issue, in relation to the official JSON schema, but that doesn't quite line up with the "Allowed values" language used in the main spec. Hopefully someone more knowledgeable can chime in with a quick answer one way or the other.

Yes, you are right. Let me fix this.

extensions/2.0/Khronos/KHR_animation2/README.md
```diff
@@ -0,0 +1,187 @@
...
+|`"/cameras/{}/perspective/znear"`                     |`"SCALAR"`   |`5126`&nbsp;(FLOAT)|Distance to the near clipping plane                           |
+|`"/materials/{}/pbrMetallicRoughness/baseColorFactor"`|`"VEC4"`     |`5126`&nbsp;(FLOAT)<br>`5120`&nbsp;(BYTE)&nbsp;normalized<br>`5121`&nbsp;(UNSIGNED_BYTE)&nbsp;normalized<br>`5122`&nbsp;(SHORT)&nbsp;normalized<br>`5123`&nbsp;(UNSIGNED_SHORT)&nbsp;normalized|The material's base color factor|
+|`"/materials/{}/pbrMetallicRoughness/metallicFactor"` |`"SCALAR"`   |`5126`&nbsp;(FLOAT)|The metalness of the material                                 |
+|`"/materials/{}/pbrMetallicRoughness/roughnessFactor"`|`"SCALAR"`   |`5126`&nbsp;(FLOAT)|The roughness of the material                                 |
+|`"/materials/{}/alphaCutoff"`                         |`"SCALAR"`   |`5126`&nbsp;(FLOAT)|The alpha cutoff value of the material                        |
+|`"/materials/{}/emissiveFactor"`                      |`"SCALAR"`   |`5126`&nbsp;(FLOAT)|The emissive color of the material                            |
```
pierremoreau, line 75: Shouldn't this one be a `"VEC3"`, as `Material.emissiveFactor` is a `number [3]`?

UX3D-nopper: Yes, correct. I just fixed it. Thank you.

pierremoreau: Thank you for the extension. I'm looking forward to using it on emissive values.

hybridherbst:
@UX3D-nopper I'm quite interested in this effort as we've been creating procedural animations in GLB format that actually also have color animations, having those as part of the file would be most welcome. I assume you're planning to (or have already?) added something like this to Gestaltor? Would be happy to provide complex test files.

UX3D-nopper:
> @UX3D-nopper I'm quite interested in this effort as we've been creating procedural animations in GLB format that actually also have color animations, having those as part of the file would be most welcome. I assume you're planning to (or have already?) added something like this to Gestaltor? Would be happy to provide complex test files.

Last Wednesday, we introduced this extension to the 3D Formats Group at Khronos.
So we hope, that this extension is accepted and ratified soon.
As soon as this happens, yes, we will add it to Gestaltor as well.

jjcasmar:
@UX3D-nopper What is the difference between EXT_property_animation and KHR_animation2?

UX3D-nopper:
> @UX3D-nopper What is the difference between EXT_property_animation and KHR_animation2?

I used EXT_property_animation as a base with some modifications and created KHR_animation2.
This should become the ratified Khronos extension. EXT would "just" be a vendor extension.

jjcasmar:
At KDAB we have a user extension hook for exporting EXT_property_animation in Blender. Does it make sense to update it to KHR_animation2 and contribute to glTF-Blender-IO?

UX3D-nopper:
> At KDAB we have a user extension hook for exporting EXT_property_animation in Blender. Does it make sense to update it to KHR_animation2 and contribute to glTF-Blender-IO?

In a first step, please wait, that we can be sure that this extension will be accepted. In a second step, yes please contribute to the I/O project for Blender.

jjcasmar:
@UX3D-nopper  Is there any progress regarding ratification of this extension?

UX3D-nopper:
> @UX3D-nopper Is there any progress regarding ratification of this extension?

During the last weeks and holiday season, there was no Khronos meeting. So, no much progress on this extension.

hybridherbst:
@UX3D-nopper would you mind giving a status update here? 
Is there any particular areas you'd like to get feedback on? Thanks!

UX3D-nopper:
I would really love to continue on this extension. However, other extensions in the queue currently have more focus.

extensions/2.0/Khronos/KHR_animation2/README.md
```diff
@@ -0,0 +1,187 @@
...
+|`"/materials/{}/emissiveFactor"`                      |`"VEC3"`     |`5126`&nbsp;(FLOAT)|The emissive color of the material                            |
+|`"/materials/{}/normalTexture/scale"`                 |`"SCALAR"`   |`5126`&nbsp;(FLOAT)|Multiplier applied to each normal vector of the normal texture|
+|`"/materials/{}/occlusionTexture/strength"`           |`"SCALAR"`   |`5126`&nbsp;(FLOAT)|Multiplier controlling the amount of occlusion applied        |
+
+##### `extras`
+
+A channel may target `extras` at any allowed JSON Pointer value but
+interpretation of the animated values is entirely application specific.
+
+## Extension compatibility and fallback behavior
```
hybridherbst, line 84: This needs some clarification: can extension properties be animated, similar to how it's specified that anything in `extras` can be animated? My understanding would be that, yes, these can be animated.

Additionally, it's not clearly specified here if implementation-specific values can be animated that end up on the nodes – an example would be animating `visibility` in three.js. Nodes in three.js have that property, thus binding it would succeed, and it could thus be animated in an implementation-specific way.

Should extensions specify which values are _supposed_ to be animated?

extensions/2.0/Khronos/KHR_animation2/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+Current and future extensions have to write against this specification to allow to animate specific values.
+
+#### Current and future restrictions
+
+The calculated values can be out of range of the minimum and maximum value, however the values have to be clamped using the properties min and max value before setting the value.
+
+##### Not animatable properties
+
+`glTFid`
```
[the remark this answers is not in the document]
hybridherbst: On further thought, I think in some cases it would actually be interesting to animate `glTFid`, but would still exclude it here:

- stop-motion like effects, where a mesh is switched to another one during a frame
  - workaround: scaling of nodes, like now
- image sprite effects, where a texture is swapped out to another one during animation
  - workaround: texture sheet + UV animation which is reasonable

extensions/2.0/Khronos/KHR_animation2/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension is based on the animation features of glTF 2.0. The structure of the schemas stay the same.  
+  
+The only major addition is, that the output values are mapped using a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901).
```
hybridherbst, line 30: I'd propose adding a snippet from the referenced spec here that I think is imporant:
> Evaluation of a JSON Pointer begins with a reference to the root
   value of a JSON document and completes with a reference to some value
   within the document.

The key point here being that one cannot move "upwards" in the path, a thing that I wondered about when reading this proposal - would it allow animating objects that are parallel or outside to the glTF scene? - with the answer from the spec being "no, needs to be within the document".

extensions/2.0/Khronos/KHR_animation2/README.md
```diff
@@ -0,0 +1,187 @@
...
+* Color factors in materials
+* Camera field of view
+
+Even using a JSON pointer, the targets and their expected behavior are explictly defined.
+
+Current and future extensions have to write against this specification to allow to animate specific values.
+
+#### Current and future restrictions
+
+The calculated values can be out of range of the minimum and maximum value, however the values have to be clamped using the properties min and max value before setting the value.
```
hybridherbst, line 46: I'm not sure I understand this part – should min and max values reflect the full range of new animation data? (I think yes) - why can they be out of range then?

hybridherbst (commented):
We took a stab at implementing this extension in UnityGltf (export) and three.js (import) to test out some ideas. The feedback here is based on open questions from that.

I think what would be valuable is if the extension would specify a "minimum set of guaranteed properties that can be animated when this extension is present". Would that be the list in "Valid target templates"? If so, would be great if the extension would make it clear that
- these are required (list might need to be amended a bit, e.g. lights are not included)
- other valid JSON Pointer references might work but that's implementation-specific

extensions/2.0/Khronos/KHR_animation2/README.md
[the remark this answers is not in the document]
donmccurdy: +1 for omitting `glTFid`, I'm worried it would be wildly complex to support in such an open-ended way. Consider: adding/removing lights means all materials must recompile, some engines must pre-compile materials, and it's impossible to predict which animations might play at the same time. 

The variant-based extensions are a good alternative for these cases.

donmccurdy:
> I think what would be valuable is if the extension would specify a "minimum set of guaranteed properties that can be animated when this extension is present".

If a minimum set were defined, I'd suggest it be much shorter than the "valid targets" list. The list is there to put an upper bound on the implementation complexity – that upper bound is still quite high. In an ideal world with perfect foresight, this extension would have been part of base glTF 2.0 and new extensions (like punctual lights) would simply opt-in to which of their properties could be animated. I see this extension's goal as setting that baseline now. It would be nice if existing extensions (e.g. KHR_lights_punctual) could be grandfathered in somehow without entirely new extensions.

extensions/2.0/Khronos/KHR_animation2/README.md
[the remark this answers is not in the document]
UX3D-nopper: That's the reason I tried to define the expclicit list what is allowed to be animated.

extensions/2.0/Khronos/KHR_animation2/README.md
donmccurdy: I would also clarify whether "before writing back" and "before setting the value" are referring to authoring tool or client behavior. I think you mean client behavior?

Not objecting to that suggestion in the extension if it's generally preferred, but we realistically can't/won't implement this clamping in three.js. Interpolators are optimized mathematical functions, not containers of state tied to a particular file format. If three.js itself supports a value (e.g. `emissiveFactor = [5.0, 0.0, 0.0]`) I can't see us imposing an additional clamp when the material or animation came from glTF as opposed to another format.

UX3D-nopper: Yes, I was meaning client behaviour. In your case, you are moving the problem to the authoring tool. In some cases, it has to sample the data.
Furthermore, this approach is from FBX.
Hmm, maybe we add this feature to the animation itself? That there explicitly one can define, that the animation is clamped in a specific range.

extensions/2.0/Khronos/KHR_animation2/README.md
donmccurdy: Agreed with @UX3D-nopper, if you find yourself keyframing `extras` it's probably time to consider a vendor extension instead.

I'd prefer that future extensions specify which properties can be animated (i.e. an "opt-in" list).

extensions/2.0/Khronos/KHR_animation2/README.md
donmccurdy: We are just talking about cases where CUBICSPLINE in/out tangents might send a value outside the "intended" range, correct? I don't think I can see why this is more of an issue for bounded values like color than for unbounding values like scale or morph weights – it's still the artist and/or authoring tool's responsibility to set the curves. 

Would it be enough to say, something like:

> Keyframe values are limited to the defined range of the animated property. When interpolation yields results individual frame values outside this range, results are undefined.

extensions/2.0/Khronos/KHR_animation2/README.md
hybridherbst: Just for my understanding: for example animating visibility or anything in extras would either 
- require a vendor extension and custom export/import implementations or
-  require a new extension that controls visibility that specifies that the property can be animated?

hybridherbst:
Would this mean that for every "animation extension - existing other extension" pair, there needs to be a separate "other extension _ animation" extension to cover which values are animated? Or would the goal be to retrospectively state for each other extensions which values can be animated?

Alternative would be to have this extension here specify something like 
> "if a viewer supports KHR_texture_transforms and KHR_animation2, the following properties can be animated:"
for the current set of KHR extensions.

Otherwise it feels like this is running into undefined behaviour pretty much immediately... for example, animating texture transforms is what I was most looking forward to from an animation extension. But now contrary to my understanding it doesn't sound like that is covered at all.

Another example would be animating emission strength: I think it would just be natural that viewers who support both KHR_animation2 and KHR_emissive_strength support animating emmisive strength, spec or not. If that isn't written anywhere it's just "random/undefined" though and nobody can rely on it.

(side note for reference, USD simply allows animating everything... not sure if that was looked into as part of this proposal)

donmccurdy:
> Would this mean that for every "animation extension - existing other extension" pair, there needs to be a separate "other extension _ animation" extension to cover which values are animated?...

No — or at least I'm hoping not. Adding an "animation" appendix to ratified extensions does not seem crazy to me, nor does adding an appendix to this extension clarifying behavior for all previously-ratified extensions. Future extensions should probably specify animation-ready properties, perhaps just as a general statement and not specifically bound to `KHR_animation2`.

> USD simply allows animating everything...

USD permits a lot of things that will not work outside its own runtime; it isn't a portable runtime format in that sense. It is focused on not losing an ounce of authoring data, as an interchange format understandably would be. 

In my opinion, it's better to set some upper bounds on complexity for clients. Animation that requires restructuring a glTF file or recompiling shaders is probably a bad thing, and the work of avoiding this should fall more on authoring workflows than on clients.

That said, the lack of a practical "update" path for glTF extensions is a particular obstacle in the case of animation. The burden of defining a list of animatable properies perfectly, on the first try, is not a desirable design constraint. 😕

UX3D-nopper:
> > Would this mean that for every "animation extension - existing other extension" pair, there needs to be a separate "other extension _ animation" extension to cover which values are animated?...
> 
> No — or at least I'm hoping not. Adding an "animation" appendix to ratified extensions does not seem crazy to me, nor does adding an appendix to this extension clarifying behavior for all previously-ratified extensions. Future extensions should probably specify animation-ready properties, perhaps just as a general statement and not specifically bound to `KHR_animation2`.
> 
> > USD simply allows animating everything...
> 
> USD permits a lot of things that will not work outside its own runtime; it isn't a portable runtime format in that sense. It is focused on not losing an ounce of authoring data, as an interchange format understandably would be.
> 
> In my opinion, it's better to set some upper bounds on complexity for clients. Animation that requires restructuring a glTF file or recompiling shaders is probably a bad thing, and the work of avoiding this should fall more on authoring workflows than on clients.
> 
> That said, the lack of a practical "update" path for glTF extensions is a particular obstacle in the case of animation. The burden of defining a list of animatable properies perfectly, on the first try, is not a desirable design constraint. 😕

Hmm, maybe we add all current KHR_* extensions and mention, what property is animatable or not.
Any future extension has to mention, if KHR_animation2 is supported, which prperty can be animated.

hybridherbst:
An example of how I think this could look:

**[KHR_lights_punctual](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_lights_punctual)**
| Property | Animatable |
|:-----------------------|:------------------------------------------|
| `name` | ❌ |
| `color` | ✅ |
| `intensity` | ✅ |
| `type` | ❌ |
| `range` | ✅ |
| `spot.innerConeAngle` | ✅ |
| `spot.outerConeAngle` | ✅ |

**[KHR_texture_transform](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_texture_transform)**
| Property | Animatable |
|:-----------------------|:------------------------------------------|
| `offset` | ✅ |
| `rotation` | ✅ |
| `scale` | ✅ |
| `texCoord` | ❌ |

**[KHR_draco_mesh_compression](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_draco_mesh_compression)**
❌ No animatable properties.

**[KHR_materials_clearcoat](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_clearcoat)**
| Property | Animatable |
|:-----------------------|:------------------------------------------|
| `clearcoatFactor` | ✅ |
| `clearcoatTexture` | ❌ |
| `clearcoatRoughnessFactor` | ✅ |
| `clearcoatRoughnessTexture` | ❌ |
| `clearcoatNormalTexture` | ❌ |

**[KHR_materials_ior](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_ior)**
| Property | Animatable |
| `ior` | ✅ |

@UX3D-nopper I could offer to complete the list above if that's helpful?

UX3D-nopper:
> An example of how I think this could look:
> 
> **[KHR_lights_punctual](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_lights_punctual)**
> 
> Property	Animatable
> `name`	❌
> `color`	✅
> `intensity`	✅
> `type`	❌
> `range`	✅
> `spot.innerConeAngle`	✅
> `spot.outerConeAngle`	✅
> **[KHR_texture_transform](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_texture_transform)**
> 
> Property	Animatable
> `offset`	✅
> `rotation`	✅
> `scale`	✅
> `texCoord`	❌
> **[KHR_draco_mesh_compression](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_draco_mesh_compression)** ❌ No animatable properties.
> 
> **[KHR_materials_clearcoat](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_clearcoat)**
> 
> Property	Animatable
> `clearcoatFactor`	✅
> `clearcoatTexture`	❌
> `clearcoatRoughnessFactor`	✅
> `clearcoatRoughnessTexture`	❌
> `clearcoatNormalTexture`	❌
> **[KHR_materials_ior](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_ior)**
> 
> Property	Animatable
> `ior`	✅
> @UX3D-nopper I could offer to complete the list above if that's helpful?

Please and do not forget to add your name!

donmccurdy:
I think a list like that would be a good thing — wherever it ends up being located. The choices above look very reasonable to me. Thanks!

UX3D-nopper:
I mean the goal is to get this forward and that things have beend dicusses and prepared. If this is final, you never know.

hybridherbst:
Think I got them all: https://github.com/ux3d/glTF/pull/17

UX3D-nopper:
> Two questions, and I apologize if either of these have already been covered here:
> 
> 1. Why the name `KHR_animation2`?  Could it be `KHR_property_animation`?
> 2. Does this animation extension allow any new behaviors with animating translate/rotate/scale?  The glTF format has a solid history of NOT allowing multiple ways to set the same setting, and I wonder, are we violating that here?  Doing so creates fragmentation, and bloats engines that have already implemented the classic animation system, so it may be worth excluding T/R/S from this new animation system and instead providing some kind of linkage to the classic one where needed.

Regarding 1:
animation2 - to express a new approach to address any animation property.

Regarding 2:
In the current spec, there is this description of the node:
"When undefined, the animated object **MAY** be defined by an extension."
https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/animation.channel.target.schema.json#L11
This fact of the specification is used to define this extension and to be backward comaptible:
https://github.com/KhronosGroup/glTF/blob/a70e8b5e7055d4d48219311362e1596d438c9390/extensions/2.0/Khronos/KHR_animation2/README.md#extension-compatibility-and-fallback-behavior

emackey:
Suggestions from today's call:

- Rename to `KHR_animations_properties`
- Add a new enum `pointer`

```
        "target" : {
            "path" : "pointer"
            "extensions": {
                "KHR_animations_properties" : {
                    "pointer" : "/materials/0/pbrMetallicRoughness/baseColorFactor"
                }
            }
        }
```

UX3D-nopper:
Thank u

lexaknyazev:
> `KHR_animations_properties `

The material extensions are using `materials` (plural) for historical reasons but I think we agreed to use singular going forward.

The suggestion was `KHR_animation_pointer`. Isn't it more precise wrt the extension functionality?

emackey:
Ah, OK, `KHR_animation_pointer` is good.

javagl:
The [`animation.sampler.schema.json` documentation of `LINEAR`](https://github.com/KhronosGroup/glTF/blob/82506c076d64d00612f077c05f78aaf4e55f151c/specification/2.0/schema/animation.sampler.schema.json#L21) currently says

> When targeting a rotation, spherical linear interpolation (slerp) \*\*SHOULD\*\* be used to interpolate quaternions.

I just looked at the latest README.md, and apparently, node rotations are (right now) the only case where a _quaternion_ is animated. But I assume that the slerp-interpolation should always be used when interpolating a quaternion. This might warrant adding a hint in the README.md as well, maybe somewhere near https://github.com/KhronosGroup/glTF/blob/a70e8b5e7055d4d48219311362e1596d438c9390/extensions/2.0/Khronos/KHR_animation2/README.md#json-pointer, which currently says

> For a `scalar` value, the JSON Pointer targets the glTF property.
> This means, the property is replaced by the interpolated value.
> 
> For any other case, the JSON Pointer targets the glTF property as well but must be an `array` property.
> This means, that the two or more elements do replace the values in the array.

and could be extended with 

> The type of the interpolation is determined by the animation sampler. If the sampler defines a `LINEAR` interpolation mode, and the interpolated value is a qaternion, then a spherical linear interpolation (slerp) \*\*SHOULD\*\* be used to compute the interpolated value.

(If this is considered to be relevant, I can also put it into a PR)

---

An aside: I'm still a bit surprised that setting/adding `"path": "pointer"` is supposed to be valid. Right now, this will cause a validation warning like
```
"code": "VALUE_NOT_IN_LIST",
"message": "Invalid value 'pointer'. Valid values are ('translation', 'rotation', 'scale', 'weights').",
"severity": 1,
"pointer": "/animations/0/channels/0/target/path"
```
This is related to https://github.com/KhronosGroup/glTF-Validator/issues/172 (where the `KHR_animation_pointer` extension might to "hook" in, to extend the set of allowed values), and raises some other low-level schema questions, but may not have to be discussed here.

emackey:
> An aside: I'm still a bit surprised that setting/adding "path": "pointer" is supposed to be valid. Right now, this will cause a validation warning

I can confirm that adding enum values does not violate JSON schema, because of the foresight of including the catch-all `string` as the last in the list.  I remember having this discussion back in 2016~2017 in the early days of VSCode's JSON schema validation integration, and future extended enum values were explicitly the reason this was added back in the day.

As for the validator's current warning, note that it is indeed just a warning (`"severity": 1`) and not an error (`"severity": 0`).  The validator can be updated to relax this when the animations extension is present.

However, I do think it wise to require this extension be listed in both `extensionsUsed` and `extensionsRequired`, similar to when the webp and/or Draco extensions are used without fallback.  Implementations will indeed require a new enum value in this field to handle the new animation, and old implementations are not guaranteed to safely ignore it (although they can and probably should be updated to ignore unknown values in all of glTF's enums, which all allow future values).

donmccurdy:
>  Implementations will indeed require a new enum value in this field to handle the new animation... 

The current specification also includes another carve-out for extensions: `When "node" isn’t defined, channel SHOULD be ignored.` This might give us some room to avoid `extensionsRequired` here.

A bit of a semantic question about how `extensionsRequired` should be interpreted by clients — I expect few implementations can practically support animation of _all_ available properties defined here. Under what circumstances are clients expected to refuse to attempt loading, when the extension is given in `extensionsRequired`?

emackey:
The `extensionsRequired` list is just a basic "My loader can parse this" requirement, not "My viewer has 100% coverage."  If they've added the `pointer` enum, then their loader can parse it, and they're good to accept it in this list.  I do like your idea about avoiding even this requirement though.

bghgary:
> Just to confirm: I assume that it is not intended to add "pointer" as a new enum value in the core spec, right?

This has to be right. We can't modify the core spec. That said, I'm not sure if it's possible to augment the schema for this enum (e.g., such that vscode will be able to show them).

emackey:
> That said, I'm not sure if it's possible to augment the schema for this enum (e.g., such that vscode will be able to show them).

Correct, it cannot go in core schema, only the extension.

VSCode has always needed a transformed copy of the glTF schema, for a number of reasons including enum/extended descriptions and direct hookup of various extension schemas into core.  There's a [script](https://github.com/AnalyticalGraphicsInc/gltf-vscode/tree/master/util#importallsh) that automates this process.  Once we see some implementations embrace this extension, I'll probably update that script to inject the new enum value, making it available to VSCode autocomplete/tooltip users.

emackey:
Continued in https://github.com/KhronosGroup/glTF/pull/2147

PR #2035, opened 2021-09-27
Refine occlusion texture description

lexaknyazev:
See https://github.com/KhronosGroup/glTF/issues/915#issuecomment-772593047.

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -1944,7 +1944,7 @@ The texture binding for normal textures **MAY** additionally contain a scalar `s
 +
 Normal vectors **MUST** be normalized before being used in lighting equations. When scaling is used, vector normalization happens after scaling.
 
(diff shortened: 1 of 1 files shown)


MiiBond:
We might also consider changing the term "ambient sources" to something else. Really, it's any non-punctual source that needs to be considered. This includes not just IBL but also area lights or light-emitting geometry. Any source where a fraction of the light could be occluded from hitting the fragment.

Maybe just "light sources"? Or "non-punctual light sources"?

donmccurdy:
> Maybe just "light sources"? 

I'd consider "ambient lights" to be a form of light source, so this might be confusing.

> Or "non-punctual light sources"?

Wouldn't an area light be a non-punctual source?

Perhaps "...receive less lighting from indirect sources"?

emackey (approved):
I think "ambient sources" is a nice tie-in to the name "ambient occlusion."  Given the timing (this is due later this week), I think we should keep the scope small here, and the one-word fix proposed in this PR should do it.  We can revisit in a later revision for deeper changes.

bghgary (approved):

Issue #2036, opened 2021-09-28
Bring back the spec readme


technik:
I just realized the spec has moved from the old readme to https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html.
While I suppose the idea is to unify the documentation with other Khronos specs, having everything directly in the repo was very convenient and easy to navigate (and contribute to). Also, Github's formatting is a lot better, and rendered the spec very easy to read. Would it be possible to bring back the readme.md as it was before? (even if the doc also exists in the Khronos registry).

technik:
I spoke too soon, the spec is still available in the repo. It's just one more click away now, here: https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/Specification.adoc. Maybe adding a link to that in the old readme.md would make it easier for other people to find the new doc, as it is not obvious if you're not used to using AsciiDoc.

javagl:
+1 for that, I also looked at the base directory yesterday, until I remembered "Oh, wait, GitHub does render adoc files...". If nobody is opposed, I'd extend the README.md with a hint like 

> "The AsciiDoc markup source file can also be viewn in GitHub, by opening the `(link)Specification.adoc`"

lexaknyazev:
@technik 
FYI: GitHub formatting is missing the auto-generated Properties Reference section and some math expressions (namely projection matrices and PBR equations).

@javagl 
Please open a PR.

emackey:
> Also, Github's formatting is a lot better, and rendered the spec very easy to read.

Maybe some style tweaks are in order?

- Our new output uses a Serif font for most text, but GitHub uses sans-serif throughout.
- We don't have any margins, even on the widest monitors.

javagl:
I have opened a PR with a short hint. I included a disclaimer about the missing references and math renderings, because otherwise (and maybe regardless of that...) people will file issues for that.

donmccurdy:
Would it be confusing (and/or a lot of trouble) to provide a fully-rendered version of just the `.adoc`-equivalent content somewhere? I understand that the spec hasn't actually become any more complex (the clarified normative language is great!), but — to me, at first glance — the new format feels overwhelming in a way that the previous markdown document somehow didn't. It's a 184-page PDF, in which 2/3rds of the pages (and 2/7ths of the TOC) are machine-generated, broken down as:

- 1. 1/3rd written content
- 2. 1/3rd property references
- 3. 1/3rd JSON schema files

[...]

PR #2037, opened 2021-09-28
Added hint about preview of AsciiDoc file

javagl:
Addressing https://github.com/KhronosGroup/glTF/issues/2036 . For a preview of the updated readme, see https://github.com/KhronosGroup/glTF/blob/b4484b3b01272ddfb170dbadb64976c9c784db1b/specification/2.0/README.md

### DIFF
--- a/specification/2.0/README.md
+++ b/specification/2.0/README.md
@@ -8,3 +8,8 @@ to AsciiDoc markup. Published versions of the Specification are now located
 in the [glTF Registry](https://www.khronos.org/registry/glTF). The
 underlying AsciiDoc markup is in the `main` branch of the [glTF
 Specification Repository](https://www.github.com/KhronosGroup/glTF).
+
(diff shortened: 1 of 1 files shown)


emackey (approved):

technik:
Commit dc1945d7baa6aabc1be4466d9b8d5ea8fc7a03e2 completely eliminated this change. I guess this is a documentation regression?

emackey:
I guess I didn't explain this, but I'd much prefer to make the published spec look as nice or nicer than GitHub's preview, as opposed to directing casual readers at the preview with its broken references and unreadable formulas.

I spent a chunk of time last week upgrading the styling on the built spec, but none of that work is published yet.  It has a nicer font, width control, TOC expanders, etc. in the build outputs now.  Actually I'm not sure what's the schedule for getting that deployed, but I'm eager to see it go live.  My hope is that casual readers won't even glance back at the broken preview after that.

Issue #1860, opened 2020-09-01
[Clarification] [bufferView -> accessor -> primitive attributes] and how to map them to vertex structure


ziocleto:
Hi, sorry if this has already been asked but I couldn't pinpoint it in here!

I'm writing a GLTF exporter based on tinyGLTF, I have a scene with a classic PBR vertex format.
`VertexFormat = { Pos,UV0,Nor,Tan,Col }`
`VBO = [VertexFormat...]`
and I have a VBO (OpenGL in this case) defined as an array with those matching vertex attributed in a classic **Array of Structures** representation.

Now from what I've understood a bufferView and accessor will handle **Structure of Arrays** IE:

`VBO= [
    Pos = [{x,y,z}, ...],
    UV0 = [{x,y}, ...], 
    ...
]`

When I define the mesh primitives I can only set an index to an accessor for the vertex component, POSITION, NORMAL, etc, the accessor can only reference a contiguous chunk of memory defined in the bufferView, I can't just declare all the vertex component at once.

So in order to convert to the VBO vertex format, I will have to do heavy processing in order to reshuffle those arrays into structs. I'm pretty sure having a PBR vertex format it's quite common? How do we serialize it to GPU memory without extra work using **Array of Structures**? 

(Looking into sparse accessors doesn't seem to be the answer)

Cheers,
Dado.

scurest:
You can use AoS layout for vert data. Look at the bufferView.byteStride and accessor.byteOffset properties.

ziocleto:
Hi, 

I can specify byteStride ok, but then accessor.byteOffset should take into account the offset within the stride, but it's not, I'm pretty confused about that. 

IE: (Simple Pos  + UV0 vertex structure, 20 bytes => 12 pos + 8 uv )

```
    "bufferViews": [
        {
            "buffer": 0,
            "byteLength": 76768,
            "byteOffset": 0,
            "byteStride": 20,
            "target": 34962
        }
    ]
```
```
"accessors": [
        {
            "bufferView": 0,
            "byteOffset": 0,
            "componentType": 5126,
            "count": 2399,
            "max": [
                0.961799,
                1.6397,
                0.539252
            ],
            "min": [
                -0.692985,
                0.0992937,
                -0.613282
            ],
            "type": "VEC3"
        },
        {
            "bufferView": 0,
            "byteOffset": 20, **<==========THIS IS WRONG==========**
            "componentType": 5126,
            "count": 2399,
[...]

PR #2038, opened 2021-09-28
Refine accessors introduction

lexaknyazev:
Fixes https://github.com/KhronosGroup/glTF/issues/1860.

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -856,7 +856,9 @@ See <<glb-file-format-specification,GLB File Format Specification>> for details 
 
 All binary data for meshes, skins, and animations is stored in buffers and retrieved via accessors.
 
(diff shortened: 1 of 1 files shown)


emackey (approved):
Seems good.  Did you want to add something strong like `bufferView.byteStride` **MUST** be undefined?

I'm fine with this either way.

javagl:
The case where the `byteStride` **MUST** be undefined (or rather **MUST NOT** be _defined_ ) is covered in https://github.com/KhronosGroup/glTF/blob/b1e60923bbc15a9d0987ad4c14aabe8c6bdc218c/specification/2.0/Specification.adoc#3611-overview :

> When a buffer view is used for vertex attribute data, it **MAY** have a `byteStride` property.  [...] Buffer views with other types of data **MUST NOT** not define `byteStride`

So this is probably fine here.

javagl (approved):

PR #2041, opened 2021-09-28
Adjust CSS styling on HTML output of glTF spec.

emackey:
This adds some CSS rules, which adjust only the HTML output, not the PDF output of the glTF spec.  Somewhat inspired by #2036, this changes the HTML version to look a bit more similar to a GitHub-style preview.  Specifically:

- The main body text uses a sans-serif font (Open Sans, which was already in use in certain inset sections), replacing a serif font.
- For wider displays, the width of the content is constrained, so it forms a more coherent column of text, as opposed to allowing words to cover a widescreen monitor side-to-side.  The PDF version already had similar constraints.
- Added the glTF logo to the HTML version (the PDF already has this too).

This also fixes a bug where section 4.4.2 shared its ID with the document header.

/cc @oddhack


emackey:
Old version:

old screenshot

New version:

new screenshot

emackey:
Build artifacts:  [spec-outputs.zip](https://github.com/KhronosGroup/glTF/suites/3898811912/artifacts/97241406)

donmccurdy:
> For wider displays, the width of the content is constrained

Thanks for this!

donmccurdy:
Copied from https://github.com/KhronosGroup/glTF/issues/2036 but certainly would not need to be addressed in this PR:

***

Would it be confusing (and/or a lot of trouble) to provide a fully-rendered version of just the `.adoc`-equivalent content somewhere? I understand that the spec hasn't actually become any more complex (the clarified normative language is great!), but — to me, at first glance — the new format feels overwhelming in a way that the previous markdown document somehow didn't. It's a 184-page PDF, in which 2/3rds of the pages (and 2/7ths of the TOC) are machine-generated, broken down as:

- 1. 1/3rd written content
- 2. 1/3rd property references
- 3. 1/3rd JSON schema files

Presumably the spec is not complete (in terms of IP) without each of those three, but from a readability perspective I'm finding myself wishing I could hide or collapse (3) and possibly even (2).

Related — when I open the (much longer) COLLADA spec PDF in Firefox, the table of contents is collapsed by default. When opening the glTF spec PDF, it's fully expanded. No idea why that is, does the PDF itself control this? 🤷‍♂️

lexaknyazev:
At the very least, the JSON schema listings and links to them in the ToC should be collapsed by default as these aren't normative anyway.

emackey:
So I found a `[%collapsible]` feature in AsciiDoctor, but it only does individual blocks within the document.  The collapsible section cannot have any section headings within, and cannot collapse part of the TOC.

If we remove all the TOC entries/subheadings inside Appendix A, we could list out all of the JSON schemas in a single collapsible block.  But, there would be no anchor tags for individual schema files in this case (although folks could link to individual files on GitHub itself).  Is that an improvement?

emackey:
Well... I put some expanders on the TOC, manually.  I think it helps.  Artifact:  [spec-outputs.zip](https://github.com/KhronosGroup/glTF/suites/3901274362/artifacts/97329991)

specification/2.0/docinfo.html
```diff
@@ -0,0 +1,129 @@
+<style>
+/*
+ * Additional style rules for the glTF 2.0 Specification.
+ * Copyright (c) 2021 Khronos Group.
```
oddhack, line 4: Should be

> Copyright 2021 The Khronos Group Inc.

"(c)" is meaningless and the actual copyright symbol is redundant with "Copyright". Using full org name.

oddhack:
General comment: I don't remember if we discussed this initially when I was doing the adoc conversion, but there's a semi-standardized khronos.css which is built off the same asciidoctor 'colony' theme that I believe glTF is using, with some minor changes. It uses the Noto sans-serif fonts. In the interests of consistency across our APIs, I recommend switching to that and then layering the still-relevant parts of these changes on top. See the internal 'cross-api/stylesheets' repo, file stylesheets/khronos.css .

specification/2.0/docinfo.html
emackey: Fixed in both docinfo.html and Specification.adoc.

emackey:
Thanks @oddhack.  I do like the khronos.css base a little better than the stock, so I've switched it.  Only minimal changes were needed to the layer added here.

Would Vulkan or other groups want to embrace the widescreen support I've added here?  And/or the TOC expanders?

javagl:
I've seen that you replaced the "tabular text" with actual tables. That certainly makes more sense than the images that I just created, but ... I'll drop them here nevertheless...

padding_mat2_1byte-buffer

padding_mat3_1byte-buffer

padding_mat3_2byte-buffer

The output with tables looks fine, and on a narrow screen, scroll bars will appear (at least in the GitHub preview - not sure what the actual rendered output will do there, but we'll see..)

emackey:
I do like the images.  If we use them, I think we should still split the bottom one into 3 parts, so that it doesn't need to fit that much text horizontally.

Also, on the images, the important text (m00..m22) is all low-contrast, medium-gray on light-gray, and the blank boxes are higher contrast.  Can this be swapped?  The data cells should be higher contrast.

If these are easy enough changes, I'm fine with dropping the tables and pushing the images here instead.

Thanks @javagl!

emackey:
And come to think of it, regardless of the tables or images version, we should also split the middle one (MAT3 1-byte) at the 0x07 mark, leaving a half-row (4 bytes) on a second line.  This will make it all look more consistent.

javagl:
The changes are easy. 

(In order ot create these images, I had inserted some hacky tweaks into the program that I originally created for creating these [data layout images](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Triangle#data-layout) for the simple sample models, and... the new changes are something that I eventually did with Inkscape on the SVG, for what it's worth... but ... they are easy...)


---

With the adjusted colors and the line breaks as requested:

**MAT2, 1-byte components**

padding_mat2_1byte-buffer

**MAT3, 1-byte components**

padding_mat3_1byte-buffer-rows-A

**MAT3, 2-byte components**

padding_mat3_2byte-buffer-rows

---

**But...** speaking of "consistency", I created these alternatives:

**MAT2, 1-byte components**

padding_mat2_1byte-buffer-rows

**MAT3, 1-byte components**

padding_mat3_1byte-buffer-rows-B

**MAT3, 2-byte components**

padding_mat3_2byte-buffer-rows

---

I'd be leaning towards the latter, but will leave the decision to others.

emackey:
Yes, let's go with the second set, it looks more aligned.  Can you add them to the `html-styling` branch here?

javagl:
I hope it's OK (and convenient enough) to open this as a PR https://github.com/KhronosGroup/glTF/pull/2050 into this branch

oddhack:
> Thanks @oddhack. I do like the khronos.css base a little better than the stock, so I've switched it. Only minimal changes were needed to the layer added here.
> 
> Would Vulkan or other groups want to embrace the widescreen support I've added here? And/or the TOC expanders?

I'll drop a tracker into the stylesheet repository and we can look at integrating this into the stylesheet factory as an option, The collapsible ToC might be a good feature for Vulkan, whose ToC is enormous.

I should also move the stylesheets repo to github soon - nothing in it is private and 99% of it is a clone of the asciidoctor factory repo.

emackey:
This is ready for merge, I think it's in a good state.

Build artifact: [spec-outputs.zip](https://github.com/KhronosGroup/glTF/suites/3920404709/artifacts/98022039)

lexaknyazev:
@javagl 
Could you share the source files for those table images?

javagl:
@lexaknyazev The images are now SVG. (They are not "nice" SVGs, because the initial version was not created manually, but I hope that does not matter - and if desired, I can replace them with versions that use actual *text* instead of vector data). The files are editable in Inkscape, for example. If you referred to something else than this "most raw (interchange format) form", please let me know.

emackey:
@lexaknyazev OK to merge?

javagl:
(If the SVGs should be changed to use text instead of curves, just drop me a note. I've tried that out with the first example. If these are individual text blocks, it does not change the file size significantly, because of all the "font style" info that is inserted for each and every string, but it might be possible to merge it into larger text blocks and save a few bytes there...)

lexaknyazev:
OK, let's merge this since it's definitely an improvement. We'll continue updating images to meet certain requirements in another PR.

donmccurdy:
If I understand correctly this is submitted but not deployed yet? Some related feedback/discussion in https://community.khronos.org/t/updated-gltf-2-0-0-specification-now-canonical/107820.

lexaknyazev:
@donmccurdy 
Yes, this will be deployed as 2.0.1 patch release very soon.

lexaknyazev:
The [glTF Registry](https://www.khronos.org/registry/glTF/) has been updated.

Issue #1823, opened 2020-06-07
Spec isn't clear that global node transform should include parents' transforms


aejsmith:
Unless I'm completely missing something, I can't find this explicitly specified in the current spec anywhere. The glTF quick reference guide says:

> The global transform of a node is given by the product of all local transforms on the path from the root to the respective node.

There is nothing along those lines in the spec itself, however.

PR #2042, opened 2021-09-28
Add node global transform definition

lexaknyazev:
Fixes https://github.com/KhronosGroup/glTF/issues/1823.

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -669,6 +669,8 @@ To compose the local transformation matrix, TRS properties **MUST** be converted
 Non-invertible transformations (e.g., scaling one axis to zero) could lead to lighting and/or visibility artifacts.
 ====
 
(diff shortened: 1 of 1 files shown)


emackey (approved):

javagl:
Hm...

> The global transformation matrix of a node is the product of its local transformation matrix and the local transformation matrices of all its parent nodes.

A node has at most _one_ parent. One could rephrase it to "...all its ancestor nodes" to avoid this. But considering the importance of the _order_ of matrix multiplications, and that the statement, when interpreted verbatim, suggests a wrong order (it's `parent.transform * node.transform`, and not `node.transform * parent.transform`), one could consider phrasing that a bit differently. Either as in the overview,

> The global transform of a node is given by the product of all local transforms on the path from the root to the respective node.

or a bit more elaborately

> The global transformation matrix of a node is the product the global transformation matrix of its parent, and its local transformation matrix.  When the node has no parent node, then its global transformation matrix is equal to its local transformation matrix.

(or even "...identical", FWIW...)

PR #2043, opened 2021-09-28
Add preview-mode-only warnings around auto-generated sections.

emackey:
This adds preview-only warning notes to the two auto-generated sections, providing links to the fully-rendered versions, and an instruction about how to suggest changes.

Also the conditional was changed from `ifdef::env-github[]`, which is GitHub-specific, to `ifndef::revdate[]` which is specific to the glTF doc build process.  This means the notes will be visible to many more previewers in addition to GitHub, such as VSCode's adoc preview window.

There should be no change to the final rendered output from this PR.  (I diffed the outputs to check).

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -2135,6 +2135,15 @@ Client implementations **SHOULD** use the following projection matrices.
 [[infinite-perspective-projection]]
 ==== Infinite perspective projection
 
+ifndef::revdate[]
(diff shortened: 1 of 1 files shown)


lexaknyazev (approved):

javagl (approved):
That looks good, and will hopefully prevent false issues.

Should we consider to add a similar disclaimer to the math formulas that are not rendered correcly?

Issue #1646, opened 2019-07-29
The evaluation order of skinning and morph targets


syoyo:
Is there an explicit definition of the evaluation order of skinning and morph targets when the glTF model contain both information?(e.g. skinning then morph, morph then skinning). 

I have glimpsed the specification but couldn't find an explicit description of the evaluation order of skinning and morph.

An expected behavior may be morph then skinning, but sometimes we want skinning then morph for corrective blendshapes https://knowledge.autodesk.com/support/maya/learn-explore/caas/CloudHelp/cloudhelp/2018/ENU/Maya-CharacterAnimation/files/GUID-B8755F24-7C24-4066-B6F8-2DE6954DC2BF-htm.html

donmccurdy:
I know that threejs applies morph before skinning, but don't think the specification mentions this, and it probably should.

> ... but sometimes we want skinning then morph for corrective blendshapes

Unfortunately I don't think there is much ability in engines for toggling this ordering around, so an option in the spec would be pretty much unsupported.

meshula:
Morph / skin / corrective is also desirable, from an art perspective...

donmccurdy:
I won't argue they couldn't be desirable, but does anything that isn't a DCC tool support them? Search results don't seem promising on [Unity](https://forum.unity.com/threads/corrective-blendshapes-in-unity.625468/) or [Unreal](https://answers.unrealengine.com/questions/662260/how-do-you-set-up-corrective-morphs.html), and I don't think we can justify adding features that only DCC tools support.

syoyo:
> I know that threejs applies morph before skinning,

@donmccurdy Thanks! So at least we are better to send PR to mention the ordering of `skin` and `morph` evaluation order.

> Search results don't seem promising on Unity or Unreal 

Hmm...  it needs more detailed design discussion how glTF-next will support complex animation/rigging system if we want to support it as an glTF standard...
(Probably we'll need to introduce modifier stack feature)

For a while, I am planning to use glTF extension to specify the ordering of morph and skin to support skinning-before-morph feature.

meshula:
It might only exist in bespoke AAA engines? Certainly we did morph/skin/corrective in ours. I would argue that Unity and Unreal as an Occam's razor is overly restrictive. At the same time, I appreciate that if other gamedevs don't speak up there's no proof that anyone besides a few high end studios need it. So I'm raising a hand.

donmccurdy:
[...]

PR #2044, opened 2021-09-28
Clarify the morph displacements order

lexaknyazev:
Fixes https://github.com/KhronosGroup/glTF/issues/1646.

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -1340,6 +1340,8 @@ Accessor type and component type for each morphed attribute semantic property **
 
 `POSITION` accessor **MUST** have its `min` and `max` properties defined.
 
+Displacements for `POSITION`, `NORMAL`, and `TANGENT` attributes **MUST** be applied before any transformation matrices affecting the mesh vertices such as skinning or node transforms.
+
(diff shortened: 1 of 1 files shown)


emackey:
Can we make an explicit reference to skinning here?  I had to re-read #1646 and read "any transformation matrices affecting the mesh vertices" about 3 times before I started to take that as meaning that morphs happen before skinning (as opposed to node transforms).  Perhaps I'm being dense, but could we spell it out for folks?

lexaknyazev:
Given that skinning is effectively defined as applying per-vertex transformation matrices (rather than per-mesh), I thought that "any" covers both cases. Spelling that out won't hurt though.

javagl (approved):
It's good to have that explicit.

(And I didn't have that on the radar. I'll try to have a look at existing implementations. A sample model for dedicatedly testing this could be nice...)

emackey:
> A sample model for dedicatedly testing this could be nice...

I never did submit my balloon test model... It looks a little wonky, but it uses morphing to inflate / deflate, and uses skinning to "flop over" when deflated.  It demonstrates well why this order is correct; I'll have to build a readme for it and post it someday.

emackey (approved):

PR #2045, opened 2021-09-28
Improve global transform language

lexaknyazev:
Follow-up as suggested in https://github.com/KhronosGroup/glTF/pull/2042#issuecomment-929602789.


javagl (approved):
I think this is clearer now.

Issue #1637, opened 2019-07-05
MikkTSpace texCoord when using second set


jjcasmar:
Lets say we have a normal map which use the second set of texture coordinates. In that case, the tangent generation should use the first or the second texture coordinate set for the MikkTSpace? This is not specified at all in the spec.

donmccurdy:
I guess I missed this comment, but it's a good point. With the core spec, the preferred answer seems clear — tangents should be generated based on whatever UV set is required by `normalTexture`. With the introduction of `KHR_materials_clearcoat`, we have a trickier problem: the model might have both `normalTexture` and `clearcoatNormalTexture`, and they might use different UVs. The spec makes no allowance for multiple `TANGENT` attributes, to say nothing of whether engines support could support it. I'm curious, now, what Blender is doing when it exports a `TANGENT` attribute for a mesh with many UVs.

Practically speaking, generating MikkTSpace tangents at runtime is nontrivial for WebGL applications (no JS or WASM implementation exists), and requires de-indexing meshes. I think the spec is correct to identify MikkTSpace as the preferred solution when tangents are missing, but there are valid reasons many implementations can't do so, or might choose not to. So, while we could try to introduce `TANGENT_0`, `TANGENT_1`, ... `TANGENT_N` syntax in some future version of the spec, I'm not sure whether that solves a real problem, or just introduces complexity for a hypothetical problem. 😕

PR #2046, opened 2021-09-29
Clarify MikkTSpace texture coordinates

lexaknyazev:
Fixes https://github.com/KhronosGroup/glTF/issues/1637.

The updated language implies that when an extension has its own normal texture, another set of tangents may need to be generated. This does not require indexing tangent attributes because the generation happens only when they are missing completely.

/cc @donmccurdy


emackey:
I do have a lingering concern here, that MikkTSpace itself might implicitly be using the opposite (FlipY) convention for texture V-coordinates compared to glTF.  There's some extended discussion of this in https://github.com/KhronosGroup/glTF-Sample-Models/issues/174.  I don't quite know what the correct phrasing should be, though.

Still, what's written here is an improvement.  +1.

lexaknyazev:
The exact wiring between glTF data and MikkTSpace implementations is not defined here; I think we'll get there eventually.

Issue #1461, opened 2018-10-06
Clarify zero-length tangent vectors


lexaknyazev:
This issue has been brought up several times in different glTF tools repos. In some cases, exporters write tangent vector as zero. 

We need the spec to clarify whether it's allowed and to provide some implementation guidelines for better ecosystem consistency.

emackey:
Also zero-length normal vectors.  The Blender exporter writes zero-length normals for degenerate triangles.

donmccurdy:
I'm OK with disallowing zero-length tangents as well.

gleblebedev:
Not unit size normal vectors could be used to store ambient occlusion information. A normal with a zero length indicates that vertex should not be lit by a light source.

emackey:
@gleblebedev For an individual game engine, tricks like that can be quite useful, but in glTF the goals include broad compatibility across a wide range of platforms and engines.  The spec already states that normals must be "normalized" (meaning unit-length), and I think it's just an oversight that tangents were not already specified that way as well.  Various open-source implementations and sample viewers for glTF will automatically normalize both these vectors to protect against invalid data.

Degenerate geometry should not be included in a glTF.  Unlit geometry should be tagged with the [KHR_materials_unlit](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_unlit) extension.  Ambient occlusion has a texture channel allocated when needed.

Issue #1639, opened 2019-07-08
Normal and tangent vectors must be orthogonal?


emackey:
Does the spec clarify that `NORMAL` and `TANGENT` vectors must be orthogonal?  And does the validator test for this?

Related, there is a question about zero-length tangents in #1461, and it's not clear to me that the spec actually calls for the `xyz` components of the tangent vector to be normalized.  We may need to clarify that those components must be normalized and must be orthogonal to the supplied normal vector (even in case of custom normal vectors that don't exactly match the geometry).

For example, [this sample code](https://github.com/KhronosGroup/glTF-Sample-Viewer/blob/341fd8f907cf49ff2f3662746c7586043628cf0d/src/shaders/primitive.vert#L104-L106) will produce a non-normalized bitangent if the two vectors are not exactly orthogonal.

```
    vec3 normalW = normalize(vec3(u_NormalMatrix * vec4(getNormal().xyz, 0.0)));
    vec3 tangentW = normalize(vec3(u_ModelMatrix * vec4(tangent.xyz, 0.0)));
    vec3 bitangentW = cross(normalW, tangentW) * tangent.w;
    v_TBN = mat3(tangentW, bitangentW, normalW);
```

This is following a glTF Slack conversation with @vpenades.

vpenades:
I believe that the only way to guarantee orthogonal normal and tangent vectors at the pixel shader, is to ensure that the normal and tangents fed to the vertex shader are orthogonal and unit length.

For example, if you have a tangent of unit length at one vertex, and a half length tangent at another vertex, it doesn't matter if the tangent and normal are orthogonal in the vertex shader.... the pixel shader interpolation will probably skew the vectors, so after re-normalization in the pixel shader, they will no longer be orthogonal.

scurest:
If morph targets are being used, even if the normals and tangents stored in each individual target are orthogonal, the final blended normal and tangent won't usually be.

emackey:
Sounds like implementations need to take care of renormalizing as needed.  That can be an implementation detail, and the spec will get tightened just for the glTF data in #2047.

PR #2047, opened 2021-09-29
Update TANGENT attribute description

lexaknyazev:
Fixes https://github.com/KhronosGroup/glTF/issues/1461.


lexaknyazev:
@emackey 
Please see if #1639 should also be addressed here (or just closed).

PR #2050, opened 2021-09-29
Replaced padding tables with images

javagl:


emackey:
Thanks @javagl.  Any chance you could double the output resolution of the image files?  They look a bit pixelated.

lexaknyazev:
Might as well use SVG instead of PNG as these images contain only simple rectangles and text.

javagl:
Of course, SVG would be preferable (I should have looked up whether they can be used directly in AsciiDoc). I'll give that a try and update this PR ASAP.

specification/2.0/Specification.adoc
```diff
@@ -1033,34 +1033,15 @@ Only the following three accessor configurations require padding.
 
 *MAT2, 1-byte components*
 
-|===
-|00|01|02|03|04|05|06|07
-|m00|m10|&#8212;|&#8212;|m01|m11|&#8212;|&#8212;
-|===
+image:figures/padding-mat2-1byte.png[pdfwidth=4in,align=left]
```
emackey, line 1036: As the first 2 images are half the width of the last, I would recommend `pdfwidth=2in` for the smaller images.

javagl: I had searched for that `pdfwidth`, and seen that _all_ images used the same value ... but will update this (together with the attempt to directly use SVG).

emackey: Yeah, I made that comment after seeing the result in the PDF.

In case you don't know this already, since these changes are happening in your own javagl fork, the GitHub actions are running there:

https://github.com/javagl/glTF/actions

Click on the topmost build, and find "spec-outputs" at the bottom of the summary page for the CI.  It will be a ZIP file with the built version of the PDF and HTML versions.

javagl:
I have replaced the PNGs with SVGs, and adjusted the widths of the first two. I assume that `pdfwidth` will have some particular effect when generating the PDF, and I haven't (yet?) set up the toolchain for doing this, so someone might want to try whether the PDF output looks as desired. In the browser, the images now appear at the right size, they can be (browser-)zoomed without pixelating, and when the window is too narrow, then the images are scaled down (but this is hardly possible with the line-wraps, because that would be somewhere below ~300 pixels...)

emackey:
Merging into #2041 (`html-styling`).

javagl:
(In hindsight, the fact that the SVGs don't store the text as text, but as curves, defeats some of the purpose (in terms of file size), but this was a result of auto-generating the first images. If I had to do them from scratch, it would be text, but ... that may not be the best investment of time for now...)

emackey:
> I haven't (yet?) set up the toolchain for doing this

That's what I was trying (badly?) to explain in my previous comment: There's no need to install any local toolchain to see the result in the final PDF.  GitHub Actions, even on the free plan, is building the final HTML and PDF from each commit you push here.  You can click on the "Actions" tab of the appropriate repo (including your own fork of glTF), and download the final output from there.

> the SVGs don't store the text as text, but as curves

In this particular case, the text is just m01 m02... I don't think anyone will need to copy it.

PR #2053, opened 2021-10-01
Clarify camera parameters

### LINKED ISSUE #1663: xmag and ymag should be clarified to be width/2 and heigth/2
Also the tutorial is misleading:
https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_016_Cameras.md

lexaknyazev:
Fixes #1663.

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -2145,6 +2145,14 @@ Client implementations **SHOULD** use the following projection matrices.
 [[infinite-perspective-projection]]
 ==== Infinite perspective projection
 
+Let
+
+- `a` be the aspect ratio (width over height) of the field of view;
+- `y` be the vertical field of view in radians;
+- `n` be the distance to the near clipping plane.
+
+Then, the projection matrix is defined as follows.
+
 ifndef::revdate[]
 [NOTE]
 .Note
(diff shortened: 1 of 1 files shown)


javagl (commented):
The clarification of the aspect ratio is good. A bit more generally, I sometimes wondered whether the `aspectRatio` of the camera or the *actual* aspect ratio of the viewport should be used. There once was a mail thread, where I sent around a test ("Re: [3dformats] Camera scale test model", 30.06.2021, 20:44), including these images:

babylonjs 002 Perspective scaled uniformly

threejs 002 Perspective scaled uniformly

showing that babylon seems to use the `aspectRatio`, whereas three.js seems to use the viewport aspect ratio. 

Given that clarification, one could file an issue in three.js, pointing to this section.

---

Regarding the definition of the variables: This makes sense, but ... is now a bit redundant. They already had been defined _after_ the equations. We might consider merging these bullet point lists, so that we only have _one_ list, with all information, like

- `a` be the aspect ratio (width over height) of the field of view, which is given as `camera.perspective.aspectRatio` or from the aspect ratio of the current viewport
- `y` be the vertical field of view in radians, given by `camera.perspective.yfov` 

... 

But I don't have a strong opinion here...

emackey:
> We might consider merging these bullet point lists, so that we only have _one_ list, with all information

I like this idea.  I'm not opposed to merging as-is, or with that change.

donmccurdy:
three.js [does use the camera's aspect ratio](https://github.com/mrdoob/three.js/blob/b1ede8160b5aefe18c7ec29f2af80f78b0ffd911/examples/js/loaders/GLTFLoader.js#L3136-L3144), but this is only going to give the expected result in a viewer if the viewport also has that same aspect ratio. It rarely does.

> ...client implementations **SHOULD NOT** crop or perform non-uniform scaling ("`stretching`") to fill the viewport.

What is this saying a client _should_ do? scale uniformly and leave part of the viewport unrendered?

lexaknyazev:
> What is this saying a client _should_ do? scale uniformly and leave part of the viewport unrendered?

I think this depends on an application. A simple scene renderer should scale uniformly and leave the extra viewport regions unused (or repurpose them for something else). An application that uses multiple active windows may resize the viewport to match the camera's aspect ratio.

javagl (approved):
I think having the formal definitions and the respective property in one block (thus, _defining_ each variable only once) is nicer.

Issue #1314, opened 2018-04-10
Animation support for 'visibility'


calrk:
I've been working with some models that use the visibility flag in their animations to show/hide meshes within the object. These models have been made in Maya, exported as fbx and tested in Unity with the flag working. This can be set as shown below from Maya:
![image](https://user-images.githubusercontent.com/2245579/38531382-df655974-3cb2-11e8-938d-5c49cd009c29.png)

I read the specification for the animation and it says `The path may be "translation", "rotation" or "scale"...`

I was wondering if it also supported the visibility property or if there were any plans to support it in future?

donmccurdy:
Currently visibility animation cannot be stored in glTF, although you could hack it with discrete/step animations of position or scale. Proper visibility animation would seem like a reasonable addition through https://github.com/KhronosGroup/glTF/pull/1301, but that will likely take some time.

calrk:
Yeah I've suggested using the position or scale, so will enforce that the models follow that. 

Thanks.

ctedin:
Visibility parameter control would be great. Scaling works pretty well though, in a pinch.

spiraloid:
I have a modular asset with lots of parts that need to be toggled on or off.  showing them all at once, even scaled to 0.001 would overfloy the polycount of the scene.  is there a workable alternative?

emackey:
There's no formal extension proposal for visibility yet.  One option is to have the renderer check for scale being non-zero before rendering each node, and skip rendering of any nodes (and their children) when zero scale is encountered.  Child nodes cannot produce any meaningful result at any scale when a parent is zero scale, so may also be skipped.

It seems likely that any future visibility extension would incur the almost same performance cost, adding a test per node at render time.  Perhaps a Boolean test would be marginally faster than three scale axes floating-point tests, but would come at the expense of new code and mechanisms built for the purpose of animating a Boolean value over time.  It seems difficult to justify a whole new category of animation import and export here, when one could get the same information by checking the scale of each node before rendering.

donmccurdy:
I've been thinking of this under a general bucket of material animation, so far. It could plausibly be handled by #1790 (defines variant "states", not timed keyframes), #1301 (arbitrary property animation), or #1889 (procedural graph-based material inputs).

andybak:
Can this issue be reopened? Or can it be closed with a proper reason (i.e. covered by another ticket, wontfix)?

Currently you need to read the entire conversation to understand what decisions have been made (and it's still not clear)

[...]

PR #2059, opened 2021-10-04
Add note about zero-scale meshes.

emackey:
Following conversation from https://github.com/KhronosGroup/glTF/issues/1314#issuecomment-754820704 and subsequent comments, I'd like to add an implementation note encouraging optimizations for zero-scale meshes.

Zero-scale meshes can be used to animate visibility of a mesh, and can enable use-cases such as fluid simulation where the computed mesh changes shape and changes number of vertices over a set of known timesteps.

/cc @donmccurdy @javagl

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -674,7 +674,7 @@ Non-invertible transforms (e.g., scaling one axis to zero) could lead to lightin
 [NOTE]
 .Implementation Note
 ====
-When the scale is zero on all three axes (by node transform or by animated scale), implementations are free to optimize away rendering of the node's un-skinned mesh, and all of the node's children's un-skinned meshes.
+When the scale is zero on all three axes (by node transform or by animated scale), implementations are free to optimize away rendering of the node's mesh, and all of the node's children's meshes. This provides a mechanism to animate visibility. For skinned meshes, this optimization would only be applicable if all joints in the skin were scaled to zero simultaneously.
 ====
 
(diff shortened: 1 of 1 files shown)


lexaknyazev (approved):
I suppose it's just a hint for developers who may skip such an optimization in their scene graph implementations, right?

scurest:
Only applies to unskinned meshes.

lexaknyazev:
Transforms of a skinned mesh are controlled by the joints' transforms that could contain zero scale as well. However, the results may be unstable when only a part of the mesh is affected.

emackey:
Is it enough to just change this to "un-skinned meshes," as I have just pushed here?

lexaknyazev:
> Is it enough to just change this to "un-skinned meshes," as I have just pushed here?

I'd say it's a bit imprecise because applying zero-scale to all joint nodes (or just their root) would lead to the same effect for skinned meshes.

emackey:
How about this:

> When the scale is zero on all three axes (by node transform or by animated scale), implementations are free to optimize away rendering of the node's mesh, and all of the node's children's meshes. This provides a mechanism to animate visibility. For skinned meshes, this optimization would only be applicable if all joints in the skin were scaled to zero simultaneously.

emackey:
Or this?

> When the scale is zero on all three axes (by node transform or by animated scale), implementations are free to optimize away rendering of the node's mesh, and all of the node's children's meshes. This provides a mechanism to animate visibility. Skinned meshes must not use this optimization unless all of the joints in the skin are scaled to zero simultaneously.

lexaknyazev (approved):
> unless all of the joints in the skin are scaled to zero simultaneously

LGTM, assuming that this includes a case when the joints get a zero-scale transform from a root node.

emackey:
Yes, that's the intention.

PR #2060, opened 2021-10-04
Add structure and property naming convention to extensions documentation

donmccurdy:
Putting in writing some of the conventions discussed during `KHR_materials_*` extension development.

We have at least one extension that doesn't quite follow this today, `KHR_materials_unlit`. It's a very simple extension so I think that's harmless, but I do think the convention is a useful one to keep in mind moving forward.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -138,6 +138,14 @@ Extensions start as a vendor extension, then can become a multi-vendor extension
 3. Names SHOULD be structured as `<PREFIX>_<scope>_<feature>`, where *scope* is an existing glTF concept (e.g. mesh, texture, image) and *feature* describes the functionality being added within that scope. This structure is recommended, but not required.
 4. Scope SHOULD be singular (e.g. mesh, texture), except where this would be inconsistent with an existing Khronos extension (e.g. materials, lights).
 
+### Structure
+
+When an extension is attached to a parent property from the core glTF 2.0 specification, properties of the extension should generally be named and defined as though they were attached directly to the parent object. For example, use `KHR_materials_clearcoat.clearcoatTexture` and NOT `KHR_materials_clearcoat.texture`. The extension object does not, in these cases, have implicit meaning of its own. Furthermore, mere presence of such an extension should not typically change the behavior of the parent property: the extension should be "disabled by default" and require at least one property before it has an effect. 
+
+When an extension is attached to the glTF root object (e.g. `KHR_lights_punctual.lights`) its purpose is generally to define a new type of resource for later reuse, and would not be affected by these recommendations.
+
+These patterns help to increase internal consistency among extensions in the glTF ecosystem, as well as increasing the likelihood that the same property names and semantics could be preserved if the extension were later promoted into the core specification.
+
 ## Promoting Extensions
 
 ### Vendor Extensions


extensions/README.md
```diff
@@ -138,6 +138,14 @@ Extensions start as a vendor extension, then can become a multi-vendor extension
 3. Names SHOULD be structured as `<PREFIX>_<scope>_<feature>`, where *scope* is an existing glTF concept (e.g. mesh, texture, image) and *feature* describes the functionality being added within that scope. This structure is recommended, but not required.
 4. Scope SHOULD be singular (e.g. mesh, texture), except where this would be inconsistent with an existing Khronos extension (e.g. materials, lights).
 
+### Structure
+
+When an extension is attached to a parent property from the core glTF 2.0 specification, properties of the extension should generally be named and defined as though they were attached directly to the parent object. For example, use `KHR_materials_clearcoat.clearcoatTexture` and NOT `KHR_materials_clearcoat.texture`. The extension object does not, in these cases, have implicit meaning of its own. Furthermore, mere presence of such an extension should not typically change the behavior of the parent property: the extension should be "disabled by default" and require at least one property before it has an effect.
```
lexaknyazev, line 143: > properties of the extension should generally be named and defined as though they were attached directly to the parent object

I think this is applicable only to extensions that are meant to "permanently" extend the core objects, i.e., as if their properties were always there. An extension that is inherently optional or domain-specific shouldn't have to follow this rule.

> the extension should be "disabled by default" and require at least one property before it has an effect

Nit: one non-default property. Also, this is our current approach for extending PBR MR model and it may be incompatible with other use cases.

donmccurdy: > I think this is applicable only to extensions that are meant to "permanently" extend the core objects, ...

Could you give a more specific counterexample? E.g. clearcoat is inherently optional, but the rule still applies.

Maybe `EXT_mesh_gpu_instancing` is a counterexample: if it were merged into the core spec it would perhaps still be a sub-object under the primitive rather than adding properties directly to the primitive. Similar to `accessor.sparse`. I'm not quite sure how to describe why that case is different though.

> Also, this is our current approach for extending PBR MR model and it may be incompatible with other use cases...

I think it's a good "best practice" in many cases, e.g. consider `KHR_texture_transform` and proposed animation extensions too. But I agree it isn't necessary (or even coherent) in other cases. Maybe a better way of framing this would be, "mere presence of such an extension should not typically change the behavior of the parent property", and leave off the overly-specific details after that?

lexaknyazev: > Could you give a more specific counterexample?

BasisU and WebP extensions both define `source` properties (that we forgot to rename to `image` during glTF 1.0->2.0 upgrade). I don't think that calling them `basisSource` or `webpSource` would be better.

To the second point, an extension that has required properties will be likely incompatible with the "mere presence" part.

donmccurdy: Agreed on the BasisU/WebP example, I don't mean to propose prefixing everything. I think the general rule of "properties of the extension should generally be named and defined as though they were attached directly to the parent object" is still a helpful exercise to think about. In other words, think about the changes you'd need to promote the extension to the glTF spec and design for them. In these examples promotion would mean allowing more image types on the existing "source" property.

> ...an extension that has required properties will be likely incompatible with the "mere presence" part.

I would say an extension with any required properties inherently satisfies the "mere presence" part. If we were re-defining `KHR_materials_unlit` today I think I would want it to have a required `"unlit": true` property.

***

Are there any changes to this wording that would improve this or make it more general? Would you prefer not to add this section?

lexaknyazev: I fully agree with the "helpful exercise to think about" approach. I'm just not convinced that we should generalize domain-specific patterns (images, textures, and materials) to the spec as a whole.

I was referring to the "disabled by default" language since an extension with a schema-required property does not technically have a default state - it's always explicit. In the unlit example, did you mean that `"unlit": false` would be a default state? If so, the property has a default value (false) and thus isn't required.

> Are there any changes to this wording that would improve this or make it more general?

- I'd split this line in two paragraphs.
- For the naming part, we should mention that the prefix rule applies when the new properties are within the domain of the core object and could have been defined directly there. So a clearcoat texture is a natural example because there are other textures already defined on materials. A counter-example is an extension that provides an alternative reference, like with texture sources.
- For the presence part, the "disabled by default" conflicts with "require at least one property" from the JSON-Schema perspective. Changing the latter to something like "require at least one property having non-default value" would solve the issue.

emackey:
While I like the spirit of this, and we're certainly following this design philosophy for PBR extensions, I'm not sure that we should generalize to all extensions like this.

donmccurdy:
Ok – that's fine with me! 👍

PR #2065, opened 2021-10-06
Expand animation sampler interpolation section

### LINKED ISSUE #1395: Must slerping between quaternions for an animation curve always go the short way round?
ie. when interpolating from q1 to q2, should q2 be replaced with whichever of {q2, -q2} is closest to q1?

Please see https://github.com/KhronosGroup/glTF-Sample-Models/issues/185.

The text in the spec is

> When targeting a rotation, spherical linear interpolation (slerp) should be used to interpolate quaternions.

Apparently most implementors have read this to mean that the short way round must be used but this is not implicit in the text.

And what about cubic interpolation?

lexaknyazev:
Provides full reference for all three interpolation modes and also fixes #1395.

/cc @scurest

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -3213,7 +3213,7 @@ v_t = \frac{sin(a(1 - t))}{sin(a)} * v_k + s * \frac{sin(at)}{sin(a)} * v_{k+1}
 [NOTE]
 .Implementation Note
 ====
-Multiplication by the dot product's sign ensures that the spherical interpolation follows the short path.
+Using the dot product's absolute value for computing stem:[a] and multiplying stem:[v_{k+1}] by the dot product's sign ensure that the spherical interpolation follows the short path along the great circle defined by the two quaternions.
 ====
 
 Implementations **MAY** approximate these equations to reach application-specific accuracy and / or performance targets.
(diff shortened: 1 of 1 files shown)


javagl (commented):
The generalization is great. 

I quickly compared the formulas to the pseudocode at https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_007_Animations.md#linear , but the crucial part (i.e. the slerp) was provided by an external contributor, and will probably have to be reviewed. (It contains a `cos` computation, for example, which I don't see in the spec, but it _might_ well be that all this collapses into something that is mathematically equivalent eventually...)


I also compared the cubic interpolation part to the pseudocode at https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_007_Animations.md#cubic-spline-interpolation , and was a bit confused: From a **quick** glance, it seems like

- "in-tangent" from the spec is called `nextTangent` in the pseudocode
- "out-tangent" from the spec is called `previousTangent` in the pseudocode

but that may be a matter of terminology, or related to the fact that `t_d` does not appear in the pseudocode at all (?)

For short: I did not _verify_ the maths in all depth (and in doubt, I'd rather assume that the pseudocode is wrong, and not the spec). But if a more detailed review of the maths was the goal of the review, I can allocate more time for that.

emackey:
> I also compared the cubic interpolation part to the pseudocode

Yes, I was staring at that too, but I think it matches.  `previousOutputTangent` in the tutorial is b<sub>k</sub> here, and `nextInputTangent` is a<sub>k+1</sub> here, if I'm not mistaken.  `deltaTime` is t<sub>d</sub> here and is included.

So... I think this is an exact match?

PR #2067, opened 2021-10-06
Clarify A2C applicability

### LINKED ISSUE #1668: Spec makes no mention of ALPHA_TO_COVERAGE
If a glTF renderer enables ALPHA_TO_COVERGE (e.g. for leaf foliage) then it will fail to render the AlphaBlendModeTest sample model in the expected way.

Does this mean that engines should always disable ALPHA_TO_COVERAGE?  If so, it should be stated in the specification.

More info at:
https://github.com/KhronosGroup/glTF-Sample-Models/issues/221

lexaknyazev:
Fixes #1668.

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -2027,7 +2027,7 @@ The `alphaMode` property defines how the alpha value is interpreted. The alpha v
 `alphaMode` can be one of the following values:
 
 * `OPAQUE` - The rendered output is fully opaque and any alpha value is ignored.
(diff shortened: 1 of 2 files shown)


emackey:
I don't think "Alpha-to-Coverage" needs code-quotes around it.

@prideout Are you still following this topic?  We have a short window for glTF spec tweaks here and we're trying to get the spec to specifically allow this.

lexaknyazev:
> I don't think "Alpha-to-Coverage" needs code-quotes around it.

It's an AsciiDoc trick to get curly quotes in the rendered output.

javagl (commented):
I don't know the technical background of this well enough, but as a clarification for people who knoe these details, this looks fine.

PR #2072, opened 2021-10-11
Final edits for the 2.0.1 spec release

lexaknyazev:
This PR bumps the document version to `2.0.1` and includes the following editorial changes.

### Specification
- Use non-breakable hyphen to keep `SPIR-V™` (in the Foreword) on one line.
- Added missing serial commas and fixed `and/or` spelling.
- All figures are numbered and have titles now.
- Use `sans-serif` as the main typeface, `monospace` for code listings.
  - **Rationale**: Current CSS file uses `Noto` which may be not a proper typeface name (it should probably have been `Noto Sans`). This font is not pre-installed by default, so removing it makes no difference for the most of users. Also, using only the generic font name ensures that the main text and the text embedded in SVG look alike, i.e., they both are using sans-serif or monospace fonts that are default on the user's OS (Arial/Courier on Windows, Helvetica/Menlo on macOS). /cc @oddhack 

### Figure Updates
- Use SVG for the glTF logo on the front page of the rendered PDF.
- Use SVG for the glTF logo in the header of the rendered HTML.
  - The embedded logo (in `docinfo.html`) was additionally optimized for file size with no visual changes.
- The image showing the file types of a glTF asset was removed as it did not contain any extra information compared to the surrounding text.
Yes, since it simply duplicates the information written in the same section.
- Updated the glTF objects hierarchy figure, use SVG for it.
  - The source figure is available in PPTX format; it was exported to SVG and further optimized.
  - The original image showed the multiplicities, but that was never explained.
- Changed the glTF coordinate system orientation figure to use white background.
  - It's a scaled-down screenshot of the `BoomBoxWithAxes` asset opened in the glTF Sample Viewer.
- Replaced the special matrix alignment figures to properly use text in SVG.
  - The source figure is available in PPTX format; it was exported to SVG and further optimized.
- The texture coordinate orientation figure was replaced with a more detailed SVG version.
  - This figure was authored directly in SVG.
- Replaced the PBR material example (Water Bottle) with an SVG figure that has a similar background image.
  - This figure was authored directly in SVG. The background image is embedded.
- Replaced GLB layout figure with an identical SVG version.
  - The source figure is available in PPTX format; it was exported to SVG and further optimized.
- Replaced PBR graph figure with an identical SVG version.
  - The source figure is available in DOT format; it was exported to SVG using GraphViz and further optimized. Specifically, extra `<tspan>` elements were needed to get correct rendering in the PDF output.

### Maintenance
- Use SVG for the glTF logo in the repo root readme.
- Removed all unused files from the `2.0/figures` directory.


javagl (approved):
- The image showing the file types of a glTF asset ( https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/figures/files.png ) in section "2.4. glTF Basics" was removed. (I assume that was intentional, because it is not really an image that "specifies" something).
 
- The original image https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/figures/dictionary-objects.png showed the multiplicities, but that may not be critical here.

PR #2079, opened 2021-10-17
Adding the KHR_animation_controller extension

UX3D-nopper:


bhouston:
In an animation system there are:
(1) tracks which have keyframes.
(2) clips, which are groups of tracks, which have a name and represent a specific action.  It can be a walk cycle, a jump, etc.
(3) controllers/actors/mixer, which is logic that schedules clips and blends them together.  These can be like player pianos where all the clips are scheduled ahead of time, or they can be dynamic controllers which schedule animations based on user inputs or scene dynamics.

This extension is currently just specifying a clip.  It isn't a controller.

Here is the controller I wrote for Three.js - notice it is all about allowing one to schedule in animation clips:

https://threejs.org/docs/#api/en/animation/AnimationMixer

I think that clips are important.  Maybe that is where we should start?  Maybe that is what you intent, but called is a controller instead?

UX3D-nopper:
In the end, I want to have a controller. And, yes, this "lego" brick is more a clip. The logic and controller will be an extensions based on this extension. Will rename it.

UX3D-nopper:
Conitnues here: https://github.com/KhronosGroup/glTF/pull/2080

PR #2080, opened 2021-10-18
Adding an KHR_animation_clip extension

UX3D-nopper:


UX3D-nopper:
We solve the feature with interactivity and a behavior node graph.

PR #2083, opened 2021-10-21
KHR_outputmapping_pq

rsahlin:
The purpose of this extension is to define a method to convert internal high range (float) values to output values that are compatible with and in the same range as thos of the perceptual qantizer.
The intended usecases for this extension is any usecase where the light contribution values will go above 1.0, for instance by using KHR_lights_punctual, KHR_emissive_strength or KHR_environment_lights

### DIFF
--- a/extensions/2.0/Khronos/KHR_outputmapping_pq/README.md
+++ b/extensions/2.0/Khronos/KHR_outputmapping_pq/README.md
@@ -1,0 +1,107 @@
+# KHR_outputmapping_pq
+
+## Contributors
+
+Copyright (C) 2021 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Exclusions
+
+
+## Overview
+
+The goal of this extension is to provide the means to map internal, usually floating point, light contribution values that may be in an unknown range to that of an attached display.  
+Output pixel values from a rendered 3D model are generally in a range that is larger than that of a display device.  
+This may not be a problem if the output is a high definition image format or some other target that has the same range and precision as the internal calculations.  
+However, a typical usecase for realtime rasterizer implementations is that the output is a light emitting display.  
+Such a display rarely has the range and precision of internal calculations making it necessary to map internal pixel values to match the characteristics of the output.  
+This mapping is generally referred to as tone-mapping, a term that we will use here.  
+
+Here the term rasterizer means a rendering engine that consits of a system wherein a buffer containing the pixel values for each frame is prepared. 
+This buffer will be referred to as the framebuffer.  
+The framebuffer can be of varying range, precision and colorspace. This has an impact on the color gamut that can be displayed.  
+  
+After completion of one framebuffer it is output to the display, this is usually done by means of a swap-chain. The details of how the swap works is outside the scope of this extension.  
+KHR_outputmapping_pq specifies one method of mapping internal pixel values to that of the framebuffer.  
+
+## Internal range
+
+When the KHR_outputmapping_pq extension is used all lighting and pixel calculations shall be done using the value 10000 (cd / m2) as the maximum ouput brightness.  
+This does not have an impact on color texture sources since they define values as contribution factor.  
+The value 10000 for an output pixel with full brightness is chosen to be compatible with the Perceptual Quantizer (PQ) used in the SMPTE ST 2084 transfer function.  
+
+When using this extension any light contribution values shall be aligned to account for 10000 cd / m2 as max luminance. This means that content creators shall be aware of 10000 cd/m2 as the maximum brightness value.  
+Values above 10000 shall be clamped.  
+
+
+## Outputmapping
+
+If the framebuffer format and colorspace is known to the implementation then a format and colorspace shall be chosen to preserve the range and precision of the SMPTE ST 2084 transfer function.  
+If the framebuffer format or colorspace is not known, or none is available that preserves range, precision and color gamut then lower range framebuffer with sRGB colorspace may be used.  
+This is to allow for compatibility with displays that does not support higher range and/or compatible colorspaces.  
+It also allows for implementations where the details of the framebuffer is not known or available.  
+
+
+
+## Schema
+
+
+## Appendix: Full Khronos Copyright Statement
+
+Copyright 2021 The Khronos Group Inc.
+
+Some parts of this Specification are purely informative and do not define requirements
+necessary for compliance and so are outside the Scope of this Specification. These
+parts of the Specification are marked as being non-normative, or identified as
+**Implementation Notes**.
+
+Where this Specification includes normative references to external documents, only the
+specifically identified sections and functionality of those external documents are in
+Scope. Requirements defined by external documents not created by Khronos may contain
+contributions from non-members of Khronos not covered by the Khronos Intellectual
+Property Rights Policy.
+
+This specification is protected by copyright laws and contains material proprietary
+to Khronos. Except as described by these terms, it or any components
+may not be reproduced, republished, distributed, transmitted, displayed, broadcast
+or otherwise exploited in any manner without the express prior written permission
+of Khronos.
+
+This specification has been created under the Khronos Intellectual Property Rights
+Policy, which is Attachment A of the Khronos Group Membership Agreement available at
+www.khronos.org/files/member_agreement.pdf. Khronos grants a conditional
+copyright license to use and reproduce the unmodified specification for any purpose,
+without fee or royalty, EXCEPT no licenses to any patent, trademark or other
+intellectual property rights are granted under these terms. Parties desiring to
+implement the specification and make use of Khronos trademarks in relation to that
+implementation, and receive reciprocal patent license protection under the Khronos
+IP Policy must become Adopters and confirm the implementation as conformant under
+the process defined by Khronos for this specification;
+see https://www.khronos.org/adopters.
+
+Khronos makes no, and expressly disclaims any, representations or warranties,
+express or implied, regarding this specification, including, without limitation:
+merchantability, fitness for a particular purpose, non-infringement of any
+intellectual property, correctness, accuracy, completeness, timeliness, and
+reliability. Under no circumstances will Khronos, or any of its Promoters,
+Contributors or Members, or their respective partners, officers, directors,
+employees, agents or representatives be liable for any damages, whether direct,
+indirect, special or consequential damages for lost revenues, lost profits, or
+otherwise, arising from or in connection with these materials.
+
+Khronos® and Vulkan® are registered trademarks, and ANARI™, WebGL™, glTF™, NNEF™, OpenVX™,
+SPIR™, SPIR-V™, SYCL™, OpenVG™ and 3D Commerce™ are trademarks of The Khronos Group Inc.
+OpenXR™ is a trademark owned by The Khronos Group Inc. and is registered as a trademark in
+China, the European Union, Japan and the United Kingdom. OpenCL™ is a trademark of Apple Inc.
+and OpenGL® is a registered trademark and the OpenGL ES™ and OpenGL SC™ logos are trademarks
+of Hewlett Packard Enterprise used under license by Khronos. ASTC is a trademark of
+ARM Holdings PLC. All other product names, trademarks, and/or company names are used solely
+for identification and belong to their respective owners.
--- a/extensions/2.0/Khronos/KHR_outputmapping_pq/README.md
+++ b/extensions/2.0/Khronos/KHR_outputmapping_pq/README.md
@@ -19,6 +19,9 @@ Written against the glTF 2.0 spec.
 ## Overview
 
 The goal of this extension is to provide the means to map internal, usually floating point, light contribution values that may be in an unknown range to that of an attached display.  
+
+The intended usecases for this extension is any usecase where the light contribution values will go above 1.0, for instance by using KHR_lights_punctual, KHR_emissive_strength or KHR_environment_lights.  
+
 Output pixel values from a rendered 3D model are generally in a range that is larger than that of a display device.  
 This may not be a problem if the output is a high definition image format or some other target that has the same range and precision as the internal calculations.  
 However, a typical usecase for realtime rasterizer implementations is that the output is a light emitting display.  
@@ -38,13 +41,17 @@ When the KHR_outputmapping_pq extension is used all lighting and pixel calculati
 This does not have an impact on color texture sources since they define values as contribution factor.  
 The value 10000 for an output pixel with full brightness is chosen to be compatible with the Perceptual Quantizer (PQ) used in the SMPTE ST 2084 transfer function.  
 
-When using this extension any light contribution values shall be aligned to account for 10000 cd / m2 as max luminance. This means that content creators shall be aware of 10000 cd/m2 as the maximum brightness value.  
+When using this extension light contribution values shall be aligned to account for 10000 cd / m2 as max luminance. This means that content creators shall be aware of 10000 cd/m2 as the maximum brightness value.  
 Values above 10000 shall be clamped.  
 
 
 ## Outputmapping
 
+To convert from internal values (wanted luminance on the display) to the non-linear output value in the range 0.0 - 1.0 the PQ EOTF shall be used.  
+
 If the framebuffer format and colorspace is known to the implementation then a format and colorspace shall be chosen to preserve the range and precision of the SMPTE ST 2084 transfer function.  
+If available, a colorspace that is compatible with the color primaries of Rec. 2020 or Rec. 2100 shall be used.  
+
 If the framebuffer format or colorspace is not known, or none is available that preserves range, precision and color gamut then lower range framebuffer with sRGB colorspace may be used.  
 This is to allow for compatibility with displays that does not support higher range and/or compatible colorspaces.  
 It also allows for implementations where the details of the framebuffer is not known or available.  


bhouston:
I see a lot of value in automatic exposure control in order to avoid over bright areas or underbought areas.  Overbright areas tend to have color shift because of the per-channel per-pixel tone mapping operators in use.

I think that we should expect viewers to just use the proper color space for their displays.  I am not sure we should put into a 3D file what its expect transfer function to that display is.

I guess we could put exposures into individual glTF files but that feels a little weird.  What happens if you open up multiple glTF files that have conflicting exposures or tone mapping parameters?

My recommendation may be that this is a viewer certification issue and we come up with guidelines for handling automatic exposure, tone mapping and transfer functions.  Tone mapping and automatic exposure controls and proper transfer functions should be part of viewers, rather than part of the individual glTF files.

I would love to see that happen!

Because to put it into a glTF file we are sort of assuming that viewers are going to be using incorrect, less than optimal OOTF, unless you use this extension.  That is a bit weird.

rsahlin:
To solve the questions of how to handle multiple loaded glTFs with or without this extension:

- Remove the option of specifying rangeExtension and gamma for the OOTF - they will always be the default values.
- The PQ (HDR or SDR) is enabled for rendering in a model that includes this extension in the root element
- Move the sceneAperture setting to the scene object - this makes it possible to set sceneAperture control on a scene level.
If no sceneAperture is explicitly set then no adjustment is done to light contribution before the PQ step.
This also makes it natural to assume that any model that is brought into this scene will be displayed using the same sceneAperture values (as the new model is inserted into an existing scene)

romainguy:
Note that PQ is used in game engines as the transfer function for HDR output (as dictated by HDR10 and Dolby Vision), it is not used as the "tone mapping" curve.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,400 @@
...
+Where this is calculated per RGB component, note that the color value in this equation must be in range 0.0 - 1.0
+It may be estimated that the display will not have the ability to output dark levels in the region of 0.0003024 in which case implementations may ignore that condition and use the same operator for the whole range 0.0 - 1.0  
+
+The incoming color value shall first be adjusted by aperture as detailed in [See Scene aperture](#scene-aperture) 
+
+
+### OETF
+
+Opto-electrical-transfer-function  
+Input signal to the PQ OETF is the scene linear light. The PQ OETF maps scene linear light to non-linear PQ signal value.
```
romainguy, line 230: Note that according to the spec, the OOTF expects values between 0 and 1 (which means values already exposed and maybe tone mapped by the camera).

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,400 @@
...
+The incoming color value shall first be adjusted by aperture as detailed in [See Scene aperture](#scene-aperture) 
+
+
+### OETF
+
+Opto-electrical-transfer-function  
+Input signal to the PQ OETF is the scene linear light. The PQ OETF maps scene linear light to non-linear PQ signal value.  
+For more information see the ITU BT.2100 specification.  
+
+After the OOTF is applied the OETF shall be applied, this will yield a non linear output-signal in the range [0.0 - 1.0] that is stored in the display buffer.
```
romainguy, line 233: I'm not sure why you want the OOTF step? The OOTF computes values in display light, and the OETF, if fed display light values, should simply yield scene light values again, but we already have scene light values in the glTF case.

sobotka:
> This will now become 10000 equals a fully exposed pixel.

This is insanity. Surely other people have pointed out how this is an untenable mess?

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,400 @@
...
+## Displaymapping
+
+This section describes how the mapping of internal, scene linear light, values to display output values in the range [0 - 1.0] shall be done.  
+
+Displaymapping will be done different depending on wether the output is an HDR or SDR capable display.  
+If the output type is unknown then it shall be considered to be SDR, and the SDR ootf shall be used.  
+
+### To HDR capable display
+
+To convert from internal values (linear scene light) to the non-linear output value in the range 0.0 - 1.0 the PQ EOTF shall be used.
```
romainguy, line 139: The EOTF will happen in the display (or display pipeline). The role of the renderer for HDR display is to output PQ encoded data using the OETF.

rsahlin:
> This is insanity. Surely other people have pointed out how this is an untenable mess?

Hi @sobotka and thanks for taking part in this discussion.

Before we go any further I would like to urge you to be respectful and provide constructive comments.
Please have a look at the code of conduct:
https://www.khronos.org/developers/code-of-conduct

Best regards
/Richard

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,400 @@
...
+### Parameters
+
+The following parameters are added by the `KHR_displaymapping_pq` extension, it is added with the extension on a scene object:  
+
+| Name                   | Type       | Description                                    | Required           |
+|------------------------|------------|------------------------------------------------|--------------------|
+| **sceneAperture**   | `object`   | Scene light adjustment setting, if no value supplied the default value of OFF will be used and scene light values will be kept unmodified.    | No |
+
+
+### Scene aperture
```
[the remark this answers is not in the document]
rsahlin: Hi Romain and thanks for your comments

Personally I feel that the meaning of the noun aperture is better suited for what this setting is doing.
Ie, the mechanism is only accounting for an opening that can be of varying size.

The word exposure is in my opinion coupled to shutter speed, ISO and aperture.

I would like to keep using aperture as a way to make a distinction from what could be a coming extension that defines a 'physical camera', with shutterspeed, ISO and aperture parameters.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: Correct.
Perhaps I should make it clear that at this stage all values are expected to be in the 0 - 10000 range, as declared by this extension?

Then, depending on how you choose to implement the OOTF and the OETF, one solution is to simply divide by 1000 before applying the OOTF.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: Good catch - that is simply a typo

sobotka:
> A simplified version can be done by providing parameters to the OOTF - however in order to avoid problems if multiple glTFs are loaded, I have excluded that support.

The OOTF defined in ITU REC. 2390 is not based on open domain tristimulus, and therefore is not applicable.

Perhaps I misread you, so for clarity, you are **not** suggesting that an entire series of open domain tristimulus values are somehow clipped to display colourimetry, are you?

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: I have clarified the wording and added implementation notes

rsahlin:
Hi Troy
I am not familiar with the term 'open domain tristimulus' - could you please elaborate what you are referring to?
Best regards/Richard
    Den torsdag 20 januari 2022 16:35:16 CET, Troy James Sobotka ***@***.***> skrev:  
 
 



A simplified version can be done by providing parameters to the OOTF - however in order to avoid problems if multiple glTFs are loaded, I have excluded that support.


The OOTF defined in ITU REC. 2390 is not based on open domain tristimulus, and therefore is not applicable.

Perhaps I misread you, so for clarity, you are not suggesting that an entire series of open domain tristimulus values are somehow clipped to display colourimetry, are you?

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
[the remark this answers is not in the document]
romainguy: Aperture would conflict more with a potential extension defining aperture/shutter speed/ISO IMO. I've also never seen the word aperture used to describe what's effectively exposure.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
romainguy: I also have issues with the requirement of bounding the scene values to the 0-10,000 range :) And dividing by 10,000 wouldn't work well for all scenes.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
romainguy: The OOTF is OETF->EOTF, I think in your case it "happens" to work as some kind of tone mapper because the OOTF is not symmetrical, but that's not its purpose. You are effectively adding a tone mapper without calling it that.

romainguy:
> Hi @romainguy and thanks for your comments
> 
> > Note that PQ is used in game engines as the transfer function for HDR output (as dictated by HDR10 and Dolby Vision), it is not used as the "tone mapping" curve (that bit is done by the "S curve" in the Lumberyard diagram for instance). PQ is also sometimes used as a log encoding for 3D LUT lookup.
> 
> Sure - at the moment this extension does not supply such a curve. A simplified version can be done by providing parameters to the OOTF - however in order to avoid problems if multiple glTFs are loaded, I have excluded that support. Engines will still have a reasonable way of exposing an API that lets you have some control over this, by exposing the rangeExtension and gamma parameters. However this is not declared by this extension.

It's still misleading to point to engines supporting PQ because they are not using it in the way you are defining here at all. They use PQ because that's the transfer function of Dolby Vision/HDR10 (some also use PQ as a log encoding function for lookup tables but that's unrelated).

romainguy:
Honestly I think at this stage a better solution to this problem is something I was discussing with @elalish: an extension that precisely define an exposure scalar factor, and a 3D LUT to be applied by the renderers to go from D65-Rec.709-Linear ("scene light") to D65-Rec.709-sRGB ("display light"), at least for SDR.

This way, content creators are free to tone map/color grade/gamut map/etc. as they see fit. This means you could use the OOTF, and others could use their preferred tone mapping solution. In addition, a 3D LUT would allow engines to support full color grading pipeline and a better interop with other tools (Photoshop, Da Vinci Resolve, etc.).

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: > I also have issues with the requirement of bounding the scene values to the 0-10,000 range :) And dividing by 10,000 wouldn't work well for all scenes.

Could you please elaborate as to why a known dynamic range of 0 - 10000 would be a problem?

Why would it not work well for some scenes?
(Remember, this factor has nothing to do with 'auto exposure', it would be an implementers way of mapping the known dynamic range into the 0 - 1 range. Thats all)

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: It is the reference OOTF consistent with ITU broadcast practises and is taken from ITU BT 2390.

> I think in your case it "happens" to work as some kind of tone mapper because the OOTF is not symmetrical, but that's not its purpose.

Why do you think that?

> You are effectively adding a tone mapper without calling it that.

I am adding a reference OOTF (from ITU BT 2390) , which is meant to provide the reference output - anything added to this would be considered artistic intent (tone mapper)

rsahlin:
> It's still misleading to point to engines supporting PQ because they are not using it in the way you are defining here at all. They use PQ because that's the transfer function of Dolby Vision/HDR10 (some also use PQ as a log encoding function for lookup tables but that's unrelated).

Well, I disagree.
I don't believe that "they are not using it in the way you are defining here at all" - it's used as a transfer function in this extension and in my opinion there are many similarities.
Afterall, I started with this extension after investigating how HDR support was handled by the gaming community.

rsahlin:
> Honestly I think at this stage a better solution to this problem is something I was discussing with @elalish: an extension that precisely define an exposure scalar factor, and a 3D LUT to be applied by the renderers to go from D65-Rec.709-Linear ("scene light") to D65-Rec.709-sRGB ("display light"), at least for SDR.

In what sense and from what perspective would that be a "better solution"?
What would be solved by such an extension that would not be solved by this?

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
bhouston: I as well think this is quite problematic.  I think that light intensities can be focused or change dynamically (grazing fresnel or via glass refractions) based on camera and scene configurations which may not be known ahead of time.   Thus a hard limit is super hard to enforce.  I am also not sure it is that needed.

sobotka:
> I don't believe that "they are not using it in the way you are defining here at all" - it's used as a transfer function in this extension and in my opinion there are many similarities.

All of those games use “tone” curves; some form of open domain tristimulus to closed domain “compression”. Full stop. Speak with any of the pipeline developers to confirm this. If you would like, I am sure someone in this thread has contacts at those houses.

It is important to separate the open domain tristimulus values, which extend from zero to infinity, from the encoded down-the-wire signal.

That is, depending on the nature of the output medium, the input tristimulus must be compressed to it. This is a non-trivial act of image formation, and must consider the output medium’s capabilities^1.

It should also be noted some foundational problems with assumptions. Specifically, it may surprise folks that even if a medium can fully represent the stimulus in terms of colourimetric definitions, it has been researched and shown that 1:1 representations _are deemed unacceptable_^2. And this is assuming our display technology could, which they are completely incapable of.

You should note that in 2390 their OOTF uses a demonstration of _previously graded_ content. Meaning the incoming signal is already a fully formed image.

Bottom line, it is absolutely critical to form the image prior to distorting a signal for down the wire delivery, or final distortions for something that complies with the previously formed down-the-wire signal for further compression. See EETF in 2390.

1. This action also spans chromaticity footprint issues, as well as intensity. Again, the depth and breadth of what this act is is _massive_, and no approach that assumes “down the wire” encoding constraints will suffice.
2. Quality of Color Reproduction, David MacAdam, Kodak. And _many_ others.

romainguy:
> Well, I disagree. I don't believe that "they are not using it in the way you are defining here at all" - it's used as a transfer function in this extension and in my opinion there are many similarities. Afterall, I started with this extension after investigating how HDR support was handled by the gaming community.

Yes, the PQ transfer function is used by those game engines when outputting to HDR formats that call for a PQ transfer function. It's not needed for SDR display, and those engines don't make use of the OOTF. It's definitely valid to use PQ to output to HDR though, I'm not contesting that.

romainguy:
> In what sense and from what perspective would that be a "better solution"? What would be solved by such an extension that would not be solved by this?

It would be a better solution because it would give content creators full control over image formation, including color grading and tone mapping. This would allow you for instance to use the PQ OOTF as your "contrast" curve, and others to use something different that suits their needs.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
romainguy: > (Remember, this factor has nothing to do with 'auto exposure', it would be an implementers way of mapping the known dynamic range into the 0 - 1 range. Thats all)

That's what tone mappers are used for. Since they are usually non-linear, they let you compress the range while preserving contrast intents, which you can't achieve with a simple scalar factor.

rsahlin:
> All of those games use “tone” curves; some form of open domain tristimulus to closed domain “compression”. Full stop. Speak with any of the pipeline developers to confirm this. If you would like, I am sure someone in this thread has contacts at those houses.

Hi @sobotka  and thanks for your comment
Please note that the purpose of this extension is not to provide color grading (tonmapping curve), LUT or any type of artistsic intent.
That is the expected behavior, full stop.

 Of course games add artistic intent - that can be added to glTF as another extension - but it is not included in this proposal.

rsahlin:
> It would be a better solution because it would give content creators full control over image formation, including color grading and tone mapping. 

That is not the purpose of this extension - the purpose of this extension is to get deterministic output, under varying lighting conditions, to HDR or SDR display.
Loading two glTF's with different 'artistic intent' curves will yield unwanted result as one of those models is sure to be displayed in a way that is not intended.

So, from that perspective it is better to do the extension this way.

Artistic intent can be added as another extension.

sobotka:
> The purpose of this extension is to enable consistent material representation under highly varying lighting conditions.

> Hi @sobotka and thanks for your comment
Please note that the purpose of this extension is not to provide color grading (tonmapping curve), LUT or any type of artistsic intent.

These two statements cannot exist together. They are in a tautological opposition.

There is no need to redefine display colourimetry for BT.2100; it’s already done.

It would be unclear what design problem this attempts to rectify.

> It is assumed that the light contribution in any given scene can vary from low values, maybe in the 100 lumen range - to values in the 100 000 lumen range.
> These, internal, values must then be matched to a range that is acceptable by displays. This is usually in the 0 - 1 range.

Scaling the BT.2100 range to 100 nits would suggest that the range you are specifying is from casual desktop display to a mere 9.11 or so EV of range.

This is a perfectly suitable display output range, but unsuitable for the range of values in a scene.

It is confusing what this attempts to do.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+### To HDR capable display
+
+To convert from internal values (linear scene light) to the non-linear output value in the range 0.0 - 1.0 the reference PQ OETF shall be used. [See OETF](#oetf)  
+This is specified in ITU BT.2100:
+https://www.itu.int/rec/R-REC-BT.2100/en  
+
+If the framebuffer format and colorspace is known to the implementation then a format and colorspace shall be chosen to preserve the range and precision of the SMPTE ST 2084 transfer function.  
+If available, a framebuffer colorspace that is compatible with the color primaries of ITU BT.2020 shall be used.  
+If this colorspace is used the source color images must be converted from linear ITU BT.709 values to ITU BT.2020 linear values, where applicable depending on texture target and source format.
```
romainguy, line 149: Note there is another solution: for a real-time renderer, there are not many benefits to rendering directly in the Rec.2020 gamut (path tracers do get some benefits in bounce lighting). You can instead render in Rec.709 and convert to Rec.2020 as a final pass before any tone mapping, transfer function, etc. This can be much easier/cheaper than having to convert every input.

romainguy:
> That is not the purpose of this extension - the purpose of this extension is to get deterministic output, under varying lighting conditions, to HDR or SDR display. Loading two glTF's with different 'artistic intent' curves will yield unwanted result as one of those models is sure to be displayed in a way that is not intended.
> 
> So, from that perspective it is better to do the extension this way.
> 
> Artistic intent can be added as another extension.

I'm all for allowing the use of HDR displays but I don't think it belongs to an extension. Otherwise, an asset could technically not be displayed in HDR on an HDR-capable display if it didn't specify the extension. Not only that, but it means that an asset using this extension would mandate a PQ encoding output, which is incompatible with HLG for instance (you could always undo the PQ encoding and re-encode with the HLG curve). To do what you are intending to do we need two things:
- Exposure, which you have under `sceneAperture` (but again I strongly disagree with this term)
- A compression curve of some kind which you could parameterize to hit a target output luminance (as long as you define the scale, and here you can get informed by what PQ does)

I would also add that you've already baked artistic intents in your proposal: "Without the reference OOTF the appearance is somewhat whitewashed"

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: I Ben and thanks for your comment

> I as well think this is quite problematic - trying to enforce a maximum intensity limit.

It does not enforce a maximum intensity limit, you should not interpret this value as maximum intensity.
It defines the known dynamic range to 0 - 10000, simply because this is the range where the perceptual quantizer will produce gradients without visible banding (assuming display is >= 10 bits per pixel)

For any case where it is desired to have a different (intensity) range the `sceneAperture` can be used to specify that.

Hope this clarifies any confusion.

rsahlin: > That's what tone mappers are used for. Since they are usually non-linear, they let you compress the range while preserving contrast intents, which you can't achieve with a simple scalar factor.

I think that you are confusing the input to the reference rendering intent (reference OOTF)  with the non-linear PQ - this takes linear scene values and 'compress' into a non-linear range intended for the output device.

Please read the sections "Perceptual Quantizer - reference OOTF" and "Perceptual Quantizer - reference OETF"

rsahlin:
> These two statements cannot exist together. They are in a tautological opposition.
> 
> There is no need to redefine display colourimetry for BT.2100; it’s already done.
> 
> It would be unclear what design problem this attempts to rectify.

And yet, this is exactly what this extension sets out to provide a solution for.
I would encourage you to make a best effort to try and understand the purpose of this extension - without it I fear we will be perpetually stuck, argumenting from different vantage points for solutions that sets out to solve different things.

As for the design problem this extension aims to solve:

In glTF, 1.0 is considered to be a fully exposed pixel.
This leads to unwanted effects when the light contribution (intensity) is above 1.0, resulting in complete white-out at very low brightness levels.

Here is an example of a scene with macbeth colored test spheres (from 3D certification process) and a directional light with low brightness (50 lumen / m2).
Rendered according to spec:

image

The hue shift (white-out) happens because the process of mapping values in a higher range to that of the display is not specified, values are simply clamped.
Clearly not PBR and not a wanted result, assuming you are looking for an image that conveys the material in a realistic manner.

Using this extension, and a directional light with intensity 10 000 the result will be:

image

sceneAperture can be setup to up or downscale light values, meaning you could easily have a scene where your directional light is 100 000 instead of 10 000

rsahlin:
I'm all for allowing the use of HDR displays but I don't think it belongs to an extension. Otherwise, an asset could technically not be displayed in HDR on an HDR-capable display if it didn't specify the extension. Not only that, but it means that an asset using this extension would mandate a PQ encoding output, which is incompatible with HLG for instance (you could always undo the PQ encoding and re-encode with the HLG curve).

I doubt this extension would prohibit display of glTFs on HDR capable displays, simply because the spec does not say anything about output.
If there is interest in HLG I can certainly look into updating the spec to be able to use framebuffer with that colorspace.
That can also be added at a later date. 
HDR10 (PQ) is chosen because it is the most widely adopted HDR format.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
lexaknyazev: Agreed. This process should not be normatively defined here.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,439 @@
...
+
+If the framebuffer format and colorspace is known to the implementation then a format and colorspace shall be chosen to preserve the range and precision of the SMPTE ST 2084 transfer function.  
+If available, a framebuffer colorspace that is compatible with the color primaries of ITU BT.2020 shall be used.  
+If this colorspace is used the source color images must be converted from linear ITU BT.709 values to ITU BT.2020 linear values, where applicable depending on texture target and source format.  
+
+For HDR output,  a range extension value of 59.5208 and gamma of 2.4 shall be used.  
+
+**Implementation notes**
+
+Conversion of color primaries to ITU BT.2020 could be done after loading of a PNG/JPEG and after the image has been gamma expanded from sRGB to linear.
```
lexaknyazev, line 184: This note should be omitted as the initial image compression and EOTF decoding is already defined in the core spec.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,439 @@
...
+
+## Scene aperture
+
+### Parameters
+
+The following parameters are added by the `KHR_displaymapping_pq` extension, it is added with the extension on a scene object:  
+
+| Name              | Type     | Description                                                                                                                                | Required |
+| ----------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------ | -------- |
+| **sceneAperture** | `object` | Scene light adjustment setting, if no value supplied the default value of OFF will be used and scene light values will be kept unmodified. | No       |
```
lexaknyazev, line 221: Since the extension object containing this property is defined on a `scene` object, the property should be called `aperture`.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,439 @@
...
+
+### Scene aperture
+
+`sceneAperture`
+
+| Name                | Type     | Description                                                                                          | Required |
+| ------------------- | -------- | ---------------------------------------------------------------------------------------------------- | -------- |
+| **apertureValue**   | `number` | Base value used to calculate aperture factor                                                         | No       |
```
lexaknyazev, line 229: Since the JSON object containing this property is defined on an `aperture` object, the property should be called `value`.

Moreover, please consider moving all members of the `aperture` object to the extension object to reduce nesting.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,439 @@
...
+The goal of this extension is to provide a way to convert the resulting (rendered) scene linear light output values to a known range so that they can be sent to a display.  
+One of the reasons for this is to retain the hue of the source materials under varying light conditions.  
+Correct representation of hue is important in order to achieve a physically correct visualization and to retain original artistic intent.  
+
+This extension does not declare user defined tone-mapping, s-curve, color lookup table (LUT) or similar as the process of applying these may be non pysically based and alter the energy conserving nature of the glTF BRDF.  
+
+Another reason is to provide means for deterministic brightness (light intensity) values.  
+According to research (Bernstein et al. 2018) the mind's perception of brightness is mostly the same, regardless of object brightness.  
+The difference in the perception of the object's brightness is based on background color.  
+This means that in order to have deterministic control of perceived brightness it is important to be able to control both object and background color, for instance by means of a set dynamic range.
```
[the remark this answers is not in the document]
rsahlin: Hi @lexaknyazev and thanks for your comments

> glTF scenes do not have any implicit or explicit background color unless the IBL is rendered.

Well, background here only relates to what an object is viewed in front of, not really stating that it is part of the glTF.

Would this be any better?
`...it is important to be able to control the scene color contribution, for instance by means of a set dynamic range.`

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,439 @@
...
+This extension does not declare user defined tone-mapping, s-curve, color lookup table (LUT) or similar as the process of applying these may be non pysically based and alter the energy conserving nature of the glTF BRDF.  
+
+Another reason is to provide means for deterministic brightness (light intensity) values.  
+According to research (Bernstein et al. 2018) the mind's perception of brightness is mostly the same, regardless of object brightness.  
+The difference in the perception of the object's brightness is based on background color.  
+This means that in order to have deterministic control of perceived brightness it is important to be able to control both object and background color, for instance by means of a set dynamic range.  
+
+It also provides the specification for using HDR compatible display outputs while at the same time retaining compatibility with SDR display outputs.  
+
+The intended usecases for this extension is any usecase where the light contribution values will go above 1.0, for instance by using KHR_lights_punctual, KHR_emissive_strength or KHR_environment_lights.
```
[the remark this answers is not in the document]
rsahlin: Yes, unless you do want your outout values to be clipped at 1.0

Would you propose a different wording?

How about:
`The intended usecase for this extension is when light contribution values will go above 1.0 - for instance by using KHR_lights_punctual, KHR_emissive_strength or KHR_environment_lights - and the output values shall be mapped to an output range that retains hue.`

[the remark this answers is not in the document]
pierremoreau: Unless KHR_materials_unlit is present, in which case those objects would still be rendered even in the absence of light sources.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: Updated according to comments.
Implementations are free to do conversion at any stage before value is written to framebuffer.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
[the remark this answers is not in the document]
rsahlin: Thanks - updated accordingly

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,439 @@
...
+The difference in the perception of the object's brightness is based on background color.  
+This means that in order to have deterministic control of perceived brightness it is important to be able to control both object and background color, for instance by means of a set dynamic range.  
+
+It also provides the specification for using HDR compatible display outputs while at the same time retaining compatibility with SDR display outputs.  
+
+The intended usecases for this extension is any usecase where the light contribution values will go above 1.0, for instance by using KHR_lights_punctual, KHR_emissive_strength or KHR_environment_lights.  
+When using this extension it is recommended that the dynamic range is kept between 0 and 10 000 lumen/m2 in order to utilize the range of the Perceptual Quantizer.  
+This can for instance be done by exporters by scaling light contributions values.  
+
+Here the term rasterizer means a rendering engine that consits of a system wherein a buffer containing the pixel values for each frame is prepared.
```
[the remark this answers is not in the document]
rsahlin: Thanks - updated.
Replaced 'rasterizer' with 'renderer'

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,439 @@
...
+### glTF asset considerations
+
+The extension affects the entire glTF asset, ie all scenes, geometry, images and textures, included in a file that is using this extension.   
+This means that the current rendered scene shall be output using the displaymapping declared by this extension whenever the usecase is relevant, ie a realtime rasterizer with output to a display.  
+
+Visualization of multiple glTF assets using this extension is supported and will produce a normative result.  
+
+If the glTF asset contains multiple scenes, each one when rendered, shall be output using this extension.  
+
+If the glTF asset contains this extension but no scene or model data then it may be treated as an enabler for displaymapping.
```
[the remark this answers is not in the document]
rsahlin: Ok, but section 3.5 of the spec states:
```
A glTF asset that does not contain any scenes SHOULD be treated as a library of individual entities
such as materials or meshes.
```
Would you prefer that the behavior of using KHR_displaymapping_pq but not containing any scenes is undefined?

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,439 @@
...
+
+[Output using extension, without scene aperture, and one directional light with an intensity of 10 000 lumen / m2](images/DisplayMappingPQ-10000.png?display=inline-block)
+
+
+
+
+
+### glTF asset considerations
+
+The extension affects the entire glTF asset, ie all scenes, geometry, images and textures, included in a file that is using this extension.
```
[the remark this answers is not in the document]
rsahlin: Yes, I mean that this extension effect the _output_ of the glTF asset :-)

Would a better wording be:
`The extension affects the output of the entire glTF asset, all scenes and nodes, included in a file that is using this extension.`

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: The current wording is:
```
Conversion of color primaries to ITU BT.2020 could be done after loading of a PNG/JPEG and after the image has been gamma expanded from sRGB to linear.
Another solution would be to perform render calculation in BT.709 colorspace and convert to target colorspace before writing out values to the framebuffer.
```
Is this ok with you?

romainguy:
> And yet, this is exactly what this extension sets out to provide a solution for. I would encourage you to make a best effort to try and understand the purpose of this extension - without it I fear we will be perpetually stuck, argumenting from different vantage points for solutions that sets out to solve different things.
> 
> As for the design problem this extension aims to solve:
> 
> In glTF, 1.0 is considered to be a fully exposed pixel. This leads to unwanted effects when the light contribution (intensity) is above 1.0, resulting in complete white-out at very low brightness levels.
> 
> Here is an example of a scene with macbeth colored test spheres (from 3D certification process) and a directional light with low brightness (50 lumen / m2). Rendered according to spec:
> 
> image
> 
> The hue shift (white-out) happens because the process of mapping values in a higher range to that of the display is not specified, values are simply clamped. Clearly not PBR and not a wanted result, assuming you are looking for an image that conveys the material in a realistic manner.
> 
> Using this extension, and a directional light with intensity 10 000 the result will be:
> 
> image
> 
> sceneAperture can be setup to up or downscale light values, meaning you could easily have a scene where your directional light is 100 000 instead of 10 000

Those images are a good example of the problem that glTF needs to solve, but there are two independent steps that need to be taken:

- Image formation (this includes exposure, tone mapping/range compression, etc.)
- Display mapping

The reason I'm arguing against this extension is that it does both. If we want an extension to output to HDR capable displays using PQ, then this extension could be that, but it should only define the OETF, output color space, etc. Exposure/scene light scaling should be part of an image formation extension along with range compression (which does not necessarily has to be artistic).

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: Thanks - updated

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: Thanks - updates

I would like to keep the aperture properties as an object that implementations may switch behavior depending on object declared or not.
Please re-open if you disagree.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -96,7 +96,7 @@ The following steps shall be performed before outputing the calculated pixel val
 This would normally be done in the fragment shader.  
 
 3: Scene aperture  
-If `sceneAperture` is defined, then the scene linear light (normally fragment shader output value) pixel value shall be adjusted as defined:    
+If `aperture` is defined, then the scene linear light (normally fragment shader output value) pixel value shall be adjusted as defined:
```
romainguy, line 99: I will be annoying here but the new name `aperture` is even more confusing as it does not related to what's normally designated by the term `aperture` when it comes to rendering/video/photography. Please consider `exposure` or `lightScale` or anything else.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
lexaknyazev: Let's agree on the intention first. The aperture object itself and all its four properties are optional, i.e., it's valid to omit them. Luminance range properties have default values. Here are two examples with their interpretation as written in the spec:

### Empty extension object
Pass scene light intact.
```json
{
  "scenes": [
    {
      "node": 0,
      "extensions": {
        "KHR_displaymapping_pq": {}
      }
    }
  ]
}
```

### Empty aperture object
Clamp scene luminance to `0.0 .. 10000.0`, **factor remains undefined**, result is undefined.
```json
{
  "scenes": [
    {
      "node": 0,
      "extensions": {
        "KHR_displaymapping_pq": {
          "aperture": {}
        }
      }
    }
  ]
}
```

Please elaborate on the expected usage of the aperture controls and I'll help with writing the correct schemas.

rsahlin:
@sobotka 

> And again, what happens when a single channel escapes the display range?

In what scenario may that be?
Are you talking about _any_ hypothetical glTF?

>You desperately need to revisit basic concepts.

Please refrain from such broad generalizations, make sure to provide constructive feedback.

Here is a link to a page that describes how to write constructive code review comments, please have a look before we continue this discussion:
https://www.michaelagreiler.com/respectful-constructive-code-review-feedback/

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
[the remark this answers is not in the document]
lexaknyazev: Since the aperture controls could be defined per-scene, I would suggest one of these two options:

### 1. Unify the controls
- Apply the PQ only to _scenes_ that have the `KHR_displaymapping_pq` extension object. Aperture controls may still be optional as they are now.
- Note that meshes not attached to nodes and nodes not attached to scenes are not rendered anyway.

### 2. Further split the controls
- Do not define any new objects for the `KHR_displaymapping_pq` extension, leave it only as an element of the `extensionsUsed` array.
- When the extension is present, it enables PQ output for all scenes of the asset.
- Note that meshes not attached to nodes and nodes not attached to scenes are not rendered anyway.
- Move aperture controls to a new extension that would require this one.

sobotka:
> > You desperately need to revisit basic concepts.

What part of revisit basic concepts is subject to tone policing?

The idea of revising basic concepts of colourimetry and image formation here are missing, as well as conflating EOTFs with those concepts.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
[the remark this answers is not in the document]
rsahlin: Hi @lexaknyazev and thanks for your comments and suggestions!

Looking at this extension from a 3D Commerce (and ikea) point of view:
What we want to achieve is to mark an asset to be rendered using this extension (retaining hue) and also be confident that a well behaved renderer (certified) will handle the asset accordingly under different usecases.

If we choose number 1:
* What would happen to an asset (nodes) that is loaded into an existing scene that does not use the extension?

If we choose number 2:
* I am thinking about going this route - the problem is that this may lead to situations where the hue shift is still present simply because users have lights above 10 000.
Do you see a solution for this scenario?
Is it as simple as specifying that light contribution values when rendering this asset must be within [0 - 10000] or do you forsee other problems?

Thanks for your constructive comments @lexaknyazev

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: Thanks for the offer @lexaknyazev :-)

 I think I understand your point know.
 
 The intention is that if `aperture` is not defined then no factoring of scene luminance will be done.
 The scene aperture step will be omitted - this is intended only for usecases where you know that your scene light is within range (0 - 10 000)
 
Do you mean that as it is specified now the scene aperture will _always_ be applied?

lexaknyazev: > Do you mean that as it is specified now the scene aperture will _always_ be applied?

I mean that the current design allows defining an empty aperture object that would not work at all because `aperture.value` and `aperture.control` do not have default values. There are two possible solutions to this issue: make them required (my preference here) or define default values for them.

rsahlin: > I mean that the current design allows defining an empty aperture object that would not work at all because `aperture.value` and `aperture.control` do not have default values.

Ah, I see - that is an unintentional error.
It should state that all values are required if the object is defined.
In the schema (gltf.aperture.schema) I have declared them as "minItems": 1

Is this enough (if I also change the table to say Yes)?

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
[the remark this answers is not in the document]
lexaknyazev: > What would happen to an asset (nodes) that is loaded into an existing scene that does not use the extension?

- Loading multiple glTF assets into an existing rendering environment is generally not defined by the core spec.
- In my opinion, it's quite unexpected for the newly added asset to change the rendering output of pre-existing assets. Should the global PQ be turned off when the asset is removed later?
- For the option 1, I'd just discourage applications from mixing PQ and non-PQ assets. Bulk-adding this extension string to the existing library is trivial enough.

>  the problem is that this may lead to situations where the hue shift is still present simply because users have lights above 10 000

This problem remains anyway because the current extension spec allows omitting the aperture controls and says:
> Light contribution values above 10000 cd / m2 are discouraged and will be clamped by implementations if not adjusted by scene aperture.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
lexaknyazev: > It should state that all values are required

As the min and max luminance properties have default values, they cannot be required.

> In the schema (gltf.aperture.schema) I have declared them as "minItems": 1

That rule refers to the length of arrays and cannot be used with numbers.
 
---

> Is this enough

Let's agree on the semantics in this thread first.

- Could any of these four properties be negative?
- Could `aperture.value` be zero?
- Could `maxLuminance` be less than `minLuminance`?

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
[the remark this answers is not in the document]
rsahlin: Good points!

>This problem remains anyway because the current extension spec allows omitting the aperture controls and says:

Correct - and I feel that this is the part of this extension that is not decided on.
The more I think about it the more I feel the need for fixed aperture control, ie option 2 with fixed values of:

value = 10 000 
minLuminance = 0
maxLuminance = naively collected on a per frame basis from light contribution. If total light > 10 000 then all contribution will be scaled according to highest value.
control = removed

Do you think this would help with the purpose of this extension?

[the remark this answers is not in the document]
rsahlin: @lexaknyazev It would be super great if you are available for a quick call on teams?
(rickard.sahlin at inter.ikea.com)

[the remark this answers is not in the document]
bhouston: I do think that exposure and tone mapping is more of a camera property than a glTF file property.  Also putting it on the camera makes it clear what to do when combining multiple GLTF files together in the same scene.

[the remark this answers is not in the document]
rsahlin: Hi @bhouston and thanks for your comment

Yes, I also think so.
Currently the aperture setting is set to a default behavior that will do nothing if values are within [0 - 10000] range - but if lit with high intensity light > 10000 then the aperture function will scale incoming light so that the highest value is 10000.
Ie meaning that if you have a light with intensity 10000 and add a new light with intensity 100000 - then they will become 10000 and 1000 respectively.

Also, the extension is turned on simply using the KHR_displaymapping_pq string in the asset.
All in order to simplify and make the extension work the same regardless of glTF model/scene.

Do you think this will suffice?

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: Resolving this according to review meeting - please re-open if relevant after new changes

bhouston:
BTW I probably won't endorse this extension as being an official Kkronos extension.  It is I believe much of it is just best practices for viewer implementations.

The exposure/aperture and maximum scene illumination isn't really something I think should be standardized in the fashion described in this extension.  I would support a camera-specific exposure setting, but not in the fashion proposed/advocated in this current draft.

rsahlin:
Hi Ben and thanks for your comment

> The exposure/aperture and maximum scene illumination isn't really something I think should be standardized in the fashion described in this extension. I would support a camera-specific exposure setting, but not in the fashion proposed/advocated in this current draft.

Why do you feel that it should not be standardized in this fashion - what would you change and how?
This extension explicitly sets out NOT to be a physical camera.

bhouston:
> Why do you feel that it should not be standardized in this fashion - what would you change and how?
> This extension explicitly sets out NOT to be a physical camera.

Physical cameras, which I've implemented a few times, have a few different aspects to them.

One area that is very useful in VFX is when you need match existing shots in terms of aperture, film size, focal length, etc.  This helps define the DoF CoC as well.  For the real-time world where the viewer's dimensions are often not fixed, these requirements do not appear often.  So I think skipping that part of VFX physical cameras is fine for now.

Another part of physical cameras is their exposure controls.  I think this is very useful.  I think that we should have an exposure control on the camera and it should be adjustable via animations and other factors.  I think we should be also able to set it to auto-mode, probably some smart sub-set of the features offered in Unreal Engine -- those are just amazing:

https://docs.unrealengine.com/4.27/en-US/RenderingAndGraphics/PostProcessEffects/AutomaticExposure/

I think that this proposal's approach isn't reflective of accepted best practices when it comes to exposure controls.  Rather it is using "aperture" instead and it is not tying it to the camera (which leads to conflicts on which data to use when) and it is also dealing with these weird lighting maximums in the scene.  This is a solution, but I don't think it is the best one.  I also think that if we conform to the current best practices in the industry it will be easier to support whatever we decide to release because the content will exist in roughly that form.

Also this extension does with HDR content and various transforms in that content pipeline which again I view are just best practices and a little bit orthogonal, or at least separable, from the auto-exposure controls.

Thus I would separate this out into KHR_camera_exposure extension that only does exposure and auto-exposure controls (do a survey of what is supported across the various main tools) and then see if there is anything else to do... I suspect there isn't anything else needed except just some discussion of best practices in terms of implementing color workflows.

I go back to the original motivation for this extension as explained in an email thread.  You had hue shifts occurring in various images as a result of a tone mapping operator as you increased the illumination in the scene that wasn't balanced by a corresponding exposure reduction.  This is the core problem.  I've run into it myself many times before.  It can be fixed by simply either adjusting manually the exposure down, or by standardizing an auto-exposure adjustment setting.  This is the simplest solution and also an incredibly standard solution to the problem.  I advocate doing that.  It is how we solved it ourselves (although we've come up with some complex solutions where we can adjust exposure on a per material basis, but that type of complexity can come later.)

rsahlin:
> Another part of physical cameras is their exposure controls. I think this is very useful. I think that we should have an exposure control on the camera and it should be adjustable via animations and other factors.

And this would be in a different extension, ie one that sets out to model a physical camera.
This is not that extension.

>I think that this proposal's approach isn't reflective of accepted best practices when it comes to exposure controls.

Exactly - it does so on purpose sincec it does not set out to model a physical camera.

>I go back to the original motivation for this extension as explained in an email thread. You had hue shifts occurring in various images as a result of a tone mapping operator as you increased the illumination in the scene that wasn't balanced by a corresponding exposure reduction. This is the core problem. I've run into it myself many times before. It can be fixed by simply either adjusting manually the exposure down, or by standardizing an auto-exposure adjustment setting.

One major issue is that glTF defines 1.0 as a fully exposed pixel.
This means that a directional light with 1 lumen/m2 intensity will result in full bright illumination - not PBR at all.
Another issue is that the glTF camera does not have any exposure controls at all. All it has is a transform and view frustum.
It really just is an observer.
The glTF spec does not even touch upon how scene linear values end up to their target - which could be anything from a 3D printer to a printed photo.

This extension solves these problems without adding a physical camera or post-processing.

I don't see that having this as best practice for viewers will work - there will surely be situations where you _want_ to do your displaymapping differently - even in a 3D commerce usecase.
In that case just drop this extension from the asset and go back to default behavior or use some other extension.

bhouston:
> One major issue is that glTF defines 1.0 as a fully exposed pixel.
> This means that a directional light with 1 lumen/m2 intensity will result in full bright illumination - not PBR at all.

All major glTF viewers implement exposure controls (usually fixed, arbitrary right now) with tone mapping usually using the VFX ACES standard and then map to the screen using sRGB color space transforms.  Thus your above statement isn't the practical reality.

Right now because the web is limited to LDR content, yes the brightest a pixel can be on screen after exposure adjustment + the tone mapping + sRGB conversion is 1.0.  But that is because that is a current output limitation.  Once you can output P3 or HDR content from a WebGL/WebGPU context, then this pipeline will change according to the capabilities of the hardware, and glTF doesn't really have to have any extensions to enforce that.

The only think I feel that should be added to the glTF is exposure controls, and explicitly auto-exposure.  Maybe the tone mapping operator as well -- there can be artistic choice here, but I am not sure it matters as the ACES is good enough for I think +90% of cases.  But the rest of the pipeline after tone mapping to display output has a well defined physically correct answer that shouldn't require any artistic choices -- it is either correct or it is incorrect for the display capabilities.  Having to specify an extension here makes no sense.

I think it isn't necessary to convince me of the value of this extension.  I've invested time into presenting this feedback.  You can choose to receive it or dismiss it.  And we can move on.

rsahlin:
Hi @emackey  and thanks for taking the time to look into this.

On the question of resolving conversations - in my experience you have one of two choices:
A: The person writing the comment is active and fair with regards to response from the person assigned to the PR. 
This shall of course include marking conversation as resolved whenever a reasonable response has been given.

B: The person assigned to the PR marks conversations as resolved whenever a reasonable response has been given - the action is now on the person that initiated the conversation to either accept or re-initiate the conversation.

Taking approach A almost inevitably leads to complete stalling of progress on the PR - hardly a wanted scenario.
This is why I have taken approach B to move forward unless there are any major blockers - naming a parameter 'exposure'  or 'aperture' is hardly blocking in my opinion.
Other conversations are resolved simply due to the fact that the solution would be to turn this extension to something that does not fit with the purpose (of it)

Yes- this is a one solution fits all - it is by design.

I would say that what happens in a usecase where a scene is taken from one glTF and nodes are insert from some other glTF is not something that is usually covered by the extension - look for instance at KHR_lights_punctual
However, I believe that the formulation of how to handle multiple assets belongs in viewer certification.
I would be happy to specify how to handle mixing of models with/without this extension. 
I'm just not sure that extensions are supposed to do this.

I agree that we shall move away from changing material behavior by mere presence - this extension does not do that!
Material properties are left unchanged, all BRDF calculations are left unchanged - this extension only covers how to handle linear scene light dynamic range and how to get these values to display.
This is an area that is currently missing.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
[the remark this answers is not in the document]
rsahlin: The parameter controls the virtual opening letting light into the observer (or the eye) with the goal of scaling down values when light contribution is higher than the dynamic range.
As such it has nothing to do with the exposure as used in a camera (aperture/shutter/ISO).

Based on this justification I feel that the most fitting term for this extension is aperture.
In my opinion exposure is more fitting for an extension that seeks out to model a physical camera having aperture/shutter and ISO.

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
rsahlin: See above comment

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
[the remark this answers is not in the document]
romainguy: Exposure *is* the scaling of values. The fact that in a camera yo use aperture/shutter/ISO to control exposure doesn't mean exposure is those values (but you can compute a photometric exposure from those settings). It is so much more confusing to use the term "aperture" since it already has a specific meaning. I'm not advocating for the use of aperture/shutter/ISO here, just for using the term exposure.

emackey:
> Taking approach A almost inevitably leads to complete stalling of progress on the PR - hardly a wanted scenario.

100% agree here.

> Yes- this is a one solution fits all - it is by design.

Ok.  If one formula solves the need for everyone, I'm glad for it.  Simpler is better.  We just need other vendors to speak up and say that this strategy works for them too.

> I would say that what happens in a usecase where a scene is taken from one glTF and nodes are insert from some other glTF is not something that is usually covered by the extension - look for instance at KHR_lights_punctual
However, I believe that the formulation of how to handle multiple assets belongs in viewer certification.
I would be happy to specify how to handle mixing of models with/without this extension.
I'm just not sure that extensions are supposed to do this.

I suppose this could be a cert issue.  For the case of KHR_lights_punctual and friends, it may not be written anywhere, but the intention (in my mind at least) was always that glTF is a "partial scene" format, contributing all its contents (not just geometry) into a larger scene.  Take for example the case of a glTF model of a street lamp, it should contain KHR_lights_punctual for the bulb in the lamp, and clearly that lamp should shine light on any other geometry or models within range in the same scene.

> Could you please elaborate what you mean that its not well targeted?

Poor phrasing on my part, apologies.  Generally I'm talking about the actual JSON structure inside the file.  Many extensions are targeted to particular objects (sometimes nodes, sometimes materials, sometimes textures, sometimes the "asset" object, depending on the nature of the extension).  The top level JSON is sort of a catch-all, for example if an extension applied to multiple JSON objects all need to share some resource, then that resource could be named with an extension at the top.  But a top-level extension on its own doesn't apply to any particular part of the glTF file structure, it merely provides a library of stuff for those deeper objects to reference.  That's kind of what I meant with my poor choice of words.  Perhaps if display mapping PQ is intended to apply to the whole scene (and all cameras within that scene), maybe the extension should be applied to the glTF `scene` object?  Generally, glTF files contain only one scene, and its contents are intended to be integrated into a larger application-supplied scene.  So I could imagine applying this extension to a glTF's default scene and expecting that to copy to the application's scene.  And if multi-scene glTFs ever become a thing that anyone uses, one could imagine a file supplying one scene with PQ and one without.

> What do you see as a problem with an extension that changes behavior by it's presence?

Here too, it's structure.  Most glTF loaders that I've been involved in are pretty quick to discard the glTF structure like a used candy wrapper, loading the tasty contents onto a hungry GPU as fast as they can be liberated from their packaging.  Often the application has some internal API or memory model that holds the good stuff without bothering to take a picture of the box it came in.  Asking these applications to take different actions based on the mere presence of a glTF extension is, in a way, putting too much emphasis on a disposable wrapper.  For example you could never have an extension "promoted to core" or be anything other than an extension with this strategy.

That said, there could certainly be an exception to that philosophy for configuration-free extensions such as unlit and PQ.  The memory model here is a simple Boolean flag, and there's hardly any point to injecting `"pq": true` in the JSON stream when you can guess what should happen just by seeing the extension's name.  This leaves little room for config options to be added later, but perhaps none are needed.  I guess it's OK, but it feels quite awkward to me.

I think the main thing that the WG should be asking for here, regardless of whatever else I've written on this PR, is some buy-in from other vendors.  Clearly the problem of overexposed glTF rendering is causing pain for users.  Can we find other vendors who will sign onto this as the best solution for all?

rsahlin:
Thanks for the reply @emackey 

> I suppose this could be a cert issue. For the case of KHR_lights_punctual and friends, it may not be written anywhere, but the intention (in my mind at least) was always that glTF is a "partial scene" format, contributing all its contents (not just geometry) into a larger scene.

I totally egree - I guess the distinction is if we should go from intention to actual specification?
In this case the extension is designed to work in a usecase where models are mixed - for instance 3D Commerce, 'Metaverse', AR and game like scenarios.

>But a top-level extension on its own doesn't apply to any particular part of the glTF file structure, it merely provides a library of stuff for those deeper objects to reference.

Ok, if this is the case we need to find a different approach to how to enable this extension.
I have deliberately put the extension outside of the `scene` object since the extension shall be applied to the whole `asset`.
Since it is non configurable I have a hard time imagining a usecase where you would want to have one scene with the extension and one without.

>For example you could never have an extension "promoted to core" or be anything other than an extension with this strategy (because there's no named setting, only the extension itself is named).

Ok, I think I would find this to be an issue in the future - putting on my 3D Commerce hat I would certainly want this extension to be promoted to core in the future.
Do you see some other solution to this _conundrum_, other than moving the extension into the `scene` object?

>Can we find other vendors who will sign onto this as the best solution for all?

(Puttin on my PBR Next hat)
We have buy-in from 3D Commerce.
Other than that, take into consideration the design goals of this extension:
* Add units to output values
* Define range of output values
* Physically correct output to display
* Backwards compatible, ie non extension models can be viewed with models using extension
* Mixing of models - 3D Commerce,  'Metaverse, AR or any usecase where models are distributet 'in the wild'
* Future proof device output behavior

Meaning that the extension may not be the 'best solution for all' - however I believe it is a good building block for a core feature that is currently missing from the spec.

romainguy:
> * `image_formation` proposes using ACES as default tonemapping. I think this isn't very neutral and causes issues with brand colors but I'm not aware of a widely accepted replacement.

ACES was chosen as a "good default" to match common existing glTF viewer behaviors. We'd have to define what "neutral" means in this particular context and it's more complex than it seems I think. To me a neutral transform should avoid hue skews and other perceptual effects like Abney, etc.

> * should the scaling factor be explicit or implicit, why, and in which situations (which I'd personally answer with "it's explicit per camera and there can be a fallback per scene")

To me it should be explicit per camera with a per-scene fallback. It could also be improved later on with an auto-exposure extension.

> 'Eye observer' rather than physical camera - focus on color perception (photopic vision)

@rsahlin While I wholeheartedly agree with this idea, I disagree that this is what's achieved here. Perception is not stimulus, and just outputting "raw" scene data to the display won't work.

donmccurdy:
@rsahlin I would strongly encourage interested 3D Commerce members to spend time exploring whether the desired "neutral" mapping and image formation goals can be expressed in terms of the exposure and LUTs provided by the `KHR_image_formation` proposal. It is well-designed. Implementations of the (more prescriptive) approach of the `KHR_displaymapping_pq` proposal would likely — by specification or not — be implemented within an imaging chain with parameters that look a lot like `KHR_image_formation`.

***

> To me it should be explicit per camera with a per-scene fallback. 

+1

rsahlin:
Hi @donmccurdy - I took the liberty of moving the reply into this thread to avoid cross-talk.

>But for an extension meant for use in the wider glTF ecosystem, the intent of KHR_displaymapping_pq is too prescriptive,

Could you please elaborate why and explain in what sense you think it's is too prescriptive?

extensions/2.0/Khronos/KHR_displaymapping_pq/README.md
```diff
@@ -0,0 +1,532 @@
...
+Imagine a background with a scene linear brightness around RGB (10.0, 10.0, 10.0) with an object of scene linear brightness around RGB (100.0, 100.0, 100.0).  
+As the mind will perceive object brightess compared to the background, ie 100 vs 10, it is vital to have a known higher dynamic range in order to achieve deterministic display output.     
+
+Apart from being widely supported and used in the TV / movie industry the perceptual quantizer is also embraced by the gaming community, with support in the engines from some of the major game companies.  
+For instance, game engines Frostbite and Lumberyard and also in specific games such as Destiny 2 and Call Of Duty.  
+
+**Implementation Notes**
+
+The extension is compatible with usecases where a different output is wanted.  
+This could for instance be a viewer or engine that implements a physical camera or a post-processing step.
```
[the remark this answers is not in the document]
rsahlin: Please have a look at the images in the sections about 'Integration points' and 'Motivation'
/images/RenderProcess.png
/images/DesignOverview.png

How the tone-mapping or LUT shall be applied together with this extension depends on in what space you want to perform the image space operations and how you have implemented this extension is your viewer or engine.

Hope this clarifies - if not please let me know

hybridherbst:
Side note, when following your steps (open "http://gltf.ux3d.io/", enable "Force Displaymapping PQ" in Advanced Settings) results in a non-clipped image that looks bad:

image

Would this be your intended result for this file?

rsahlin:
> Adding the entire chain until PQ-encoded data is sent through the wire to a display as part of this extension will, if I'm not understanding these proposals wrong, for the foreseeable future not be possible in web viewers. 

Sure, and then web-viewers will use the SDR target until they can get access to swapchain format and colorspaces.
I don't see how this would be a problem.

>It could also happen that a viewer ends up implementing HDR support (for example, converting floating point Rec709 to Rec2020 and encoding with PQ to send to a HDR10 display seems to be what Unity is doing in HDR mode) and that this ends up "double-encoding" with PQ, effectively breaking any intended rendering effects.

Then it seems that they have already implemented the HDR part of this extension. 
Without knowing exactly how they have implemented the HDR support I would guess that they just need to adapt to using the OOTF + OETF when this extension is used.

Why would they do "double-encoding"?

rsahlin:
>Would this be your intended result for this file?

Absolutely not, this seems to be a bug in the Sample-Viewer use of the transmission or volume extension.

Try without the displaymapping extension enabled, not using 'image based' and change the "compensatory light intensity" - you will notice that nothing changes.
I guess the reason being light values are clipped somewhere - never going above 1.

rsahlin:
> I think ultimately [KHR_image_formation](https://github.com/elalish/glTF/tree/khr_image_formation/extensions/2.0/Khronos/KHR_image_formation) and [KHR_displaymapping_pq](https://github.com/KhronosGroup/glTF/tree/KHR_outputmapping_pq/extensions/2.0/Khronos/KHR_displaymapping_pq) in their current forms have the same goal: to specify a scaling factor and a neutral range mapping to make sure assets display perceptually similar in different viewers under different lighting conditions, instead of the current situation where this is entirely up to viewers. Please @elalish, @romainguy and @rsahlin correct me if I'm already wrong in that assumption.

This assumption is wrong.

KHR_displaymapping_pq - is designed to be compatible with a number of usecases that mix several glTF models, can have varied and changing light conditions and at the same time render with expected result (to both SDR and HDR displays)
Metaverse, AR, 'Furnish the home' and 3D Commerce.
This is why this extension has no configurable parameters and avoids doing any model or scene based filtering

KHR_image_formation has one usecase and that is the 'Main product picture' type of usecase - showing one model under fixed light conditions and at the same time applying color-grading (a filter)
It is not compatible with the other usecases.

rsahlin:
>Also note that this extension here could be implemented by https://github.com/KhronosGroup/glTF/pull/2128 with the appropriate LUT.

Not while retaining compatibility with different usecases such as Metavers, AR, 'Furnish the home' or 3D Commerce.
It will only work for a very limited use of the 'Main product picture' usecase - you would not even be able to change the light setup without breaking the look.
That extension as it is simply not an alternative to KHR_displaymapping_pq

I have already proposed to change that extension into one that defines a physical camera with aperture/shutter/ISO - nothing else - to be used together with this extension.

I think that is the best way forward.

PR #2090, opened 2021-10-31
KHR_materials_volume: Corrections to intervals of volumetric properties

donmccurdy:
For `thicknessFactor`, `thicknessTexture`, and `attenuationDistance`, the intervals given in the specification either don't match the schema, or don't seem correct based on context. I've updated them to match the schema and added the range in interval notation within the Properties table.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_volume/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_volume/README.md
@@ -90,7 +90,7 @@ If the extension is combined with `KHR_materials_transmission`, a refractive mic
 The volumetric properties are defined by adding the `KHR_materials_volume` extension to any glTF material. A non-zero thickness switches from thin-walled to volumetric behavior. This requires a manifold/closed mesh. The properties of the medium are not texturable, it is assumed to be homogeneous.
 
 ```json
-materials: [
+"materials": [
     {
         "extensions": {
             "KHR_materials_volume": {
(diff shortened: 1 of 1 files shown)


extensions/2.0/Khronos/KHR_materials_volume/README.md
```diff
@@ -109,14 +109,14 @@ The extension defines the following parameters to describe the volume.
...
+The thickness of a volume enclosed by the mesh is typically quite difficult to compute at run-time in a rasterizer. Since glTF is primarily used with real-time rasterizers, this extension allows for the thickness of the volume to be baked into a thickness map. Thickness is given in the coordinate space of the mesh. Any transformations applied to the mesh's node will also be applied to the thickness. Thickness is an absolute value and defined as the product of thickness factor and thickness texture value. The thickness factor is defined in the range \[0, inf), whereas the thickness texture value range is limited to \[0,1]. An exemplary configuration could set the thickness factor to the longest edge of the bounding box of a mesh, and the texture scales this value so that it corresponds to the actual thickness underneath each surface point.
```
emackey, line 119: @proog128 Is it legal to have `thicknessTexture` drop to zero in some locations?  Obviously we can't switch to thin-walled per-fragment like this.

proog128: Thickness of zero is fine also in the volumetric case. If thickness is zero, attenuation distance d and color c do not have an effect anymore when computing transmittance:

T(x) = e<sup>-σ<sub>t</sub>x</sup> = e<sup>(log(c) / d) * x</sup>
T(0) = 1

lexaknyazev (approved):
Thanks!

PR #2091, opened 2021-11-04
Remove duplicate 1-max(F0.rgb) in KHR_materials_specular

proog128:
In #2022 the duplicate (1-F0) term was removed from `c_diff` in the core spec. This PR removes it also from the KHR_material_specular extension.


bghgary (approved):

bghgary:
Good catch!

lexaknyazev (approved):

PR #2102, opened 2021-11-18
Fix link in KHR_materials_transmission

### LINKED ISSUE #2100: KHR_materials_transmission broken link in readme
In the Reference section of the readme, the link to Enterprise PBR is broken (404).
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_transmission/README.md#reference

It points to a spec using a date in the URI, which is probably not great for continuity over time as they get deprecated. Perhaps it would be better to use the User Guide page, since presumably it would be a bit longer-lasting?
https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/user_guide.md.html#parameters/basecomponent/transparency

lexaknyazev:
Fixes #2100.

The 2020x version does not seem to exist.


proog128 (approved):
Thanks!

Issue #2126, opened 2022-03-03
Add MX (Matrix.org Foundation) Vendor Prefix


robertlong:
He, I'd like to request the `MX` prefix for the [Matrix.org Foundation](https://matrix.org/). This prefix will be used for our metaverse platform, [Third Room](https://github.com/matrix-org/thirdroom).

We also plan on working with Khronos and the Open Metaverse Interoperability Group as much as possible. This vendor prefix will mostly be used while extensions are under development or if there are any platform-specific glTF extensions needed.

Issue #2125, opened 2022-02-25
Add SNAP vendor prefix


rbsheth:
Hi! I'd like to request the `SNAP` vendor prefix for [Snap, Inc.](https://snap.com)

PR #2129, opened 2022-03-07
Added MX and SNAP vendor prefixes

javagl:
Fixes https://github.com/KhronosGroup/glTF/issues/2126
Fixes https://github.com/KhronosGroup/glTF/issues/2125

@lexaknyazev There is a statement

> Once a prefix is approved, it should be added to the [glTF Validator's list of known prefixes](https://github.com/KhronosGroup/glTF-Validator/blob/master/lib/src/ext/extensions.dart).

in the `Prefixes.md`. Should that just be a plain PR into the validator repository, or is there any further process required for "approving" a prefix?


lexaknyazev:
I usually just sync the "known prefixes" when issuing a new validator release. No extra PR is needed.

lexaknyazev (approved):

PR #2139, opened 2022-04-08
Fix direction of X axis in Specification.adoc

Squeakrats:
The +X axis points to the right.

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -657,7 +657,7 @@ For property names, glTF usually uses camel case, `likeThis`.
 [[coordinate-system-and-units]]
 == Coordinate System and Units
 
-glTF uses a right-handed coordinate system. glTF defines +Y as up, +Z as forward, and -X as right; the front of a glTF asset faces +Z.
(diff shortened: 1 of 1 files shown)


lexaknyazev:
With +Z pointing forward, +X points to the left (hence -X points to the right) since it's a right-handed coordinate system. Please see the inlined image.

javagl:
Despite having followed that discussion, I just looked at the change and the spec and _for a short moment_ thought: "Wait, shouldn't that really be _+X_ in the text?"

The confusion could probably alleviated by using some sort of "Avatar" (instead of the BoomBox), maybe even with the hands labeled 'Left'  and 'Right':

glTF Coordinates

"Ah, yes, that's the left hand!"

Squeakrats:
I originally interpreted forward as you would in image editing software. 
"Bring To Front", "Send to Front", or "Bring Forward" will bring the shape to the front. That was what I thought "**+Z as forward**" was meaning to convey. From that perspective (the viewer perspective) right is still right. 

After reading the response and re-reading the text I realized it was intended to be relative to the object, which probably makes more sense.

PR #2147, opened 2022-04-21
KHR_animation_pointer

UX3D-nopper:


UX3D-nopper:
@lexaknyazev If accepted, squash and merge it.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -0,0 +1,282 @@
...
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension is based on the animation features of glTF 2.0. The structure of the schemas stay the same.  
+  
+The only major addition is, that the output values are mapped using a [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901).
+  
+### Motivation
```
rsahlin, line 35: Could you please elaborate on the usecases for this extension and provide some samples that shows the benefits of adding this extension?

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -0,0 +1,282 @@
...
+
+At this point of time, one can only target the transformation or weight data of a node.  
+With this extension, one can technically target any value (`scalar`, `vec2`, `vec3`, `vec4` and scalars in an `array`) in glTF e.g.
+
+* Color factors in materials
+* Camera field of view
+
+Even using a JSON pointer, the targets and their expected behavior are explictly defined.
+
+Current and future extensions have to write against this specification to allow to animate specific values.
```
rsahlin, line 45: I am afraid of this type of wording in the extension, as much as possible I believe that extensions shall be self contained - this puts a burden on all new extensions to take this extension into consideration.

I would rather prefer that this extension adresses the needed targets for the usecase(s) it sets out to solve.

rsahlin:
Please elaborate on why this extension is needed and what usecases it sets out to solve.

If this extension is intended as a 'technology enabler' I think care should be taken to confine it to only the core spec, not making it affect current and future extensions by default.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
hybridherbst: The usecase is being able to animate glTF files beyond the current very strict limitations (colors, UVs, camera FOV, etc. can't be animated).
This includes current and future extensions. Extensions can specify what data is supposed to be animatable.

hybridherbst:
Uploaded some first work-in-progress sample models + videos as a Draft PR:
https://github.com/KhronosGroup/glTF-Sample-Models/pull/353

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -56,7 +56,7 @@ It is not allowed to animate a glTFid property, as it does change the structure
 
 ### Valid target templates
 
-|`path`                                                |Accessor Type|Component Type(s)  |Description                                                   |
+|`pointer`                                             |Accessor Type|Component Type(s)  |Description                                                   |
```
hybridherbst, line 59: Just wanted to note that this (together with a couple more places) was also part of my PR here in case you missed it: 
https://github.com/ux3d/glTF/pull/20

UX3D-nopper: Oh, yes, I missed it.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -0,0 +1,318 @@
...
+|`"/materials/{}/normalTexture/scale"`                 |`"SCALAR"`   |`5126`&nbsp;(FLOAT)|Multiplier applied to each normal vector of the normal texture|
+|`"/materials/{}/occlusionTexture/strength"`           |`"SCALAR"`   |`5126`&nbsp;(FLOAT)|Multiplier controlling the amount of occlusion applied        |
+
+It is valid, to set the `pointer` to a property, which is not stored in the glTF.  
+This is the case, when the default value is omitted.  
+However, the parent object and/or extension **must** exist.  
+
+##### `extras`
+
+A channel may target any allowed JSON Pointer value, so values in `extras` can be targeted as well, but interpretation of the animated values is entirely application specific.
```
fire, line 92: I wish to animate an extras property that is representing a new material or a new material image? How can this be done?

UX3D-nopper: With a JSON Pointer, you can reference any element in a glTF:
E.g. `/materials/0/extras/myValue` is pointing to the first material having an individual 'myValue' in `extras`.

hybridherbst:
Question that came up while implementing export for this extension:

1. When a material extension property (such as iridescenceFactor) is animated, 
- it seems obvious that `KHR_materials_iridescence` needs to be added to extensionsUsed on the glTF root.
- does it also need to be added to the extensions of the material itself?
- what should implementations do when a material doesn't specify the extension, but its values are animated? My guess would be that most implementations would then not properly animate those properties (as e.g. the shader may not be the right one at that point in time).

UX3D-nopper:
> Question that came up while implementing export for this extension:
> 
> When a material extension property (such as emissiveStrength) is animated,
> 
> * it seems obvious that `KHR_materials_emissive_strength` needs to be added to extensionsUsed on the glTF root.
Yes, correct.
> * does it also need to be added to the extensions of the material itself?
Yes, and this independent of the animation pointer. All glTF extensions fow work like this.
> * what should implementations do when a material doesn't specify the extension, but its values are animated? My guess would be that most implementations would then not properly animate those properties (as e.g. the shader may not be the right one at that point in time). Should we make a suggestion here?
This is an invalid glTF. You are accessing a value with the JSON pointer of an extension, which does not exist in the glTF.

UX3D-nopper:
It is there:

> It is valid, to set the pointer to a property, which is not stored in the glTF.
> This is the case, when the default value is omitted.
> However, the parent object and/or extension must exist.

hybridherbst:
Updated with an explicit reference to the JSON pointer RFC – I think this could be made more clear (that there's a difference to RFC6901 in this proposal)

UX3D-nopper:
In general, in any extension or whenever we use a JSON Pointer:
The original glTF should not be altered and/or in some way generate data.

Furthermore, with the current animation system, one is targeting the node transforms even these are not stored in the glTF. That is alllowed, as the default values should not be present.

If someone uses the schema, the values are defined, but not present.

issacclee:
Hi Guys,

Just out of curious, is there anything particular that has stopped this extension from getting ratified? 
It is flagged as ready for testing, is there anything the community can do to help testing this extension? Such as produce more gltf example model with this extension?

bhouston:
Let's move this forward!  We at glTF Interactivity group, we think this is a great extension.

bhouston:
There appears to be implementations for Three.js, Babylon.js, and also Unity's glTFast (PR linked above.)

donmccurdy:
I would request that we not consider three.js a vendor implementation quite yet — Needle Tools has implemented this extension for three.js, and it's great to see that work, but I'm not sure the three.js maintainers have arrived at an opinion yet.

bghgary:
I will also note that the Babylon.js implementation has two unresolved issues for this extension.
1. We don't support dynamic updates of alpha cutoff values. It is possible to implement, but no one has ever asked for this and thus we are not sure it's that's useful.
2. Orthographic camera animations are not working. This could just be a bug in the implementation.

bhouston:
@donmccurdy  wrote:

> I would request that we not consider three.js a vendor implementation quite yet — Needle Tools has implemented this extension as an addon for three.js, and it's great to see that work, but I'm not sure the three.js maintainers have arrived at an opinion yet. Complexity of implementation is a major concern.

What is really nice about this extension is that it is a universal interface to modifying a 3D model/scene that works across engines.  This enables one to write behavior/logic once and have it work across different engines.  It is clearly the future, although I do not know how strange the Needle Tools implementation was, maybe it is just that implementation that is a problem.

I think the next generation of glTF extensions are making behavior portable.  This extension is the foundation for portable behavior.

Hackn0214:
We are developing a 3dsmax version of the glTF import/export plugin.(It is available for download from the Project Explorer page)
We are implementing the KHR_animation_pointer extension and are using AnimateAllTheThings.gltf and gltf-viewer-86cef.web.app to check the functionality.
Are there any other tools that would be useful to check the functionality?
Thanks,

Hackn0214:
It looks like there are a few settings that need to be made when outputting the animation, but I'll give it a try!
Thanks.

bghgary:
> Are there any other tools that would be useful to check the functionality?

You can use the [Babylon sandbox](https://sandbox.babylonjs.com/) to check also. There are [two issues](https://github.com/KhronosGroup/glTF/pull/2147#issuecomment-1520800128) I mentioned earlier.

Hackn0214:
Thank you, All.
This extension has not been ratified yet, but it would be nice to know the status of support for the various tools and libraries.

issacclee:
> Thank you, All. This extension has not been ratified yet, but it would be nice to know the status of support for the various tools and libraries.

I think the most comprehensive implementation of this extension belongs to the NeedleTool

BeRo1985:
> > Are there any other tools that would be useful to check the functionality?
> 
> You can use the [Babylon sandbox](https://sandbox.babylonjs.com/) to check also. There are [two issues](https://github.com/KhronosGroup/glTF/pull/2147#issuecomment-1520800128) I mentioned earlier.

The GLTF support in my PasVulkan project do also support KHR_animation_pointer, see [https://www.youtube.com/watch?v=yLIyjX81CB4](https://www.youtube.com/watch?v=yLIyjX81CB4), so if you do want a precompiled Windows or Linux binary of the PasVulkan gltfviewer example project, then just contact me, then I'll give you some binaries.

mikeskydev:
What is needed to move this forward? Are we being held up because it's not implemented in something freely available like Blender for authoring purposes? Unreal? More examples needed?

My company is looking to use this extension in production for 3D commerce, hence my work in adding it to glTFast. It would be great to get this ratified to encourage more tools to adopt it.

I already think there's a decent amount of implementations for a pre-ratified extension. 2 implementations for Unity (Covering import and export), three.js (not in main, but at least as a plugin), Babylon, Gestaltor, 3dsMax, various other independent implementations. glTF-Transform is awaiting ratification before implementing.

Hackn0214:
The 3dsMa version of the import/export plugin has also finished implementing KHR_animation_pointer and will be released once Khronos announces ratification of this extension.
We too would like to know the reason for the lack of progress in releasing this feature.

aaronfranke:
In my opinion, ratification should not be a requirement before implementation. I have interest in implementing KHR_animation_pointer in Godot Engine, and while I haven't done it yet, this is just due to other priorities, not the lack of ratification. When I do get around to it I'd be happy to provide detailed feedback on the spec (to be clear I'm not at all saying that ratification should be delayed for my or other people's feedback, Khronos should proceed how they see fit).

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -0,0 +1,318 @@
...
+* Bryce Hutchings, Microsoft [@brycehutchings](https://twitter.com/brycehutchings)
+* Don McCurdy
+* Ed Mackey, AGI @emacke
+* Felix Herbst, prefrontal cortex [@hybridherbst](https://twitter.com/hybridherbst)
+* Gary Hsu, Microsoft [@bghgary](https://twitter.com/bghgary)
+* Jamie Marconi, Microsoft [@najadojo](https://twitter.com/najadojo)
+* Joe Herdman, [@jbherdman](https://github.com/jbherdman)
+* Lewis Weaver, Microsoft
+* Marco Hutter
+* Norbert Nopper, UX3D [@UX3DGpuSoftware](https://twitter.com/UX3DGpuSoftware)
```
aaronfranke, line 16: Lists should be formatted using `-` per the Khronos recommended Markdown formatting style. See https://github.com/KhronosGroup/glTF/pull/2315 for discussion.

```suggestion
- Alexey Knyazev
- Ben Houston, Threekit
- Bryce Hutchings, Microsoft [@brycehutchings](https://twitter.com/brycehutchings)
- Don McCurdy
- Ed Mackey, AGI @emacke
- Felix Herbst, prefrontal cortex [@hybridherbst](https://twitter.com/hybridherbst)
- Gary Hsu, Microsoft [@bghgary](https://twitter.com/bghgary)
- Jamie Marconi, Microsoft [@najadojo](https://twitter.com/najadojo)
- Joe Herdman, [@jbherdman](https://github.com/jbherdman)
- Lewis Weaver, Microsoft
- Marco Hutter
- Norbert Nopper, UX3D [@UX3DGpuSoftware](https://twitter.com/UX3DGpuSoftware)
```

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -0,0 +1,318 @@
...
+```javascript
+"animations" : [
+    {
+        "channels" : [
+            {
+                "name" : "Targeting x, y, z, w for a rotation of node at index 0."
+                "sampler" : 0,
+                "target" : {
+                    "path" : "pointer"
+                    "extensions": {
```
aaronfranke, line 248: This example JSON is marked as the wrong language, is not valid JSON (missing commas), and has inconsistent formatting.

```suggestion
```json
"animations": [
    {
        "channels": [
            {
                "name": "Targeting x, y, z, w for a rotation of node at index 0.",
                "sampler": 0,
                "target": {
                    "path": "pointer",
                    "extensions": {
```

extensions/2.0/Khronos/KHR_animation_pointer/schema/glTF.KHR_animation_pointer.schema.json
```diff
@@ -0,0 +1,16 @@
...
+    "title": "KHR_animation_pointer glTF extension",
+    "type": "object",
+    "description": "The descriptor of the animated property.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+      "pointer": {
+        "type": "string",
+        "description": "JSON pointer of the object to target. The value of the path from animation.channel.target **must** be `pointer`."
+      },
+        "extensions": { },
```
aaronfranke, line 12: The schema file has inconsistent indentation.

aaronfranke (commented):
What is the reason that the OP in this PR has no description? It's a good practice to provide some context for what this is without having to read through the whole spec. At a minimum, it could link to [the previous PR](https://github.com/KhronosGroup/glTF/pull/2033).

Where can I find example files of KHR_animation_pointer? This PR does not include any. You have provided a snippet of one file that was modified to use KHR_animation_pointer, but not the whole file.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -0,0 +1,318 @@
...
+
+```javascript
+"animations" : [
+    {
+        "channels" : [
+            {
+                "name" : "Targeting x, y, z, w for a rotation of node at index 0."
+                "sampler" : 0,
+                "target" : {
+                    "path" : "pointer"
```
aaronfranke, line 247: This is not valid glTF. You can't use `"pointer"` here inside of `"path"`.

The [Animation Channel Target schema](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/animation.channel.target.schema.json) explicitly only allows `"translation"`, `"rotation"`, `"scale"`, and `"weights"` as values for `"path"`.

Any new values need to exclusively be present inside of the extension. Take a look at how [`EXT_texture_webp`](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_texture_webp#using-without-a-fallback) works. For example, this could work:

```json
"target": {
    "extensions": {
        "KHR_animation_pointer": {
            "pointer": "/nodes/0/rotation"
        }
    }
}
```

extensions/2.0/Khronos/KHR_animation_pointer/schema/glTF.KHR_animation_pointer.schema.json
```diff
@@ -0,0 +1,16 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "KHR_animation_pointer glTF extension",
```
aaronfranke, line 3: Since this extension is only allowed to be placed on the `animation.channel.target.schema.json` data structure, the title here should be `"KHR_animation_pointer glTF Animation Channel Target Extension"`, and the file should be named  `animation.channel.target.KHR_animation_pointer.schema.json`.

See these PRs for discussion: https://github.com/KhronosGroup/glTF/pull/2313 and https://github.com/KhronosGroup/glTF/pull/2314

aaronfranke (commented):
Based on the example, it looks like this extension is only allowed to be placed on the `animation.channel.target.schema.json` data structure. However, this is not explicitly stated in the README. It should be stated.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
hybridherbst: As far as I'm aware "anyOf" allows extensions to add future string values, and implementations are supposed to ignore ones they don't understand. New fields need to go into an extension.

See https://github.com/KhronosGroup/glTF/pull/899#discussion_r110480173

cc @lexaknyazev, would appreciate your input.

hybridherbst:
@aaronfranke read support for this extension has been implemented in glTFast, Babylon and three.js so far. Write support is in UnityGltf and I believe also in a PR for Blender 4. 
Thanks for the additional comments!

Most implementations are based on the test models I provided at the top of this PR, you may have missed that since GitHub folds earlier comments:
https://github.com/KhronosGroup/glTF/pull/2147#issuecomment-1132251826

I'm happy to create more test models if needed.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
javagl: I already brought that up in https://github.com/KhronosGroup/glTF/pull/2033#issuecomment-1104262883 (see bottom of that comment), with some responses by Ed and Don, but maybe Alexey still wants to chime in here.

aaronfranke: I apologize, I was mistaken. Ed is correct, one of the values is `{ "type": "string" }` so this is fine. I would click the resolve conversation button but it seems I don't have that button.

donmccurdy:
While the glTF core specification does define morph target animation, it's unfortunately restricted to animating _all or none_ of the available morph targets. Implications:

1. In a mesh with many morph targets, animation intended to target just one of them must store a large sparse accessor.
2. While any morph target animation is playing, nothing else can change the morph target weights.

This issue has come up in three.js discussion threads a few times. My understanding from this section of the proposed `KHR_animation_pointer` language ...

>  For clarification, it is not allowed to animate a single element in an array

... would be that `KHR_animation_pointer` does not lift the restriction. Is that something we should reconsider?

I have been tempted to change the three.js implementation, scanning the animation channel output for morph weights that are 0 throughout playback and removing them from the animation track. But that's a potentially costly scan, and involves some guesswork as to the intent of the model's author. 😕

lexaknyazev:
This extension is being refactored to decouple specific properties from the general design of animating their values using JSON pointers.

The updated specification for accessing glTF properties at runtime could be found at
- https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/ObjectModel.adoc

hybridherbst:
@lexaknyazev can you clarify what that means, and why?

lexaknyazev:
That document will be the centralized place for listing standard pointers to runtime properties; it will be regularly updated with future extensions.

The separation is needed because the pointer syntax and semantics are going to be used not only by keyframed animations.

hybridherbst:
Thanks, that clears it up and makes sense!

I think what I was looking for is guidance for future-specific extensions and vendor- or application-specific extensions.
For example, in the KHR_animation_pointer proposal it was noted that float/float2/float3/float4/float[] are usually animatable.

elalish (approved):
LGTM, thanks!

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -0,0 +1,318 @@
+# KHR\_animation_pointer
```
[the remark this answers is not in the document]
elalish: Did this fix get dropped?

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -0,0 +1,251 @@
...
+- Lewis Weaver, Microsoft
+- Marco Hutter
+- Norbert Nopper, UX3D [@UX3DGpuSoftware](https://twitter.com/UX3DGpuSoftware)
+
+Copyright 2024 The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
```
elalish, line 23: Release Candidate?

extensions/2.0/Khronos/KHR_animation_pointer/README.md
elalish: I believe the expanded examples below address this. When we get a model in the Sample Assets repo it'll be even more clear - maybe something with an animated texture transform or color.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
lexaknyazev: Will update after the WG approves.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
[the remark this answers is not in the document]
lexaknyazev: It's fine as is. Some other extensions also have this backslash in the title. We could eventually align all of them if needed.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
lexaknyazev: The new animation pointer extension language only defines its JSON structure and compatible accessor types. Actual list of mutable properties is defined in a separate document, which will also be used by interactivity extensions.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
bghgary: We should also revive the samples before going to RC.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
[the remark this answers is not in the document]
emackey: For consistency I think it should have both backslashes:  `KHR\_animation\_pointer`  It seems odd for it to have only one.

When extension names appear in Markdown outside of code-quotes, the backslash helps distinguish the middle of the name from italic text.

extensions/2.0/Khronos/KHR_animation_pointer/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension provides a standardized way of animating arbitrary glTF properties according to the **glTF 2.0 Asset Object Model**.
```
emackey, line 31: Can this be a link to the document?

aaronfranke:
The example models here https://github.com/KhronosGroup/glTF-Sample-Models/pull/353 were not merged, and glTF-Sample-Models was archived. The new repo, glTF-Sample-Assets, is missing these examples. https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models

Please ensure that the example files are not lost, we should merge them into the sample assets repo.

DRx3D:
@aaronfranke : The models are not lost. Updated versions of these models have been submitted as PR #106 to SampleAssets. That PR cannot be merged because the quality checks have failed and the PR is marked as DRAFT.

aaronfranke:
@DRx3D Ah thank you, I forgot to check the PRs on that repo. https://github.com/KhronosGroup/glTF-Sample-Assets/pull/106

aaronfranke:
Godot 4.4 supports KHR_animation_pointer: https://godotengine.org/releases/4.4/

https://docs.godotengine.org/en/latest/classes/class_gltfobjectmodelproperty.html#class-gltfobjectmodelproperty

PR #2163, opened 2022-05-22
Missing hard break

dmurdoch:
"unsigned byte" and "unsigned short" should be on separate lines; sometimes they are rendered as

```
unsigned byte unsigned
short
```

which is confusing.


lexaknyazev:
We'll include this fix in the next spec release.

PR #2170, opened 2022-06-08
KHR_anchors_plane

bhouston:
This draft extension aims to make glTF equivalent with USDZ in terms of its ability to internally specify plane anchor points within the home, such as floor/ground, ceiling and wall.  It only supports these vertical or horizontal planes, it does not support slopes.


donmccurdy:
Are you aware of the `preferredSurfaces` property available in the [`model3d` XMP namespace](https://github.com/KhronosGroup/3DC-Metadata-Recommendations/blob/main/model3d.md#model-3d-namespace)? This is [available in glTF now](https://twitter.com/donrmccurdy/status/1529448113597034498). I don't know whether AR viewers have implemented support or whether modelviewer can currently pass it through when autogenerating a USDZ file for iOS devices.

In my opinion, the advantages of using XMP for AR-related metadata are:

1. AR metadata can be generalized beyond glTF
2. More practical versioning system than extensions / easier updates for rapidly-evolving AR use cases
3. Possibly easier pass-through of XMP metadata in processing tools that aren't AR-related and might not implement an extension

The main limitation of the XMP approach is that while XMP packets can be _attached_ to glTF objects like scenes or nodes, a packet cannot _reference_ other nodes conveniently, e.g. by index.

bhouston:
I think that the XMP information is good for the basic AR use case for planes/surfaces.  Let's go in that direction for this extension.  We just have to better publicize it.  I think a blog post would be useful on this topic to really push it forward.

The "face" option is incorrect and should be removed from this list.  The face is not a surface.  It requires additional information to specify putting something on a face correctly.  ( I've done a bunch of face try-on work but it isn't public, but if you scroll down there are face virtual try-on examples in gif form: https://www.threekit.com/immersive-shopping-suite.  I've tried to codify the simplest method of face landmark try-on here, but I guess it should probably now be converted into XMP data instead: https://github.com/KhronosGroup/glTF/pull/2171 )

donmccurdy:
Sorry, I don't know the answers to that. My guess would be that this repository is still the right place to discuss anchoring-related additions to the XMP namespace.

PR #2190, opened 2022-07-27
Move REUSE Compliance into a separate action

lexaknyazev:


javagl (approved):
I don't know all of the technical background of GH actions, REUSE, and the Khronos Docker container. It looks like the license check that was previously run from _within_ the docker container is now run as a separate GH action (and additionally, it is run on pull requests - which would have avoided some confusion earlier today). So ... as far as I can tell, this looks fine to me...

lexaknyazev:
- Having REUSE as a separate GitHub Action leads to separate entries in the "checks" section below, making them more visible to contributors.
- Instead of hard-coding the REUSE check command, we rely on the GitHub Action provided by FSFE.

Issue #948, opened 2017-05-07
glTF 2.0: KHR_materials_displacements extension


McNopper:
Regarding displacement, I suggest the following extension. This one is for geometry displacement. If another displacement technique needs to be defined e.g. parallax, a displacementParallax could be defined.

```
{
  "materials": [
    {
      "displacementGeometry": {
        "displacementGeometryFactor": 1.0,
        "displacementGeometryTexture": 0,
        "displacementGeometryOffset": 0.0
      }
    }
  ]
}
```

From the texture, the r channel is used. The factor is used to multiply the channel. If no texture is present, the value is take as it is. Finally, the offset is added to the above value.

emackey:
+1 for displacementParallax 😃

McNopper:
I think, even the parameters can be reused:
```
{
  "materials": [
    {
      "displacementParallax": {
        "displacementParallaxFactor": 1.0,
        "displacementParallaxTexture": 0,
        "displacementParallaxOffset": 0.0
      }
    }
  ]
}
```

Factor is also know as the height scale. Also an offset would make sense.

pjcozzi:
@McNopper how eager are you to finish this extension?  Should we focus on the lights, environment map, and perhaps static lightmap extensions before this one to best support the Blender exporter?

Anteru:
Wouldn't it make sense to allow 3D displacement as well, in which case the RGB channels are used, and not just R?

McNopper:
@pjcozzi It is also implemented experimental in the Blender exporter.
However, I suggest we focus on and finalize the light and common material before end of July.
Regarding the displacement, environment, lightmap and possible other extensions, we should give us some time and e.g. finalize them before GDC 2018.
For the "postponed" extensions, we still can implement them in an experimental way in the Blender exporter.

For common and light extension, I recomend we should fully finalize them including schema and so on.

@Anteru The original idea having only the R channel was to displace the geometry depending on the normals. One reason for also postponing this extension is, that we somehow can encode the displacement, that the engine can finaliy decide, if it wants to "really" displace the geometry by using tessellation or "just" displace it by using e.g. parallax mapping. The RGB channels, do you want to use them as the displacement factor for each axis?

Also, as many more textures would be included, we should give us some time on how we could pack them. Futhermore, we should evaluate, how all the major engines normally "expect" the displacement data.

Anteru:
[...]

PR #2196, opened 2022-08-18
KHR_parallax_mapping

maierlars:
Hi, before I continue working on this, I would like to know if there is any chance that this will be accepted.
There is [this](https://github.com/KhronosGroup/glTF/issues/948) thread (5yrs old by the way) that ends up in every google search and people are still asking, but it doesn't look like anything has been done.

The thread also mixes parallax mapping and actual displacement mapping. While I think both is nice to have, I don't think we want to have one extension that does everything. And since I'm more interested in parallax mapping, I went for it.

I'm doing this without being attached to any company, I hope this is not a problem.
Any help is welcome!

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_parallax_mapping/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_parallax_mapping/README.md
@@ -1,5 +1,5 @@
 <!--
-Copyright 2015-2021 The Khronos Group Inc.
+Copyright 2015-2022 The Khronos Group Inc.
 SPDX-License-Identifier: CC-BY-4.0
 -->
 
@@ -49,8 +49,11 @@ This extension does _not_ require the geometry of the object this material is ap
 
 ## Known Implementations
 
-None.
+Although no implementations of this extension are known, these techniques are used widely by the industry.
 
 ## Resources
 * [Introduction](https://en.wikipedia.org/wiki/Parallax_mapping)
 * [Discussion about different implementations](https://learnopengl.com/Advanced-Lighting/Parallax-Mapping)
+* [GPU Gems 2 - Per-Pixel Displacement Mapping with Distance Functions](https://developer.nvidia.com/gpugems/gpugems2/part-i-geometric-complexity/chapter-8-pixel-displacement-mapping-distance-functions)
+* [GPU Gems 3 - Relaxed Cone Stepping for Relief Mapping](https://developer.nvidia.com/gpugems/gpugems3/part-iii-rendering/chapter-18-relaxed-cone-stepping-relief-mapping)
+* [SIGGRAPH Asia 2021 - Tessellation-Free Displacement Mapping for Ray Tracing](https://perso.telecom-paristech.fr/boubek/papers/TFDM/)
--- a/extensions/2.0/Khronos/KHR_materials_parallax_mapping/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_parallax_mapping/README.md
@@ -1,0 +1,56 @@
+<!--
+Copyright 2015-2021 The Khronos Group Inc.
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# KHR\_materials\_parallax\_mapping
+
+## Contributors
+
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+Many materials provide a height-map that can either be used for real geometry displacement or for parallax-mapping.
+This extension defines a way to specify a texture as a height map.
+
+## glTF Schema Updates
+
+The height at a given position is calculated by sampling the given texture and using the red channel. The sampled value is then multiplied with the `heightFactor`.
+The result is used in an implementation defined way to offset all other texture lookups. (See Resources)
+
+This extension does _not_ require the geometry of the object this material is applied to to be modified.
+
+```json
+{
+    "materials":[
+        {
+            "extension": {
+                "KHR_material_parallax_mapping": {
+                    "heightTexture": 0,
+                    "heightFactor": 0.5,
+                }
+            }
+        }
+    ]
+}
+```
+
+### JSON Schema
+
+- [parallaxMapping.schema.json](schema/parallaxMapping.schema.json)
+
+## Known Implementations
+
+None.
+
+## Resources
+* [Introduction](https://en.wikipedia.org/wiki/Parallax_mapping)
+* [Discussion about different implementations](https://learnopengl.com/Advanced-Lighting/Parallax-Mapping)
--- a/extensions/2.0/Khronos/KHR_materials_parallax_mapping/schema/parallaxMapping.schema.json
+++ b/extensions/2.0/Khronos/KHR_materials_parallax_mapping/schema/parallaxMapping.schema.json
@@ -1,0 +1,23 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "parallaxMapping",
+    "type": "object",
+    "description": "Reference to a heightmap for parallax mapping.",
+    "allOf": [ { "$ref": "glTFChildOfRootProperty.schema.json" } ],
+    "properties": {
+        "heightTexture": {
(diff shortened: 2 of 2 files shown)


BruceCherniak:
I too would be interested in such an extension as a way to greatly increase apparent model detail without a large increase in model size/complexity.

The approach in GPU Gems 2 and many other places is not without caveats, but provides PM in a lightweight way:
https://developer.nvidia.com/gpugems/gpugems2/part-i-geometric-complexity/chapter-8-pixel-displacement-mapping-distance-functions

The "Relaxed Cone Stepping for Relief Mapping” in GPU Gems 3 is also interesting as a “cheap” displacement map technique:
https://developer.nvidia.com/gpugems/gpugems3/part-iii-rendering/chapter-18-relaxed-cone-stepping-relief-mapping

Also interesting is the "Tessellation-Free Displacement Mapping for Ray Tracing” technique presented at SIGGRAPH Asia 2021, although more difficult to implement and not suitable for all backends:
https://perso.telecom-paristech.fr/boubek/papers/TFDM/

Cheers,
Bruce

On Aug 18, 2022, at 8:32 AM, Lars Maier ***@***.******@***.***>> wrote:


Hi, before I continue working on this, I would like to know if there is any chance that this will be accepted.
There is this<https://github.com/KhronosGroup/glTF/issues/948> thread (5yrs old by the way) that ends up in every google search and people are still asking, but it doesn't look like anything has been done.

The thread also mixes parallax mapping and actual displacement mapping. While I think both is nice to have, I don't think we want to have one extension that does everything. And since I'm more interested in parallax mapping, I went for it.

I'm doing this as a private person, I hope this is not a problem.
Any help is welcome!

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/KhronosGroup/glTF/pull/2196

Commit Summary

  *   0f08674<https://github.com/KhronosGroup/glTF/pull/2196/commits/0f08674d165198976d7b83d61c8f87d17af51a6c> First draft.

File Changes

(2 files<https://github.com/KhronosGroup/glTF/pull/2196/files>)

  *   A extensions/2.0/Khronos/KHR_materials_parallax_mapping/README.md<https://github.com/KhronosGroup/glTF/pull/2196/files#diff-0f8b396663224e5db193eef38ab2355027f243757f7849ae62127dcb679589fb> (56)
  *   A extensions/2.0/Khronos/KHR_materials_parallax_mapping/schema/parallaxMapping.schema.json<https://github.com/KhronosGroup/glTF/pull/2196/files#diff-6e1f8d3244eea8899049b8fe49458e8384d174728164f7f0d6ceaee85cbc637f> (23)

Patch Links:

  *   https://github.com/KhronosGroup/glTF/pull/2196.patch
  *   https://github.com/KhronosGroup/glTF/pull/2196.diff

maierlars:
Thanks for providing additional resources.

maierlars:
Almost 3 weeks later. How do we proceed with this pull request? @UX3D-nopper is there an internal discussion about this? Is there anything I can provide to you?

UX3D-nopper:
There is a dedicated group at Khronos, called PBR TSG, which is taking care of material extensions. As there are currently so much pull requests and suggestions, this will probably take a while.

maierlars:
Thanks for your response! Then I will be patient. 🙂

donmccurdy:
I'd just add — one helpful way to contribute on proposed extensions can be to research the relevant feature support in (1) authoring tools, and (2) viewing engines. Standards are typically not designed "from scratch". Knowing the similarities and differences among popular existing or upcoming software helps to move things forward.

maierlars:
Engine Support:
- [Unity](https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@7.1/manual/SGNode-Parallax-Occlusion-Mapping.html)
- [Unreal](https://docs.unrealengine.com/udk/Three/DevelopmentKitGemsParallaxOccludedMapping.html)
- [CryEngine](https://docs.cryengine.com/display/SDKDOC2/Parallax+Occlusion+Mapping)

In fact, if an engine provides support for custom shaders, parallax occlusion mapping is only limited by the rendering backend.

Authoring Tools:
- Substance Designer outputs a heightmap by default.
- Most texture packs available only usually come with a height map.
- Blender supports baking height maps.

I think, that this is a common technique and is widely supported.

extensions/2.0/Khronos/KHR_materials_parallax_mapping/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Overview
+
+Many materials provide a height-map that can either be used for real geometry displacement or for parallax-mapping.
+This extension defines a way to specify a texture as a height map.
+
+## glTF Schema Updates
+
+The height at a given position is calculated by sampling the given texture and using the red channel. The sampled value is then multiplied with the `heightFactor`.
+The result is used in an implementation defined way to offset all other texture lookups. (See Resources)
```
nicopap, line 27: It might be worthwhile to specify the direction of the offset.  (is white the topmost height or bottommost height?) Afaik Godot has white goes down, and calls the texture a "depth map"

nicopap:
As #948 can attest, there is a lot of interest for the addition of height maps to the standard PBR material. As noted earlier in this discussion, parallax mapping is widely supported in existing engines, furthermore, modelling tools like Blender are well capable of exporting height maps (in any case, if a tool is capable of exporting a normal map, it is likely capable of exporting a height map), so the implementation burden for API consumers seem minimal.

Some additional context from #948:
* [There's more uses you can do out of a height/bump channel, such as calculating more accurate occlusion and shadowing.](https://github.com/KhronosGroup/glTF/issues/948#issuecomment-844978729)
* [I now think that in the long run, as hardware improves, displacement mapping may "win" over parallax mapping, for most cases. So we might only need a displacement extension, but we should consider whether that extension is asking viewers to subdivide or tessellate the base mesh, and whether that can be combined with subdivision surfaces.](https://github.com/KhronosGroup/glTF/issues/948#issuecomment-936736624)
* [To that end, the burden of converting all bump maps to tangent-space normal maps is intentionally placed on the glTF creation/export toolchain, not on the receiving client implementation.](https://github.com/KhronosGroup/glTF/issues/948#issuecomment-612992282)

Note that all the questions around the materials_displacement extension was about the specific [displacement mapping implementation](https://en.wikipedia.org/wiki/Displacement_mapping) **which is not relevant for this proposition**. Here, we are talking about [parallax mapping](https://en.wikipedia.org/wiki/Parallax_mapping), a technique that **doesn't require any additional geometry**.

If the [gltf rust crate](https://crates.io/crates/gltf) were to support this as an extension, would it be enough of a proof of concept to start turning the gears of getting this into the standard?

extensions/2.0/Khronos/KHR_materials_parallax_mapping/README.md
maierlars: Good point! I will add that. Would it make sense to fix a certain behavior (i.e. 0 is lowest, 1 is highest) or have it configurable. Translating one to the other is trivial.

Alternatively, one could allow the `heightFactor` to be negative.

Not sure what is best. Any opinions/ideas?

nicopap: I spent the two last days on the question lol. Given it's a **height** map, of course white should be top (when height is at high value, it should be tall etc.). This goes against Godot's implementation (my own, and I imagine most people's), but it's just a question of doing `1.0 - height` in the shader, which I doubt would introduce an unacceptable overhead, especially given that parallax mapping algorithm oftentimes sample multiple times at different positions the height map per fragment.

nicopap:
I've looked in depth at [mmoeller's implementation for Blender](https://devtalk.blender.org/t/parallax-occlusion-mapping/15774) and I have one additional comment:

- `heightFactor` is a bad name, I know it follows the gltf established convention, but it doesn't make sense. It should express the units of displacements (probably world units I guess?)

maierlars:
Another 1.5 yrs later. Does anyone know if there is something going on inside khronos with respect to parallax mapping? Has this proposal or any variant of it been considered?

PR #2215, opened 2022-10-13
`KHR_lights_punctual`: Clarify relationship between intensity and color. Also some grammar.

will-ca:
In #2214 I ended up also making explicit the presumed separation between light intensity and light colour. (I.E. Light colour does not get normalized to match the intensity.) I also did some minor grammar and formatting changes/fixes.

This `cherry-pick`s those changes.

This is not meant to change what the specification says, only make clearer and explicit what I believed was already presumed.


emackey (approved):
We talked about this on the PBR TSG call today and it looks like a useful clarification, thanks!

emackey:
@UX3D-haertl We have a few more clarifications in the works, but this PR adds the following text:

> The `intensity` represents the luminous intensity that the light would emit if it were colored pure white (`[1.0, 1.0, 1.0]`). The `color` property acts as a wavelength-specific multiplier.

I think this should help clarify the lumens / watts conversation.  Clearly we weren't the first to have this issue.

PR #2221, opened 2022-10-31
WIP: KHR_lights_photometric

BruceCherniak:
This extension enables the use of photometric light profiles as light sources within a scene.  Photometric light sources model the distribution of measured light intensity or of relative illuminating power.  Two popular formats for expressing photometric lights are IES and EULUMDAT.  Intensity distributions may also be generated procedurally.

This extension creates a glTF json schema that conveys the lighting distribution information, agnostic of the the file format; and will provide sample code to create simple file importers and converters.

### DIFF
--- a/extensions/2.0/Khronos/KHR_lights_photometric/README.md
+++ b/extensions/2.0/Khronos/KHR_lights_photometric/README.md
@@ -1,0 +1,169 @@
+# KHR_lights_photometric
+
+## Contributors
+
+* Bruce Cherniak, Intel, bruce.cherniak@intel.com
+* Timothy Blut, NVIDIA, tblut@nvidia.com
+* Lutz Kettner, NVIDIA, lkettner@nvidia.com
+* ... (many more)
+
+Copyright (C) 2018-2022 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
+<!--TODO: Draft or Stable-->
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension enables the use of photometric light profiles as light sources within a scene.  Photometric light sources model the distribution of measured light intensity or of relative illuminating power.  Two popular formats for expressing photometric lights are IES and EULUMDAT.  Intensity distributions may also be generated procedurally.
+
+This extension creates a glTF json schema that conveys the lighting distribution information, agnostic of the the file format; and provides sample code to create simple file importers and converters.
+
+
+(EULUMDAT is a data file format used for specification of photometric data especially intensity distributions from light sources such as lamps and luminaries. The file extension is .ldt. The format was proposed by Axel Stockmar (Light Consult Inc., Berlin) in 1990. The format is the European equivalent to the IES file format specified in IESNA LM-63.)<sup>1</sup>
+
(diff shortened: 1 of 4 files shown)


paularden-onshape:
These comments are made on the basis that the schema used here and description should be able to stand on its own without incorporating by reference anything from the IESNA LM-63-* or other de-facto standards. Of course the intent is to provide conversions for those, however in theory someone should be able to use it independently.

Some of the things I mention might be intended to be added in the TODO parts, so apologies of that is already planned. I think some of these are questions that need to be resolved before completing those however.

The assumptions around where photometric zero is located with respect to the glTF coordinate system need to be documented fully, if not people will make differing assumptions. I would suggest that like `KHR_lights_punctual` the nadir (vertical 0°) point in the `-Z` direction (then the wording can be borrowed from that extension as well). There is no equivalent for horizontal 0° since you can't have asymmetry in that extension, however I would suggest the `+X` direction makes the most sense for this based on what I've seen of other implementations.

This would mean the `X/Y` plane is considered the horizontal photometric plane and the `X/Z` plane is considered the vertical photometric plane for all angular data specified.

Furthermore the direction in which the angles proceed needs to be defined (CW or CCW). This is obviously only relevant for horizontal angles. In IES-63-* this direction differs depending on the photometric type (A, B or C). I'd suggest choosing the direction used for type C as it is the most common (counter-clockwise). This will be important for understanding how the `orientation` property should be applied as well as how IESNA and EULUMDAT data should be re-oriented during conversion.

On that topic, is this representation to be considered fully unwrapped (i.e., no symmetry)? If so it should be made clear that the angular data must cover the [0 - 360] range of horizontal and [0-180] range of vertical angles. Will there be a minimum number of data points? Given only a count of the number of horizontal and vertical angles is provided I am assuming the intent is that it is unwrapped (which is much simpler for implementers to use for sure).

I'm definitely in favour of this representation being unwrapped since a huge part of the complexity of dealing with IES files is interpreting the symmetry and different orientation types. If not unwrapping symmetries then symmetry properties would also need to be stored. Ideally candela values should be copied directly by symmetry when unwrapping in conversion and not re-sampled on a regular grid or anything like that.

Currently it is not specified which order the angles are specified in (i.e., whether it is horizontal x vertical or vertical by horizontal) so this should also be made explicit as well. Is the flat structure intended primarily for easy mapping from existing formats? If not then perhaps it makes sense to use a two dimensional array to reflect the nature of the data. If this were done, in theory the sizes wouldn't be needed either.

Are the data points intended to be candela values or unit-less normalized values (if the later, normalized to what, maximum luminous intensity or something else)? Defining this affects the definition of the multiplier and ultimately how united quantities would be obtained if that is a use case being considered.

When using LM-63 based photometry it is common practice to use the various multipliers, lamp lumens, ballast factor etc to manipulate the total flux of the luminaire to some known total output power. Without knowing the original flux, this isn't possible with normalized values since information is lost (at least not without first integrating the normalized power over the full range of angles). So either that would need to be stored as additional data or instead of a normalized distribution, a distribution which sums to unity could be used instead (with the advantage of giving a ready made PDF for sampling if needed).

All that is to say I think more definition is required around what is actually being stored in the data and whether we care about units. If the only use cases being targeted are "pretty" light sources then an arbitrary normalised set of values and unit-less multiplier are of course fine. Like a HDRI environment, which may or may not be radiometrically calibrated (usually not), luminous intensity data could be seen as arbitrary or calibrated.

emackey:
At the Nov 7th PBR TSG meeting, there was general agreement that the JSON array `data` here would be better expressed using a glTF `accessor` and `bufferView` pair.

BruceCherniak:
I have closed this PR as I will have little/no time for further pursue such extension.

Issue #2222, opened 2022-11-03
Add MAGNUM vendor prefix


mosra:
Hi, I'd like to request the `MAGNUM` vendor prefix for the Magnum Engine. We're working on extensions for creating multi-draw-optimized glTF files.

Prefix: `MAGNUM`
Name: Magnum Engine
URL: https://magnum.graphics/
Contact: info@magnum.graphics

Thank you! :)

PR #2224, opened 2022-11-05
Add MAGNUM prefix

javagl:
Fixes https://github.com/KhronosGroup/glTF/issues/2222


javagl:
If this was merged, I could add https://github.com/KhronosGroup/glTF/issues/2230 . I could also add the other prefix in this PR, but in any case, someone will have to hit the "Reviewed this" button...

lexaknyazev (approved):

PR #2225, opened 2022-11-06
Documenting the extension development process

javagl:
There is a short description about the process for developing extensions in the [current `extensions/README.md`, "About glTF Extensions"](https://github.com/KhronosGroup/glTF/blob/757d942a89810840645e62254b09357dbf767fdc/extensions/README.md#about-gltf-extensions) section. This description only mentions a few aspects of the process (mainly some specific _technical_ ones). 

The process could be described more clearly, and on a _methodological_ level. This PR is an early draft for such a description. 

Direct preview: https://github.com/javagl/glTF/blob/extension-process-documentation/extensions/README.md


javagl:
The actual extension development process is pretty much: "Create a PR, and see what happens next". Although I think that there are several points that _should_ be clarified, it's unlikely that this is going to happen, and even less likely that it is going to happen _here_, so there probably isn't much use in keeping this open.

PR #2228, opened 2022-11-20
EXT_node_lod extension proposal

takahirox:
Working repository: https://github.com/takahirox/EXT_node_lod

I would like to propose `EXT_node_lod` extension that allows LODs in glTF.

I know already [`MSFT_lod`](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/MSFT_lod) extension exists for LOD. The new extension is less flexible but simpler than `MSFT_lod` so it should be easier for the handlers (implementations) to handle.

**JSON structure and examples**

`EXT_node_lod` can be defined at the `node` level.

```json
"nodes": [{
    "name": "High_LOD",
    "mesh": 0,
    "extensions": {
        "EXT_node_lod": {
            "lod": [
                {"node": 1, "coverage": 0.5},
                {"node": 2, "coverage": 0.2},
            ],
        },
    },
},
{
    "name": "Medium_LOD",
    "mesh": 1,
},
{
    "name": "Low_LOD",
    "mesh": 2,
}]
 ```

**Difficulty of handling `MSFT_lod`**

While I was writing [a handler for `MSFT_lod`](https://github.com/takahirox/three-gltf-extensions/tree/main/loaders/MSFT_lod) I realized that `MSFT_lod` has some difficulties to handle.

* Both `node` and `material` can define LOD. The behavior is unstated in the specification if `material` containing LOD is referred by `node` containing LOD.
* A `node` can specify other `node`s so LOD `node` can form nest or even infinity loop. And the behavior for them is unstated in the specification.
* Threshold for switching LOD is defined in `extra` as optional. Unclear how to switch LOD for applications/viewers/engines if it is undefined because they don't know the qualities of LODs.

**Simplification ideas**

I would like to suggest to simplify the LOD specification. It will lose some flexibilities but be simpler to write an implementation. The differences from `MSFT_lod` specification are

* Allow to define LOD only at node level and drop material LOD. This change can avoid the case where both node and material have LOD. Material LOD capability is still covered by node LOD (refer to the same geometry attributes and refer to different materials).
* Add/Clarify some LOD node restrictions for simplicity.
  * Nested LOD nodes are disallowed
  * Child nodes are part of that LOD
  * LOD nodes are regarded as part of the scene node graph
* Require threshold values. This can avoid unknown threasholds for applications/viewers/engines.

**Additional notes**

* With the simplification suggestions, the extension will be easier to be handled so I think it would be good to be a multi-vendor extension (`EXT_` prefix) rather than a vendor extension.

**Discussions**

I'm still thinking how threshold values should be, screen coverage, distance from camera, or others. In the current PR, I chose screen coverage for now just to follow `MSFT_lod` spec. Probably compatibility is important so we may need to choose the generic one that can be handled consistently across applications/viewers/engines.

https://github.com/takahirox/EXT_node_lod/issues/1

**Related threads**

* https://github.com/KhronosGroup/glTF/issues/1045
* https://github.com/KhronosGroup/glTF/issues/1743
* https://github.com/KhronosGroup/glTF/issues/1952
* https://github.com/KhronosGroup/glTF/issues/1953

### DIFF
--- a/extensions/2.0/Vendor/EXT_node_lod/README.md
+++ b/extensions/2.0/Vendor/EXT_node_lod/README.md
@@ -1,0 +1,234 @@
+# EXT\_node\_lod
+
+## Contributors
+
+* Takahiro Aoyagi, Mozilla, [@takahirox](https://github.com/takahirox)
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This `EXT_node_lod` extension adds the ability to specify various Levels
+of Detail (LOD) to a glTF asset. An implementation of this extension can
+use the LODs for various scenarios like rendering a different LOD based 
+upon the distance of the object or progressively loading the glTF asset
+starting with the lowest LOD.
+
+This extension allows an LOD to be defined at `node` level. The LODs can
+specify different geometry and materials across various levels.
+
+The extension is added to the highest LOD level. The extension defines an
+`lod` property which is an array containing the objects including the
+`node` property that points to the index of the `nodes` as LOD node and
+the screen coverage number as a hint to determine the switching threshold
+for the LOD.
+
+Every `node` index in the array of the extension points to the index of
+another `node` object which should be used as the LOD that is lower level
+than the previous array element.
+
+An implementation of this extension can parse through the `lod` array of
+the extension to create the list of LOD levels available for an asset.
+If an asset containing this extension is loaded in a client that doesn't
+implement the extension, then the highest LOD level will be loaded and
+all lower LODs will be ignored.
+
+## Examples
+
+```json
+"nodes": [
+    {
+        "name": "High_LOD",
+        "mesh": 0,
+        "extensions": {
+            "EXT_node_lod": {
+                "lod": [
+                    {"node": 1, "coverage": 0.5},
+                    {"node": 2, "coverage": 0.2},
+                ],
+            },
+        },
+    },
+    {
+        "name": "Medium_LOD",
+        "mesh": 1,
+    },
+    {
+        "name": "Low_LOD",
+        "mesh": 2,
+    },
+]
+```
+
+In this node, *node[0]* is the highest LOD. The first element in the `lod` array,
+*node[1]*, is the next lower LOD (or the Medium_LOD). The second element in the
+`lod` array, *node[2]*, is the lowest LOD.
+
+## Screen coverage
+
+The `coverage` property in the `lod` array specifies a screen coverage
+value that can be used as a hint to determine the switching threshold for the
+various LODs.
+
+Screen coverage is a ratio between `0.0` and `1.0` representing how big the
+mesh's rendered screen surface area is in the total screen surface area.
+
+The highest LOD `coverage` value is implied to `1.0`. The `coverage` value
+must be between `0.0` and `1.0`, and must be between the next higher LOD
+`coverage` value and the next lower LOD `coverage` value.
+
+The LOD is expected to be selected to be rendered if the current screen
+coverage is between the LOD's `coverage` value and the next lower LOD
+`coverage` value (or 0 if it is the lowest LOD).
+
+In the example above the screen coverage values can be used as follows:
+
+* LOD0(high) Rendered from 1.0 to 0.5
+* LOD1(medium) Rendered from 0.5 to 0.2
+* LOD2(low) Rendered from 0.2 to 0
+
+## LOD node restriction
+
+For simplicity there are some restrictions for LOD nodes.
+
+### Nested LOD nodes are disallowed
+
+A node referred by a `node` of a `lod` array element of the `EXT_node_lod` extension
+or its descendant nodes **MUST NOT** define the `EXT_node_lod` extension.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by vpenades** [comment]
Question: at which point a node would stop being visible?

Based on the current implementation, I guess the last LOD, the one with smallest coverage, would point to an empty node?

Otherwide, I would add a property called "minCoverage"  that would define at which point the node should not be rendered at all.

minCoverage should be smaller than the smallest coverage of any LOD, or 0 by default

**Comment by takahirox** [comment]
> I guess the last LOD, the one with smallest coverage, would point to an empty node?

Yes because I think the implementation would be simple regardless of whether including invisible LOD like the following pseudo code.

```
const lod = new LOD();
lod.addLevel(loadNode(nodeIndex));
for (let i = 0; i < extension.lod.length; i++) {
  lod.addLevel(loadNode(lod[i].node, lod[i].coverage));
}
```

> Also, How coverage is calculated? I guess it would be based on spheres? in which case maybe it could be useful having a Radius as a hint?

Perhaps sphere or box.

This LOD extension switches node so to calculate the spade the implementation needs to traverse descendant nodes. It may be troublesome so adding a hint (as required?) may be good. Or distance based coverage may be good because no calculation is needed.

**Comment by vpenades** [comment]
> Yes because I think the implementation would be simple regardless of whether including invisible LOD like the following pseudo code.
> 
> ```
> const lod = new LOD();
> lod.addLevel(loadNode(nodeIndex));
> for (let i = 0; i < extension.lod.length; i++) {
>   lod.addLevel(loadNode(extension.lod[i].node), extension.lod[i].coverage);
> }
> ```

My concern is that currently, glTF validator gives warnings on empty nodes, and many tools not aware of the extension might "optimize" the models by removing empty nodes. My feeling is that from one reason or another, glTF discourages empty nodes and preffers explicit and concise definitions.

> 
> > Also, How coverage is calculated? I guess it would be based on spheres? in which case maybe it could be useful having a Radius as a hint?
> 
> Perhaps sphere or box.
> 
> This LOD extension switches node so to calculate the node sphere/box the implementation needs to traverse descendant nodes. It may be troublesome so adding a hint (as required?) may be good. Or distance based coverage may be good because no calculation is needed.

The whole logic of the extension sits on top of the logic to switch between LODs, and this logic needs to be precisely defined to avoid multiple engines behaving diferently because they did different interpretations.

So if you use "coverage"... what it means, coverage of what? the whole geometry? a bounding sphere?  if two engines calculate the sphere in a different way, the end result would be that the coverage would be different in both engines, and as a consequence, the two engines would trigger the LOD switch at different times. Tha'ts why I think that if Coverage is used, it should be defined in the extension to guarantee consistency across engines.

Distance would be an alternative option. But distance based LODs are sensitive to the screen size, so when rendering very large images would trigger lower LODs to soon, resulting in a loss of quality.  If anything I would use Distance divided by screen height, to ensure similar behavior for different image sizes. Again, the extension need to define this to ensure consistest behavior across multiple engines

**Comment by takahirox** [comment]
> My concern is that currently, glTF validator gives warnings on empty nodes, and many tools not aware of the extension might "optimize" the models by removing empty nodes

Some potential options in my mind are

1. The extension doesn't care about it. Empty node should be valid in the spec (, shouldn't it?). Encourage tools for allowing empty nodes. @donmccurdy Any thought about this from tooling end (gltf-transform) ?
2. The extension doesn't allow empty node. Users may use a node with a mesh that has like (0,0,0), (0,0,0), (0,0,0) positions for invisible LOD.
3. The extension allows invisible LOD in the extension definition like `"lod": [{"node": null, "coverage": 0.5}]` or minCoverage as you suggested.

> So if you use "coverage"... what it means, coverage of what?

The goal is to clearly define coverage perhaps with fomula.

**Comment by takahirox** [comment]
I need a double check because I'm not an expert of 3D CG algebra but perhaps screen coverage ratio can be calculated with this formula? Assuming sphere.

```
radius = length((projectionMatrix * viewMatrix * modelMatrix * vec4(radius, 0, 0, 1.0)).xyz);
ratio = radius * radius * pi / 4.0;
```

**Comment by takahirox** [comment]
Thinking if different radiuses per LOD level should be allowed and where bounding sphere radiuses for LODs should be defined.

Radiuses for lower LODs can be defined in elements of `EXT_node_lod.lod` array like

```
"nodes": [{
    "extensions": {
        "EXT_node_lod": {
            "lod": [
                {"node": 1, "coverage": 0.5, radius: 10.0},
                {"node": 2, "coverage": 0.2, radius: 10.0},
            ],
        },
    }
},
...
]
```

But where for the highest LOD. `EXT_node_lod.radius` may be another option (if we don't allow different radius per LOD level).

```
"nodes": [{
    "extensions": {
        "EXT_node_lod": {
            "lod": [
                {"node": 1, "coverage": 0.5},
                {"node": 2, "coverage": 0.2},
            ],
            "radius": 10.0,
        },
    }
},
...
]
```

takahirox:
If glTF vs [glXF](https://github.com/KhronosGroup/glXF) will end up with self-contained vs scene (composition), defining the extension not in glTF but in glXF might be more straightforward and intuitive.

PR #2235, opened 2022-12-22
EXT_texture_avif

leon:
This extension adds support for .avif textures.

I was really hoping for jpeg xl to win the race, but google pulled it from chrome.
so the next best image format that is getting good support in browsers
https://caniuse.com/avif
is AVIF

This extension is a copy of the webp extension, but I've rewritten parts of the readme, and created a simple sample with and without png fallback.

I don't know if there are any special considerations we need to take regarding avif.

let me know if you would like me to rework anything.

### DIFF
--- a/extensions/2.0/Vendor/EXT_texture_avif/README.md
+++ b/extensions/2.0/Vendor/EXT_texture_avif/README.md
@@ -2,12 +2,13 @@
 
 ## Contributors
 
-- Omar Shehata, Cesium, [@OmarShehata](https://github.com/OmarShehata)
-- Pascal Massimino, Google, [@Skal65535](https://github.com/skal65535)
+- Leon Radley, [@leon](https://github.com/leon)
+- Don McCurdy, [@donmccurdy](https://github.com/donmccurdy)
+- Alexey Knyazev, [@lexaknyazev](https://github.com/lexaknyazev)
 
 ## Status
 
-Complete
+Draft
 
 ## Dependencies
 
@@ -17,8 +18,7 @@ Written against the glTF 2.0 spec.
 
 This extension allows glTF models to use AVIF as a valid image format. A client that does not implement this extension can ignore the provided AVIF image and continue to rely on the PNG and JPG textures available in the base specification. Defining a fallback texture is optional. The [best practices](#best-practices) section describes the intended use case of this extension and the expected behavior when using it without a fallback texture.
 
-AVIF is an image format developed by Netflix that uses the AV1 VideoCodec [specification](https://aomediacodec.github.io/av1-avif/).
-The format has gained popularity and you can see in what browsers it is supported here [caniuse](https://caniuse.com/avif)
+AVIF is an image format developed by the The Alliance for Open Media.
 
 ## glTF Schema Updates
 
@@ -78,8 +78,8 @@ When used in the glTF Binary (.glb) format the `images` node that points to the 
 
 ## Best Practices
 
-Since AVIF is not as widely supported as JPEG and PNG, it is recommended to use this extension only when transmission size is a significant bottleneck. For example, in geospatial applications the total texture payload can range from gigabytes to terabytes of data. In those situations, AVIF textures can be used without a fallback to improve storage and transmission.
-
+Since AVIF is not as widely supported as JPEG and PNG, until [browser support](https://caniuse.com/avif) is good you will have to create a separate version of your asset with this extension and then do feature detection to load the avif version.
+AVIF has support for image sequences much like GIF, but this extension is only targeting using AVIF as a replacement for jpeg and png.
 When a fallback image is defined, this extension should not be present in `extensionsRequired`. This will allow all clients to render the glTF correctly, and those that support this extension can request the optimized AVIF version of the textures.
 
 ### Using Without a Fallback
@@ -106,4 +106,5 @@ If a glTF contains a AVIF with no fallback texture and the browser does not supp
 
 - [AVIF specification](https://aomediacodec.github.io/av1-avif/)
 - [AVIF reference implementation](https://github.com/AOMediaCodec/libavif/).
+- [Can I Use Browser Support](https://caniuse.com/avif)
 - [Sharp](https://sharp.pixelplumbing.com/) is a NodeJS library for fast encode/decode of AVIF using native modules (built on top of libvips which has support for AVIF.
--- a/extensions/2.0/Vendor/EXT_texture_avif/README.md
+++ b/extensions/2.0/Vendor/EXT_texture_avif/README.md
@@ -1,0 +1,109 @@
+# EXT_texture_avif
+
+## Contributors
+
+- Omar Shehata, Cesium, [@OmarShehata](https://github.com/OmarShehata)
+- Pascal Massimino, Google, [@Skal65535](https://github.com/skal65535)
+
+## Status
+
+Complete
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension allows glTF models to use AVIF as a valid image format. A client that does not implement this extension can ignore the provided AVIF image and continue to rely on the PNG and JPG textures available in the base specification. Defining a fallback texture is optional. The [best practices](#best-practices) section describes the intended use case of this extension and the expected behavior when using it without a fallback texture.
+
+AVIF is an image format developed by Netflix that uses the AV1 VideoCodec [specification](https://aomediacodec.github.io/av1-avif/).
+The format has gained popularity and you can see in what browsers it is supported here [caniuse](https://caniuse.com/avif)
+
+## glTF Schema Updates
+
+The `EXT_texture_avif` extension is added to the `textures` node and specifies a `source` property that points to the index of the `images` node which in turn points to the AVIF image.
+
+The following glTF will load `image.avif` in clients that support this extension, and otherwise fall back to `image.png`.
+
+```
+"textures": [
+    {
+        "source": 0,
+        "extensions": {
+            "EXT_texture_avif": {
+                "source": 1
+            }
+        }
+    }
+],
+"images": [
+    {
+        "uri": "image.png"
+    },
+    {
+        "uri": "image.avif"
+    }
+]
+```
+
+When used in the glTF Binary (.glb) format the `images` node that points to the AVIF image uses the `mimeType` value of `image/avif`.
+
+```
+"textures": [
+    {
+        "source": 0,
+        "extensions": {
+            "EXT_texture_avif": {
+                "source": 1
+            }
+        }
+    }
+],
+"images": [
+    {
+        "mimeType": "image/png",
+        "bufferView": 1
+    },
+    {
+        "mimeType": "image/avif",
+        "bufferView": 2
+    }
+]
+```
+
+### JSON Schema
+
+[glTF.EXT_texture_avif.schema.json](schema/glTF.EXT_texture_avif.schema.json)
+
+## Best Practices
+
+Since AVIF is not as widely supported as JPEG and PNG, it is recommended to use this extension only when transmission size is a significant bottleneck. For example, in geospatial applications the total texture payload can range from gigabytes to terabytes of data. In those situations, AVIF textures can be used without a fallback to improve storage and transmission.
+
+When a fallback image is defined, this extension should not be present in `extensionsRequired`. This will allow all clients to render the glTF correctly, and those that support this extension can request the optimized AVIF version of the textures.
+
+### Using Without a Fallback
+
+To use AVIF images without a fallback, define `EXT_texture_avif` in both `extensionsUsed` and `extensionsRequired`. The `texture` node will then have only an `extensions` property as shown below.
+
+```
+"textures": [
+    {
+        "extensions": {
+            "EXT_texture_avif": {
+                "source": 1
+            }
+        }
+    }
+]
+```
+
+If a glTF contains a AVIF with no fallback texture and the browser does not support AVIF, the client should either display an error or decode the AVIF image at runtime (see [resource](#resources) for decoding libraries).
+
+## Known Implementations
+
+## Resources
+
+- [AVIF specification](https://aomediacodec.github.io/av1-avif/)
+- [AVIF reference implementation](https://github.com/AOMediaCodec/libavif/).
+- [Sharp](https://sharp.pixelplumbing.com/) is a NodeJS library for fast encode/decode of AVIF using native modules (built on top of libvips which has support for AVIF.
--- a/extensions/2.0/Vendor/EXT_texture_avif/schema/glTF.EXT_texture_webp.schema.json
+++ b/extensions/2.0/Vendor/EXT_texture_avif/schema/glTF.EXT_texture_webp.schema.json
@@ -1,0 +1,15 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "title": "EXT_texture_avif glTF extension",
+  "type": "object",
+  "description": "glTF extension to specify textures using the AVIF image format.",
+  "allOf": [{ "$ref": "glTFProperty.schema.json" }],
+  "properties": {
+    "source": {
+      "allOf": [{ "$ref": "glTFid.schema.json" }],
+      "description": "The index of the images node which points to a AVIF image."
+    },
+    "extensions": {},
+    "extras": {}
+  }
+}


leon:
I have done a PR to three to get this added there.
https://github.com/mrdoob/three.js/pull/25173

leon:
I have also done a PR for Babylon
https://github.com/BabylonJS/Babylon.js/pull/13370

extensions/2.0/Vendor/EXT_texture_avif/README.md
```diff
@@ -0,0 +1,109 @@
+# EXT_texture_avif
+
+## Contributors
+
+- Omar Shehata, Cesium, [@OmarShehata](https://github.com/OmarShehata)
+- Pascal Massimino, Google, [@Skal65535](https://github.com/skal65535)
```
lexaknyazev, line 6: The content of this section should not have been copy-pasted.

extensions/2.0/Vendor/EXT_texture_avif/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Status
+
+Complete
```
lexaknyazev, line 10: The extension must start as a draft.

extensions/2.0/Vendor/EXT_texture_avif/README.md
```diff
@@ -0,0 +1,109 @@
...
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension allows glTF models to use AVIF as a valid image format. A client that does not implement this extension can ignore the provided AVIF image and continue to rely on the PNG and JPG textures available in the base specification. Defining a fallback texture is optional. The [best practices](#best-practices) section describes the intended use case of this extension and the expected behavior when using it without a fallback texture.
+
+AVIF is an image format developed by Netflix that uses the AV1 VideoCodec [specification](https://aomediacodec.github.io/av1-avif/).
+The format has gained popularity and you can see in what browsers it is supported here [caniuse](https://caniuse.com/avif)
```
lexaknyazev, line 21: Change these two lines to
> AVIF is an image format developed by the The Alliance for Open Media.

The "Can I use..." link should be put in the Resources section.

extensions/2.0/Vendor/EXT_texture_avif/README.md
```diff
@@ -0,0 +1,109 @@
...
+]
+```
+
+### JSON Schema
+
+[glTF.EXT_texture_avif.schema.json](schema/glTF.EXT_texture_avif.schema.json)
+
+## Best Practices
+
+Since AVIF is not as widely supported as JPEG and PNG, it is recommended to use this extension only when transmission size is a significant bottleneck. For example, in geospatial applications the total texture payload can range from gigabytes to terabytes of data. In those situations, AVIF textures can be used without a fallback to improve storage and transmission.
```
lexaknyazev, line 81: Replace the copy-pasted WebP rationale with something specific to AVIF. The section should mention concrete glTF use-cases that benefit from using this extension.

extensions/2.0/Vendor/EXT_texture_avif/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+When a fallback image is defined, this extension should not be present in `extensionsRequired`. This will allow all clients to render the glTF correctly, and those that support this extension can request the optimized AVIF version of the textures.
+
```
lexaknyazev, line 84: Given the flexibility of the format (auxiliary images, sequences, etc.), the extension spec must define an appropriate scope that is compatible with glTF texture usage.

lexaknyazev (changes requested):

extensions/2.0/Vendor/EXT_texture_avif/README.md
lexaknyazev: The core specification already defines color management. Other extra features that could be present in AVIF files should be addressed here.

extensions/2.0/Vendor/EXT_texture_avif/README.md
leon: I guess we are the authors?
I added myself, Don and you :)

extensions/2.0/Vendor/EXT_texture_avif/README.md
leon: done

extensions/2.0/Vendor/EXT_texture_avif/README.md
leon: Done

extensions/2.0/Vendor/EXT_texture_avif/README.md
leon: I changed it a bit, but copy is now my strong suite, feel free to update it to what you think is better.

For me the main advantage while developing mobile open gl is transfer size on 4g.
beeing able to only have to download 2mb instead of 20 is a big win.

extensions/2.0/Vendor/EXT_texture_avif/README.md
leon: I added a part about this only being a replacement for jpeg / png.

I added support for the extension in gltf-pipeline today
https://github.com/leon/gltf-pipeline/tree/avif

an .avif file that is an image starts with ftypavif
whilst image sequences has another byte sequence,
I only added support for the image version for now.

extensions/2.0/Vendor/EXT_texture_avif/README.md
lexaknyazev: That's only a win on network transmission / disk storage. AVIF decoding complexity is higher than with other formats and GPU memory usage is the same as JPEG / PNG / WebP. For example (don't copy this literally), a 3D application that features real photos / paintings may prioritize transmission size over other other metrics.

In a case of glTF, images are used as textures, for which GPU throughput is usually more important, so our general recommendation is to use `KHR_texture_basisu`.

This section should provide glTF use-cases that prioritize transmission / storage size and are not bound by GPU memory or throughput.

donmccurdy:
Experimental implementation in [glTF Transform](https://github.com/donmccurdy/glTF-Transform), for testing purposes:

```bash
npm install --global @gltf-transform/cli@avif

gltf-transform avif input.glb output.glb
```

Related:

- https://github.com/donmccurdy/glTF-Transform/pull/771
- https://github.com/donmccurdy/glTF-Transform/pull/752

zeux:
Out of curiosity, what is the motivation for including both WebP and AVIF as glTF extensions? Is there specific guidance on when an application  should use WebP vs AVIF, and how they compare to BasisU?

donmccurdy:
I'll copy my thoughts from [this thread](https://github.com/KhronosGroup/glTF-Blender-IO/issues/1308#issuecomment-1427227408), they're probably more relevant here anyway —

> I expect that any sufficiently large graphics application — particularly at the scale and complexity of a game — is best served by GPU texture formats like KTX2. And often in much smaller applications too, for exactly the reasons you mention. In the kind of "virtual world software" described by the OP that may well be the best choice.
> 
> However, the web has a wide range of other content patterns. Consider VFX on a fancy brand's landing page, or 3D graphic illustrations on a news article. Visitors quickly leave a site if it doesn't load almost instantly, time to [first-contentful-paint](https://web.dev/fcp/) should be <2s, and page size budgets are commonly measured in kilobytes, not megabytes. Under those constraints I think web developers often want the smallest transmission size they can get, and would expect to use relatively little of their VRAM budget.
> 
> Other obstacles include the comparatively easy configuration of good lossless or near-lossless compression in WebP and AVIF formats, compared to more careful tuning often required in KTX2. Raising awareness around the VRAM differences in these formats is an important and ongoing challenge, too.

I would like to find better comparisons on WebP vs. AVIF for graphics purposes, but in general my rule of thumb would probably be, in order of preference:

1. Use KTX2 when VRAM or texture GPU upload time are critical
2. Use AVIF when bundle size is critical and VRAM is not, if device supports AVIF
3. Use WebP when bundle size is critical and VRAM is not, if device doesn't support AVIF
4. Use PNG and JPEG for portability between applications

It isn't quite cut-and-dried: there are certainly cases where WebP outperforms AVIF. But I think that is the general shape of things.

zeux:
> It isn't quite cut-and-dried: there are certainly cases where WebP outperforms AVIF. But I think that is the general shape of things.

It feels a little odd to have two independent formats, AVIF and WebP, that seem to perform a similar function. This bloats the spec and results in increased complexity in the entire ecosystem, since now multiple formats of each type need to be supported. It makes sense to support one of AVIF or WebP but it feels odd to support both? Maybe the ship has sailed because WebP is already supported, and AVIF seems to compress a little better.

bghgary:
> Maybe the ship has sailed because WebP is already supported

WebP is not ratified though.

donmccurdy:
My feeling is that the KHR_ prefix should represent extensions that Khronos believes are net-positive for the ecosystem, with a healthy respect for the increased complexity of supporting multiple similar alternatives. EXT_ extensions should represent only that "at least two vendors have chosen to implement this thing", without a value judgment from Khronos. Certainly I would not expect the entire ecosystem to implement all EXT_ extensions. In three.js we're trying to support all KHR_ extensions, but we've intentionally decided to implement some EXT_ extensions and not others, on a case by case basis. 

I see a pretty strong advantage with AVIF, compared to WebP, PNG, and JPEG. Compared to KTX2 there are pros and cons. For three.js I would like to offer all of those options. We can rename this to a THREE_ prefix if there are no other interested vendors.

aaronfranke:
@zeux I am interested in implementing EXT_texture_webp in Godot Engine, but I can't do the same with AVIF yet because Godot Engine does not support AVIF. It will take time before AVIF is a commonly enough adopted format, in the meantime WebP has been around for a decade, is more widely supported, and is still superior to PNG and JPG.

I don't see the harm in having both available, if only WebP then we're holding back progress for devices that support AVIF (so that they can't reach their full potential), and if only AVIF then we are holding back progress for devices that don't yet support AVIF (so that they are still stuck using PNG or JPG).

donmccurdy:
> I was really hoping for jpeg xl to win the race, but google pulled it from chrome.

Since it's mentioned above – some signs of life for JPEG XL:

https://webkit.org/blog/14205/news-from-wwdc23-webkit-features-in-safari-17-beta/#images

donmccurdy:
> > Use AVIF when bundle size is critical and VRAM is not
>
> The extension spec should provide specific examples.

Proposal:

- Website landing pages displaying a 3D "hero" illustration
- Authoring contexts in which lossless compression is required
- Small "islands" of 3D content within a traditional site

There's also a practical angle here — users may not have the time or ability to tune compression settings and UV layouts for optimized KTX2 compression. Platforms hosting user-generated content may have difficulty generalizing KTX2 settings without a human in the loop. But I don't think those are concerns the spec needs to call out.

I think it would also be fair to provide counter-examples — Content intended for immersive AR/VR applications should almost certainly use KTX2.

bowald:
Hi.

We at IKEA are interesting to see this extension move on to ratification. We are looking into switching to AVIF textures for some parts of our E-commerce viewers, first and foremost to display a single article at the product page. In this particle case we value transfer size (or rather time to first rendered frame) as well as visual fidelity. We will no be VRAM constraint, So we favor a AVIF or WebP over BasisU for this use case. This is similar to the use case @Donmccurdy identified in [this reply](https://github.com/KhronosGroup/glTF/pull/2235#issuecomment-1578912224) regarding a "hero" asset. 

We have looked into how AVIF compares to WebP and in short AVIF retains intended quality of the input textures with a lower file size, the difference is most notable for Data textures such as Occlusion-roughness-metalness and Normal maps.

In our test we targeted a lower file size for AVIF compare to WebP and compare the visual result.

We picked a number of [glTF sample assets](https://github.com/KhronosGroup/glTF-Sample-Assets) and generate a WebP and AVIF version using [gltf-transform](https://gltf-transform.dev/) CLI, with the commands below. Note that we are using lossy compression for both of the formats.
```
gltf-transform avif {input} {output} --quality 53
gltf-transform webp {input} {output} --quality 75
```

These are the file sizes we got, and as seen all the AVIF versions are lower in transfer size:

| Model | Original Textures | WebP Textures | AVIF Textures |
|-------|-------------------|---------------|--------------|
| AnisotropyBarnLamp | 7.4 MB | 355.9 kB | **231.1 kB** |
| BoomBox | 10.4 MB | 375.7 kB | **223.3 kB** |
| DamagedHelmet | 3.2 MB | 1.0 MB | **803.8 kB** |
| FlightHelmet | 90.3 MB | 6.4 MB | **5.1 MB** |
| GlamVelvetSofa | 3.0 MB | 313.8 kB | **321.0 kB** |
| SciFiHelmet | 26.6 MB | 1.2 MB | **1.1 MB** |
| SheenChair | 3.0 MB | 329.3 kB | **221.0 kB** |

We compared the converted textures against the original using [idiff](https://openimageio.readthedocs.io/en/latest/idiff.html), in [OpenImageIO](https://github.com/AcademySoftwareFoundation/OpenImageIO), built with [libheif](https://github.com/strukturag/libheif) and [libwebp](https://github.com/webmproject/libwebp). We tracked the mean error between the source and WebP as well as source and AVIF. From this test we saw that AVIF in general performing better then Webp, especially for the data textures. (Which is not describe in the table, but I can add more data for a specific model if interest). There are some color textures where WebP marginally outperformed AVIF, however in general both formats perform good when it comes to color. Lower error is better, and the "winner" for each texture is marked with **bold** in the table.

| Model | Tex 1 WEBP-Orig | Tex 1 AVIF-Orig | Tex 2 WEBP-Orig | Tex 2 AVIF-Orig | Tex 3 WEBP-Orig | Tex 3 AVIF-Orig | Tex 4 WEBP-Orig | Tex 4 AVIF-Orig | Tex 5 WEBP-Orig | Tex 5 AVIF-Orig | Tex 6 WEBP-Orig | Tex 6 AVIF-Orig | Tex 7 WEBP-Orig | Tex 7 AVIF-Orig | Tex 8 WEBP-Orig | Tex 8 AVIF-Orig | Tex 9 WEBP-Orig | Tex 9 AVIF-Orig | Tex 10 WEBP-Orig | Tex 10 AVIF-Orig | Tex 11 WEBP-Orig | Tex 11 AVIF-Orig | Tex 12 WEBP-Orig | Tex 12 AVIF-Orig | Tex 13 WEBP-Orig | Tex 13 AVIF-Orig | Tex 14 WEBP-Orig | Tex 14 AVIF-Orig | Tex 15 WEBP-Orig | Tex 15 AVIF-Orig | Tex 16 WEBP-Orig | Tex 16 AVIF-Orig | Tex 17 WEBP-Orig | Tex 17 AVIF-Orig | Tex 18 WEBP-Orig | Tex 18 AVIF-Orig | Tex 19 WEBP-Orig | Tex 19 AVIF-Orig | Tex 20 WEBP-Orig | Tex 20 AVIF-Orig | Tex 21 WEBP-Orig | Tex 21 AVIF-Orig | Tex 22 WEBP-Orig | Tex 22 AVIF-Orig | Tex 23 WEBP-Orig | Tex 23 AVIF-Orig | Tex 24 WEBP-Orig | Tex 24 AVIF-Orig | Tex 25 WEBP-Orig | Tex 25 AVIF-Orig | Tex 26 WEBP-Orig | Tex 26 AVIF-Orig | Tex 27 WEBP-Orig | Tex 27 AVIF-Orig | Tex 28 WEBP-Orig | Tex 28 AVIF-Orig | Tex 29 WEBP-Orig | Tex 29 AVIF-Orig | Tex 30 WEBP-Orig | Tex 30 AVIF-Orig |
|-------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|---------------|
| AnisotropyBarnLamp | 0.00472 | **0.00194** | 0.00786 | **0.00646** | 0.00191 | **0.00156** | 0.00846 | **0.00421** |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| BoomBox | **0.00361** | 0.00374 | **0.00010** | 0.00011 | 0.00419 | **0.00308** | 0.01091 | **0.00510** |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| DamagedHelmet | 0.01153 | **0.01102** | 0.00431 | **0.00379** | **0.00038** | 0.00042 | 0.01273 | **0.00881** | 0.00559 | **0.00390** |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| FlightHelmet | 0.00856 | **0.00826** | 0.00607 | **0.00560** | 0.00848 | **0.00768** | 0.01238 | **0.01238** | **0.06881** | 0.12270 | 0.00607 | **0.00560** | 0.00806 | **0.00558** | 0.00938 | **0.00675** | 0.01238 | **0.01238** | 0.02024 | **0.01246** | 0.01303 | **0.00922** | **0.06881** | 0.12270 | 0.00131 | **0.00131** | 0.00771 | **0.00486** | 0.00848 | **0.00768** | 0.00825 | **0.00558** | 0.01606 | **0.00762** | 0.00856 | **0.00826** | 0.01263 | **0.00663** | 0.01367 | **0.00859** | 0.01263 | **0.00663** | 0.00806 | **0.00558** | 0.00825 | **0.00558** | 0.02024 | **0.01246** | 0.00131 | **0.00131** | 0.01367 | **0.00859** | 0.00938 | **0.00675** | 0.01606 | **0.00762** | 0.01303 | **0.00922** | 0.00771 | **0.00486** |
| GlamVelvetSofa | 0.04140 | **0.02156** | 0.00535 | **0.00459** |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| SciFiHelmet | 0.25458 | **0.00422** | 0.25536 | **0.00543** | 0.26321 | **0.00790** | 0.01429 | **0.00780** |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
| SheenChair | 0.62889 | **0.62833** | 0.01563 | **0.01381** | 0.00876 | **0.00851** | 0.42792 | **0.42743** | **0.00295** | 0.00306 | 0.00842 | **0.00631** | **0.48547** | 0.48580 |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |

Comparing the the visual result for the [ScFi Helmet](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/SciFiHelmet), its rendered in [BabylonJS SandBox](https://sandbox.babylonjs.com/).
We can see the scratches in the front plate of the visor is better preserved in the AVIF version, as well as the three bevels on the right side of the helmet is better preserved in the AVIF version
POWERPNT_74jrwZpZ8m

Looking at the [Anisotropy Barn Lamp](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/AnisotropyBarnLamp) we can see that the details on the knurled socket, displayed in the section above the text, are better preserved in the AVIF version of the asset.
POWERPNT_FSaSxczSwe

To summarize, we see value in this extension for one of our use cases, and would like to see this move on to ratification. 

I see that there are some parts of the text that may need to be updated, such as the support for AVIF in the Best Practices section. [AVIF is now supported in 96.19% of all browsers](https://caniuse.com/avif).

lexaknyazev:
> We at IKEA are interesting to see this extension move on to ratification.

Just to clarify, are you interested in moving this feature forward in general or do you need it to be put under Khronos IP (that's what "ratification" means)?

bowald:
> > We at IKEA are interesting to see this extension move on to ratification.
> 
> Just to clarify, are you interested in moving this feature forward in general or do you need it to be put under Khronos IP (that's what "ratification" means)?

We would like to see this being put under Khronos IP, similar to [EXT_texture_webp](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_texture_webp). 

Looking at the WebP vs AVIF we think that there is value putting this under Khronos IP as it outperforms WebP. Main reason for us is to make sure the glTF we deliver is using ratified extensions, to make sure its supported in as many places as possible. 

This extension, even though not ratified, is already implemented in ThreeJS, BabylonJS, gltf-transform among others, and it would be valuable to see more support for it, one example is in [glTF-Blender-IO](https://github.com/KhronosGroup/glTF-Blender-IO/issues/2625).

If we can help out in anyway to move this proposal along, we are happy to do so.

leon:
I wonder how JXL compares to avif. 

I have been patiently waiting for it to land in all the browsers and it looks like it is finally in the works in most of them. 

I think once jxl is available everywhere it will become the next jpg.

So that should also be a candidate for being included as a plugin in gltf. 

But it's not supported yet. So fully understand that IKEA would want to move forward with something that is already out there 😁

trvb:
Hello Khronos folks,

I'd like to share some feedback on our experience at [Renault Group](https://github.com/grouperenault), as we would very much like to see this extension ratified.

As part of our [openR Link](https://www.renault.co.uk/renault-connect/openr-link.html) infotainment system, we are developing the next generation of our Around View Monitor (AVM), or 360-degree camera system, for our customers. Our solution is built on a bespoke rendering engine that runs on [Android](https://developer.ampere.cars/en/) and many other platforms.

Our AVM displays a realistic scene that includes a vehicle avatar. This avatar is fully interactive based on vehicle sensors: its doors, wheels, and lights are all animated. More importantly, it is a very accurate physical representation of the vehicle, including its chassis, body, paint, model range, and customization options. Our customers get to see their vehicle on screen as it is in real life, which we think is a really neat immersive experience.

Because we support a highly diverse vehicle range and deploy our system firmware as a single image through FOTA updates, the size of our 3D assets became a concern. It was critical for us to keep them as small as possible, so that we could support the full range of vehicle customization on our avatar without blowing up our system image size.

We quickly realized that using lossless 2048x2048 textures would result in hundreds of megabytes of assets, which we simply could not afford, so it became necessary for us to implement compression.

We benchmarked several solutions for both mesh and texture compression, and when it came to textures, it became clear that there were only a handful of alternatives:
- WebP
- AVIF
- BasisU

Because GPU VRAM usage was never an issue for us, and BasisU is somewhat harder to implement across the wide variety of platforms we support, we did not pursue it. We also had concerns that we would not reach the same level of compression efficiency as when using general-purpose image encoders.

So we evaluated WebP against AVIF using a sample of vehicle textures representative of our use case and different encoder settings. We quickly realized that even though AVIF was more efficient than WebP (as @Bowald also pointed out), WebP had inherent limitations that prevented us from reaching our target quality.

WebP in lossy mode is limited to 8-bit 4:2:0 chroma subsampling. No matter which encoder settings we used, areas containing fine, high-contrast details, such as headlights, were getting damaged. This kind of texture damage was outright rejected by our design experts, who take great care to calibrate their assets for optimal customer perception.

On the other hand, AVIF is much better at preserving finer details when configured for 8-bit 4:4:4 chroma sampling, even with aggressive compression settings. We confirmed this through blind testing at customer-typical display resolutions, using different encoder settings.

Here is a quick comparison to demonstrate the point using [Megane](https://www.renault.fr/vehicules-electriques/nouvelle-megane.html). For this benchmark, we used fairly aggressive compression settings and sampled areas of interest from the original 2048x2048 texture. We tuned the AVIF and WebP quality settings to produce files of nearly identical sizes, so in essence we are comparing codec performance for the same storage budget.

Benchmark:
- PNG lossless: resolution = 2048x2048, file size = 2.57 MB
- WebP lossy: sampling = 8-bit 4:2:0, alpha Q = 100, Q = 78, file size = 130 KB
- AVIF lossy: sampling = 8-bit 4:4:4, alpha Q = 100, Q = 70, file size = 127 KB

Rear lights:
BCB_rear_lights_strip
Here, WebP fails to preserve fine details, causing an odd blurring effect along the lines. The visual difference between AVIF and the PNG reference is imperceptible.

Rear logo:
BCB_rear_logo_strip
The WebP version shows typical compression artifacts, while the AVIF version shows few, if any, and remains close to the reference.

Side skirt:
BCB_side_skirt_strip
Here, WebP suffers from severe artifacts. AVIF also shows some distortion, perhaps resembling a moire pattern, but remains reasonably close to the PNG reference overall.

Front speaker:
BCB_front_speaker_strip
This is the most challenging test. The speaker in the front door panel occupies a very small area and contains fine, grid-like patterns. WebP is unable to preserve the pattern, while AVIF is not perfect but remains quite good overall.

We had the same results with other textures: AVIF is a clear winner for quality at equivalent file sizes, which makes sense considering it is based on a more recent and advanced image codec than WebP. What really put the nail in the coffin of WebP was that even when encoding with very high quality settings like Q = 100, negating a considerable amount of the storage savings, the degradation of the finer details would never completely go away with WebP.

Another advantage of AVIF is its broad support across a wide range of devices. We also have fast decoders like [dav1d](https://code.videolan.org/videolan/dav1d) that run incredibly fast on embedded devices. During our testing, we were not able to measure any decode-time penalty when using AVIF instead of PNG, which is important as you expect an in-vehicle camera system to be available quickly when needed.

To sum up, we would not have been able to go into production with a WebP-based solution. Its compression algorithm is not advanced enough for our requirements, and it does not give the same level of control as AVIF. Performance and adoption are really good with AVIF, so for us it really was an obvious choice.

We are more than happy to answer any questions you may have :-)

PR #2250, opened 2023-01-29
Add: EXT_texture_qoi

spnda:
This adds an extension for the [QOI image format](https://qoiformat.org/), which is an incredibly simple format that losslessly compresses images to a similar size of PNG, while offering 20x-50x faster encoding and 3x-4x faster decoding compared to libpng and stbi.

I came across this format recently and thought it'd make a great addition to glTF to vastly increase load times while still being supported by virtually any platform just like PNG and JPEG are. The reference decoder and encoder is only 300 loc in C, making QOI a format that's incredibly easy to implement. I've noted that my own glTF library now supports the extension (albeit on [another branch](https://github.com/spnda/fastgltf/tree/EXT_texture_qoi)).

### DIFF
--- a/extensions/2.0/Vendor/EXT_texture_qoi/README.md
+++ b/extensions/2.0/Vendor/EXT_texture_qoi/README.md
@@ -1,0 +1,97 @@
+# EXT_texture_qoi
+
+## Contributors
+
+* Sean, [@spnda](https://github.com/spnda)
+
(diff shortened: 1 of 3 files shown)


heysokam:
Any updates on the acceptance of this extension?
Its status has been frozen for three years. Why is that the case?

PR #2251, opened 2023-01-30
KHR_materials_transmission: Minor fix to implementation note.

abwood:
This is a minor fix to an implementation note in the KHR_materials_transmission extension. I believe there's a missing negative sign in our computation for the transmission half vector (must use dot(N,-L)).

This is already done correctly in the sample viewer: [punctional.glsl](https://github.com/KhronosGroup/glTF-Sample-Viewer/blob/9693a0380cd268f18e7523a3e2b961cda58b3c02/source/Renderer/shaders/punctual.glsl#L83)

I also confirmed that EnterprisePBR also uses the correct dot(N,-L). See equation 5 in [EnterprisePBR](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2022x.md.html#components/core/dielectricbsdffortransparentsurfaces)

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_transmission/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_transmission/README.md
@@ -199,7 +199,7 @@ $$
 using the transmission half vector $H_T$
 
 $$
-H_T = \text{normalize}(V + 2 \left( N \cdot -L \right) N + L)
+H_T = \text{normalize}(V - 2 \left( N \cdot L \right) N + L)
 $$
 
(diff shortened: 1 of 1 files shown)


proog128:
Good catch, in addition we should change to round brackets, otherwise the change has no effect:

$$
H_T = \text{normalize}(V + 2 \left( N \cdot -L \right) N + L)
$$

abwood:
Yes of course. Thanks @proog128 for catching that.

proog128 (approved):

emackey (approved):
LTGM. Also got verbal approval from @MiiBond during the PBR TSG meeting.

PR #2277, opened 2023-03-31
MPEG Avatar Extension

haudiobe:


aaronfranke:
How do MPEG avatars compare to other glTF avatar formats, like VRM or RPM? What is the justification for creating a new format, what does MPEG do better than the existing formats?

PR #2286, opened 2023-05-05
EXT_mesh_manifold

elalish:
As the long ago spec editor of [3MF](https://3mf.io/3mf-specification/), I realized that a small glTF extension would get us full interoperability with solid-modeling workflows. I'm thinking `EXT` makes sense here since the point is certainly to transfer these files between different software, but this extension is a bit niche as pure rendering workflows have little reason to care. However, implementation is also nearly trivial on the consuming side, as this extension puts the onus on the producer to create a more constrained file.

I don't have my Google hat on for this one - this is related to my side project, [Manifold](https://github.com/elalish/manifold), an open-source robust mesh geometry kernel. I have [implemented](https://github.com/elalish/manifold/blob/master/bindings/wasm/examples/manifold-gltf.js) reading and writing of this extension using @donmccurdy's excellent `gltf-transform` library. You can test it live [here](https://manifoldcad.org/model-viewer.html).

Though I've been working with Khronos here for several years, I've never proposed an extension myself before, so I'd love any feedback on how best to proceed with this.


fire:
I am quite interested in this. Will give it a look.

emackey:
Interesting.  Does there need to be any harmony between this extension and the manifold requirements placed by `KHR_materials_volume` on the mesh?

https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_volume#extending-materials

We already have a ratified extension with a parameter (`thicknessFactor`) that indicates when the mesh is expected to be manifold as opposed to thin-walled.  That one doesn't go into nearly the detail shown here on what exactly that means for the asset. But I wouldn't want this extension and `volume` to disagree on whether a given mesh was supposed to be manifold or not.  Just something to think about.

elalish:
@emackey Yeah, I think they'll work just fine together. In the volume extension I believe manifoldness is not required (or even defined), but simply recommended. This extensions would be a nice pairing to ensure that manifoldness in a strict way if desired. Having recently tested a bunch of glTF sample models, I can tell you most of the volume ones don't actually pass this manifold check, which isn't really a surprise or a problem per se.

emackey:
@elalish You're right, the volume extension doesn't define what "manifold" means.  But it is a normative requirement that when `thicknessFactor` is nonzero, the polygons of any mesh using that material are no longer considered a loose collection of thin-walled triangles, and instead are considered to be a watertight boundary around a 3D volume of material.  Behavior of gaps in the boundary is undefined.  I've thought about asking the glTF validator to check such meshes for water-tightness, but that may add too much complexity and cost to validation, I would guess.  At the very least, `doubleSided` has no meaning for such meshes, and that's specified normatively too.

elalish:
Yeah, that seems fine. I'd put the distinction this way: the volume extension wants the mesh to be watertight *geometrically*, since that's all renderers care about. This extension requires manifoldness *topologically* since that's what's required for computational algorithms. The classic case is a sharp crease in the normals - the triangles butt up against each other, so it's perfectly fine for volume, but since the verts themselves are duplicated and not associated, it can't be topologically manifold without this extension.

fire:
If gltf transform can apply the manifold extension to valid gltfs in both optional and required mode, that would be amazing for the use case of ensuring a manifold supply-chain from input to output.

elalish:
@lexaknyazev Would you be so kind as to check my extension spec? I've verified that it's working well for round-tripping the data, but I'm not so familiar with JSON schema.

extensions/2.0/Vendor/EXT_manifold/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+This extension allows authors to denote that the contained mesh data is in fact manifold, and adds a small amount of extra data to losslessly recover this manifold mesh by leveraging sparse accessors. 
+
+## Extending Meshes With Manifold Data
+
+Manifoldness is implied by adding the `EXT_manifold` extension to a glTF `mesh`. The `primitives` of the `mesh` **should** use the same set of attributes and each attribute sematic **must** consistently reference the same accessor. The `indices` **must** be present and the referenced `accessor` **must** in turn reference a single `bufferView` for the set of primitives.
```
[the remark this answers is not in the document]
elalish: Done.

extensions/2.0/Vendor/EXT_manifold/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+A mesh is considered manifold when it topologically represents a solid object with well-defined volume, sometimes known as water-tight. Manifoldness has many definitions, so we are specifically using the same definition as the 3MF spec (3D Manufacturing Format) for maximum interoperability: an oriented 2-manifold triangle mesh.
```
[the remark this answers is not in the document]
elalish: Done.

elalish:
@lexaknyazev Do you know why the test failed? I have no idea what it's checking. Also, this is ready for a second review.

extensions/2.0/Vendor/EXT_manifold/README.md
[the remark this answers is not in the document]
lexaknyazev: Was "the same set of attributes" requirement removed intentionally?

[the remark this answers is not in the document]
elalish: Yes, it is not a requirement after all - and is particularly important when combing objects that have different attributes. Here's an example where only part of the mesh has vertex colors: https://twitter.com/ManifoldCAD/status/1669371574145908736

elalish:
Okay, I've updated my [gltf-io.ts](https://github.com/elalish/manifold/blob/master/bindings/wasm/examples/gltf-io.ts) with documentation and it's now able to handle different numbers of properties per primitive and such. I should probably publish it to npm. 

Meanwhile, that is now allowing [manifoldCAD.org](https://manifoldcad.org) to export GLBs with this extension that all pass the Khronos validator. For a round-trip, you can try https://manifoldcad.org/make-manifold where you can upload a GLB with this extension to verify it is in fact manifold, or a GLB without this extension and it will find any meshes that are close to manifold and `Merge` and re-export them with this extension. e.g. DragonAttenuation:
image
Becomes:
image
The dragon itself is close enough to manifold to successfully `Merge`, but the cloth backdrop is not.

elalish:
My understanding is that as soon as there are two independent implementations of an `EXT` extension it can be merged (there is no ratification process like for `KHR`). @fire @hrgdavor @tpotancok @alteous are any of you interested in making a second implementation? I've kicked the tires on this enough that I'm confident the extension works quite well.

elalish:
I also just saw https://github.com/KhronosGroup/glTF/tree/main/extensions#naming - should I rename this to `EXT_mesh_manifold`?

javagl:
> My understanding is that as soon as there are two independent implementations of an `EXT` extension it can be merged (there is no ratification process like for `KHR`). 

There are some elements of the process that are not perfectly pinned down yet, and I'm omitting some details - the following is an attempt of a _very_ short **tl;dr**:

The _intention_ of the `KHR` prefix originally was that it is used for ratified extensions. But if there is a `VENDOR_...` or `EXT_...` extension, then it should still be possible to ratify it, _without_ having to rename it. So the prefix cannot indicate the 'ratification state' here. (Now, `KHR_` basically means that it was proposed by the Khronos group - omitting some details and vaguenesses here)

Whether an extension specification is **merged** depends more on the "maturity" of the extension, and whether it has been thoroughly _reviewed_ (some of this is hard to quantify...)

The requirement for _"two independent implementations"_ is indeed a condition for an extension being **ratified** (but this could be done for `EXT_`- and even `VENDOR_`-extensions, eventually)

> I would also likely use gltf-transform, hope that does not disqualify it as second impl. 

I'm not aware of a clear definition of what 'a second implementation' means. But I could imagine that an implementation of the extension in glTF-Transform **could** count as an 'implementation' - **iff** it is not solved with manual twiddling in raw JSON objects, but indeed an [extension implementation](https://gltf-transform.dev/extensions) in the spirit of the library. (This concept of 'two implementations' should be defined more clearly...)

elalish:
I wouldn't push this extension for ratification - glTF is primarily about rendering, while this extension is for CAD/3D printing workflows that are not normally the core users of glTF.

>Whether an extension specification is merged depends more on the "maturity" of the extension, and whether it has been thoroughly reviewed (some of this is hard to quantify...)

It would be great to have some "next steps" on this. There aren't yet two independent implementations, but there is an open source one that reads and writes with a bunch of tests. Pretty much the only concern I have left about it is the name. It would be nice to point people to a published spec to encourage them to implement.

alteous:
@elalish, I'd be interested in implementing this.

elalish:
@alteous Great, let me know if you have any questions for me. How do you plan to make use of it?

emackey:
> Pretty much the only concern I have left about it is the name.

+1 for `EXT_mesh_manifold` unless the implementers here have already shipped stable implementations of this.  If it doesn't break too much, it would be great to rename it before more implementations show up.

alteous:
> @alteous Great, let me know if you have any questions for me. How do you plan to make use of it?

I should have clarified, we would export the extension data from our modelling app: https://kittycad.io/modeling-app

elalish:
Okay, I updated the name here, along with my implementation: https://github.com/elalish/manifold/pull/630 since I'm pretty sure we're the only one so far. 

@emackey, merging is only blocked until there is an approving review - would you feel comfortable?

elalish:
@hrgdavor we should just publish my `manifold-gltf.ts` and `gltf-io.ts` on npm and you can grab them as dependencies, especially if you're already using `manifold-3d`. Let me know if you're interested.

hrgdavor:
> @hrgdavor we should just publish my `manifold-gltf.ts` and `gltf-io.ts` on npm and you can grab them as dependencies, especially if you're already using `manifold-3d`. Let me know if you're interested.

Yes I am interested, I think at least few more people that are experimenting with manifoldjs outside of jscad.

extensions/2.0/Vendor/EXT_manifold/README.md
```diff
@@ -0,0 +1,171 @@
+# EXT\_manifold
```
emackey, line 1: Name change here at the top of the README.  Additionally, please rename this entire folder from `EXT_manifold` to `EXT_mesh_manifold`.

extensions/2.0/Vendor/EXT_manifold/schema/glTF.EXT_manifold.schema.json
```diff
@@ -0,0 +1,48 @@
+{
```
emackey, line 1: Please rename this file to `mesh.EXT_mesh_manifold.schema.json`.  (So, "mesh" instead of "glTF" at the front, and add "mesh" again to the extension name).

extensions/2.0/Vendor/EXT_manifold/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Extending Meshes With Manifold Data
+
+A glTF mesh is denoted as a manifold by adding an `EXT_mesh_manifold` extension object to it and obeying the following restrictions:
+
+- all primitives **MUST** use the `TRIANGLES` topology type;
+- same attributes of different mesh primitives **MUST** reference the same accessor;
+- the `indices` property **MUST** be defined for all primitives and the referenced accessors with index data **MUST** reference the same buffer view.
+
+The `manifoldPrimitive` added in the extension is an alternative `primitive` with additional restrictions:
```
lexaknyazev, line 33: Conventions: update the line to
> The required `manifoldPrimitive` property added ...

extensions/2.0/Vendor/EXT_manifold/README.md
```diff
@@ -0,0 +1,171 @@
...
+- it **MUST** use the `TRIANGLES` topology type;
+- it **MUST** have only the `POSITION` `attribute`, which **MUST** reference the same position `accessor` as the other primitives use; 
+- its `indices` **MUST** reference an `accessor` which **MUST** in turn reference the same `bufferView` as the other primitives' `indices`;
```
lexaknyazev, line 36: Conventions: update the line to:
> its `indices` property **MUST** reference an accessor, which **MUST** in turn reference the same buffer view as indices accessors of other primitives of the same mesh

extensions/2.0/Vendor/EXT_manifold/README.md
```diff
@@ -0,0 +1,171 @@
...
+- its `indices` `accessor` **MAY** include a `sparse` accessor that defines which vertices are to be merged, but every index merged **MUST** have identical position values for the before and after vertices referenced, so that the geometry is unchanged;
```
lexaknyazev, line 37: Conventions: update the line to:
> when its `indices` accessor includes a `sparse` property, the sparse data also defines which vertices are to be merged...

extensions/2.0/Vendor/EXT_manifold/schema/glTF.EXT_manifold.schema.json
```diff
@@ -0,0 +1,48 @@
...
+        "manifoldPrimitive": {
+            "allOf": [
+                {
+                    "$ref": "mesh.primitive.schema.json"
+                },
+                {
+                    "material": false
+                },
+                {
+                    "targets": false
```
emackey, line 21: Sadly this doesn't seem to override these values in VSCode, for what it's worth.  After Googling a bit I'm not sure if it's even possible in a JSON schema to use `allOf` and then override some of the imported properties like this.  @lexaknyazev Do you know if one can import a schema (with `allOf` ... `$ref`) and then exclude two of the properties defined in the referenced schema?

extensions/2.0/Vendor/EXT_manifold/schema/glTF.EXT_manifold.schema.json
```diff
@@ -0,0 +1,48 @@
...
+                }
+            ],
+            "description": "An alternative single primitive that contains only POSITION, providing the same geometry as all of the normal primitives, but manifold."
```
emackey, line 24: Likewise in VSCode, this `description` and any `title` added here is completely overridden by the ones defined in `mesh.primitive.schema.json`.  I don't think this should hold up the PR, but it's unfortunate if we wanted VSCode support for this.

For example here's a hover-tooltip with the incorrect description applied:

screenshot

extensions/2.0/Vendor/EXT_manifold/schema/glTF.EXT_manifold.schema.json
```diff
@@ -0,0 +1,48 @@
...
+                    "$ref": "glTFid.schema.json"
+                }
+            ],
+            "description": "The index of the accessor that contains the vertex sparse values for merging into a manifold."
+        },
+        "extensions": {},
+        "extras": {}
+    },
+    "required": [
+        "manifoldPrimitive"
```
emackey, line 46: The rest of this file seems to work fine in VSCode, so that's good.

extensions/2.0/Vendor/EXT_manifold/README.md
```diff
@@ -0,0 +1,171 @@
...
+- it **MUST NOT** contain a material;
+- it **MUST NOT** contain morph targets, though the other primitives can.
+
+If the `indices` `accessor` is sparse, then `mergeIndices` and `mergeValues` **MUST** be included and reference `accessors` which are equivalent to the sparse indices and values. This small amount of JSON is duplicated here because sparsity is generally considered an internal detail of compression, but here it also represents the relationship between the separate materials and the unified manifold.
```
lexaknyazev, line 41: To what extent should `mergeIndices` and `mergeValues` be equivalent to internal sparse data?
1. Precisely referring to the same data (buffer views, offsets, and component types) or just resolving to equivalent values no matter where they come from?
2. Presumably, the number of elements must match anyway.

---

> the relationship between the separate materials and the unified manifold

What do "separate materials" mean here?

---

What does it mean for a `manifoldPrimitive` to use a non-sparse indices accessor?

extensions/2.0/Vendor/EXT_manifold/schema/glTF.EXT_manifold.schema.json
```diff
@@ -0,0 +1,48 @@
...
+            "allOf": [
+                {
+                    "$ref": "mesh.primitive.schema.json"
+                },
+                {
+                    "material": false
+                },
+                {
+                    "targets": false
+                }
```
lexaknyazev, line 22: This has no effect on the schema.

elalish (commented):
Thanks for the helpful reviews - I believe I've addressed all the comments.

emackey (approved):
Looks like all of my feedback was addressed here.

extensions/2.0/Vendor/EXT_mesh_manifold/README.md
```diff
@@ -0,0 +1,171 @@
...
+This extension allows authors to denote that the contained mesh data is in fact manifold, and adds a small amount of extra data to losslessly recover this manifold mesh by leveraging sparse accessors. 
+
+## Extending Meshes With Manifold Data
+
+A glTF mesh is denoted as a manifold by adding an `EXT_mesh_manifold` extension object to it and obeying the following restrictions:
+
+- all primitives **MUST** use the `TRIANGLES` topology type;
+- same attributes of different mesh primitives **MUST** reference the same accessor;
+- the `indices` property **MUST** be defined for all primitives and the referenced accessors with index data **MUST** reference the same buffer view.
+
```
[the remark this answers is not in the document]
elalish: done.

extensions/2.0/Vendor/EXT_mesh_manifold/README.md
```diff
@@ -0,0 +1,171 @@
...
+- its `indices` property **MUST** reference an `accessor`, which **MUST** in turn reference the same `bufferView` as the indices accessors of other primitives of the same mesh;
+- when its `indices` `accessor` includes a `sparse` property, the sparse data also defines which vertices are to be merged, but every index merged **MUST** have identical position values for the before and after vertices referenced, so that the geometry is unchanged;
+- it **MUST NOT** contain a material;
+- it **MUST NOT** contain morph targets, though the other primitives can.
+
+If the `indices` `accessor` is sparse, then `mergeIndices` and `mergeValues` **MUST** be included and reference `accessors` which are equivalent to the sparse indices and values, by referencing the same buffer views, offsets, component type, and count. This small amount of JSON is duplicated here because sparsity is generally considered an internal detail of compression, but here it also represents the relationship between the separate primitives and the unified manifold. If the mesh contains only primitives that are already disjoint and independently manifold, then no sparsity or merge `accessors` are needed.
```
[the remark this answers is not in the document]
elalish: Yeah, that makes sense. I had to rework things a bit so it makes sense to read. Look good?

extensions/2.0/Vendor/EXT_mesh_manifold/README.md
```diff
@@ -0,0 +1,177 @@
...
+      "count": 8592,
+      "bufferView": 3,
+      "byteOffset": 6912
+    }
+  ],
+```
+
+## Usage
+
+While this extension places additional restrictions on how a mesh is stored, it is still a valid glTF 2.0 and thus will render properly even on software that does not implement support for this extension. For software that needs a manifold mesh rather than a rendering mesh, the simplest path is to simply replace the set of `primitives` with the `manifoldPrimitive`, which is still a valid glTF, but is simply missing all non-position `attributes`.
```
lexaknyazev, line 173: >  For software that needs a manifold mesh rather than a rendering mesh, the _simplest_ path is to _simply_ replace the set of `primitives` with the `manifoldPrimitive`, which is still a valid glTF, but is _simply_ missing all non-position `attributes`.

consider _simplifying_ this to

> Software that needs a manifold mesh rather than a rendering mesh **SHOULD** ignore the original mesh primitives and use the `manifoldPrimitive` instead.

elalish: Good call - and I managed to remove another superfluous paragraph while I was at it.

lexaknyazev (approved):

elalish:
Looks like we're official now - thanks all! Let's exchange some solid geometry!

Issue #1365, opened 2018-06-08
EXT_hdr_textures


MiiBond:
I'm writing an extension that adds the ability to reference high dynamic range images from within glTF and I'd like some feedback. For context, we (Adobe) are interested in using this in conjunction with the KHR_environments extension being discussed #946. I suppose it's also possible to use these as emissive textures in your materials.

There are two main possibilities for HDR storage:
1. An existing HDR file format like .hdr, .exr or .heif 
1. Packing HDR data into something that glTF already supports; .png. 

Option 1 Pros: 
1. Best quality 
1. Doesn't require shoe-horning data that the format wasn't designed for. 

Option 2 Pros: 
1. Smaller download size 
1. Potentially significantly smaller in-memory size (though this depends on the encoding and runtime choices) 
1. glTF already understands PNG. 
1. Can be combined with MSFT_dds_texture extension and future format extensions to encode HDR is other file types. 

I'm leaning more towards option 2 but, really, we could define an extension that supports both. To encourage adoption, however, I'm going to suggest that we keep the number of possibilities low. 

```
"images": [ 
  { 
    "uri": "my_hdr_image.png", 
    "mimeType": "image/png", 
    "extensions": { 
      "EXT_hdr_textures": { 
        "encoding": "RGBD", 
        "maxValue": 256 // optional? We can certainly decide on recommended values for RGBM and RGBD
      } 
    } 
  } 
] 
 ```

I'm thinking of limiting the encoding types to just **RGBM** and **RGBD**. **RGBE** and **LogLuv** are also possible, and are higher-quality, but they don't interpolate well so need to be unpacked to float textures in memory to be useful. This makes them okay delivery formats but eat up a lot of GPU memory. **RGBM** and RGBD also don't interpolate correctly, of course, but it's far less noticeable and, in my experience, isn't a major concern. 

Another possible hit against **RGBE** and **LogLuv** is that they look garbled when just viewed straight-up in the browser. Is that something we should be concerned about? I believe SketchFab used to (and maybe still does) use **RGBE** and that's also what I've used in the past. Though, my personal choice landed on **RGBE** because I found that **RGBM** lacked the range I needed. I've later learned that I probably wasn't using it correctly as it's important that the packing and unpacking happen in gamma space, not linear. 

[...]

PR #2288, opened 2023-05-12
NV_materials_mdl

tblutNV:
The NV_materials_mdl extensions enables glTF assets to store and transfer materials defined in the NVIDIA Material Definition Language (MDL). The intention of this extension is to enable higher quality rendering based on MDL materials instead of glTF 2.0 PBR materials. The material definitions are mutually exclusive, therefore, if a renderer does not support this extension, then closely matching standard PBR materials can be used as a fallback to achieve a representative rendering of the glTF asset.

Two additional vendor extensions, NV_image_exr and NV_image_vdb, that enable images to be stored and transferred in the OpenEXR and OpenVDB formats, respectively, are added as well. The intention of defining these extensions only on images and not textures, like other image format extensions do, is that these formats are only needed for consumption in MDL and there is no obvious use-case for them in regular glTF textures.

We're looking forward to further discussions.


donmccurdy:
> The intention of defining [EXR and VDB] only on images and not textures, like other image format extensions do, is that these formats are only needed for consumption in MDL and there is no obvious use-case for them in regular glTF textures.

One notable reason for extending the texture rather than the image: it provides a fallback mechanism. The example below would require support for OpenEXR, while allowing clients to optionally choose a KTX2 version instead, for purposes of upload time or VRAM cost.

```jsonc
{
    "asset": {
        "version": "2.0"
    },
    "extensionsUsed": [
        "KHR_texture_basisu",
        "NV_texture_exr"
    ],
    "extensionsRequired": [
        "NV_texture_exr"
    ],
    "textures": [
        {
            "extensions": {
                "NV_texture_exr": {
                    "source": 0
                },
                "KHR_texture_basisu": {
                    "source": 1
                },
            }
        }
    ],
    "images": [
        {
            "mimeType": "image/x-exr",
            "bufferView": 1
        },
        {
            "mimeType": "image/ktx2",
            "bufferView": 2
        }
    ]
}
```

The example above is hypothetical at the moment — new extensions would be required to use float16 or float32 KTX2 files in glTF, as well. But this is worth considering, particularly if any of the the proposals here evolved into EXT_ or KHR_ extensions.

[EXT_lights_image_based](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_lights_image_based), [KHR_materials_emissive_strength](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_emissive_strength), and [KHR_lights_environment](https://github.com/KhronosGroup/glTF/pull/1956) (draft proposal) extensions could plausibly benefit from OpenEXR textures<sup>1</sup>. Moreover, implementations of NV_materials_mdl might plausibly choose to support KTX2 textures with fallback to OpenEXR textures, for faster upload of texture data to the GPU.

tblutNV:
The fallback mechanism is a valid point. This would also allow us to use the OpenEXR images that are referenced in MDL materials in the standard PBR materials which we use a fallback if NV_materials_mdl isn't supported. OpenVDB doesn't have any obvious use case outside of NV_materials_mdl though, given that it's volume data. In MDL we use it for 3D textures which are not supported by glTF 2.0 currently, but this might change in the future of course, so it wouldn't really hurt to allow OpenVDB resources to be used in textures, I guess?

fire:
Does gltf have a extension for loading openexr images? That could be a good addition similar to BPTC bc6 and webp extensions for hdr and regular images.

donmccurdy:
I'm not aware of a previous EXT_texture_exr proposal. For realtime rendering, I would really lean toward KTX2's float16, float32, or BC6H capabilities though.

donmccurdy:
Correction to my earlier post — one related proposal:

- https://github.com/KhronosGroup/glTF/issues/1365

I believe @hybridherbst and Needle Tools might also be using OpenEXR textures in glTF extensions through a vendor extension, for lightmaps.

hybridherbst:
That's correct. Ours is modelled pretty much 1:1 after [EXT_texture_webp](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_texture_webp/README.md). The reason for going that route is that we're definitely seeing future use of EXR textures as emissive textures, besides the usage as IBL, lightmap or reflection probe (our current usage for the extension).

I agree with what @donmccurdy mentioned regarding the possibility of defining fallbacks for textures, too.

<details>
<summary><em>Example glTF structure</em></summary>
<code><pre>
{
  "asset": {
    "generator": "Needle Engine",
    "version": "2.0"
  },
  "extensionsUsed": [
    "NEEDLE_gameobject_data",
    "EXT_texture_exr",
    "NEEDLE_lightmaps",
    "NEEDLE_lighting_settings"
  ],
  ...
  "images": [
    {
      "mimeType": "image/exr",
      "bufferView": 6,
      "name": "Skybox"
    }
  ],
  ...
  "textures": [
    {
      "sampler": 0,
      "source": 0,
      "name": "Skybox",
      "extensions": {
        "EXT_texture_exr": {
          "source": 0
        }
      }
    }
  ],
  ...
}</pre></code>
</details>

I can put up a proposal PR for `EXT_texture_exr` if that helps. I believe the main point of discussion last time was that EXR is a relatively complex format, and while the "core" of it for "I just want a single layer of floating point data" is pretty agreed upon there's numerous edge cases that one wouldn't really want to impose on loaders.

tblutNV:
We decided to change the image extensions to texture extensions. I'm assuming our proposal for NV_texture_exr looks very similar to your extension @hybridherbst ? We don't explicitly mention any edge cases though. Our only requirement is basically that OpenEXR images are allowed to be listed in the images array for consumption by NV_materials_mdl.

lexaknyazev:
glTF images are barely resource pointers so adding new media types there does not hurt because core glTF textures are allowed to refer only to PNG and JPEG images. On the other hand, glTF textures are referenced by materials (both core and extended) and must conform to various context-dependent rules.

Conceptually, glTF textures should be regarded only as a tuple of a 2D image (with mip levels in case of KTX) and an associated sampler state. The fallback mechanism for WebP and BasisU extensions ensures that new formats do not change anything beyond storage and/or VRAM consumption.

The original MDL proposal was well-scoped because only MDL internal objects would be allowed to use EXR and VDB images. Now with both formats proposed as texture extensions, there are multiple cases that dramatically increase the scope because now any texture slot in any material may refer to texture objects with EXR and/or VDB payloads. Therefore, these texture extensions now have to meticulously define EXR and VDB for non-MDL use cases.

tblutNV:
NV_texture_exr also defines a fallback mechanism like the WebP extension does @lexaknyazev. However, NV_texture_vdb doesn't since it isn't possible to fall back to a 2D PNG or JPG for volumetric data. If the extensions must exactly define how the features of the formats should be used in textures, then we'd rather stick to our original proposal of only extending image and keeping everything contained in MDL internal objects, given our tight time constraint. We won't be able to invest too much time into this at this moment. We would be willing to help define a EXT_texture_exr in the future though, but wouldn't be able lead the initiative.

donmccurdy:
> glTF textures should be regarded only as a tuple of a 2D image (with mip levels in case of KTX) and an associated sampler state

I'm not sure I agree on this point. Should 3D textures, array textures, or animated textures someday have a purpose in glTF, I would expect to find them in this array.


> ...there are multiple cases that dramatically increase the scope because now any texture slot in any material may refer to texture objects with EXR and/or VDB payloads. Therefore, these texture extensions now have to meticulously define EXR and VDB for non-MDL use cases.

This doesn't need to be the case, correct? The `*_texture_exr` extension could say, for example, _"This extension may not be not be attached to any texture unless that texture is unused. Other extensions may loosen this requirement."_

hybridherbst:
@lexaknyazev in our case we exactly based the extension on how EXT_texture_webp is defined. Would you say that does it wrong too? Personally I agree with @donmccurdy here, some notes on expected usage would work but having different formats defined very differently feels wrong from a consistency standpoint.

My takeaway here is that I'll open a PR for `EXT_texture_exr` so that the discussion here isn't derailed too much and we can continue there.

lexaknyazev:
> Should 3D textures, array textures, or animated textures someday have a purpose in glTF, I would expect to find them in this array.

That would be true for glTF 1.0 that directly binds textures to custom shaders.

Texture objects of glTF 2.0 have to remain compatible with materials that use them. Since there is no unambiguous way to use, e.g., a texture object with VDB payload as a material's base color, a VDB texture extension has either to define it, which is clearly out of NV scope, or completely ban it, which makes putting such payloads in glTF's texture objects much less valuable.

> Different entrypoints for texture formats, with different fallback mechanisms, seems to needlessly complicate client implementation from my perspective.

They are not so different after all
- Actual images go to the `images` array anyway and MDL objects would refer to them directly without being restricted by glTF's textures.
- A restricted EXR texture extension for some other use cases would still be possible and it may even refer to the same images as MDL objects do as long as the images are compatible.

donmccurdy:
> A restricted EXR texture extension for some other use cases would still be possible and it may even refer to the same images as MDL objects do as long as the images are compatible.

That's a very interesting idea... in that case, possibly the `NV_materials_mdl` extension does not need to be paired with these VDB or OpenEXR extensions at all, it could just reference normal images containing these file formats? Does MDL also define sampler and texCoord properties, if those are needed?

Then if we someday need glTF materials referencing OpenEXR textures, `EXT_texture_exr` or another extension would provide that access point.

tblutNV:
> That's a very interesting idea... in that case, possibly the `NV_materials_mdl` extension does not need to be paired with these VDB or OpenEXR extensions at all, it could just reference normal images containing these file formats?

I don't think simply including such file formats in an asset without specifying extensions for them is a valid option. How would the loader know which formats are required to be supported otherwise?

> Does MDL already define sampler and texCoord properties, if those are needed?

Yes, texture coordinates and certain sampler properties can be defined in MDL. The texture coordinates of the glTF mesh can be accessed in MDL, but there is no restriction on which values are passed to the texture lookup functions since they are simply float2/float3 values.

tblutNV:
We decided to update the `NV_materials_mdl` extension with your suggestion @lexaknyazev, so support for recognizing OpenEXR and OpenVDB image objects is now described in `NV_materials_mdl` and the `NV_image_*` extensions have been removed.

lkettnerNV:
Hi everybody. There was no time in yesterdays call to discuss this extension. However, I think we are pretty far along with this one and we adopted the improvements suggested here on how to handle the image format extensions. I would be happy if the can finalize this pull request in this thread also without meeting.  I can also be available for the next call next week, but would not be available after that until late August.   We are of course available and happy to answer all questions you might have here, and very grateful for all the great feedback we received here.

emackey:
@lexaknyazev When you get a chance can you post the last round of review comments for the JSON schemas here?  We'll status again at the PBR meeting on Monday.  Thanks!

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+]
+```
+
+## Defining MDL Modules
+All modules that are referenced by the function calls in the `functionCalls` array must be listed in the `modules` array. A `module` has the following properties:
+
+| Property | Type | Description | Required |
+|:---------|:-----|:------------|:---------|
+| `uri` | `string` | The URI (or IRI) of the MDL module. | No |
+| `bufferView` | `integer` | The index of the bufferView that contains the MDL module. | No |
```
lexaknyazev, line 73: > The index of the bufferView

Please use plain text when not using formatting, i.e.:
>The index of the buffer view

Also in the text below.

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+| `mimeType` | `string` | The media type of the bufferView's data. | No |
```
lexaknyazev, line 74: Are other media types except `application/vnd.mdl` expected here? If not, this property seems unnecessary.

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+| `modulePath` | `string` | Relative path of the module. | No |
+| `name` | `string` | glTF name of the module. | No, Default: `""` |
```
lexaknyazev, line 76: The `name` property of glTF's top-level array elements does not have a default value.

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+
+Each module defines either the `uri` property, which contains the URI (or IRI) of an external MDL module file or a data-URI with embedded data, or the `bufferView` property, which points to a bufferView object that contains an MDL module. If the `bufferView` property is used then `mimeType` must also be defined and have the value `application/vnd.mdl`. If a data-URI or bufferView object is used then the `modulePath` property must be defined, which specifies a path relative to the glTF file that is used for importing by other modules, e.g. if an MDL module that has the path `./my/embedded/module.mdl` relative to the glTF file is embedded into a bufferView object then the `modulePath` property must have this path as its value.
```
lexaknyazev, line 78: - nit: Missing comma after 'e.g.'
- Mention that data-URI must have the `application/vnd.mdl` media type.
- Is it allowed to have both `uri` and `modulePath` defined at the same time?

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+
+| Property | Type | Description | Required |
+|:---------|:-----|:------------|:---------|
+| `module` | `integer` | The containing module's index. | No |
+| `functionName` | `string` | The unqualified function name. | :white_check_mark: Yes |
+| `type` | `Type` | The function's return type. | :white_check_mark: Yes |
+| `arguments` | `array` | A list of function arguments. | No, Default: `[]` |
+| `name` | `string` | glTF name of the function call. | No, Default: `""` |
+
+For built-in functions, e.g. the `float3` and `texture_2d` constructors, the `module` property must not be defined. The operator functions as defined in Section 12.10 of the [MDL Language Specification](https://github.com/NVIDIA/MDL-SDK/tree/master/doc/specification) are valid built-in functions.
```
lexaknyazev, line 138: nit: Missing comma after 'e.g.'

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+```
+
+## Defining MDL Function Calls
+Function calls are listed in the `functionCalls` array. A `functionCall` can have the following properties: 
+
+| Property | Type | Description | Required |
+|:---------|:-----|:------------|:---------|
+| `module` | `integer` | The containing module's index. | No |
+| `functionName` | `string` | The unqualified function name. | :white_check_mark: Yes |
+| `type` | `Type` | The function's return type. | :white_check_mark: Yes |
```
lexaknyazev, line 134: Consider using a more specific identifier for the type of the `type` property (2nd column). E.g., `functionCall.type` instead of `Type`.

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+| `value` | `any` | The literal value. | No |
+
+Function arguments can either be value or call arguments, depending on if the `functionCall` or `value` properties are specified. If the `value` property is specified then the `type` property must also be specified. A call argument references a function call who's return value is used as the argument's value. A value argument's value is defined directly as a literal. The following types are supported as value arguments and can be specified as literals:
+
+* Built-in scalar, vector, and matrix types
+* `string` and `color`
+* Built-in and user-defined enumeration types
+* Arrays of the above types
+
+For the vector and matrix types each component must be specified seperately and for a `color` value the `r`, `g` and `b` values must be specified. Note that matrices in MDL follow column-major conventions, e.g. `float2x3` has 2 columns and 3 rows, and thus, matrix values are to be specified as an array with the values in column-major order. For enumerations, the value name should be specified as a string, e.g. `intensity_radiant_exitance` for the built-in enumeration type `intensity_mode`.
```
lexaknyazev, line 171: - typo: separately
- nit: Missing comma after 'e.g.'

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+        {
+            "name": "value2",
+            "functionCall": 2
+        }
+    ]
+}
+```
+
+### Member Accessor
+Member accessors (the *dot operator*) are named after the member being accessed preceded by the structures name followed by the dot operator (e.g. `texture_return.tint` or `float3.x`). The function has a single argument `s` which is the instance of the structure type from which the member should be accessed and the return type is the type of the member. Unlike most other special functions the member accessor is defined in the module in which the structure type is defined.
```
lexaknyazev, line 250: nit: Missing comma after "Unlike most other special functions"

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+                "typeName": "color_pair"
+            },
+            "functionCall": 1
+        }
+    ]
+}
+```
+
+### Array Length Operator
+The name of the array length operator function is `operator_len` and has a single argument `a` which must be an array type. The return type is `int`.
```
lexaknyazev, line 294: Given that the number of array elements is known at compile time, why is this operator needed?

extensions/2.0/Vendor/NV_materials_mdl/README.md
tblutNV: Yes, `uri` and `modulePath` must be defined at the same if `uri` contains a data-URI, since `modulePath` acts as a "virtual file path" for the cases where no real file path is provided.

extensions/2.0/Vendor/NV_materials_mdl/README.md
tblutNV: fixed

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+
+## Defining MDL Function Calls
+Function calls are listed in the `functionCalls` array. A `functionCall` can have the following properties: 
+
+| Property | Type | Description | Required |
+|:---------|:-----|:------------|:---------|
+| `module` | `integer` | The containing module's index. | No |
+| `functionName` | `string` | The unqualified function name. | :white_check_mark: Yes |
+| `type` | `Type` | The function's return type. | :white_check_mark: Yes |
+| `arguments` | `array` | A list of function arguments. | No, Default: `[]` |
```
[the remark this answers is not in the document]
tblutNV: fixed

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+A fully conforming implementation must also support the `EXT_lights_ies` extension.
+
+## Overview
+
+This extension provides the ability to use materials in glTF that are defined in the NVIDIA Material Definition Language (MDL) as defined in the [MDL Language Specification](https://github.com/NVIDIA/MDL-SDK/tree/master/doc/specification). The intention of this extension is to enable higher quality rendering based on MDL materials instead of glTF 2.0 PBR materials.
```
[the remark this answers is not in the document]
tblutNV: We reworded this section and believe it better describes the intention of this extension now, which is to enable high-quality material definitions based on MDL in addition to the glTF 2.0 PBR model. Thanks for pointing this out.

extensions/2.0/Vendor/NV_materials_mdl/schema/functionCall.schema.json
```diff
@@ -0,0 +1,38 @@
...
+            "description": "The return type of the function."
+        },
+        "arguments": {
+            "type": "array",
+            "items": {
+                "$ref": "functionCall.argument.schema.json"
+            },
+            "default": [],
+            "minItems": 1,
+            "description": "A list of named value and/or function call arguments."
```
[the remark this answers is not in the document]
tblutNV: I added the `uniqueItems` property, so this case should be covered now as well. Thanks.

extensions/2.0/Vendor/NV_materials_mdl/schema/functionCall.argument.schema.json
```diff
@@ -0,0 +1,91 @@
...
+                {
+                    "type": "array",
+                    "items": {
+                        "type": "number"
+                    },
+                    "minItems": 4,
+                    "maxItems": 16
+                },
+                {
+                    "type": "array",
```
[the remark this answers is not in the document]
tblutNV: Yes, this is actually intended since zero-sized arrays in MDL are allowed.

extensions/2.0/Vendor/NV_materials_mdl/schema/glTF.NV_materials_mdl.schema.json
```diff
@@ -0,0 +1,44 @@
...
+            "description": "The list of all MDL modules."
+        },
+        "userTypes": {
+            "type": "array",
+            "items": {
+                "$ref": "userType.schema.json"
+            },
+            "minItems": 1,
+            "description": "The list of all user-defined types."
+        },
```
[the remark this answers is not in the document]
tblutNV: This was an oversight by me. Thanks for pointing it out :)

extensions/2.0/Vendor/NV_materials_mdl/README.md
tblutNV: I'm not really sure how to choose a more specific identifier here since we are referring to the object type defined in `type.schema.json`. The same object type is used in function call arguments, so `functionCall.type` doesn't seem right. Any suggestions on how to handle this better? Maybe rename `type.schema.json` to `functionCall.type.schema.json`, but still use it in `functionCall.argument.schema.json`? Not sure if this is good practice or not.

extensions/2.0/Vendor/NV_materials_mdl/README.md
```diff
@@ -0,0 +1,511 @@
...
+}
+```
+
+The `name` argument of the MDL resource type constructors refer to the following resource arrays:
+| Resource Type | glTF Array | Description |
+| --------------|------------|-------------|
+| `texture_2d` | `images` | The core glTF images array. |
+| `texture_3d` | `images` | The core glTF images array. |
+| `texture_cube` | `images` | The core glTF images array. |
+| `texture_ptex` | `images` | The core glTF images array. |
```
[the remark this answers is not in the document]
tblutNV: The MDL Spec doesn't define which image formats can be used for each texture type, that's why we don't do it here either.

extensions/2.0/Vendor/NV_materials_mdl/README.md
lexaknyazev: Besides data-URI, is it allowed to have `modulePath` and `uri` simultaneously?

extensions/2.0/Vendor/NV_materials_mdl/schema/functionCall.schema.json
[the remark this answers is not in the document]
lexaknyazev: `uniqueItems` checks for full-object equality, so it would allow two arguments with the same name but different types.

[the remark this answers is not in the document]
tblutNV: So, the only way to "enforce" this is to write into the description that two arguments with the same name are not allowed?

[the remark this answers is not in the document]
lexaknyazev: Yes, with normative language, i.e., `**MUST NOT**`.

extensions/2.0/Vendor/NV_materials_mdl/README.md
tblutNV: If the URI isn't a data-URI, then they shouldn't be allowed simultaneously. Or in other words, the implementation should ignore `modulePath` then.

extensions/2.0/Vendor/NV_materials_mdl/README.md
tblutNV: I opted to rename `type.schema.json` to `functionCall.type.schema.json` and reference `functionCall.type` for the `type` properties, as you suggested.

extensions/2.0/Vendor/NV_materials_mdl/README.md
tblutNV: This operator is useful for composing function call graphs in authoring tools. Even if the array size is known, querying it makes it more convenient in some cases when authoring the graphs. For completeness, we want to be able to represent all possible expressions in the call graph. We don't want to require having different call graphs during import and export, due to replacing function calls to the array size operator with constants. This choice should be left to the authoring tools.
Granted, the example for this operator in the README could have easily been replaced with a constant. I changed the example, so it calls the operator on the return value of a function call, which is a more realistic use-case.

extensions/2.0/Vendor/NV_materials_mdl/schema/functionCall.schema.json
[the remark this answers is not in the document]
tblutNV: I removed `uniqueItems` again and used normative language, as you suggested. I also clarified some other things in the schema with normative language.

extensions/2.0/Vendor/NV_materials_mdl/README.md
tblutNV: I clarified this in the README and schema. Now it should be clear that `modulePath` must not be defined if `uri` is defined and isn't a data-URI.

extensions/2.0/Vendor/NV_materials_mdl/schema/functionCall.schema.json
```diff
@@ -0,0 +1,37 @@
...
+    "type": "object",
+    "description": "Function call with its list of arguments. Can represent the entry point into a function call graph or be a node in such a graph.",
+    "allOf": [ { "$ref" : "glTFChildOfRootProperty.schema.json" } ],
+    "properties": {
+        "module": {
+            "allOf": [ { "$ref": "glTFid.schema.json" } ],
+            "description": "The ID of the containing module.  This field **MUST NOT** be defined if a built-in function is specified."
+        },
+        "functionName": {
+            "type": "string",
```
lexaknyazev, line 14: Are empty strings allowed here?

extensions/2.0/Vendor/NV_materials_mdl/schema/functionCall.type.schema.json
```diff
@@ -0,0 +1,49 @@
...
+    "type": "object",
+    "description": "MDL type describing either a built-in or user-defined type, or an array of a built-in or user-defined type.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "module": {
+            "allOf": [ { "$ref": "glTFid.schema.json" } ],
+            "description": "The ID of the containing module.  This field **MUST NOT** be defined if a built-in type is specified."
+        },
+        "typeName": {
+            "type": "string",
```
lexaknyazev, line 14: Are empty strings allowed here?

extensions/2.0/Vendor/NV_materials_mdl/schema/functionCall.type.schema.json
```diff
@@ -0,0 +1,49 @@
...
+                {
+                    "type": "string",
+                    "const": "varying"
+                },
+                {
+                    "type": "string",
+                    "const": "uniform"
+                },
+                {
+                    "type": "string"
```
lexaknyazev, line 37: This case allows any string value. Is it intentional?

extensions/2.0/Vendor/NV_materials_mdl/schema/glTF.NV_materials_mdl.schema.json
```diff
@@ -0,0 +1,36 @@
...
+            "description": "The list of all function calls."
+        },
+        "bsdfMeasurements": {
+            "type": "array",
+            "items": {
+                "$ref": "bsdfMeasurement.schema.json"
+            },
+            "minItems": 1,
+            "description": "The list of all BSDF measurements."
+        },
```
lexaknyazev, line 32: `functionCall` is required in the material extension object. Should `functionCalls` array be also required here? A counter-example is using the MDL extension only to embed BSDF measurements in a glTF asset without referring to them.

extensions/2.0/Vendor/NV_materials_mdl/schema/functionCall.type.schema.json
tblutNV: Good catch, I fixed it here and for `functionCall.functionName`.

extensions/2.0/Vendor/NV_materials_mdl/schema/glTF.NV_materials_mdl.schema.json
tblutNV: I don't think that `functionCalls` should be required for the exact reason you stated. The `functionCall` field in the material is required, but the extension object itself isn't. So, I believe not requiring any of the arrays in the glTF extension object is fine.

lexaknyazev (approved):

PR #2293, opened 2023-05-31
Add KHR_interactivity draft

lexaknyazev:
Rendered version: https://github.com/KhronosGroup/glTF/blob/interactivity/extensions/2.0/Khronos/KHR_interactivity/Specification.adoc

---

<sub>Edited by javagl: For those looking at this PR in the future, the link is now https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_interactivity/Specification.adoc , with the exact version from this PR being https://github.com/KhronosGroup/glTF/blob/b11fb686120f822afc85b89960ab8e7e8f51b026/extensions/2.0/Khronos/KHR_interactivity/Specification.adoc</sub>


bhouston:
Great work!  What sections are next?  Also what could we release as a draft extension for SIGGRAPH 2023?

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,834 @@
...
+| Output value sockets | `float value` | 2.718281828459045
+|===
+
+===== Pi
+
+[cols="1h,1,2"]
+|===
+| Type | `math/pi` | Ratio of a circle's circumference to its diameter
+| Output value sockets | `float value` | 3.141592653589793
+|===
```
aaronfranke, line 279: If we have Pi, we should also have Tau. https://tauday.com/tau-manifesto The value Tau is the circle constant (equal to 2*Pi, or rather, Pi is half of Tau). Using Tau usually results in more readable code. Tau is supported in many programming languages such as [C#](https://learn.microsoft.com/en-us/dotnet/api/system.math.tau?view=net-7.0), [Python](https://docs.python.org/3/library/math.html#math.tau), [GDScript](https://docs.godotengine.org/en/stable/classes/class_@gdscript.html#constants), [Unreal Blueprints](https://docs.unrealengine.com/5.3/en-US/BlueprintAPI/Math/Trig/GetTAU/), and more, so it's useful for interoperability with other languages, especially Unreal Blueprints which is conceptually similar.

```
===== Tau

[cols="1h,1,2"]
|===
| Type | `math/tau` | The circle constant, the circumference of the unit circle in radians.
| Output value sockets | `float value` | 6.2831853071795862
|===
```

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -1835,6 +1835,48 @@ The graph execution **MAY** be terminated when the amount of pending activations
...
+[cols="1h,1,2"]
+|===
+| Type | `variable/get` | Get a custom variable value
+| Configuration
+| `int variable` | The custom variable index
+| Output value sockets
+| `T value`  | The custom variable value
+|===
+
+This node gets a custom variable value using the variable index provided by the `variable` configuration value. The type `T` is determined by the referenced variable. The variable index **MUST** be non-negative and less than the total number of custom variables, otherwise the node is invalid.
```
aaronfranke, line 1851: Why are variables accessed by index instead of by name? It would make more sense to me to set a variable called `"money"` than one called `27`.

aaronfranke (changes requested):
https://www.khronos.org/blog/gltf-interactivity-specification-released-for-public-comment

In this recent blog post, it says that interactivity is ready for public comment, review, and feedback, and ratification is soon. I have multiple questions.

* The blog post says to give feedback by "raising an issue on the [...] pull request". Do they mean leaving a comment or a review?
* What is the time period for public comment, review, and feedback?
    * It says ratification may happen in the third quarter of 2024, so is the public comment period shorter than this?
* How will Khronos respond to public comment, review, and feedback?
    * I left comments on this PR last year, when will they have a response?
* The blog post says "This initial specification will be updated over time to include additional specific action triggers, and Khronos is also actively working on approaches to incorporating audio and video into interactive glTF assets."
    * When will Khronos give attention to [KHR_audio_emitter](https://github.com/KhronosGroup/glTF/pull/2137)? If the goal is to use it from within KHR_interactivity, surely it should be ratified and merged first?
* This PR is missing JSON schemas, where are they?
* This PR is missing example files, where can I find a glTF file with KHR_interactivity inside of it?
* Where can I find an example implementation of KHR_interactivity?
    * Which game engines have an implementation of KHR_interactivity, even experimental/unmerged?
    * I see there is the "[DCC React App for Authoring glTF Assets with KHR_interactivity Extension](https://github.com/KhronosGroup/glTF-InteractivityGraph-AuthoringTool/tree/initial-work-merge)", but an authoring tool by itself is not enough, we need multiple implementations that can read and execute the interactivity graph as evidence that this standard is portable and interoperable.
* Why doesn't the description of this PR contain this information?

I'm not trying to be rude, but this seems like it's still in a relatively immature state, as there are many things missing. It does not seem proper for Khronos to jump from "missing important pieces" to "ratified and final" in such a short timespan.

Please, take the time to build out the specification more before finalizing it. There is an important lesson to be learned from Collada: It's not enough to have *a* standard, it needs to be a *good* standard. And we can't assess if it's a good standard unless we have battle-tested it in many game engines, and give a much longer time for public review, perhaps a year.

madjin:
I'd like to echo @aaronfranke's comment above regarding the current state of the `KHR_interactivity` spec. There are some critical issues that need addressing:

### Key Concerns

1. **Incomplete / Missing Files:**
   - More time, example files, and implementations are needed for a spec like this, especially given its ties to other extensions like physics and audio. The quality bar for this spec should be above average for this reason.

2. **Public Comment Process:**
   - Clarification is needed on the feedback process mentioned in the blog post. Specifically:
     - What is the exact time frame for public comment?
     - When can we expect responses to previous comments left on the PR?

3. **Development:**
   - Attention should be given to `KHR_audio_emitter` if it is to be used within `KHR_interactivity`.
   - JSON schemas and example files are missing from the current PR.
   - Specify example implementations and supported game engines

Rushing to ratification could be detrimental or harm public perception of the standards process. At the very least can we extend the public review period for thorough testing and validation to ensure `KHR_interactivity` becomes a robust and well-developed standard? Thanks!

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
aaronfranke: In case more justification is needed, the KHR_interactivity document currently contains Degrees-To-Radians (`math/rad`) and Radians-To-Degrees (`math/deg`). These allow converting between degrees and radians. Tau can be used for a similarly important purpose, dividing or multiplying by this number allows converting between turns and radians, where 1 turn is τ radians. Tau can be used for more than just this, but the point is, if `math/rad` and `math/deg` are justified as a part of KHR_interactivity, then `math/tau` is also justified.

javagl (commented):
A quick review pass

- This is based on the AsciiDoc preview (i.e. not checking the formulas)
- It is a _quick_ pass - i.e. I did not think through all conditions, implications, and corner cases
- I have not been toooo deeply involved in large parts of the discussion recently
- I did already provide some thoughts and feedback when the discussion started...

---

The type system can be a rabbit hole of endless philosophical discussions. The current state is pretty much focussed (one could say "narrowly focussed") on what is supposed to be supported _currently_ within glTF. But there **will** be the question of whether there should also be a `float3x3`, or `intN` types, or `doubleN` types, or whether a "value type" of a socket could also be a `string`, or whether there is any consideration for how to handle _arrays_. 
(I'm not saying what is "good or bad" or "right or wrong" here - just that one should be prepared for these questions...)

---

From skimming over the set of nodes, most of them look like pragmatic, straightforward 1:1 mappings of the functionality of JavaScript or common programming languages. Most room for discussion can probably be found in the ones that refer to the execution model.

---

Section "3.6.3.1.7. Multi Gate": 

This looks oddly specific, mainly due to that `isRandom` flag. 

On a brainstorming level: I wonder whether it could make sense to split this into "Multi Gate" and "Random Multi Gate". Yes, the `isRandom` flag can be toggled _at runtime_. And if there were _arrays_, one could build a that "random" functionality based on something like a "Multi Gate" that received an _array_ of `int[]` values (containing the order of the outputs), and some "Shuffle" node...

In general, I avoid "random" like the plaque, and think that it should **never** appear, **anywhere** in software development, **unless** there is a "random seed". Maybe the `isRandom` could be replaced by a "random seed" (with 0=not random or so). 

This may have to be thought through more thoroughly.

But even if nothing is changed there: It is not clear (i.e. not specified) whether the order will be _the same_ in each loop pass, or whether it will be randomized for each pass.

---

Section "3.6.3.1.8. Wait All":

In that context, I wondered whether there should be some sort of "(count down) latch": Something that activates an _output_ if an input was activated `N` times. (Sure, all of this could probably be "emulated" with other nodes - more on that at the bottom...)

---

Section "3.6.3.2.1. Set Delay":

This looks subtle, and raises a bunch of questions. The first one is: This (obviously) introduces _asynchronicity_. The implications of this are hard to grasp, and should be thought through _very_ carefully. 

An important, high-level question: How much precision (in terms of _timing_) is expected from an implementation here? 

Imagine a graph like this:
```
    /--- setDelay(1.0) ------------------------------ OP0
for< 
    \--- setDelay(0.5) --- setDelay(0.4999999999) --- OP1
```
Is it expected that `OP1` will be (guaranteed to be) executed _before_ `OP0`? (How much?)

---

Section "3.6.4.2. Object Model Access":

Some link to https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/ObjectModel.adoc could fit in here...

---


Section "3.6.4.2.3. Pointer Interpolate"

Too complex for me to think it through right now, but a high-level question: Couldn't this be emulated with a custom graph and just setters? (There would need to be a `slerp` operation for `float4` as quaternion). The point is whether this is an _integral_ function, or just a convenience building block for a common use-case...

---

(Have to stop here, will try to continue later at https://github.com/KhronosGroup/glTF/blob/interactivity/extensions/2.0/Khronos/KHR_interactivity/Specification.adoc#364-state-manipulation-nodes )

---

Another high-level question: 

In how far should the specification address things like "compound nodes"?

Let's be honest: There are *hundreds* of half-baked flow/execution-graph based formats and libraries out there (and I actually considered to implement a `KHR_interactivity` editor based on one of mine, namely https://github.com/javagl/Flow ). And one could talk endlessly about the trade-off between granularity and convenience.

For example: There currently is a `length` function for `float3`. One could just omit that, One could simply assemble this, based on `extract`, `mul`, `add`, and `sqrt`. Of course, that would be _horribly_ inconvenient. But even the _current_ level will be perceived to be inconvenient very quickly, *unless* there is a mechanism to define aggregate operations. 

One could _consider_ adding the option to define "sub-graphs" or "custom nodes" that assemble other nodes, so that commonly used functionality can be used, like a library, without having to repeatedly go though the tedious process of building the same graph structure. (I'm very well aware of the complexity that this entails. That's why I'm not _proposing_ to do this. I just wanted to bring it up...)

lexaknyazev:
> But there **will** be the question of whether there should also be a `float3x3`, or `intN` types, or `doubleN` types, or whether a "value type" of a socket could also be a `string`, or whether there is any consideration for how to handle _arrays_.

- Matrices other than 4x4 are trivially possible but adding them in advance would further increase required test/sample coverage so we'd like to have specific usecases first. Same with integer vectors.
- `doubleN` is not needed as all floats here always have double precision.
- Strings and arrays are not included to avoid dynamic memory allocations in the type system for now.

---

> Section "3.6.3.1.7. Multi Gate":
> 
> On a brainstorming level: I wonder whether it could make sense to split this into "Multi Gate" and "Random Multi Gate". Yes, the `isRandom` flag can be toggled _at runtime_.

It cannot. Node configuration is static, i.e., it's a compile-time flag. 

> But even if nothing is changed there: It is not clear (i.e. not specified) whether the order will be _the same_ in each loop pass, or whether it will be randomized for each pass.

It certainly does not have to be the same each time. We'll clarify whether it's explicitly randomized.

---

> Section "3.6.3.1.8. Wait All":
> 
> In that context, I wondered whether there should be some sort of "(count down) latch": Something that activates an _output_ if an input was activated `N` times.

We decided that implementing that with a custom variable and a branch is simple enough to not have a dedicated node. The "Wait All" is warranted because implementing it without arrays is too burdensome.

> Section "3.6.3.2.1. Set Delay":
> 
> An important, high-level question: How much precision (in terms of _timing_) is expected from an implementation here?

TBD, subject to implementors feedback.

---

> Section "3.6.4.2.3. Pointer Interpolate"
> 
> Too complex for me to think it through right now, but a high-level question: Couldn't this be emulated with a custom graph and just setters?

Having this node allows a graph to easily start multiple interpolations without manually tracking them. Implementing the same logic with a custom graph would be very inconvenient without arrays.

---

> In how far should the specification address things like "compound nodes"?

TBD, likely in the next revision.

aaronfranke:
> Maybe the isRandom could be replaced by a "random seed" (with 0=not random or so).

0 is a valid seed, so using 0 to mean "not random" should not be a thing. If anything should be changed, separating out the random and non-random versions makes the most sense to me.

> Section "3.6.3.2.1. Set Delay":
>
> An important, high-level question: How much precision (in terms of timing) is expected from an implementation here?

This would be an unreasonable requirement to impose on implementers, and would produce unexpected results for glTF asset authors. If I delay for 0.5 seconds in a glTF, I would expect this to delay 0.5 seconds from the point in time when the delay block was executed, not 0.5 seconds after the last delay.

As a solution to the problem of a race condition between 1.0 and 0.5 + 0.49999, I would suggest encouraging glTF asset authors to write behavior graphs that are robust to race conditions and will function in either case. Even if race conditions caused by delays were resolved in some convoluted solution, there is also the problem of race conditions in general. Trying to resolve all race conditions at the glTF implementation level is an unsolvable problem on the level of the halting problem; there are an infinite amount of cases to consider on an infinite amount of systems.

> > In how far should the specification address things like "compound nodes"?
>
> TBD, likely in the next revision.

Being able to build "functions" out of other nodes seems like a highly useful feature, with wide-reaching implications for how assets and implementations interact with each other (such as if one glTF provides a "library" for another to use, probably with both glTF files as part of a unified experience in a glXF file). It may be important enough to have in v1.

In a way, this would be similar to how glTF itself defines in its base specification how it can be extended using `"extensions"` as an explicitly allowed part of the schema. The big difference being that the implementation of the behavior graph "extension" *could* be itself provided by another behavior graph.

Yet it may also be highly useful to define a function at the glTF implementation level, to allow implementers to expose new functionality for the behavior graph to use. I did a search of the KHR_interactivity document for "exten" and it seems that only extending the type system is defined in "4.2. Types", but KHR_interactivity does not define any other recommended ways to extend KHR_interactivity. I could imagine a glTF extension defining its own `mynamespace/myblock` behavior graph node, but it may be better if KHR_interactivity recommends a pattern for this, such as `ext/KHR_something/some_func` mirroring glTF's own `"extensions"`.

javagl:
> Matrices other than 4x4 are trivially possible but adding them in advance would further increase required test/sample coverage so we'd like to have specific usecases first. Same with integer vectors.

Looking beneath the surface, the question refers to what I already brought up early in the process, when the discussion about `KHR_interactivity` started. Namely, how powerful, generic, and extensible the type system should be. Right now, we have `float4x4`. We could add `float3x3`, `float2x2`, `float4x3`, and later maybe `int2`, `int3`, `int4` (and find some relief in the fact that this still revolves around JSON, and the necessity to differentiate between (at least) _eight_ different types `(u)int(8/16/32/64)` is ruled out from the beginning). 
The deeper question is: At which point <strike>would</strike> will the effort of adding a new type become prohibitively large? Depending on how many types will have to be added, there _will_ inevietably be a point where the _initial_ difficulties that come with trying to define a type system that includes things like `matrix<rows, cols, componentType>` will outweigh the effort of adding a new type in a system where this triggers a combinatorial explosion.

But to emphasize that: I'm **NOT** advocating for generalizing the type system to include such "parameterized types" (The effort is so high that even most "real" programming languages don't get this right...). I just want to make sure that people are aware of the possible limitations of the current approach in terms of evolvability.

---

> > I wonder whether it could make sense to split this into "Multi Gate" and "Random Multi Gate". Yes, the isRandom flag can be toggled at runtime.
> 
> It cannot. Node configuration is static, i.e., it's a compile-time flag.

That sounds like it could be beneficial to make the distinction between "random" and "not random" _explicit_ via different nodes. At least, I could imagine that the implementation for both of them could be easier (and maybe more elegant and efficient) than trying to squeeze that "(optional) randomness" into this single node. (But of course, all that is a gut feeling for now - allocating enough time to actually try and _implement_ all that may be a challenge...)

@aaronfranke 

> 0 is a valid seed, so using 0 to mean "not random" should not be a thing. 

I expected that :-) And "giving a special meaning to a certain value" is something that I usually like to avoid. It was just an attempt to describe some sort of "middle ground"/"compromise" for the two main points of that comment:

- The "optional randomness" may warrant a new node type (and you seem to not be opposed to that in general)
- Whenever there is anything "random", there should definitely be a "random seed". 
 
I know, there is no built-in functionality for the latter in JavaScript. But this is one point where the quirks and shortcomings of this "language" should not leak through. Proper testing and debugging with a seedless 'random' is a pain in the back (or even impossible in many cases). 

---

> Trying to resolve all race conditions at the glTF implementation level is an unsolvable problem on the level of the halting problem

Yes! And that's the main (broader) point of that (suggestive) example of `1.0`-vs-` 0.5 + 0.49999`. Another way of phrasing it (similarly suggestive, and completely made-up "pseudocode") would be a test case for an execution engine:

```
var value = 0;
setDelay(1.0).decrement(value);                   // Happens last
setDelay(0.5).setDelay(0.49998).increment(value); // Happens first
setDelay(0.9999).assert(value == 1);              // Happens in-between
```

Again: I'm not proposing (or advocating for) anything. I just wanted to bring up that question, and if the answer is ~"there are no specific timing guarantees", then this _could_ be fine, but might have to be elaborated further in the specs. Otherwise, someone could build a graph that resembles the pseudocode above (even unintentionally, due to _assumptions_ that are made about the exact execution order), and may receive some erroneous result (like "divide by zero" or whatnot).

---

> Being able to build "functions" out of other nodes seems like a highly useful feature, with wide-reaching implications for how assets and implementations interact with each other (such as if one glTF provides a "library" for another to use, probably with both glTF files as part of a unified experience in a glXF file). It may be important enough to have in v1.

It would certainly be useful. And it could be "important" (to keep the graphs and their complexity manageable in the long run). But as you said: It opens a whole can of worms, and should be thought through _very_ carefully. I do have some of the related questions on the radar, but will abstain from further comments until it is supposed to be addressed explicitly. 

The meta-options of 
- "adding it in v1" vs. "adding it later in a v2" (implying a "breaking change"?!?)
- omitting it in v1, but adding some `KHR_interactivity_compound_node_library` extension later

will probably be part of the considerations here.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,2914 @@
...
+
+On the other hand, behavior graphs are a superset of trigger-action lists, meaning that the former can support everything that trigger-action lists can, and more. Behavior graphs support “Queries”, “Logic” and “Control Flow” nodes, making them more expressive and capable of creating more sophisticated behaviors. This makes behavior graphs the preferred method of choice for high-end game engines, as it offers an identical safety model as trigger-action lists while being more expressive.
+
+=== Turing Completeness
+The execution model and node choices for this extension mean that it is Turing-complete. This means that an implementation of this can execute any computation and it is also hard to predict if it will run forever (e.g. halt or not.)
+
+While this may present security implications, it is not a major hindrance and can be safely mitigated so that any implementation does not become susceptible to denial of services by badly behaving behavior graphs, whether intention or not.
```
eoineoineoin, line 194: Nit: should read " whether intentional or not."

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,2914 @@
...
+== Graphs
+
+A behavior graph is a JSON object containing _nodes_. It **MAY** also contain custom variables and custom events.
+
+Behavior graphs are directed graphs with no directed cycles.
+
+[[nodes]]
+== Nodes and Sockets
+
+A _node_ is a JSON object, which represents an executable item. Each node is defined by its _type_ and a set of _sockets_. There are four kinds of sockets.
```
aaronfranke, line 224: This is kind of a nitpick, and I've been debating whether or not to bring this up, but I figure I should.

I don't like that elements of the KHR_interactivity behavior graph are called "nodes". We already use the word "node" for something *very specific* in glTF [here](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/node.schema.json). It would be more confusing than necessary to overload this word with another meaning.

Instead, I recommend changing the name to "block", short for "script block". It is common in visual scripting systems to call their elements "visual script blocks". The term "block" is common, descriptive, and most importantly, is not already used for something else in glTF. It would be a clearer term to use.

madjin:
Meant to post these notes sooner, got some notes from X spaces that we hosted couple weeks ago with @bhouston and other Khronos contributors + OMI group:

## X Space 6-20-24

https://x.com/open_metaverse/status/1803902289742303743

• The interactivity spec originated from Adobe's work on trigger-action lists for Adobe Aero and USDZ.
• There was debate between trigger-action lists and behavior graphs approaches, with behavior graphs ultimately chosen.
• The spec development process involved studying existing systems like Unreal Blueprints and Unity Visual Scripting.
• The goal was to create a "boring" standard that consolidates best practices rather than innovating.
• The spec introduces concepts like the glTF object model, events, and variables that can be built upon by future extensions.
• It's designed with a layered approach, allowing different levels of capability and security.
• There are concerns about the spec being incomplete, lacking examples, JSON schemas, and other expected components. :star: 
• The current implementation is limited in terms of data types and capabilities compared to full programming languages.
• There's discussion about potential future work, including adding more complex data types and operations.
• Security considerations include limiting node execution time and restricting memory allocation.
• There's debate about whether the spec could support compilation from text-based languages to behavior graphs.
• The spec is seen as a foundation for future developments, potentially including WASM integration.
• There are questions about how the spec will be implemented in various engines and viewers.
• The community is encouraged to try the reference implementation and provide feedback.
• There are plans for more discussions and spaces to gather community input before ratification.
• The spec is part of a larger ecosystem of glTF extensions, including physics, audio, and procedural materials.

## X Space 6-21-24

https://x.com/glTF3D/status/1803548289134109034

• The genesis of the interactivity spec came from Adobe's work on trigger action lists for USDZ and Adobe Aero.
• There was debate between trigger-action lists, behavior graphs, and WASM approaches, with behavior graphs ultimately chosen.
• Security considerations were a major factor in the design, leading to a more constrained system than arbitrary JavaScript.
• The spec builds on the glTF object model and the Animation Pointer extension.
• Custom events allow GLTFs to send and receive messages, potentially enabling communication between nested GLTFs.
• The spec is designed to be flexible but with performance considerations in mind.
• Implementations are expected to limit execution time to maintain performance.
• There are ongoing efforts to finalize related extensions like audio and physics.
• Google is working on implementing the spec for use in Google Maps and other products.
• There are concerns about the timeline feeling rushed and the need for more example assets and supporting materials.
• Godot devs expressed interest in the spec but also raised concerns about other needed features, like consistent UUIDs for nodes across exports.
• There was discussion about the challenge of maintaining unique identifiers for nodes when optimizing or merging assets.
• The Blender team is exploring how to integrate the spec with their geometry nodes system.
• There are some concerns about the expansion of glTF's scope and potential performance implications.
• The importance of having multiple implementations before ratification was emphasized.
• The community was encouraged to contribute to the implementation efforts, particularly in projects like three.js.
• The spec is not intended to replace game engines but to enable interactivity for simpler use cases.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,2914 @@
...
+  }
+],
+"nodes": [
+  {
+    "type": "variable/set",
+    "configuration": [
+      {
+        "id": "variable",
+        "value": [0]
+      }
```
boguscoder, line 2882: are configuration values implicitly typed in JSON schema or they should have "type" key as well?

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,2914 @@
...
+
+===== Pointer Get
+
+[cols="1h,1,2"]
+|===
+| Type | `pointer/get` | Get an object model property value
+| Configuration
+| `string pointer` | The JSON pointer or JSON pointer template
+| Input value sockets
+| `int <segment>` | The JSON pointer template path segment to be substituted at runtime
```
boguscoder, line 2338: Since input value sockets MUST be connected, is `segment` optional socket(s)? or its omnipresent ? What if pointer is not templated, what shall `socket` be connected to, random const value (like 0) ?

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,2914 @@
...
+    "signature": "float"
+  }
+]
+```
+
+The signature value **MUST** be one of the value types defined in this extension specification or `"custom"`. In the latter case, the type semantic **MUST** be provided by an additional extension.
+
+Non-custom signature **MUST NOT** appear more than once.
+
+== Events
```
aaronfranke, line 2750: Can we get a high-level explanation of how this might work with a trigger volume as defined by `KHR_physics_rigid_bodies`? Am I reading correctly that the graph would have a script block (graph node) with type `event/receive` that points to an item in the document-level `"events"` array? And this event could have an ID of something like `"trigger/bodyEntered"`, which is defined by the extension to emit when a body enters the trigger volume?

What would the output value type of that custom event be? It would need to be something that allows us to reference which object did the entering. A glTF node index comes to mind, but then that wouldn't work for interacting between scenes. Probably the only option here would be a runtime-determined ID for each node in the scene (could just be the memory address), which allows for many glTF files to interact with each other. However, this would then mean that in order to use this value, the `KHR_interactivity` graph would need a script block (graph node) that can access an object using the same runtime-provided IDs, similar to how object model properties can be accessed using JSON pointers.

For example, let's say we want to build a trigger volume that teleports any object that enters it to another location. Assume that the trigger volume and the destination location are already known in the glTF file. How can the interactivity graph allow objects from other glTF scenes to be teleported, or even non-glTF objects provided by the game engine at runtime, including the player object itself? If a player walks into a glTF teleporter, I expect that the glTF should be able to detect that, and set the position of the player.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
aaronfranke, line 2338: What does it mean to have the segment as an input integer value? Using the below example pointer `"/nodes/{myId}/scale"`, what does 0 mean? Is it referring to `/nodes`, the 0th item, and therefore 1 would refer to the `{myId}`? Or is it referring to the first instance of `{}`, so if there was `"/nodes/{myId}/{myProp}"` then 0 refers to `{myId}` and 1 would refer to `{myProp}`? But then what does the number mean, is it saying that only one of these will be substituted? Or is it saying to substitute `0` for `{myId}`? But then what's the text below about "the node has the `myId` input value socket"? If we had `"/nodes/{myId}/{myProp}"` then would the node need both `myId` and `myProp` inputs? If so, then we can perform the substitution with only the template and the inputs, and I don't know why we would ever read the `int <segment>` value. I genuinely don't understand what this means.

boguscoder: I'm not the author but they way I read is the following:
if you have path template configured as `/nodes/{myId}/foo/{myProp}` then your node has two input sockets
`int myId` and `int myProp` respectively, so connecting node to values 42 and 100500 gives you `/nodes/42/foo/100500`. `int <segment>` is just not most readable way to convey idea of dynamic socket number, 
there is no 0s not 1s involved, curious how did you arrive at them?

aaronfranke: Oh! I completely missed that. So that is just a placeholder, and there is no input socket called segment. So for actual cases, there might be no segment input value sockets if it's not a template.

boguscoder: I think I replied to myself below, so #solved

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,2956 @@
...
+.. activate the `done` output flow linked to the current entry
+.. skip the next steps.
+4. Using the cubic Bézier spline information, compute the output progress position _q_ based on the _t_ value. This step implies that latexmath:[t \in [0; 1\]].
+5. Set the linked Object Model property to the new value computed as a linear interpolation between the original and the target property values using the output progress position _q_ as the interpolation coefficient.
+
+If the Object Model property is a quaternion, spherical linear interpolation expression **SHOULD** be used.
+
+Intermediate output progress values **MAY** be less than zero or greater than one.
+
+==== Animation Control Nodes
```
boguscoder, line 2486: can it be explicitly called out if animation from interactivity graph takes precedence over built-in animations if they are using same pointer or its up to implementation

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
Hackn0214: This is not a response to the above, but I have an additional question.
The Behave Graph Authoring tool allows the creation of a "node/OnSelect" node, but there is no definition of this node in the Interactivity Extension Specification.
Is this simply a documentation mistake, or is it related to KHR_node_selectability?

Hackn0214:
I apologize if this has already been mentioned somewhere.
I am currently checking the Interactivity Extension Specification.
How are p1 and p2 parameters used in node pointer/interpolate?

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
boguscoder: closing this one as done, this will be in the text soon

hybridherbst:
A couple of practical questions based on my experiments:

1. For looping an animation, is the expected approach to set endTime to positive infinity?

3. For preventing animations overlap when multiple animations target the same node, is the expected approach to just stop all animations that might be playing, or to add an internal state of which animation was last played and then stop that?
4. For interpolating rotations, how would the most-used quaternion interpolation – slerp – be implemented? I'm not sure a bezier curve (p1/p2) can be mapped to a spherical interpolation.

warrenm:
> 3. For interpolating rotations, how would the most-used quaternion interpolation – slerp – be implemented? I'm not sure a bezier curve (p1/p2) can be mapped to a spherical interpolation.

My understanding is that the control points specified by an interpolation node are to be used as the control points of a Bézier spline that is used as an _easing function_, _à la_ [cubic Bézier easing functions](https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function#cubic_bézier_easing_function) in CSS and elsewhere. 

If that's correct, then the implementation would pass the normalized time parameter (on [0, 1]) to the easing function to get the eased progress fraction, then use that as the "time" parameter of the lower-level interpolation routine (whether lerp or slerp).

hybridherbst:
Thanks! I found it now. [The spec](https://github.com/KhronosGroup/glTF/blob/interactivity/extensions/2.0/Khronos/KHR_interactivity/Specification.adoc#36423-pointer-interpolate) states that it's linear interpolation but also has this remark:
> If the Object Model property is a quaternion, spherical linear interpolation expression SHOULD be used.

Hackn0214:
Thank you for your response.
These parameters seem a bit challenging for designers to handle.

Hackn0214:
Regarding the question I asked earlier, it seems it might have gotten buried and become hard to notice, so I’d like to ask again.
I couldn't find node/OnSelect(event/OnSelect) in the specification, but is it possible to handle events when an object is selected?

lexaknyazev:
> I couldn't find node/OnSelect(event/OnSelect) in the specification, but is it possible to handle events when an object is selected?

This event is defined in the `KHR_node_selectability` extension, see #2422.

Hackn0214:
> > I couldn't find node/OnSelect(event/OnSelect) in the specification, but is it possible to handle events when an object is selected?
> 
> This event is defined in the `KHR_node_selectability` extension, see #2422.

Oh, I thought this extension was an attribute to specify whether an object can be selected. I’ll read through the draft. Thanks!

Hackn0214:
How would a connected graph with multiple output value sockets, as shown in the diagram, be represented in glTF?
gltfnode

Hackn0214:
It's not a self-solution, but is there any issue with this kind of definition?

      "nodes": [
        {
          "type": "math/sign",
          "values": [
            {
              "id": "a",
              "node": 2,
              "socket": "1"
            }
          ]
        },
        {
          "type": "math/sign",
          "values": [
            {
              "id": "a",
              "node": 2,
              "socket": "0"
            }
          ]
        },
        {
          "type": "math/extract2",
          "values": [
            {
              "id": "a",
              "type": 1,
              "value": [
                0.0,
                0.0
              ]
            }
          ]
        }
      ],

hybridherbst:
I believe this matches the output of https://github.com/KhronosGroup/glTF-InteractivityGraph-AuthoringTool/tree/initial-work-merge (minus the missing type declarations in your partial file).

Hackn0214:
Thank you.
I'm comparing the behavior of my custom tool with the authoring tool as a reference.
By the way, is there a way to delete a node once it's placed on the graph in this tool? Currently, I'm restarting the tool to clear the graph.

boguscoder:
@Hackn0214 backspace seems to remove just fine?

Hackn0214:
> @Hackn0214 backspace seems to remove just fine?
Ah... I was pressing the Delete key.

Hackn0214:
When defining a custom event as shown below, if event/receive and event/send nodes refer to this custom event.

     "events": [
        {
          "id": "MyEvent",
          "values": [
            {
              "id": "Val_1",
              "type": 0
            },
            {
              "id": "Val_2",
              "type": 0
            }
          ]
        }
      ],

I think the generated node will have sockets as shown below, but what do you think?
gltfnode

Hackn0214:
Is it possible to return the value pointed to by a pointer with the behavior graph?
For example, I would like to change the base color of a selected object's material.

/nodes/{nodeIdx}/mesh -> meshIdx
/meshes/{meshIdx}/primitibes/0/material -> materialIdx
/materials/{materialIdx}/pbrMetallicRoughness/baseColorFactor

boguscoder:
If configuration value is array of length 2,3,4 or 16 it could be ambiguous if its vec2/vec3/vec4/vec4x4 or int[], this might not be a problem for JS but is a problem for strongly-typed implementations, is there any recommendation for them how to disambiguate?

Hackn0214:
For example, how about specifying the type at the time of the call?
In fact, I would also like to have a feature that returns the number of elements as shown below.
/materials -> total material number in the scecne
/meshes/0/primitives -> Total primitive number

boguscoder:
@Hackn0214  you already can get number of materials when using appropriate pointer , check out [core  pointers](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/ObjectModel.adoc#4-core-pointers)

Hackn0214:
That's wonderful!
Thank you!

Hackn0214:
I would appreciate it if you could add /meshes/{}/primitives.length

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -2317,187 +2670,438 @@ This node has no internal state.
...
-- Interpolation duration value converted to the implementation-defined high precision time type
-- Object Model property value at the time of the successful node activation
-- Information needed for cubic Bézier spline evaluation derived from the node's input values
-- Target property value
+If the `value` input value or any intermediate interpolated value are not valid for the resolved property, the effective property value becomes implementation-defined and subsequent `pointer/get` evaluations of the property **MAY** return any value of the corresponding type until the property is updated with a valid value. This is not an error. Implementations **MAY** generate runtime warnings in this case as deemed possible.
+
+[NOTE]
+.Example
+====
+If the resolved glTF property is `"/materials/0/pbrMetallicRoughness/metallicFactor"`, its current value is zero, and the interpolation target value is two, the effective emissive factor becomes undefined when the interpolated value is greater than one.
```
hybridherbst, line 3040: Wrong word: effective ~~emissive~~ metallic factor

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -687,30 +860,59 @@ If any input value is _NaN_, the output value is false.
...
+
+The value of the output value socket `value` **MUST** be initialized to a random number on the first access. Any two accesses of the output value socket `value` **MUST** return the same value if there were no flow socket activations (of other nodes) between them.
+
+[NOTE]
+.Example
+====
+This means that, e.g., a `math/eq` node with both its input value sockets connected to the same `math/random` node always returns true.
+====
+
+The value of the output value socket `value` **SHOULD** be updated when accessed as a result of a new flow socket activation, including self-activations.
```
hybridherbst, line 895: Why is this a SHOULD and not a MUST? Feels weird that this is undefined behaviour. If the goal was to say "the new value can be the same as the old value" that could be added instead.

lexaknyazev: SHOULD is not undefined, its normative definition from RFC 2119 is:
> This word, or the adjective "RECOMMENDED", mean that there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course.

That said, I think it's safe to use MUST here.

hybridherbst: Thanks! I didn't mean to say that "should" is undefined, just that graph authors can't rely on the behaviour of `math/random` if it only "should" be updated in the case described here.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,4196 @@
...
+A node is executed when one of its input flow sockets is activated, when one of its output value sockets is accessed by another node, when an operation-specific event occurs. A node **MAY** repeatedly activate its own input flow sockets during the execution.
+
+Usually, the node execution includes evaluating its input value sockets (if any), processing its own logic, and activating any number (including zero) of output flow sockets.
+
+[[nodes-sockets]]
+=== Sockets
+
+There are four kinds of sockets.
+
+_Output value sockets_ represent data initialized by the node or produced during its execution. For example, it could be results of math operations or parts of the node's internal state. Accessing these sockets either triggers computing the return value on the fly by executing the node or returns a value based on the node's internal state. Exact behavior depends on the node's operation. As a general rule, output value sockets **MUST** retain their values until a node having any number of flow sockets is executed.
```
RaananW, line 235: The last sentence is a bit hard to understand. Would it be right to say "until a node with any number of flow sockets is executed." ?

lexaknyazev:
> Is there a (reasonably up-to-date) machine-processable representation of this? (I.e. something like a schema or a "repository" of node structure descriptions)?

Not yet, this document is the only normative source for now. We'll publish regular JSON schemas after confirming that the current early implementations are aligned with them.

That said, JSON schemas alone are not enough since most operations have their own predefined socket types and ids. Suggestions on the machine-readable node spec representations are welcome.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,4196 @@
...
+[[validation-node-object]]
+== Node Object Validation
+
+1. _Assert_ that the element of the `nodes` array is a JSON object ("`the node`").
+2. _Assert_ that "`the node`" object has the `declaration` property that is a JSON index.
+3. If the `declaration` property value is not less than the `declarations` graph array length,
+.. _reject the graph_.
+4. If "`the node`" object has the `configuration` property:
+.. _assert_ that the `configuration` property is a non-empty JSON object;
+.. _assert_ that each property of the `configuration` JSON object is a JSON array.
```
RaananW, line 4127: Each property of the configuration object should have a `value` property that is an array

lexaknyazev: Good catch! Fixed.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
lexaknyazev: Done.

javagl:
What is the timeline for this and its review (beyond "as quickly as possible")? I have some smaller "pending" review comments, but am not sure when I find the time to continue. Is the preferred way "one complete review", or "comments as soon as they come up"?

javagl (commented):
OK, some minor comments inlined (I hope that I can continue here soon)

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,4301 @@
...
+| Operation | `math/clamp` | Clamp operation
+.3+| Input value sockets
+| `floatN a` | Value to clamp
+| `floatN b` | First boundary
+| `floatN c` | Second boundary
+| Output value sockets
+| `floatN value` | latexmath:[\min(\max(a, \min(b, c)), \max(b, c))]
+|===
+
+This node relies on `math/min` and `math/max` nodes defined above.
```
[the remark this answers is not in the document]
lexaknyazev: It should say "is defined in terms of".

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,4301 @@
...
+| `float3 value` | Vector stem:[a] rotated around vector stem:[b] counter-clockwise by stem:[c]
+|===
+
+If the vector stem:[b] is not unit, rotation results **MAY** be undefined.
+
+===== Transform
+
+[cols="1h,1,2"]
+|===
+| Operation | `math/transform` | Vector transformation
```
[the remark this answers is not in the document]
lexaknyazev: > This will cause trouble and ambiguities.

Please elaborate, i.e., propose an example graph that would be ambiguous but still conforming to the spec.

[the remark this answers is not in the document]
javagl: There currently does not seem to be a reasonable way to _create_ a graph to begin with (except for a text editor)

The ambiguity mainly refers to the definitions and implementations of the nodes, and the authoring process, and maybe the validation. When there is a `math/transform` node, then it is not clear which type of node this is. 

While the validation section contains a statement

> (6.7, (7))
> If the set of input value sockets defined by the declaration referenced by the declaration property does not match the set of input value sockets defined by the values property with regard to socket ids and types, reject the graph.

the declaration does not carry sufficient type information. Using `math/transform` with a `float2` and a `matrix3x3` is not valid, but it is not made _explicit_. 

I know, this is one of the "back to the drawing board" comments, and it's unlikely to be addressed. So feel free to mark it as "resolved". (I'm looking forward to see how the authoring/tooling support will handle all this, eventually).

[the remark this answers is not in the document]
lexaknyazev: > the declaration does not carry sufficient type information

The TSG explicitly decided that such information for the core ops must be stored in implementations rather than assets.

So a declaration like:
```json
{ "op": "math/transform" }
```
is interpreted as any of
```
"math/transform", (float2 a, float2x2 b) -> (float2 value)
"math/transform", (float3 a, float3x3 b) -> (float3 value)
"math/transform", (float4 a, float4x4 b) -> (float4 value)
```

Thus a node that uses `float2` and `matrix3x3` input sockets is definitively invalid.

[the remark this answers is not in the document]
lexaknyazev: Only nodes that logically represent the same operation _and_ have identical socket ids (ignoring types) are "overloaded". So `math/add`, `math/transpose`, `math/transform`, etc. share the same names for all variants while `math/combine*`, `math/extract*`, `math/rotate*`, etc. do not.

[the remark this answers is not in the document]
lexaknyazev: These are valid concerns and we expect that authoring UI/UX would be human-oriented rather than JSON-oriented, thus accidentally creating an invalid `math/transform` node would not be possible.

On top of this, engines have to perform at least minimal graph validation (as described in the dedicated spec section) and thus they would be able to catch invalid nodes early and safely.

> In this specific case, one could see an analogy to _function signatures_.

Yes, and many programming languages support arbitrary overloads as well as templates, generics, variadic arguments, optional parameters, etc. In our case the "variety" of such overloads is strictly limited to socket types.

javagl:
To not further distract from the core of this pull request (which is _the specification itself_) with considerations about possible JSON-based representations of nodes, I just opened https://github.com/KhronosGroup/glTF/issues/2472 . This contains a few first thoughts and some experimental outputs of auto-generating JSON-based node descriptions from the AsciiDoc.

pfcDorn:
Hey, would be good when we have some more Matrix Nodes, which would help a lot witch TRS Matrices, like

- Compose TRS Matrix( Inputs: position, quaternion-rotation, scale,  Output: Matrix4x4)
- Decompose TRS Matrix (Input: Matrix4x4, Output: Position, Rotation, Scale)

This both would help a lot with the use of the core pointer "/nodes/{}/globalMatrix"

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,4554 @@
...
+
+[cols="1h,1,2"]
+|===
+| Operation | `animation/start` | Start playing an animation
+| Input flow sockets
+| `in` | The entry flow into this node
+.4+| Input value sockets
+| `int animation` | Animation index
+| `float startTime` | Start time
+| `float endTime` | End time
```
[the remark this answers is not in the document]
lexaknyazev: They are not normalized. You're right that pointer access is needed to play an animation to completion finite number of times if the animation length is not statically known.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,4554 @@
...
+3. If stem:[T] is not equal to zero, let stem:[s] be the current iteration number computed as follows:
+latexmath:[s=\begin{cases}
+            \left\lceil \dfrac{r-T}{T} \right\rceil & \text{if } r \gt 0 \\
+            \left\lfloor \dfrac{r}{T} \right\rfloor & \text{if } r \le 0 \\
+           \end{cases}]
+4. Now for each _requested input timestamp_ stem:[r], the corresponding _effective input timestamp_ is
+latexmath:[t=\begin{cases}
+            r - s * T & \text{if } T \ne 0 \\
+            0 & \text{if } T=0 \\
+           \end{cases}]
```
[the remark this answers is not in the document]
lexaknyazev: If $T=2$ and $r=1$, then:

$s=\left\lceil\dfrac{r-T}{T}\right\rceil=\left\lceil\dfrac{1-2}{2}\right\rceil=\left\lceil-0.5\right\rceil=0$

$t=r - s * T=1-0*2=1$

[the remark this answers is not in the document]
javagl: I think that the point is the 'ceiling' operator. The result will not be `s = -0.5`, but `s=0` in this case
`console.log(Math.ceil((1-2)/2)); // Prints 0`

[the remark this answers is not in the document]
javagl: OK, I shuld have refreshed the page. When I wrote this, the previous two comments had defnitely not been there ...

pfcDorn:
Hey, 
i already discuss this with @RaananW and with @mattmacf98, and we have all the same opinion that we need a official Log Node. This would help a lot with very technical test files for the Interactivity implemantation. 
I have prepared a Debug Log Node proposal for this:
**[Gist Spec Proposal](https://gist.github.com/pfcDorn/22bbd104aa8296608d4d1b279fd83f91)**

Such official Log would be very helpfull for engine implementations, especially with the more complex implementations like Matrix calculations. So we can create test files which are also be able to log math results for faster comparison and not only visual results (which are only fine for fast detecting if it's working as indented).

pfcDorn:
Would it not be better to rename all Matrix Math nodes with a "mat" prefix? For more consistent with "matCompose", "matMul" and "matDecompose"?

hybridherbst:
@lexaknyazev I think we discussed this a couple of times in the WG but I can't find clear language in the spec in regards to events forming loops. My understanding is that the intent of events is that they are _never_ executed immediately and are supposed to be executed in the next tick, to prevent endless loops from forming. Is that correct? How can I derive that from the current spec?

aaronfranke:
I know this is a minor complaint in the grand scheme of things, but I would like to bring up again that the name "node" is already taken in glTF for an object in the scene hierarchy that has a transform, and therefore I recommend we avoid using the name "node" again for something else in KHR_interactivity to avoid confusion. I know [Unity uses "node"](https://docs.unity3d.com/Packages/com.unity.visualscripting@1.8/manual/vs-nodes.html), but Unity's scene objects are "GameObject". My suggestion is "block" since this is a term in visual scripting languages.

Can we get a Khronos response on this? Anything from "yes, let's switch the name" to "no, and here is why".

lexaknyazev:
> While writing unit tests I noticed that variable/interpolate has no caveat for types that have no "intermediary" state.

From the operation description:
> If the referenced variable is integer or boolean, the node is invalid and the graph **MUST** be rejected.

pfcDorn:
What's the status with the quaternion math op specs :)

hybridherbst:
@lexaknyazev If you would be able to update this PR with the quaternion nodes that would be well appreciated! Thanks

robertdorn83:
hey @lexaknyazev, is there a plan for adding something like quatLookRotation? or do i missing a workaround with the new quat nodes?

jsorel:
Feedback from a GLTF user and, GLTF java binding developer: 

While I can understand the choice of graph structures to define the operations, unfortunately I believe it is a short/mid term solution : 
- because graphs gets unreadable when they get larger
- increasing boxes/operation types over the versions.
- bad integration in IDE

Has a more developer version been studied ? like LUA <ins>**with a dedicated function set**</ins> ?
lua has been in the game development industry for very long, and is also used in many tools and international standards because of it's easy learning curve and simple integration in existing tools.
Using the lua syntax would allow a direct interaction in IDE, many developers and game designer knowing it already.

I understand adding script support in the format is raising the bar a lot, but I believe it's still better then a cheap graph solution.

robertdorn83:
@lexaknyazev , was there any discussion in the past about some color nodes, like create a color with a HSV node and vice versa.  Can also be very practical for interactivity stuff :)

taye:
Thank you to everyone working on this proposal. I'm building an AR effects library and need a system for programming interactivity. While the [WASM](https://github.com/KhronosGroup/glTF/issues/2485) proposal is compelling, for my own project I'm more interested in using a format that's easier to inspect without running a debugger and that's generally more intuitive to people who aren't experienced programmers. Visual scripting systems keep appearing in video game and AR editors, and I'd prefer to follow a standard than to make yet another system from scratch.

I've implemented KHR_interactivity for glTF-Transform and I've made a partial runtime implementation for Three.js using [behave-graph](https://github.com/bhouston/behave-graph). I've also made a custom extension which depends on of interactivity for face landmark effects. You can find the [code here on Github](https://github.com/miru-media/miru/tree/main/packages/gltf-ar-effects). I'd like to develop my Three.js runtime further, but I'm not sure what are the current states of behave-graph and [glTF-InteractivityGraph-AuthoringTool](https://github.com/KhronosGroup/glTF-InteractivityGraph-AuthoringTool).

taye:
@robertdorn83 There's a [proposal for procedural textures](https://github.com/KhronosGroup/glTF/pull/2381), but it seems to be independent from interactivity graphs. It would be great to have only one graph-based scripting language spec. Perhaps it makes sense to split the scripting language specification from the platform interface specification.

javagl:
> There's a https://github.com/KhronosGroup/glTF/pull/2381, but it seems to be independent from interactivity graphs. It would be great to have only one graph-based scripting language spec. 

I do/did not have the details of the procedural texture specs on the radar. But if does _not_ use the same structures as this one, then this indeed raises a bunch of questions.

> Perhaps it makes sense to split the scripting language specification from the platform interface specification.

Due to the lack of familiarity with the other spec, I cannot say where the right place might be for this "split". But given that most of the spec _here_ is about the graph, its operations, and its execution (and only the last, tiny step is to ~"write some value into some property"), it sounds like there could be a _huge_ overlap between this spec and the procedural materials one. Fragmentation on the level of the graph implementation could increase the implementation effort tremendously. 

Note: I cannot say whether this is really technically feasible. But **iff** we could define a graph-based system that was powerful and versatile enough to describe interactivity **and** procedural materials, then it would be very likely also be powerful enough to handle _any_ possible future extensions (like procedural geometry or whatnot...)

lexaknyazev:
> there could be a _huge_ overlap between this spec and the procedural materials one

Procedural textures extension has different interoperability goals and constraints than this extension. Although it may sound convenient to unify them, there are many subtle operational discrepancies that could prevent that.

emackey:
> > there could be a _huge_ overlap between this spec and the procedural materials one
> 
> Procedural textures extension has different interoperability goals and constraints than this extension. Although it may sound convenient to unify them, there are many subtle operational discrepancies that could prevent that.

Agreed.  Material graphs and behavioral graphs are different things.  A material graph does not use control flow nodes or local variables, for example.  A material graph gets inputs from the vertex attributes and textures of the actual mesh being rendered, and is evaluated at render time for that mesh, not during a physics/interactivity/heartbeat cycle between animation frames.

Igor-dvr:
Is there interactivity Blender extension? 
Any plans to implement the extension?

robertdorn83:
@lexaknyazev I have some question about the animation/play node. 
For a looping playback, "start" would be 0, and "end" would be positive infiniy.  right?
But what if a animation should played reversed? Especially with looping. 
Should it be "start" = animation.length and "end"= negative infinity? 
But then, the spec says : 
"If the startTime or endTime input values are NaN **or if the startTime input value is infinite**, **activate the err output flow** and skip the next steps."

lexaknyazev:
> But what if a animation should played reversed? Especially with looping.

To get infinite looping in reverse direction, set end to negative infinity and start to 0 (or any other finite value).

robertdorn83:
@lexaknyazev 
image

the term "index" for the input is not specific enough here. Which variable index? From the config array? Or the graph variable index?
Let's say whe have an config here with [3,6,9]. Would be the input names  "3", "6", "9"  **OR**  "0", "1", "2" ?

lexaknyazev:
@robertdorn83 The operation details located after the table make it quite clear that input value socket ids correspond to graph variable indices, that is 3, 6, and 9 in your example. We'll update the table for clarity.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
8vama: @aaronfranke The rationales I see here are (1) string is not a supported data socket type in the Interactivity extension. (2) Accessing by index also enables bulk updates by iterating a range of variables. (3) It feels consistent with the glTF conventions and avoids duplicating the variable names in json. 

The authoring tool can always allow users to query variables by name, and convert to index under the hood.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
lexaknyazev: Additional interactivity extensions like `KHR_node_selectability` or `KHR_node_hoverability` can define their own operations and events.

Cross-asset object linking is out of scope of this extension for now but that topic will be explored in future.

robertdorn83:
@lexaknyazev 
Currently the readonly pointer for activeCamera contains only position and rotation. Would it make also sense to add matrix to it ? I know we can build a matrix out of position and rotation because its global space, but just for consistency with the node transforms.

Also, and we already talked about partly when i remember correctly, but it would be also more consistent to add the other gltf camera properties such as activeCamera.orthographic.xmag, activeCamera.orthographic.zfar,  activeCamera.perspective.aspectRatio etc. These are all known standard gltf properties.
Also maybe adding activeCamera.isOrtho, activeCamera.viewMatrix, activeCamera.projectionMatrix.
My wishlist for christmas this year ;)

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -0,0 +1,5222 @@
...
+.4+| Output value sockets
+d| `float3 translation` | Translation vector
+d| `float4 rotation` | Rotation quaternion
+d| `float3 scale` | Scale vector
+d| `bool isValid` | True if the input matrix is decomposable; false otherwise
+|===
+
+The output values are computed as follows:
+
+1. If the first three elements of the fourth row of stem:[A] are not zeros or if the last element of the fourth row of stem:[A] is not close to positive one within an implementation-defined threshold, set the `isValid` output value to false and goto to the step 11.
```
Jmangles: "goto to the step 11." Reads a bit odd. This step also instructs you to set isValid to false while Step 11 also instructs you to do the same

lexaknyazev: The step 11 checks if `isValid` is false; it doesn't set it.

lexaknyazev: The step 11 checks if `isValid` is false; it doesn't set it.

robertdorn83:
@lexaknyazev 
hey alexey, why have you added a "ref event" output on the onStart and onTick event? I have nothing in mind why we could need a reference to  the event here. In Unity - and im sure in a lot of other engines too - such events don't even have a Event-Object to pass through.

robertdorn83:
@lexaknyazev  Some questions about the ref types:
- I wasn't able to find some information how to resolve a Ref (let's say a Node) to the graph node index. (can also be usefull to check if a ref belongs to the own gltf file) 
would it be  with
 **pointer/get (template: `/node/{nodeRef}/index`, type: int)**  
or
  **pointer/get (template: `/node/{nodeRef}/`, type: int)** ?

- what when i want to get a mesh from a node: is `/node/ []{} /mesh` returning a index or ref now? Or depends it now how i setup the type in the node config? so `type=Ref` would give me the ref to the mesh, `type=Int` would give me the index? But the index would only work/is usefull, when the mesh belongs to the same gltf file as the graph.  

- initial Values for Variables from type Ref or Input Sockets from type Ref:  so we use here as a value just a string with `/node/6/`? 

These pointer access cases and ref initial values seems to bemissing in the specs currently .. ?

lexaknyazev:
> how to resolve a Ref (let's say a Node) to the graph node index

That's impossible. Only the opposite direction is currently supported.

> what when i want to get a mesh from a node: is `/node/ []{} /mesh`

All such pointers return refs, the Object Model is going to be updated.

> initial Values for Variables from type Ref or Input Sockets from type Ref

Their syntax and semantics are defined in the Variables section (currently 5.3). When not set, the value is null; otherwise it must be a static (no parameters) JSON pointer string that will be resolved to a runtime reference value (if resolvable) or to null (if unresolvable) when the graph is loaded.

robertdorn83:
I updated the Samples and Tests with the new Ref spec changes. They are currently in a separate branch: [gltf Interactivity Assets](https://github.com/KhronosGroup/glTF-Test-Assets-Interactivity/tree/ref-type-spec-changes)
Was e bit more work than expected, so sorry for the delay :)

robertdorn83:
this onSelect part requires an update  > index to ref changes

image

robertdorn83:
@lexaknyazev 

can you add some pointer samples to get Ref-Objects via json pointers?   like  pointer/get > /animations/0, to get the animation reference.  

and: is "/animations/0" the correct format to get the reference from animation-index 0? or would it be "/animations/0/"?

emackey:
Can we merge or rebase main here?

lexaknyazev:
> Can we merge or rebase main here?

That will happen automatically because there are no merge conflicts.

emackey:
> That will happen automatically because there are no merge conflicts.

Sure, but I'm trying to see what will happen to the CI itself, there are several CI workflow diffs here.

lexaknyazev:
> Sure, but I'm trying to see what will happen to the CI itself, there are several CI workflow diffs here.

The workflows run successfully and produce correct outputs, see the Checks tab.

emackey (approved):
I checked the CI outputs, and this appears safe to merge.  I have not reviewed the Interactivity spec itself in fine-grain detail, but many others here have been doing that for many months.  A discussion earlier today revealed no objections to merging, with the understanding that further refinements are likely as a wider audience begins to embrace this.

PR #2304, opened 2023-07-24
Mark several recently ratified extensions as complete, update registry

emackey:
The following extensions have been ratified this year:

- KHR_materials_anisotropy
- EXT_mesh_gpu_instancing
- EXT_meshopt_compression
- EXT_texture_webp

Additionally, some wording was adjusted to allow EXT extensions to become ratified, and to clarify when _required_ status should be marked on an extension.

(Also, some links to glTF 1.0 extensions removed. Anyone who needs those should click on the relevant folder, no need to advertise them anymore).

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -15,6 +15,7 @@ The following extensions have been ratified by the Khronos Group:
 
 * [KHR_draco_mesh_compression](2.0/Khronos/KHR_draco_mesh_compression/README.md)
 * [KHR_lights_punctual](2.0/Khronos/KHR_lights_punctual/README.md)
+* [KHR_materials_anisotropy](2.0/Khronos/KHR_materials_anisotropy/README.md)
 * [KHR_materials_clearcoat](2.0/Khronos/KHR_materials_clearcoat/README.md)
 * [KHR_materials_emissive_strength](2.0/Khronos/KHR_materials_emissive_strength/README.md)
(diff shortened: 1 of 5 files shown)


extensions/README.md
```diff
@@ -97,22 +98,9 @@ extensions, Khronos extensions, or inclusion in a future version of the glTF spe
...
-### Khronos extensions for glTF 1.0
-
-* [KHR_binary_glTF](1.0/Khronos/KHR_binary_glTF/README.md)
-* [KHR_materials_common](1.0/Khronos/KHR_materials_common/README.md)
-
-### Vendor extensions for glTF 1.0
-
-* [CESIUM_RTC](1.0/Vendor/CESIUM_RTC/README.md)
-* [WEB3D_quantized_attributes](1.0/Vendor/WEB3D_quantized_attributes/README.md)
-
```
lexaknyazev, line 115: Why was this section removed? This readme file technically applies to both 1.0 and 2.0 subfolders.

emackey: I think it's time to stop calling attention to the 1.0 folders.  The folders still exist, so if any legacy users still need access, they can look in those folders as needed.  But I'm certain that Khronos no longer needs to call them out in a registry.

lexaknyazev (approved):

bghgary (approved):
LGTM

PR #2313, opened 2023-08-23
Fix schema file names of Khronos extensions so they match their scope

aaronfranke:
Most of the schema files in this repo have their names matching the scope they reside in. However, a few files were not. The `KHR_materials_*` extensions were not following this naming convention. This PR fixes that. This PR also fixes `KHR_texture_transform` which had the scope after the extension name instead of before like most schemas.

The convention that most of this repo follows is that `glTF.<ext>.schema.json` files extend the base document, while `<scope>.<ext>.schema.json` files extend that scope. For example, `material.<ext>.schema.json` extends `material.schema.json`. I like this convention, but regardless I am not here to question this, just make it consistent. In this PR, I only touched the Khronos extensions in the 2.0 folder.

For any websites these files are being hosted on as-is, redirects may be required.


aaronfranke:
Thanks, good catch. I also noticed `KHR_materials_sheen` and `KHR_materials_unlit` were missing links to the schema files. I have updated the PR to add links to the README files of those extensions.

lexaknyazev (approved):

PR #2314, opened 2023-08-23
Use friendly unique names for all Khronos extension schemas

aaronfranke:
This PR improves the schema titles. This is mostly cosmetic, but it may have practical benefits, such as improved human and machine parsability, and it gives us the ability to uniquely identify a schema file from its title alone.

When making this PR, I used the following rules:

* All titles should be in Title Case where words start with a capital letter except for stylized `glTF`.
* All extension schemas should include the extension name at the start. For example, all `KHR_lights_punctual` schemas should start the title with "KHR_lights_punctual".
    * Previously the light properties schema was just titled `"light"`. This is not unique enough to be helpful without the context of knowing you are dealing with `KHR_lights_punctual`.
* Schemas that extend a specific part of glTF should specify that part in the format "glTF \<something\> Extension".
    * If the extension is applied on the base document, use the text "glTF Document Extension".
    * Note that `KHR_xmp_json_ld.schema.json` can be applied anywhere, so it does not apply to a specific part, so I kept it as just "glTF Extension".

In this PR, I only touched the Khronos extensions in the 2.0 folder.

### DIFF
--- a/extensions/2.0/Khronos/KHR_draco_mesh_compression/schema/mesh.primitive.KHR_draco_mesh_compression.schema.json
+++ b/extensions/2.0/Khronos/KHR_draco_mesh_compression/schema/mesh.primitive.KHR_draco_mesh_compression.schema.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-04/schema",
-    "title": "KHR_draco_mesh_compression extension",
+    "title": "KHR_draco_mesh_compression glTF Mesh Primitive Extension",
     "type": "object",
     "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
     "properties": {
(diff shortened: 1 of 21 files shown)


javagl:
While I agree to the _goals_ of this improvement, one has to be careful: If I remember correctly, [wetzel](https://github.com/CesiumGS/wetzel) (the tool that is used for generating the schema reference) uses the `title` property in a ... very special way, namely as a sort of "identifier" for the types that are defined. 

This is just a hint for now: I don't know how actively wetzel is used for generating the MD of _extensions_. I think it should at least be _possible_ to do this, but on top of that, I don't know whether the changes here might cause wetzel to choke on the schema. 

@emackey Maybe you know from the tip of your head whether these changes could be "critical" or cause wetzel to fail for things for which it previously worked? 
(Otherwise, I could give it a try with the _old_ and _new_ state, but ... would have to allocate some time for that...)

emackey (approved):
Thanks for the ping @javagl.  I'm fine with the changes here.  Currently we are not running wetzel on any extension schemas, only core, so checking wetzel compatibility is out-of-scope for this PR.  It looks like the changes here add consistency and clarification.  LGTM.

lexaknyazev (approved):
LGTM. Would need to be rebased on #2313.

emackey:
> Would need to be rebased on #2313.

Looks like #2313 needs more work.  Let's merge this one first.

PR #2315, opened 2023-08-24
Fix inconsistent list formatting in Khronos extension README.md files

aaronfranke:
In the current repo, the extensions have inconsistent list formatting. This PR fixes that.

Most of the files use *, but about a quarter of the list items use -. I've changed all of them to *. If instead - is desired, let me know. I personally prefer * but I'm not here to push for one or the other, I just want to make it consistent.

This PR also fixes `KHR_materials_anisotropy` having an ordered list instead of unordered, and fixes `KHR_materials_clearcoat` and `KHR_materials_transmission` not having the references formatted as a list.

Also, some trailing space characters were removed. These do nothing (note: they *were* doing something in the references, but now that they are using list formatting, the trailing spaces are no longer needed).

In this PR, I only changed the Khronos extensions in the 2.0 folder.


javagl:
That also bugs me (not only in this repo, but _everywhere_ (where MD is used). I subjectively lean towards `-`, because the `*` is already "overloaded" a bit, via its use in `*italic*`, `**bold**`, and `***italic and bold***` (!) in MD, but don't have a strong preference  - let's see what the main spec maintainers say.

(From a quick look at the changes, you also seem to have changed the _numbered_ lists (`1.`, `2.`...) to _unnumbered_ lists (`*`). This is a change that affects the output, and may therefore be seen with more scrutiny, but again, no strong opinion on my side)

aaronfranke:
> (From a quick look at the changes, you also seem to have changed the numbered lists

The only file where I did this is `KHR_materials_anisotropy`, because that makes more sense as an unordered list, and is consistent with the references in other files.

emackey:
> I'm happy to go with either * or - depending on what the maintainers prefer.

I think `-` might be official (for Markdown, which itself is not particularly official I guess).  Also some of the automatic TOC generators (that we haven't run since the previous decade) also output `-` for the TOC.  Let's have all the lists use minus signs.

> > (From a quick look at the changes, you also seem to have changed the numbered lists
> 
> The only file where I did this is `KHR_materials_anisotropy`, because that makes more sense as an unordered list, and is consistent with the references in other files.

As long as the reference isn't mentioned by number, I'm fine with that.

emackey (approved):
Thanks @aaronfranke.

lexaknyazev (approved):

PR #2320, opened 2023-09-06
Update overview image to 2.0.0d

javagl:
The overview image had been version 2.0.0b, while the actual overview had been updated to 2.0.0d in the meantime. (The change only affected the list of extensions on the last page).

**NOTE**: Updating the binary files may bloat the repository a bit (with 5.5MB each time). I considered to just change the links to point to the releases at https://github.com/javagl/gltfOverview/releases . This could be more economical in the long run, and equally "stable" (considering that they are hosted on GitHub anyhow). The only caveat would be that the main repo README would then no longer be "standalone" (i.e. it would directly inline an image that is _not_ contained in the repo itself). 

Does someone (maybe @lexaknyazev ) have a preference for one solution or the other?


lexaknyazev:
As a general rule, OK if the linked image is in another repo of the same GitHub organization, e.g., in `KhronosGroup/glTF-Tutorials`.

In this case, I'd suggest just running the new images through some PNG optimizer(s). It may be possible to considerably reduce file sizes with no visual difference.

javagl:
@lexaknyazev I wasn't aware that PNG could be compressed _so_ considerably (Probably because it is mainly text and few colors). There is some dithering visible in the 'Duck', but the file size was reduced from 4.5MB to 1.3MB (and the 'small' version from 1.3MB to 290KB), and ... that's worth it, I guess.

And at the risk at embarrassing myself: I _think_ that if I had updated _this_ PR, the large file would still appear in the history after merging - maybe depending on whether squashing is used or not? To be safe, I opened https://github.com/KhronosGroup/glTF/pull/2325 as a replacement for this.

PR #2324, opened 2023-09-11
Draft of the glTF Asset Object Model

lexaknyazev:
This is a prerequisite for the animation pointer and interactivity extensions.

### DIFF
--- a/specification/2.0/ObjectModel.adoc
+++ b/specification/2.0/ObjectModel.adoc
@@ -1,0 +1,442 @@
+// Copyright 2023 The Khronos Group Inc.
+//
+// SPDX-License-Identifier: CC-BY-4.0
+
+// :regtitle: is explained in
+// https://discuss.asciidoctor.org/How-to-add-markup-to-author-information-in-document-title-td6488.html
+= glTF{tmtitle} 2.0 Asset Object Model Specification
+:tmtitle: pass:q,r[^™^]
+:regtitle: pass:q,r[^®^]
+The Khronos{regtitle} 3D Formats Working Group
+:data-uri:
+:icons: font
+:toc2:
+:toclevels: 10
+:sectnumlevels: 10
+:max-width: 100%
+:numbered:
+:source-highlighter: coderay
+:title-logo-image: image:../figures/glTF_RGB_June16.svg[Logo,pdfwidth=4in,align=right]
+:docinfo: shared-head
+:stem:
+
+// This causes cross references to chapters, sections, and tables to be
+// rendered as "Section A.B" (for example) rather than rendering the reference
+// as the text of the section title.  It also enables cross references to
+// [source] blocks as "Listing N", but only if the [source] block has a title.
+:xrefstyle: short
+:listing-caption: Listing
+
+ifndef::revdate[]
+:toc-placement!:
+
+[NOTE]
+.Note
+====
+Khronos posts the AsciiDoc source of the glTF specification to enable community
+feedback and remixing under CC-BY 4.0. Published versions of the Specification
+are located in the https://www.khronos.org/registry/glTF[glTF Registry].
+====
+endif::[]
+
+// Table of contents is inserted here
+toc::[]
+
+:leveloffset: 1
+
+[[foreword]]
+= Foreword
+
+Copyright 2023 The Khronos Group Inc.
+
+This Document is protected by copyright laws and contains material proprietary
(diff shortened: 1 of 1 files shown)


javagl (commented):
Looks fine for me so far.

(...with the usual disclaimers: I'm not up to date with the latest states of the animation pointer and interactivity efforts. Having such a common, clearly defined object model as the basis for both is certainly a resonable approach. It's hard to foresee the implications that this may have for viewers that want to support it. But it looks like something that could be implemented generically in many cases)

In terms of the actual document, the only thing that ... surprised me a bit was the _explicit_ listing of the "Extension Pointers" in section 5. It looks pretty repetitive. When there is a new extension that uses a "texture", then all the properties will have to be added there. 

I think it could/should be possible to describe the effect of `KHR_texture_transform` in a more generic way - namely, that it exposes the `scale` in each `NormalTextureInfo`, and the `offset/rotation/scale` in each `TextureInfo`.

(Are there exceptions to that?)

So while I see the point of having the whole list _explicit_ in the document, but wondered whether this could be summarized as something like

`/materials/{}/**/<anyTextureInfo>/extensions/KHR_texture_transform/offset`

(Maybe in addition to the explicit lists, in a dedicated section about the `KHR_texture_transform` or so...)

lexaknyazev:
> Having such a common, clearly defined object model as the basis for both is certainly a reasonable approach. It's hard to foresee the implications that this may have for viewers that want to support it.

This document is supposed to only standardize pointer syntax and resolution. It explicitly omits many details that need to be specialized for different use cases.

> I think it could/should be possible to describe the effect of `KHR_texture_transform` in a more generic way - namely, that it exposes the `scale` in each `NormalTextureInfo`, and the `offset/rotation/scale` in each `TextureInfo`.

Minor correction: `scale` is a core part of `NormalTextureInfo` and it's a different parameter than texture transform's `scale`.

> It looks pretty repetitive.

It's intended (at least for now) so that the combinatorial effect of the `KHR_texture_transform` extension is more obvious to implementors. Although the current draft has been written manually, it would be reasonable to automate generation of pointer lists for the future.

elalish (approved):
Thanks!

dwrodger (approved):
Reviewed in the Interactivity DTSG

specification/2.0/ObjectModel.adoc
```diff
@@ -0,0 +1,497 @@
...
+
+
+[[motivation]]
+== Motivation and Design Goals (Informative)
+
+Although <<gltf,glTF 2.0 Specification>> does not define any runtime behavior, there is a growing demand for manipulating glTF assets at runtime and querying their state in a portable way.
+
+Upon loading, applications usually parse glTF JSON into implementation-specific internal structures and therefore do not have a common protocol to manipulate asset contents.
+
+This document defines a set of <<json-pointer,JSON pointers>> that glTF implementations are expected to support for identifying asset properties that could be manipulated at runtime.
```
rsahlin, line 115: I think we must be careful of the wording here.
`glTF JSON` is a JSON document (RFC4627) that contains the glTF model data.
When applications load this data we have no way of knowing what format the data is in, it could be fully converted into typed arrays and objects in an application that has no knowledge of JSON.

I would suggest using a wording that ties the object model to a runtime (format agnostic) behavior where the important aspect is to be able to locate the data properties declared in the glTF 2.0 spec in Chapter 5 (Properties Reference)

As the spec reads now it is assumed that the pointers are resolved from the JSON Document since this is how `JSON Pointers` work (see RFC 6901)

Imagine a usecase where a glTF model is loaded into a solution that uses some other fileformat (MaterialX, USD) by converting the glTF to it's fileformat. 
The JSON document is not present but the implementation knows where to find the data, for instance declared by glTF property `glTF.nodes.rotation`

I just want to make sure that the wording is not tightly coupled to the JSON fileformat but rather the glTF properties.

lexaknyazev: > When applications load this data we have no way of knowing what format the data is in, it could be fully converted into typed arrays and objects in an application that has no knowledge of JSON.

An app would have to keep some sort of mapping between original glTF JSON properties and their actual runtime representation anyway.

> Imagine a usecase where a glTF model is loaded into a solution that uses some other fileformat (MaterialX, USD) by converting the glTF to it's fileformat.

This would be an implementation detail out of the spec's scope. If an app converts incoming glTF data into something else, it should be able to convert pointers as well.

Feel free to open a PR adjusting the language.

Issue #2326, opened 2023-09-15
EXT_meshopt_compression dummy buffer byteLength can be any size


andreasplesch:
EXT_meshopt_compression decompresses bufferViews of a compressed buffer. It also requires a reference to a corresponding uncompressed fallback buffer which can be an empty dummy but has to have a byteLength.

Currently, the [extension spec.](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_meshopt_compression/README.md#overview) does not seem to place any requirements on the byteLength of a dummy buffer. Most generators of compressed gltf such as gltfpack probably assign the total uncompressed byteLength of the buffer after decompression of all bufferViews which reference this dummy buffer. This is very helpful to implementations which may want to decompress a compressed bufferView into an arrayBuffer of the provided byteLength.

But such implementations cannot currently rely on the provided byteLength to be compatible with the uncompressed length of all compressed bufferViews since the byteLength is allowed to be any size.

A spec. sentence such as

> The byteLength property of a fallback buffer _**MUST**_ be sufficiently large to contain all uncompressed bufferViews referencing it.

in the [fallback buffer section](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_meshopt_compression/README.md#fallback-buffers) may address this concern. Since gltfpack likely already behaves in this way and gltfpack is the only current generator (not anymore, see below), full backward compatibility seems to be guaranteed.

Let me add here as well that the properties of the parent bufferView of the extension object would also be expected to be consistent with the uncompressed data but does not seem to be actually required to be so, eg. cannot be relied on to view a dummy buffer populated with uncompressed data.

donmccurdy:
I'm fine with this clarification, but we shouldn't assume no other implementations exist. glTF Transform does the same:

https://github.com/donmccurdy/glTF-Transform/blob/41d1e13eff32789418f67f4b3d40646cf52b7fe4/packages/extensions/src/ext-meshopt-compression/meshopt-compression.ts#L386-L421

andreasplesch:
Thanks, indeed, let's not assume that, and it is great that glTF Transform does what for now is perhaps implied but not required. So adding the clarification would be more important in case other generators come along.

[...]

PR #2327, opened 2023-09-15
clarify dummy buffer length

andreasplesch:
see #2326

### DIFF
--- a/extensions/2.0/Vendor/EXT_meshopt_compression/README.md
+++ b/extensions/2.0/Vendor/EXT_meshopt_compression/README.md
@@ -72,7 +72,7 @@ Each `bufferView` can contain an extension object with the following properties:
 For the extension object to be valid, the following must hold:
 
 - When parent `bufferView` has `byteStride` defined, it matches `byteStride` in the extension JSON
-- Buffer view length is equal to `byteStride` times `count`
+- The parent `bufferView.byteLength` is equal to `byteStride` times `count`
 - When `mode` is `"ATTRIBUTES"`, `byteStride` must be divisible by 4 and must be <= 256.
 - When `mode` is `"TRIANGLES"`, `count` must be divisible by 3
 - When `mode` is `"TRIANGLES"` or `"INDICES"`, `byteStride` must be equal to 2 or 4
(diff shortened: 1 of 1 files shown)


extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -118,6 +120,8 @@ While the extension JSON specifies a separate buffer to source compressed data f
 { "byteLength": 1432878 }
 ```
 
+The `byteLength` property of such a dummy buffer **MUST** be sufficiently large to contain all uncompressed bufferViews referencing it.
```
lexaknyazev, line 123: Please replace "dummy" with "placeholder" here and in the next paragraph.

lexaknyazev (changes requested):

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
andreasplesch: Thanks for quickly looking at this. While this wording was the original choice of the extension authors (likely based on earlier discussions), let me make this change for better readability.

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
```diff
@@ -118,7 +120,9 @@ While the extension JSON specifies a separate buffer to source compressed data f
 { "byteLength": 1432878 }
 ```
 
-When stored in a GLB file, the dummy buffer should have index 1 or above, to avoid conflicts with GLB binary buffer.
+The `byteLength` property of such a placeholder buffer **MUST** be sufficiently large to contain all uncompressed bufferViews referencing it.
```
lexaknyazev, line 123: nit: please use "buffer views" instead of "bufferViews" because the latter is not a valid symbol reference here (there are other examples of inconsistent symbol usage in the extension text but let's keep this PR focused on one issue).

andreasplesch: Let me do that ( symbol reference in a prose ? ).

lexaknyazev (approved):

extensions/2.0/Vendor/EXT_meshopt_compression/README.md
andreasplesch: Done. I think I understand , bufferViews is a formal property but we are referring here to members of the array, not the array.

Issue #1542, opened 2019-01-24
Consider removing support for multiple scenes in a single asset


donmccurdy:
Although GLTF technically allows multiple scenes, it isn't a common intention, and frankly I've never seen it used that way. In every known case, the file will contain a single scene — or perhaps none at all, e.g. for a material library — and loaders can consider that scene object a "top-level node" for application use.

The fact that glTF allows more than one scene seems to confuse or surprise users, who generally think of "scene formats" and "model formats", and multi-scene formats aren't a meaningful concept. Unless there's a strong need for the feature, I think we could consider disallowing use of multiple scenes in some future version of the spec, and/or discouraging it now.

lexaknyazev:
Keep in mind that we have IBL extension that is defined on `scene`.

stevenvergenz:
For our use case, we treat a scene loaded from a glTF as equivalent to a Unity prefab. So being able to pack multiple variations of a prefab that use the same assets is useful. Admittedly, multi-scene export isn't common, but that might change.

vpenades:
Actually, my intention was to use multiple scenes support as a cheap way of having LOD levels, where each scene is a LOD

donmccurdy:
> Actually, my intention was to use multiple scenes support as a cheap way of having LOD levels, where each scene is a LOD...

I'd be fully supportive of having an official LOD extension at some point. Microsoft uses one internally, and there's some discussion of bringing that in: https://github.com/KhronosGroup/glTF/issues/1045.

zellski:
I’m on the fence about this. It’s worthwhile to reduce complexity, and it’s meaningful to look at what people are actually doing — I agree in the vast majority of cases, the concept of “scenes” is more bewildering than helpful.

And yet the concept of being able to embed distinct scene graph roots in one GLB that can take advantage of glTF’s powerful reference mechanic to minimise duplication is really powerful — especially, as we’ve seen here, in conjunction with other extensions.

Some use cases, like LOD, can almost certainly work better at node or mesh levels. But nodes are extremely general (and already bound up in complexity such as skeletons). A simple extension that is only meaningful at the root is forced to target nodes, because we’ve removed scenes, would need to somewhat awkwardly communicate this constraints in English in the extension specification, and the constraint can’t be expressed in a schema.

[...]

PR #2329, opened 2023-09-18
GODOT_single_root

aaronfranke:
Preview: https://github.com/aaronfranke/glTF/tree/GODOT_single_root/extensions/2.0/Vendor/GODOT_single_root

Many games engines import glTF scenes as objects with a single root node. Unity has prefabs with a single root GameObject, and Godot has scenes with a single root node. Since glTF allows defining multiple root nodes, engines will insert the glTF root nodes as children of the "real" root node, which makes it difficult for a glTF file to define the properties of the "real" root node. This is important for things like physics bodies and character controllers, which are defined on the root node, so that all child nodes are moved with the body. Aside from physics, a single root node would avoid an extra node in the tree when round-tripping between glTF and these engines.

The `GODOT_single_root` glTF extension solves this problem by imposing additional constraints on the glTF scene to ensure it can be parsed into one of these objects with a single root node. Implementations can detect `GODOT_single_root` and import the single root node as the object's "real" root node in the scene/prefab/etc. The extension contains no data, it only restricts behavior and gives a hint to importers.

If there is interest from other vendors, this could be made into a multi-vendor extension. The restrictions imposed here could also be added to the base glTF spec in a future version (could be, say, "2.1" since this can be done without breaking compatibility as long as we require the scene/scenes boilerplate).

EDIT: Also note, the restrictions imposed here are a superset of https://github.com/KhronosGroup/glTF/issues/1542.


extensions/2.0/Vendor/GODOT_single_root/GODOT_single_root.md
```diff
@@ -0,0 +1,83 @@
...
+
+The `GODOT_single_root` glTF extension solves this problem by imposing additional constraints on the glTF scene to ensure it can be parsed into one of these objects with a single root node. Implementations can detect `GODOT_single_root` and import the single root node as the object's "real" root node in the scene/prefab/etc. The extension contains no data, it only restricts behavior. The restrictions are as follows:
+
+* The document `"scenes"` array MUST have exactly one scene.
+* The document `"scene"` index MUST be set to 0, the index of the first scene in the `"scenes"` array.
```
lexaknyazev, line 24: These restrictions contradict the language above:
> import glTF scenes as objects with a single root node

and

> constraints on the glTF scene to ensure it can be parsed into one of these objects with a single root node

Either the overview needs to be adjusted to provide rationale for disallowing multiple scenes, which is orthogonal to root nodes, or these two restrictions should be lifted.

extensions/2.0/Vendor/GODOT_single_root/GODOT_single_root.md
```diff
@@ -0,0 +1,83 @@
...
+* The scene MUST have exactly one node, the single root node.
+* The scene's node index MUST be set to 0, the index of the first node in the `"nodes"` array.
```
lexaknyazev, line 26: Order of elements in glTF object arrays has no specific meaning (except a GLB buffer reference for historical reasons). Extensions should not change that unless there's a very strong reason to.

extensions/2.0/Vendor/GODOT_single_root/GODOT_single_root.md
```diff
@@ -0,0 +1,83 @@
...
+* The single root node MUST NOT have `"translation"`, `"rotation"`, `"scale"`, or `"matrix"` properties.
```
lexaknyazev, line 27: These properties have default values so they may be always defined in some environments. Consider changing to:
> The ... properties of the single root node MUST either be undefined or have default values.

extensions/2.0/Vendor/GODOT_single_root/GODOT_single_root.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension imposes constraints on glTF scenes to ensure they have a single root node.
+
+Many games engines import glTF scenes as objects with a single root node. Unity has prefabs with a single root GameObject, and Godot has scenes with a single root node. Since glTF allows defining multiple root nodes, engines will insert the glTF root nodes as children of the "real" root node, which makes it difficult for a glTF file to define the properties of the "real" root node. This is important for things like physics bodies and character controllers, which are defined on the root node, so that all child nodes are moved with the body. Aside from physics, a single root node would avoid an extra node in the tree when round-tripping between glTF and these engines.
```
lexaknyazev, line 19: > which makes it difficult for a glTF file to define the properties of the "real" root node

What are these properties and why cannot scene objects be used for defining them?

extensions/2.0/Vendor/GODOT_single_root/GODOT_single_root.md
```diff
@@ -0,0 +1,83 @@
...
+
+The `GODOT_single_root` glTF extension solves this problem by imposing additional constraints on the glTF scene to ensure it can be parsed into one of these objects with a single root node. Implementations can detect `GODOT_single_root` and import the single root node as the object's "real" root node in the scene/prefab/etc. The extension contains no data, it only restricts behavior. The restrictions are as follows:
```
lexaknyazev, line 21: The extension should mention whether it restricts nodes that do not belong to any scene.

lexaknyazev (changes requested):

javagl:
Just as a cross-reference: https://github.com/KhronosGroup/glTF/issues/1542

extensions/2.0/Vendor/GODOT_single_root/GODOT_single_root.md
aaronfranke: Properties defined in glTF nodes, especially extensions. I don't think it makes sense to duplicate all of the node logic to apply to scenes, both import and export. You wouldn't add a `"mesh"` to a scene, right?

extensions/2.0/Vendor/GODOT_single_root/GODOT_single_root.md
aaronfranke: The specific reason is to make the `"scene"` and `"scenes"` properties redundant, so the glTF file does not have any properties that specify which node is the root node. Since the goal is to have only root node, a list of root nodes is unnecessary and undesired, and is only required to keep for compatibility with glTF.

aaronfranke:
Yes, the restrictions here are very much related to #1542. For further explanation of the problem and proposed solution, see my comment at the bottom of that issue: https://github.com/KhronosGroup/glTF/issues/1542#issuecomment-1675792594

emackey (approved):
Since this is a vendor extension used in production vendor software, we should probably merge this (and generally any other PRs in similar condition).  My approval here is not an endorsement for use of this in other software, just an acknowledgement that we should allow vendor-specific documentation of its use.

aaronfranke:
What is the Khronos stance on merging vendor extensions? This is finalized and actively used in production.

javagl:
Maybe an explicit ping @lexaknyazev can help. (I skimmed over the requested changes, and saw that they had been marked as 'resolved', but not by the _reviewer_ - so this PR will at least need another approval, to make sure that the changes actually are resolved _from the standpoint of the reviewer_). 

(No opinion from my side. The _goal_ of the extension itself sounds reasonable, as a somewhat standardized, non-breaking way of solving https://github.com/KhronosGroup/glTF/issues/1542 , and the goal of ~"making `scene/scenes` obsolete" sounds reasonable as well, but the devil is in the detail, and I trust Alexey to ensure that there is no breaking change or inconsistency hidden in the specification text)

donmccurdy:
Hey all, I feel it's important that vendor extensions are approved and merged regularly (in the absence of egregious quality issues). If there's not bandwidth for a thorough review from working group members, that's entirely understandable! But perhaps the default in that case should be to merge extensions as-is, and perhaps even automatically. GitHub "labels" on extension PRs could be helpful as well, to indicate general PR status and next steps.

emackey:
/ping @lexaknyazev This PR is actually blocked by your earlier request for changes. None of us have GitHub permission to merge this with that request still present here.  Can you remove the block?

fire:
@lexaknyazev Can you reply and remove the block or at least state reasoning for keeping the block?

extensions/2.0/Vendor/GODOT_single_root/examples/dynamic_body_root.gltf
lexaknyazev, line 1: Update the example asset to use only this extension and add a link to it to the extension document OR just remove the example asset since the inline example present in the README is enough for showing the extension design.

aaronfranke:
The example asset has been removed as requested.

aaronfranke:
@lexaknyazev This PR is blocked from merging by your request for changes.

extensions/2.0/Vendor/GODOT_single_root/README.md
```diff
@@ -0,0 +1,89 @@
...
+- The document `"scene"` index MUST be set to 0, the index of the only scene in the `"scenes"` array.
+- The scene MUST have exactly one node, the single root node.
+- The scene's node index MUST be set to 0, the index of the first node in the `"nodes"` array.
+- The `"translation"`, `"rotation"`, `"scale"`, and `"matrix"` properties of the single root node MUST either be not defined or have default values.
+
+These restrictions make the glTF `"scenes"` and `"scene"` properties fully redundant, so that implementations using `GODOT_single_root` do not need to parse the `"scenes"` or `"scene"` properties. Since there is only one root node, node index 0 is reserved as the single root node. However, the `"scenes"` and `"scene"` properties must be kept in the glTF file for compatibility with implementations that do not support this extension. This means that the glTF file must contain properties equivalent to this:
+
+```json
+	"scene": 0,
+	"scenes": [{ "nodes": [0] }],
```
lexaknyazev, line 33: Use spaces for indentation.

extensions/2.0/Vendor/GODOT_single_root/README.md
```diff
@@ -0,0 +1,89 @@
...
+	"extensionsUsed": [
+		"GODOT_single_root"
+	],
+	"nodes": [
+		{
+			"name": "SingleRootNode"
+		}
+	],
+	"scene": 0,
+	"scenes": [{ "nodes": [0] }]
```
lexaknyazev, line 64: Use spaces for indentation.

extensions/2.0/Vendor/GODOT_single_root/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+```json
+	"scene": 0,
+	"scenes": [{ "nodes": [0] }],
+```
+
+Note: Other properties set on the scene JSON, such as the name etc, are allowed, but not required or expected. If the scene does not have a name, use the single root node's name as the scene name.
+
+The single root node acts as the root object that represents the entire glTF file, since all other nodes in the tree are descendants, and all resources such as meshes are either used by this node or by descendants. This requires the glTF file to only contain one scene, with one root node. Multiple scenes per file is not allowed, since it goes against this goal, and [is usually not a meaningful concept](https://github.com/KhronosGroup/glTF/issues/1542).
```
lexaknyazev, line 38: Remove this link; an extension should not make such comments about the base spec.

extensions/2.0/Vendor/GODOT_single_root/GODOT_single_root.md
lexaknyazev: Nodes not belonging to any scene would violate the "single root" idea so the extension should be explicit about them.

Without addressing this question, the extension becomes completely redundant since explicitly doing
```js
if (gltf.scene === 0 &&
    gltf.scenes.length == 1 &&
    gltf.scenes[0].nodes.length == 1 &&
    gltf.scenes[0].nodes[0] === 0 &&
    gltf.nodes[0].hasDefaultTransform())
{
   singleRoot = true; // actually single root
}
```

is not more difficult but much more reliable than something like

```js
if (gltf.extensionsUsed.contains("GODOT_single_root "))
{
   singleRoot = true; // maybe single root
}
```

aaronfranke:
@lexaknyazev Updated per your latest comments. Clarified that nodes are allowed to not be in the scene.

> Without addressing this question, the extension becomes completely redundant since explicitly doing

Those are all prerequisite conditions to being compliant with `GODOT_single_root`, but those checks alone would not be sufficient to have the desired behavior. Consider the following case:

```
glTF Scene
    glTF Node A
        glTF Node B
```

With `GODOT_single_root`, the imported root is "glTF Node A". Without, the root is "glTF Scene". We can't automatically make "glTF Node A" be the root for consistency with this case:

```
glTF Scene
    glTF Node A
        glTF Node B
    glTF Node C
```

In this case, we have no choice but to make "glTF Scene" be the root. We don't want the existence or non-existence of "glTF Node C" to affect the behavior of "glTF Node A" as the scene root or not. Therefore, we make an explicit flag for files to ask for "glTF Node A" to be the root, and this flag is only valid when the prerequisite conditions are true.

extensions/2.0/Vendor/GODOT_single_root/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+The `GODOT_single_root` glTF extension solves this problem by imposing additional constraints to ensure the entire glTF file can be parsed into one of these objects with a single root node. Therefore, all other nodes are either descendants of this node, or are not in the glTF scene node tree. Implementations can detect `GODOT_single_root` and import the single root node as the object's "real" root node in the scene/prefab/etc. The extension contains no data, it only restricts behavior. The restrictions are as follows:
```
lexaknyazev, line 21: Does the updated language mean that out-of-scene nodes are silently ignored?

extensions/2.0/Vendor/GODOT_single_root/README.md
```diff
@@ -0,0 +1,89 @@
...
+These restrictions make the glTF `"scenes"` and `"scene"` properties fully redundant, so that implementations using `GODOT_single_root` do not need to parse the `"scenes"` or `"scene"` properties. Since there is only one root node, node index 0 is reserved as the single root node. However, the `"scenes"` and `"scene"` properties must be kept in the glTF file for compatibility with implementations that do not support this extension. This means that the glTF file must contain properties equivalent to this:
+
+```json
+    "scene": 0,
+    "scenes": [{ "nodes": [0] }],
+```
+
+Note: Other properties set on the scene JSON, such as the name etc, are allowed, but not required or expected. If the scene does not have a name, use the single root node's name as the scene name.
+
+The single root node acts as the root object that represents the entire glTF file, since all other nodes in the tree are descendants, and all resources such as meshes are either used by this node or by descendants. This requires the glTF file to only contain one scene, with one root node. Multiple scenes per file is not allowed, since it goes against this goal, and is usually not a meaningful concept.
```
lexaknyazev, line 38: Please remove the "and is usually not a meaningful concept" part.

lexaknyazev:
> Therefore, we make an explicit flag for files to ask for "glTF Node A" to be the root

So the goal is to explicitly flag assets that _should_ be imported without creating an extra "Scene" node assuming that they are compatible with such workflows. In other words, the main point of the extension is not to add some restrictions but to signal the intended asset usage that needs those restrictions.

Following the example above, the behaviour should then be:

| Node C | Single root flag | Action |
|---|---|---|
| ❌ | ❌ | Import scene as root |
| ✔️ | ❌ | Import scene as root |
| ❌ | ✔️ | Import Node A as root |
| ✔️ | ✔️ | Asset is invalid |

The first sentence of the extension overview should highlight this key detail (feel free to tweak the language):
> This extension allows flagging glTF assets as having a single root node to enable more efficient and accurate workflows with various DCC tools, e.g., by directly importing the asset's node hierarchy without creating additional scene nodes. This flag imposes certain constraints on glTF assets as described below.

extensions/2.0/Vendor/GODOT_single_root/README.md
aaronfranke: Aren't they already silently ignored with base glTF, unless some other extension uses them?

javagl: ^ That _might_ have referred to _skeletons_, but I haven't read the details, so that might be completely off-track. 

Iff it referred to skeletons, then maybe 

> all other nodes are either descendants of this node

could be

> all other _non-skeleton_ nodes are either descendants of this node

or so, but all this is to be confirmed - maybe I mistunderstood something here

aaronfranke: I don't want to impose a restriction either which way for non-scene nodes. They should behave the same as a vanilla glTF asset.

lexaknyazev: Ack, assuming that non-scene node presence does not lead to creating a synthetic "Scene-for-non-Scene-nodes" root.

lexaknyazev (approved):
LGTM since the extension intention is clearly stated now.

PR #2335, opened 2023-10-15
Fix apparent FB_geometry_metadata copy-paste typo

toji:
Corrects the extension property name and content description.


emackey (approved):
Thanks @toji.

PR #2338, opened 2023-10-17
Add SVG support for textures in GLTF specification

fire:
We believe it is possible to have two implementations: one in `three.js` and another in `Godot Engine`.

I am opening this proposal for discussion. Please feel free to share your comments.

### DIFF
--- a/extensions/2.0/Vendor/EXT_texture_svg/README.md
+++ b/extensions/2.0/Vendor/EXT_texture_svg/README.md
@@ -1,0 +1,120 @@
+# EXT_texture_svg
+
+## Contributors
+
+- K. S. Ernest (iFire) Lee, Godot Engine contributor, V-Sekai contributor.
+- Aaron Franke, The Mirror Megaverse Inc.
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension allows glTF models to use SVG as a valid image format. A client that does not implement this extension can ignore the provided SVG image and continue to rely on the PNG and JPG textures available in the base specification. Defining a fallback texture is optional. The [best practices](#best-practices) section describes the intended use case of this extension and the expected behavior when using it without a fallback texture.
+
+SVG is a vector image format that can be scaled to any size without loss of quality. SVG is supported by all major game engines and web browsers.
+
+SVG images are ideal for assets with simple shapes or text. For instance, a glTF asset like a billboard displaying a logo or a sign with substantial text would benefit from SVGs. This allows for clear visibility at any distance without the need for large raster images that increase file size.
+
+For non-vectorizable artwork, PNG and JPEG raster images can be embedded into SVG to allow for mixing vector shapes and text with complex detailed raster images.
+
+## glTF Schema Updates
+
+The `EXT_texture_svg` extension is added to the `textures` node and specifies a `source` property that points to the index of the `images` node which in turn points to the SVG image.
+
+The following glTF will load `image.svg` in clients that support this extension, and otherwise fall back to `image.png`.
+
+```
+"textures": [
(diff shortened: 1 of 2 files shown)


donmccurdy:
Is "SVG as texture" something that is done in engines today? three.js does have an SVG loader, but it generates triangulated geometry, usually as a one-time operation. While we could then draw that geometry to a texture, it may be inefficient as a way to solve the stated goals of this extension. Streaming greater detail at closer distances sounds difficult — more in the territory of OpenSubDiv — and I'm not sure how to implement that on SVG.

Crisp, legible text in a 3D scene is an important problem, and not easily handled today. If that's a goal for glTF, I think it would be important that we either (a) offer quality and performance _at least as good_ as SDF/MSDF fonts, or (b) provide the source text and layout hints so that implementations are free to improve over time. I'm worried that transmitting text as SVG would not satisfy either of those goals.

fire:
Storing the SVG allows for decomposition via importer-time scaling to a raster image, curve conversion to mesh geometry or (multichannel) signed distance fields. These method can match SDF/MSDF fonts in quality and performance. However, raw svg is resource-intensive and may not be feasible for real-time applications or systems with limited resources compared to its decomposed forms.

Making mesh geometry efficient is outside the scope of this pr. See meshoptimizer, msft lod and etc. However, it is possible to use a series of LODs to provide performance without using dynamically subdivision on the mesh.

While SVG might not be the optimal format for delivering beziers and raster embeds, its wide availability makes it a practical choice. Another format that is available is lottie for vector graphics animation.

For providing source text and layout hints for future improvements, storing intermediary SVG inputs might not be feasible, especially when they are procedural generated or involve fonts that may not be available on all systems. PDF/A archive ensures its files are self contained.

Having yet another 2d vector standard inside of GLTF is possible too.

- https://github.com/MerlinVR/Unity-MSDF-Fonts
- https://github.com/godotengine/godot/pull/60513
- https://docs.unity3d.com/Manual/com.unity.textmeshpro.html
- https://docs.unrealengine.com/4.27/en-US/Basics/Actors/3DText/
- https://en.wikipedia.org/wiki/PDF/A
- https://github.com/zeux/meshoptimizer
- https://www.thorvg.org/ (lottie and svg)

fire:
Lottie is in the process of standardization with the Linux Foundation group.

aaronfranke:
@donmccurdy SVG stands for Scalable Vector Graphics. It is also a 2 dimensional format. While you could triangulate it as a mesh, and another extension could do that, the simplest and most natural way to interpret a 2D graphic is an image.

elalish:
Personally, I find this most interesting in conjunction with animated textures - this is why we started supporting Lottie animations in `<model-viewer>`: https://modelviewer.dev/examples/scenegraph/#animatedTexturesExample

For static textures, vector graphics have all the problems @donmccurdy points out, since with traditional rendering you would rasterize the graphic at a fixed resolution ahead of time. However, for animated textures, you are re-rasterizing the textures every frame anyway, which gives potential to also adjust the resolution on the fly. It'll still be resource intensive though. 

I know various parties have interest in a video texture extension for glTF - maybe that would be a good place to start talking about the potential for vector formats.

fire:
I have greatly support @hermet on thorvg, he is working on standardizing lottie. Perhaps there is interested in a lottie gltf extension.

hermet:
Hello, in most practical scenarios, rendering SVG/Lottie scenes requires several stages for composition, including blending, masking, and applying layered effects, etc which is significantly more complex than text / video drawing. This means that even if we attempt to draw the vector contents directly on the main surface, it would necessitate generating one or more intermediate frame buffers(textures) for these compositions. From my experience, as vector contents complexity increases, retaining the scene (texture) for Lottie/SVG can significantly improve performance among the applications.

PR #2339, opened 2023-10-22
EXT_materials_bump

bhouston:
This is a custom extension that we are using.  Figured it may be of interest to others.

### DIFF
--- a/extensions/2.0/Vendor/EXT_materials_bump/README.md
+++ b/extensions/2.0/Vendor/EXT_materials_bump/README.md
@@ -1,0 +1,128 @@
+# EXT\_materials\_bump
+
+## Contributors
+
+- Ben Houston, Threekit [@bhouston](https://github.com/bhouston)
+
+Copyright 2023 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft, Ratified by the Khronos Group
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Exclusions
+
+- This extension must not be used on a material that also uses `KHR_materials_pbrSpecularGlossiness`.
+- This extension must not be used on a material that also uses `KHR_materials_unlit`.
+
+## Overview
+
+This extension adds one parameters to the metallic-roughness material: `bump`.
+
+`bump` allows users to perturb the surface normal via a bump map. The bump map is a grayscale texture, where the value of each texel represents the height of the surface at that texel.  The height is then used to perturb the normal direction.
+
+## Extending Materials
+
+The `KHR_materials_bump` extension can be specified in the `extensions` property of the material:
+
+```json
+{
+    "materials": [
+        {
+            "extensions": {
+                "KHR_materials_bump": {
+                    "bumpFactor": 0.5
+                }
+            }
+        }
+    ]
+}
+```
+
+Factor and texture are combined by multiplication to describe a single value.
+
+| |Type|Description|Required|
+|-|----|-----------|--------|
+| **bumpFactor** | `number` | The strength of the specular reflection. | No, default: `1.0`|
+| **bumpTexture** | [`textureInfo`](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | A texture that defines the strength of the specular reflection, stored in the alpha (`A`) channel. This will be multiplied by bumpFactor. | No |
+
+The `bump` parameters affect the normal used in `dielectric_brdf` of the glTF 2.0 metallic-roughness material.
+
+## Implementation
+
+*This section is non-normative.*
+
+The `bump` parameter is used to perturb the normal direction. The normal direction is perturbed by the value of the bump map at the texel corresponding to the fragment. The perturbed normal is then used in the `dielectric_brdf` function of the glTF 2.0 metallic-roughness material.
+
+** TODO Derive Normal from Bump **
+
+In order to properly blend the bump map with a normal map, please follow the recommendation of https://blog.selfshadow.com/publications/blending-in-detail/.
+
+```glsl
+vec3 t = originalNormal.xyz * vec3( 2,  2, 2 ) + vec3( -1, -1, 0 );
+vec3 u = bumpNormal.xyz * vec3( -2, -2, 2 ) + vec3( 1,  1, -1 );
+vec3 r = normalize( t * dot(t, u) / t.z - u ); // not required if t and u are pre-normalized
+vec3 blendedNormal = r * 0.5 + 0.5;
+```
+
+## Schema
+
+- [material.KHR_materials_bump.schema.json](schema/material.KHR_materials_bump.schema.json)
+
+## Appendix: Full Khronos Copyright Statement
+
+Copyright 2023 The Khronos Group Inc.
+
+Some parts of this Specification are purely informative and do not define requirements
+necessary for compliance and so are outside the Scope of this Specification. These
+parts of the Specification are marked as being non-normative, or identified as
+**Implementation Notes**.
+
+Where this Specification includes normative references to external documents, only the
+specifically identified sections and functionality of those external documents are in
+Scope. Requirements defined by external documents not created by Khronos may contain
+contributions from non-members of Khronos not covered by the Khronos Intellectual
+Property Rights Policy.
+
+This specification is protected by copyright laws and contains material proprietary
+to Khronos. Except as described by these terms, it or any components
+may not be reproduced, republished, distributed, transmitted, displayed, broadcast
+or otherwise exploited in any manner without the express prior written permission
+of Khronos.
+
+This specification has been created under the Khronos Intellectual Property Rights
+Policy, which is Attachment A of the Khronos Group Membership Agreement available at
+www.khronos.org/files/member_agreement.pdf. Khronos grants a conditional
+copyright license to use and reproduce the unmodified specification for any purpose,
+without fee or royalty, EXCEPT no licenses to any patent, trademark or other
+intellectual property rights are granted under these terms. Parties desiring to
+implement the specification and make use of Khronos trademarks in relation to that
+implementation, and receive reciprocal patent license protection under the Khronos
+IP Policy must become Adopters and confirm the implementation as conformant under
+the process defined by Khronos for this specification;
+see https://www.khronos.org/adopters.
+
+Khronos makes no, and expressly disclaims any, representations or warranties,
+express or implied, regarding this specification, including, without limitation:
+merchantability, fitness for a particular purpose, non-infringement of any
+intellectual property, correctness, accuracy, completeness, timeliness, and
+reliability. Under no circumstances will Khronos, or any of its Promoters,
+Contributors or Members, or their respective partners, officers, directors,
+employees, agents or representatives be liable for any damages, whether direct,
+indirect, special or consequential damages for lost revenues, lost profits, or
+otherwise, arising from or in connection with these materials.
+
+Vulkan is a registered trademark and Khronos, OpenXR, SPIR, SPIR-V, SYCL, WebGL,
+WebCL, OpenVX, OpenVG, EGL, COLLADA, glTF, NNEF, OpenKODE, OpenKCAM, StreamInput,
+OpenWF, OpenSL ES, OpenMAX, OpenMAX AL, OpenMAX IL, OpenMAX DL, OpenML and DevU are
+trademarks of The Khronos Group Inc. ASTC is a trademark of ARM Holdings PLC,
+OpenCL is a trademark of Apple Inc. and OpenGL and OpenML are registered trademarks
+and the OpenGL ES and OpenGL SC logos are trademarks of Silicon Graphics
+International used under license by Khronos. All other product names, trademarks,
+and/or company names are used solely for identification and belong to their
+respective owners.
--- a/extensions/2.0/Vendor/EXT_materials_bump/schema/material.KHR_materials_bump.schema.json
+++ b/extensions/2.0/Vendor/EXT_materials_bump/schema/material.KHR_materials_bump.schema.json
@@ -1,0 +1,24 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_materials_bump glTF Material Extension",
+    "type": "object",
+    "description": "glTF extension that defines a bump which perturbs the surface normal.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "bumpFactor": {
+            "type": "number",
+            "description": "The strength of the bump displacement.",
+            "default": 1.0,
+            "minimum": 0.0,
+            "maximum": 1.0,
+            "gltf_detailedDescription": "This parameter scales the amount of normal perturbation by the bump map on the surface normal."
+        },
+        "bumpTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "A texture that defines the displacement in the alpha channel.",
+            "gltf_detailedDescription": "A texture that defines the displacement in the alpha channel. This will be multiplied by bumpFactor."
+        },
+        "extensions": { },
+        "extras": { }
+    }
+}


bhouston:
Example glTF that uses this extension exported via ThreeJS's glTF exporter is here: https://drive.google.com/file/d/1NX_LGF0Yp_F5ceWhStcoYbH_UGpuhFqz/view?usp=share_link. You can also import it into the ThreeJS editor via the ThreeJS glTF importer.  I can contribute this example to glTF-SampleAssets if desired.

extensions/2.0/Vendor/EXT_materials_bump/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Overview
+
+This extension adds one parameters to the metallic-roughness material: `bump`.
+
+`bump` allows users to perturb the surface normal via a bump map. The bump map is a grayscale texture, where the value of each texel represents the height of the surface at that texel.  The height is then used to perturb the normal direction.
+
+Benefits of bump maps are primarily these two:
+
+1. Bump maps are much more compact that normal maps and lead to smaller glTF files as well as less GPU memory requirements.  This is because the normal map has to store a full 3D vector per texel, while the bump map only needs to store a single scalar value per texel.  Thus bump maps require 3 times less data than normal maps for transmission and GPU storage.  Normal maps are also very sensitive to texture compressions methods, while bump maps are less sensitive.
```
donmccurdy, line 31: For purposes of glTF users, reducing GPU memory would require use of GPU compression formats, correct? If I'm reading `KHR_texture_basisu` correctly, that implies that `bumpTexture` must use the red channel, like `occlusionTexture`

extensions/2.0/Vendor/EXT_materials_bump/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+2. Bump maps are useful for capturing small details in combination with a normal map for large scale structures.  For example, when rendering human faces, there are usually two layers of surface pertubation detail, the unwrapped normal map is used for representing wrinkles, and then a repeating bump map is used for skin pore details.  This is also the case for many other materials such as wood, metal, and stone, where there is both macro and micro details.
```
donmccurdy, line 33: I'll just note that this "second layer for small details" is not unique to bump maps, and could also be achieved with a second normal map. Common use cases that benefit from a second layer of normals *might* also require a second layer `baseColorTexture` or `metallicRoughnessTexture` ... should that be part of the proposal?

Support for MaterialX-based subgraphs as inputs to material texture sockets would also solve these issues, though that is a more complex direction.

extensions/2.0/Vendor/EXT_materials_bump/README.md
```diff
@@ -0,0 +1,142 @@
...
+## Exclusions
+
+- This extension must not be used on a material that also uses `KHR_materials_pbrSpecularGlossiness`.
+- This extension must not be used on a material that also uses `KHR_materials_unlit`.
+
+## Overview
+
+This extension adds one parameters to the metallic-roughness material: `bump`.
+
+`bump` allows users to perturb the surface normal via a bump map. The bump map is a grayscale texture, where the value of each texel represents the height of the surface at that texel.  The height is then used to perturb the normal direction.
```
donmccurdy, line 27: I think "grayscale" will need clarification here. If this is a single-channel texture, can we specify which channel the implementation must sample from, and allow other channels to be used by other material texture slots? Probably that should be the red channel - see below.

extensions/2.0/Vendor/EXT_materials_bump/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+The `bump` parameter is used to perturb the normal direction. The normal direction is perturbed by the value of the bump map at the texel corresponding to the fragment. The perturbed normal is then used in the `dielectric_brdf` function of the glTF 2.0 metallic-roughness material.
+
+### Blending Normal and Bump Perturbations
+
+In order to properly blend the bump map with a normal map, please follow the recommendation of https://blog.selfshadow.com/publications/blending-in-detail/.
+
+```glsl
+vec3 t = originalNormal.xyz * vec3( 2,  2, 2 ) + vec3( -1, -1, 0 );
+vec3 u = bumpNormal.xyz * vec3( -2, -2, 2 ) + vec3( 1,  1, -1 );
```
donmccurdy, line 74: The reference in Stephen Hill's blog uses normal maps for both the base and detail layers, as Unity does. We would need to make adjustments to this code, if our implementation will instead use grayscale bump maps.

bhouston: I think that as long as you derive the normal from the bump map, you can use the same blending equation.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: We should specify the channel.  I was thinking it may be possible to pack in into another channel (like the ORM packing) so I sort of held off on that.  But if we want it to be its own map, red makes sense.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: I would prefer to keep this extension simple and mapping to how existing materials work in popular WebGL renders so that it is minimal effort to implement.  Note that both ThreeJS has support for [both a bump and a normal map](https://github.com/mrdoob/three.js/blob/dev/src/materials/MeshStandardMaterial.js#L34) as does BabylonJS has [support for both a bump and normal map] (https://github.com/BabylonJS/Babylon.js/blob/master/packages/dev/core/src/Materials/PBR/pbrBaseMaterial.ts#L495).

Right now glTF has the artificial limitation that it cannot represent what Three.js and Babylon.js can do.

I know that some may be purists and what to ensure that bump maps are relegated to history.  I am okay with that.  This is why I didn't suggest this as a KHR_ extension rather just an industry extension.

I am trying to get glTF to parity with what these engines can do.

(BTW Unity has support for 2 maps as well, the second is called a detail map - scroll down to the bottom: https://docs.unity3d.com/Manual/StandardShaderMaterialParameterNormalMap.html. Unity converts all bump maps to normal maps because they do not care about transmission size or GPU memory usage.)

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: If we do not pack the bump map into another channel of an existing common glTF texture map then yes, we would need to use a compressed texture if we want to reduce GPU memory requirements.

bhouston:
@donmccurdy would it make you feel better if I rename this to be EXT_materials_displacement?  And then I can add in text that says this can be rendered either as normal perturbation method or as a surface displacement?  I think the main difference is that if I call it EXT_materials_displacement, the displacementScale factor can go above 1.  Let me know...

extensions/2.0/Vendor/EXT_materials_bump/README.md
```diff
@@ -0,0 +1,142 @@
...
+## Contributors
+
+- Ben Houston, Threekit [@bhouston](https://github.com/bhouston)
+
+Copyright 2023 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft, Ratified by the Khronos Group
```
lexaknyazev, line 12: An extension cannot be ratified and draft at the same time.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: I don't see this extension as hurting anyone because if you export a gLTF from Babylon or ThreeJS right now, they just simply discard the bump map - so exporting is a lossy operation.  All this does is no longer discard it.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: fixed!

extensions/2.0/Vendor/EXT_materials_bump/README.md
elalish: I agree that the concept of combining different resolutions of maps is useful, and applies equally to all maps. Personally, I think the MaterialX-style procedural texture maps (which is already in progress for glTF) would be ideal for this, particularly if they can combine input images in addition to procedural generation. 

The hurt we need to be aware of is industry fragmentation. The big value of glTF is there is a "right way" to do PBR, which unifies artist workflows. We should consider carefully if we have two possible approaches that accomplish the same goal.

extensions/2.0/Vendor/EXT_materials_bump/README.md
elalish: Can I see the evidence that bump maps are less sensitive to compression artifacts than normals maps? I'm surprised because producing a normal map from a bump map is effectively taking a derivative, and derivatives amplify noise.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: > The big value of glTF is there is a "right way" to do PBR.  We should consider carefully if we have two possible approaches that accomplish the same goal.

Remember @elalish this is not a KHR_ extension, but rather just an industry pragmatic one, notice the EXT_ prefix.

I purposely was not aiming to get glTF committee blessing as I figured it wouldn't get blessing because bump maps are is viewed as the old way to do things.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: > Can I see the evidence that bump maps are less sensitive to compression artifacts than normals maps?

Normal maps can store information that bump maps struggle to store, but for things that bump maps can store, they work well.  Normal maps are more general purpose and historically been a struggle to reconstruct the floating point normals with high accuracy from compressed textures that just store 8-bit representations.  https://developer.download.nvidia.com/whitepapers/2008/real-time-normal-map-dxt-compression.pdf

Bump maps do not claim to be able to do the beautiful slow moving large scale smooth curves and shapes that normal maps do.  Rather they are for fine detail.  Using normal maps for fine detail is overkill in terms of memory usage.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: In order to get an EXT_ accepted into the Khronos registry of extension (there is no need for a vote because this isn't a KHR_ extension), one only needs to 2 implementations to adopt it.  Threekit is adopting this as we speak, thus only one other implementation is required.

donmccurdy:
My main concern would be (potentially) causing bump maps to become much more common in the ecosystem. There is currently no other way to achieve detail maps in glTF, so tools would have to implement an outdated feature (bump maps) in order to support a very useful modern feature (detail maps). Or if Blender and Sketchfab began writing bump maps with `EXT_materials_bump`, just because it's what the user uploaded, rather than converting to normal maps, that'd seem unfortunate.

Perhaps I'm overthinking the risk. I don't need to block this if others are happy with the proposal! I see displacement and parallax maps more positively.

***

Aside: I'd be very interested in the MaterialX story for vertex displacement and animation. Does that mean MaterialX input to a `displacementTexture` socket, or is it a separate thing? But maybe off topic here. 😅

bhouston:
@donmccurdy wrote:
> Or if Blender and Sketchfab began writing bump maps with EXT_materials_bump, just because it's what the user uploaded, rather than converting to normal maps, that'd seem unfortunate.

You do realize that right now Sketchfab just discards the bump map right if it exists?  They also support, like ThreeJS both a normal and a bump.  So we currently just have information loss.  This addresses that.

bhouston:
> My main concern would be (potentially) causing bump maps to become much more common in the ecosystem. 

It is possible to convert on load to a normal map from a bump map.  So you can stick with normal mapping through out your engine if you so desired.

>  I'd be very interested in the MaterialX story for vertex displacement and animation. 

That is great too!

I guess I do not see this as a step backwards, but rather just supporting what is already in the ecosystem.

bhouston:
> so tools would have to implement an outdated feature (bump maps) in order to support a very useful modern feature (detail maps)

BTW MaterialX supports bump maps: https://materialx.org/docs/api/class_height_to_normal_node_glsl.html. Pixar Renderman also has bump map support: https://rmanwiki.pixar.com/display/REN24/PxrBumpMixer (which was all reworked based on new 2020-era research outlined here: https://jcgt.org/published/0009/03/04/)

echadwick-artist:
As far as packing goes, I compiled a list of which channels are still unused by extensions which pack textures. However there's the downside of potentially incurring more compression noise from having disparate data in neighboring channels. /shrug

KHR_materials_anisotropy 	
RGB = anisotropyTexture
**Alpha = unused**

### KHR_materials_clearcoat
RGB = clearcoatNormalTexture
Red = clearcoatTexture
Green = clearcoatRoughnessTexture
**Blue = unused
Alpha = Unused**

### KHR_materials_iridescence
Red = iridescenceTexture
Green = iridescenceThicknessTexture 
**Blue = unused
Alpha = Unused**

### KHR_materials_sheen
RGB = sheenColorTexture
Alpha = sheenRoughnessTexture

### KHR_materials_specular		
RGB = specularColorTexture
Alpha = specularTexture

### KHR_materials_transmission
Red = transmissionTexture
### KHR_materials_volume
Green = thicknessTexture
**Blue = unused
Alpha = Unused**

extensions/2.0/Vendor/EXT_materials_bump/README.md
elalish: Okay, for 8-bit lossless images I see your point. I was thinking of lossy compression mostly, but we can't usually make use of that for normal maps anyway.

extensions/2.0/Vendor/EXT_materials_bump/README.md
elalish: Ah, thanks, I was actually wondering what the bar was for an EXT. Fair enough.

lexaknyazev:
> If the plan is to pack the bump into a shared texture, I would prefer we use the Blue channel, since there are three textures where it can fit.

Adding an unrelated blue channel to already uncorrelated two-component textures would greatly increase compression artifacts, in some cases preventing use of compressed data altogether.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: There are specific texture compression formats that are suitable for normal maps - such as those mentioned in the paper I linked in my previous comment.

But because bump map are used for micro details, generally you do not magnify them much at all.  At that size on screen you do not really seem compression artifacts.  Think of a noise texture that you would put on a piece of concrete or a fine ridged bump map you would put on commercial-grade carpet, or a bump map on shirt fabric - the details are so fine that you can not see any compression artifacts.

Whereas with normal maps they are usually representing larger scale details, large smooth surfaces, like the wrinkles on the shirt, and you do magnify them, thus the compression artifacts are much more visible.

The high frequency of the detail bump maps make it harder to see compression artifacts unlike the low frequency details in the normal map.

bhouston:
@lexaknyazev Threekit now supports this extension and mrdoob over at ThreeJS is also supportive - will merge my PR to ThreeJS one it is merged here: https://github.com/mrdoob/three.js/pull/27036#issuecomment-1780755602. This means that we have two implementations.

How do we get this merged into Khronos as a community standard then?
 
I can also commit to adding it to the @donmccurdy gltf-transform tool as an optional extension there as well if he let's me.

bhouston:
> Adding an unrelated blue channel to already uncorrelated two-component textures would greatly increase compression artifacts, in some cases preventing use of compressed data altogether.

I suggest we just use the red channel and do not merge it with any other textures for the time being.  I think it doesn't fit well with any of the ones @echadwick-artist mentions - it isn't a natural fit with those extensions anyhow.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: I will mark this as resolved if that is okay.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: I agree, I have made the default channel A as you suggest.

extensions/2.0/Vendor/EXT_materials_bump/README.md
bhouston: I have changed the extension to use the red channel per @donmccurdy's suggestion.

extensions/2.0/Vendor/EXT_materials_bump/README.md
lexaknyazev: > I agree, I have made the default channel A as you suggest.

Did you mean `R`?

donmccurdy:
> > Or if Blender and Sketchfab began writing bump maps with EXT_materials_bump, just because it's what the user uploaded, rather than converting to normal maps, that'd seem unfortunate.
> 
> You do realize that right now Sketchfab just discards the bump map right if it exists? They also support, like ThreeJS both a normal and a bump. So we currently just have information loss.

I did not; that is unfortunate. 😭

> ...would it make you feel better if I rename this to be EXT_materials_displacement? And then I can add in text that says this can be rendered either as normal perturbation method or as a surface displacement?

I think it's probably best to keep these as two separate techniques.

***


I don't feel strongly enough to say more than I have, and will bow out here. It would be helpful if other vendors could give some rough sense of positive/neutral/negative signal — however you feel is best for your applications and communities. 🙂

bhouston:
BTW Alban over at Sketchfab liked the extension when I mentioned it to him: it:https://twitter.com/benhouston3d/status/1716640855455412305

bhouston:
This is now officially supported in ThreeJS, was merged today.  Thus it is supported in 2 separate systems - ThreeJS and Threekit. :)

extensions/2.0/Vendor/EXT_materials_bump/README.md
```diff
@@ -0,0 +1,148 @@
...
+Written against the glTF 2.0 spec.
+
+## Exclusions
+
+- This extension must not be used on a material that also uses `KHR_materials_pbrSpecularGlossiness`.
+- This extension must not be used on a material that also uses `KHR_materials_unlit`.
+
+## Overview
+
+This extension adds one parameters to the metallic-roughness material: `bump`.
```
lexaknyazev, line 25: Typo: one **parameter**

extensions/2.0/Vendor/EXT_materials_bump/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+`bump` allows users to perturb the surface normal via a bump map. The bump map is a grayscale texture, where the value of each texel represents the height of the surface at that texel.  The height is then used to perturb the normal direction.
```
lexaknyazev, line 27: Please change "grayscale" to "single-channel".

extensions/2.0/Vendor/EXT_materials_bump/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+Bump maps were introduced into computer graphics in 1978 by Blinn in "Simulation of Wrinkled Surfaces, while normal maps were introduced in 1989 by Cohen et al, "Appearance-Preserving Simplification."
```
lexaknyazev, line 29: Please split the rest of this section (starting from this line) into a new section called "Motivation" with the following text prepended:
> *This section is non-normative.*

extensions/2.0/Vendor/EXT_materials_bump/README.md
```diff
@@ -0,0 +1,148 @@
...
+    ]
+}
+```
+
+Factor and texture are combined by multiplication to describe a single value.
+
+| |Type|Description|Required|
+|-|----|-----------|--------|
+| **bumpFactor** | `number` | The strength of the specular reflection. | No, default: `1.0`|
+| **bumpTexture** | [`textureInfo`](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | A texture that defines the strength of the specular reflection, stored in the red (`R`) channel. This will be multiplied by bumpFactor. | No |
```
lexaknyazev, line 64: Descriptions are clearly from a different extension.

extensions/2.0/Vendor/EXT_materials_bump/schema/material.EXT_materials_bump.schema.json
```diff
@@ -0,0 +1,24 @@
...
+            "description": "The strength of the bump displacement.",
+            "default": 1.0,
+            "minimum": 0.0,
+            "maximum": 100.0,
+            "gltf_detailedDescription": "This parameter scales the amount of normal perturbation by the bump map on the surface normal."
+        },
+        "bumpTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "A texture that defines the displacement in the red channel.",
+            "gltf_detailedDescription": "A texture that defines the displacement in the red channel. This will be multiplied by bumpFactor."
```
lexaknyazev, line 19: "bumpFactor" -> "the bump factor"

extensions/2.0/Vendor/EXT_materials_bump/schema/material.EXT_materials_bump.schema.json
```diff
@@ -0,0 +1,24 @@
...
+    "description": "glTF extension that defines a bump which perturbs the surface normal.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "bumpFactor": {
+            "type": "number",
+            "description": "The strength of the bump displacement.",
+            "default": 1.0,
+            "minimum": 0.0,
+            "maximum": 100.0,
+            "gltf_detailedDescription": "This parameter scales the amount of normal perturbation by the bump map on the surface normal."
```
lexaknyazev, line 14: Please prepend the detailed description with the short description from L10.

lexaknyazev (changes requested):

extensions/2.0/Vendor/EXT_materials_bump/schema/material.EXT_materials_bump.schema.json
```diff
@@ -0,0 +1,24 @@
...
+    "type": "object",
+    "description": "glTF extension that defines a bump which perturbs the surface normal.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "bumpFactor": {
+            "type": "number",
+            "description": "The strength of the bump displacement.",
+            "default": 1.0,
+            "minimum": 0.0,
+            "maximum": 100.0,
```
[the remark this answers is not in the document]
DanielSturk: I'd suggest removing the minimum as well to allow for negative scales, like the normalTexture.

PR #2340, opened 2023-10-22
KHR_materials_dispersion

bhouston:
A first draft of something we are supporting.


emackey:
Hi Ben, this proposal sparked a long conversation in the PBR TSG today.  General comments and feedback:

- This looks like a great candidate, and we think it should have the `KHR_` prefix.
- The Abbe number appears in [Enterprise PBR](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2022x.md.html) (section 1.9, Dispersion), and in [OpenPBR](https://academysoftwarefoundation.github.io/OpenPBR/) (section 3.3.3), indicating compatibility with other PBR systems, which we like.
- The draft should be more clear about its dependence on `KHR_materials_volume` and its effect on transmissive volumetric materials such as gemstones.
- It would be good to mention (in a non-normative "Implementation Notes" section) that there's a straightforward realtime implementation, for systems that are already processing volumetric IOR, this just adjusts the IOR per color channel.
- The default value should make no change to the core material.  In other words, adding an empty JSON object `"KHR_materials_dispersion" : { }` should not change the material at all.  Perhaps the number 0 indicates no dispersion, and it becomes the default.  This is in line with all of our other PBR extensions: The PBR exists outside of the extension (in Blender and in ThreeJS etc), and so the glTF JSON structure has no influence on the material by itself.
- It would be nice to name references.  Maybe we can name the references that OpenPBR names?

Overall this looks like a great start!

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -1,4 +1,4 @@
-# EXT\_materials\_disperson
+# KHR\_materials\_disperson
```
lexaknyazev, line 1: nit: typo

bhouston: fixed

bhouston:
Hi @emackey!  Thanks for the quick feedback!

> This looks like a great candidate, and we think it should have the KHR_ prefix.

Done!

> The Abbe number appears in [Enterprise PBR](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2022x.md.html) (section 1.9, Dispersion), and in [OpenPBR](https://academysoftwarefoundation.github.io/OpenPBR/) (section 3.3.3), indicating compatibility with other PBR systems, which we like.

I've referenced both of those now in the spec and also was clear on their parameter naming.  OpenPBR represents dispersion via two separate numbers in order to make it easier for artistic control (Abbe number and scale factor), but they still resolve to a single value in the end, thus I suggest we stick with just Abbe number in the glTF material model.

> The draft should be more clear about its dependence on KHR_materials_volume and its effect on transmissive volumetric materials such as gemstones.

I've added an image of a gemstone as well as referencing KHR_material_volume as the extension this is enhancing.

> It would be good to mention (in a non-normative "Implementation Notes" section) that there's a straightforward realtime implementation, for systems that are already processing volumetric IOR, this just adjusts the IOR per color channel.

Done.

> The default value should make no change to the core material. In other words, adding an empty JSON object "KHR_materials_dispersion" : { } should not change the material at all. Perhaps the number 0 indicates no dispersion, and it becomes the default. This is in line with all of our other PBR extensions: The PBR exists outside of the extension (in Blender and in ThreeJS etc), and so the glTF JSON structure has no influence on the material by itself.

Done!  Good idea.

> It would be nice to name references. Maybe we can name the references that OpenPBR names?

I've referenced OpenPBR and its references!

> Overall this looks like a great start!

Nice!  Thank you!

MiiBond:
Very cool! Thanks for defining this.
One thing to note is that the dispersion parameter in OpenPBR, Autodesk Standard Surface and Adobe Standard Material are all varying (i.e. texturable). I know that complicates things but, for the sake of future-proofing, maybe it's worth considering?
Another thing to consider is that the Abbe number doesn't scale in a perceptually-linear way (in terms of the visible amount of dispersion). But the reciprocal of the Abbe number _is_ proportional to the change in the IOR across the spectrum and the angle of refraction varies approximately linearly with the index at these scales. Also, higher Abbe numbers result in less dispersion which might be somewhat counter-intuitive. 
I'm not suggesting that being intuitive to the artist is really that important for glTF since most people aren't modifying the glTF parameters directly so take this for what it's worth.
In ASM, we define the `dispersion` parameter as 20/V so that a value of 1.0 is equivalent to V=20, which is about the lowest Adde number for real materials. Values over 1.0 are still valid for artists that want to exaggerate the effect. Decreasing values lower the amount of dispersion down to 0.0.
Aside from being more intuitive for artists, this mapping also has the added benefit of being more easily defined with a texture.

bhouston:
@MiiBond I can see us using the 20/V parametrization, it makes sense.  I can make that change.  I had noticed that in the paper.

@MiiBond does OpenPBR/ASS/ASM allow for ior to vary around a material and do they allow for absorption to vary across the surface of a material?  glTF has not adopted these textures in part because it the surface parameterization is hard to extrapolate through the volume correctly and clearly.  I know one could technically use a generalized Barycentric coordinates (e.g. Warren coordinates.). I figured we were staying away from this use case.  If you want to make the case that we should vary Dispersion on the surface of the material via a map unlike Absorption and IOR, please do.

emackey:
@MiiBond to confirm, if we invert the value (20/V), does that make it correspond more linearly to the visual effect of dispersion?  For example, if the user supplied a value of `0.1` it shows visually very small amounts of dispersion, but a value of `0.9` or `1.0` shows a larger visual separation of colors?  If so then yes, @bhouston if you don't mind adjusting this draft to use this parameterization, that would be the PBR TSG's preference.  In the JSON schema, The `default` and `minimum` could be `0.0` (with 20 / 0.0 being the "infinite" Abbe number), and `maximum` could be omitted since there's no hard stop at 20 / 1.

On the PBR call today we also talked about the possibility of texturing this parameter, but decided against, at least for now.  We previously ratified a non-texturable IOR setting on the idea that it's volumetric and not a surface-texture property, and it wouldn't make sense to reverse that decision here for dispersion.

emackey:
Sample model time.  This one presumes the 20/V change, where `dispersion` ranges from zero to one.

[DispersionIorTest.zip](https://github.com/KhronosGroup/glTF/files/13272908/DispersionIorTest.zip)

MiiBond:
Sorry, I've been incapacitated by the flu since Thursday.
@emackey, yes, that's correct. 0.0 is no-effect and 1.0 would be the maximum, real-world amount of dispersion.
@bhouston, IOR and dispersion are varying in OpenPBR but I believe that it's still on the TODO list to define what that means for volumetric properties. They're also varying in Standard Surface, I think, so maybe that documentation has some insights. In ASM, we made both IOR and dispersion uniform to avoid this issue.

As you guys said, though, it probably doesn't make sense to change things in glTF at this point.

lexaknyazev (changes requested):

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+| |Type|Description|Required|
+|-|----|-----------|--------|
+| **dispersion** | `number` | The strength of the dispersion effect, specified as 20/Abbe number. | No, default: `0`|
+
+The default value of 0 has a special meaning in that no dispersion should be used.  This is the default value for backwards compatibility.  Any value zero or larger is considered to be a valid dispersion value, although the range between [0 , 1] is the range of realistic values.
+
+Here is a table of some material dispersion Abbe numbers, including the outlier of Rutile which is a very high dispersion material:
+
+| Material | Abbe Number (V) |
```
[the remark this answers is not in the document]
bhouston: addressed

MiiBond:
My Babylon.js implementation of this extension has been merged.
https://github.com/BabylonJS/Babylon.js/pull/14501

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+- [material.KHR_materials_dispersion.schema.json](schema/material.KHR_materials_dispersion.schema.json)
+
+## Reference
+
+### Theory, Documentation and Implementations
+
+[Abbe Number - Wikipedia](https://en.wikipedia.org/wiki/Abbe_number)
+
+[Abbe Number - Wolfram Formula Repository](https://resources.wolframcloud.com/FormulaRepository/resources/Abbe-Number#:~:text=The%20Abbe%20number%2C%20also%20known,of%20V%20indicating%20low%20dispersion.)
```
lexaknyazev, line 108: Consider dropping the fragment part (starts with `#`) of the link.

emackey:
New sample model available!  I've switched from gems to prisms, and updated the background cloth.  Also, the range of values goes up to 5.0 in the model now.

Here's the GLB:

[DispersionIorTest_v2.zip](https://github.com/KhronosGroup/glTF/files/13641603/DispersionIorTest_v2.zip)

And some screenshots, rendered by https://sandbox.babylonjs.com/ using Mike's newly-merged PR.  I'm using the "Studio" IBL with default settings here.

If someone could path-trace this model for comparison, that would be much appreciated!

Dispersion_screenshot1

Dispersion_screenshot2

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
emackey: Fixed.

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+Optical dispersion is represented in terms of the Abbe number parameterization \( $V$ \).  The dispersion effect is a result of the wavelength-dependent index of refraction of a material.  Dispersion is a widely adopted parameter in modern PBR models.  It is present in both OpenPBR (as *transmission_dispersion_abbe_number*) and the Dassault Enterprise PBR Shading Model (as $V_d$).
+
+The Abbe number \( $V$ \) is computed from the index of refraction at three wavelengths of visible light: 486.1 nm (short wavelength blue, $n_F$), 587.6 nm (central wavelength, $n_d$), and 656.3 nm (long wavelength red, $n_C$).  The Abbe number makes the simplifying assumption that the index of refraction variance is linear:
+
+$$V = \frac{n_d - 1}{n_F - n_C}$$
+
+To calculate the index of refraction at a specific wavelength \( $\lambda$ \), given an Abbe number \( $V$ \) and the central index of refraction as specified by the ```KHR_materials_ior``` extension (assumed to be at the central wavelength, \( $N_d$ \)):
```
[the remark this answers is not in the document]
emackey: Thanks @proog128.  Great to see you here!

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Overview
+
+This extension adds one parameter to the metallic-roughness material: `dispersion`.
+
+`dispersion` enables configuring the strength of the angular separation of colors (chromatic aberration) transmitting through a relatively clear volume.  It is an enhancement to the default ```KHR_materials_volume``` transmission model which assumes no dispersion.
+
+Optical dispersion is represented in terms of the Abbe number parameterization \( $V$ \).  The dispersion effect is a result of the wavelength-dependent index of refraction of a material.  Dispersion is a widely adopted parameter in modern PBR models.  It is present in both OpenPBR (as *transmission_dispersion_abbe_number*) and the Dassault Enterprise PBR Shading Model (as $V_d$).
+
+The Abbe number \( $V$ \) is computed from the index of refraction at three wavelengths of visible light: 486.1 nm (short wavelength blue, $n_F$), 587.6 nm (central wavelength, $n_d$), and 656.3 nm (long wavelength red, $n_C$).  The Abbe number makes the simplifying assumption that the index of refraction variance is linear:
```
bsdorra, line 32: Suggest small reformulation to clarify that this is the Abbe definition for visible light (others exist), and to add info on the origin of the three wavelength constants. 

```suggestion
The Abbe number \( $V$ \) for visible light is computed from the index of refraction at three wavelengths of Fraunhofer's spectral lines: 486.1 nm (short wavelength blue, $n_F$), 587.6 nm (central wavelength, $n_d$), and 656.3 nm (long wavelength red, $n_C$). The Abbe number makes the simplifying assumption that the index of refraction variance is linear:
```

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,170 @@
...
+And then:
+
+$$
+A = n_d - \frac{B}{\lambda_d^2}
+$$
+
+And finally:
+
+$$
+n(\lambda) = A + \frac{B}{\lambda^2}
```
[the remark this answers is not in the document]
bsdorra: To be precise, the first two terms of the general Cauchy equation. We should probably add this info as well?
$$
\eta(\lambda) = a + \frac{b}{\lambda^2} + \frac{c}{\lambda^4} + ...,
$$

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Overview
+
+This extension adds one parameter to the metallic-roughness material: `dispersion`.
+
+`dispersion` enables configuring the strength of the angular separation of colors (chromatic aberration) transmitting through a relatively clear volume.  It is an enhancement to the default ```KHR_materials_volume``` transmission model which assumes no dispersion.
+
+Optical dispersion is represented in terms of the Abbe number parameterization \( $V$ \).  The dispersion effect is a result of the wavelength-dependent index of refraction of a material.  Dispersion is a widely adopted parameter in modern PBR models.  It is present in both OpenPBR (as *transmission_dispersion_abbe_number*) and the Dassault Enterprise PBR Shading Model (as $V_d$).
+
+The Abbe number \( $V$ \) is computed from the index of refraction at three wavelengths of visible light: 486.1 nm (short wavelength blue, $n_F$), 587.6 nm (central wavelength, $n_d$), and 656.3 nm (long wavelength red, $n_C$).  The Abbe number makes the simplifying assumption that the index of refraction variance is linear:
```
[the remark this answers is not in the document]
emackey: The TSG agreed on the $V_d$ change. Done.

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+And finally:
+
+$$
+n(\lambda) = A + \frac{B}{\lambda^2}
+$$
+
+![Dispersion on a Gem](./figures/Dispersion.jpg)
+
+In this extension, we store a transformed dispersion instead of the Abbe number directly.  Specifically we store $20/V$ so that a value of 1.0 is equivalent to $V=20$, which is about the lowest Abbe number for normal materials. Values over 1.0 are still valid for artists that want to exaggerate the effect. Decreasing values lower the amount of dispersion down to 0.0.  This is the same transform used by both Adobe Standard Material and OpenPBR.
```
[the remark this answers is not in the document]
emackey: The group talked about this, but decided that this mapping is in common use in other systems (Adobe ASM, OpenPBR has both this and raw Abbe available, etc).  For people looking up physical values in a table, this is a simple calculation to apply, and doesn't make it any less "physical."  For example one may convert Fahrenheit to Celsius without loosening the relationship with a physical temperature.  And for artists wishing to take control with creative license, having a simple linear control is clearly the better choice.  So, I think we will leave $20/V_d$ in place.

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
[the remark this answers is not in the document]
emackey: I'm unsure how to actually add this to the draft.  Can someone write up the exact change(s) wanted here?

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,201 @@
...
+And then:
+
+$$
+A = n_d - \frac{B}{\lambda_d^2}
+$$
+
+And finally:
+
+$$
+n(\lambda) = A + \frac{B}{\lambda^2}
```
bsdorra, line 56: > I'm unsure how to actually add this to the draft. Can someone write up the exact change(s) wanted here?


```suggestion
To calculate the index of refraction at a specific wavelength \( $\lambda$ \), we use the first two terms of Cauchy's equation:

$$
n(\lambda) = A + \frac{B}{\lambda^2}
$$

Given an Abbe number \( $V_d$ \) and the central index of refraction as specified by the ```KHR_materials_ior``` extension (assumed to be at the central wavelength, \( $n_d$ \)), the coefficients are defined as:

$$
B = \frac{n_d - 1}{V_d \left( {\lambda_F^{-2}} - {\lambda_C^{-2}} \right)}
$$

And:

$$
A = n_d - \frac{B}{\lambda_d^2}
$$

By substituting the symbolic wavelength constants $n_d$, $n_F$ and $n_C$ with the values provided above we get:

$$
n(\lambda) = \max(n_d + \frac{n_d-1}{V_d} \left( \frac{523655}{\lambda^2} - 1.5168 \right), 1)

```

emackey: Thanks!

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,208 @@
...
+$$
+A = n_d - \frac{B}{\lambda_d^2}
+$$
+
+By substituting the symbolic wavelength constants $n_d$, $n_F$ and $n_C$ with the values provided above we get:
+
+$$
+n(\lambda) = \max(n_d + \frac{n_d-1}{V_d} \left( \frac{523655}{\lambda^2} - 1.5168 \right), 1)
+
+$$
```
lexaknyazev, line 64: This block does not render on GitHub. Maybe try removing the empty line?

lexaknyazev (changes requested):

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
emackey: Fixed, thanks!

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,207 @@
...
+A = n_d - \frac{B}{\lambda_d^2}
+$$
+
+By substituting the symbolic wavelength constants $n_d$, $n_F$ and $n_C$ with the values provided above we get:
+
+$$
+n(\lambda) = \max(n_d + \frac{n_d-1}{V_d} \left( \frac{523655}{\lambda^2} - 1.5168 \right), 1)
+$$
+
+In this extension, we store a transformed dispersion instead of the Abbe number directly.  Specifically we store $20/V_d$ so that a value of 1.0 is equivalent to $V_d=20$, which is about the lowest Abbe number for normal materials. Values over 1.0 are still valid for artists that want to exaggerate the effect. Decreasing values lower the amount of dispersion down to 0.0.  This is the same transform used by both Adobe Standard Material and OpenPBR.
```
lexaknyazev, line 65: Please put the last sentence into a NOTE block as it's not normative.

```
> [!NOTE]  
> This is the same transform ...
```

emackey: Added.  This didn't display in VSCode's markdown preview, but works fine on GitHub.

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,210 @@
...
+B = \frac{n_d - 1}{V_d \left( {\lambda_F^{-2}} - {\lambda_C^{-2}} \right)}
+$$
+
+And:
+
+$$
+A = n_d - \frac{B}{\lambda_d^2}
+$$
+
+By substituting the symbolic wavelength constants $n_d$, $n_F$ and $n_C$ with the values provided above we get:
```
lexaknyazev, line 59: This should likely say `\lambda_d`, `\lambda_F`, and `\lambda_C`.

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+This extension adds one parameter to the metallic-roughness material: `dispersion`.
+
+`dispersion` enables configuring the strength of the angular separation of colors (chromatic aberration) transmitting through a relatively clear volume.  It is an enhancement to the default ```KHR_materials_volume``` transmission model which assumes no dispersion.
+
+Optical dispersion is represented in terms of the Abbe number parameterization \( $V_d$ \).  The dispersion effect is a result of the wavelength-dependent index of refraction of a material.  Dispersion is a widely adopted parameter in modern PBR models.  It is present in both OpenPBR (as *transmission_dispersion_abbe_number*) and the Dassault Enterprise PBR Shading Model (as $V_d$).
+
+The Abbe number \( $V_d$ \) for visible light is computed from the index of refraction at three wavelengths of Fraunhofer's spectral lines: 486.1 nm (short wavelength blue, $n_F$), 587.6 nm (central wavelength, $n_d$), and 656.3 nm (long wavelength red, $n_C$). The Abbe number makes the simplifying assumption that the index of refraction variance is linear:
+
+$$V_d = \frac{n_d - 1}{n_F - n_C}$$
```
bsdorra, line 39: @lexaknyazev  
- Change wording wrt. Fraunhofer spectral lines
- Define (more precise) lambda symbols 

```suggestion
With the simplifying assumption that the index of refraction variance is linear,
the Abbe number \( $V_d$ \) for visible light is computed as:
$$V_d = \frac{n_d - 1}{n_F - n_C}$$

where $n_F$, $n_d$ and $n_C$ are the indices of refraction at the wavelengths of the Fraunhofer F, d, and C spectral lines: 
$$\lambda_F = 486.13 nm$$
$$\lambda_d = 587.56 nm$$
$$\lambda_C = 656.27 nm$$

```

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,210 @@
...
+And:
+
+$$
+A = n_d - \frac{B}{\lambda_d^2}
+$$
+
+By substituting the symbolic wavelength constants $\lambda_d$, $\lambda_F$, and $\lambda_C$ with the values provided above we get:
+
+$$
+n(\lambda) = \max(n_d + \frac{n_d-1}{V_d} \left( \frac{523655}{\lambda^2} - 1.5168 \right), 1)
```
bsdorra, line 62: @lexaknyazev With the changed precision of the lambdas the numbers should be ok now.

lexaknyazev (approved):

extensions/2.0/Khronos/KHR_materials_dispersion/README.md
```diff
@@ -0,0 +1,218 @@
...
+$$\lambda_C = 656.27 nm$$
+
+
+To calculate the index of refraction at a specific wavelength \( $\lambda$ \), we use the first two terms of Cauchy's equation:
+
+$$
+n(\lambda) = A + \frac{B}{\lambda^2}
+$$
+
+Given an Abbe number \( $V_d$ \) and the central index of refraction as specified by the ```KHR_materials_ior``` extension (assumed to be at the central wavelength, \( $n_d$ \)), the coefficients are defined as:
```
proog128, line 55: `\( $n_d$ \))` -> `\( $n_d$ \)`

lexaknyazev: Why? The closing parenthesis is needed to balance `(assumed`.

emackey: Is there any easy way to avoid nested parenthesis here?  Or should we just leave it?

lexaknyazev: We could just remove all parentheses around in-text variables. It's an editorial choice.

proog128: Right, sorry. Looked awkward at first, but it's correct.

BeRo1985:
I've implemented support of KHR_material_dispersion also in my PasVulkan project now:

Screenshot_20240210_000201

donmccurdy:
Adding read/write/edit support in glTF Transform as well:

- https://github.com/donmccurdy/glTF-Transform/pull/1262

zeux:
gltfpack support:

- https://github.com/zeux/meshoptimizer/pull/660

Issue #2311, opened 2023-08-22
Define conventions and best practices for naming and data layout in glTF extensions


aaronfranke:
When designing a new glTF extension, it would be nice to have guidelines for how to structure the data "the glTF way". These guidelines would be helpful to everyone for both Khronos extensions and vendor extensions.

Let me give an example. Does Khronos recommend using a `"type"` string enum to describe the behavior of an object, or are boolean flags preferred? For example, [glTF cameras](https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_016_Cameras.md) are defined using `"type": "perspective"` or `"type": "orthographic"`.  
Is this the recommended way extensions should define types? What about `"isOrthographic"` or `"isPerspective"` booleans?  
Context: There is ongoing debate on whether physics body motion should have its behavior type defined with a string-based enum like `"type": "dynamic"` or `"type": "kinematic"`, or if it should use an `"isKinematic"` boolean.

Another example, type-specific data. From looking at glTF cameras, it seems like the glTF way is to combine a type enum with sub-JSON, like `{ "type": "perspective", "perspective": { ... } }`. However, I have seen other proposed layouts, including key-only `{ "perspective": { ... } }` and flat `{ "type": "perspective", ... }`. Which is preferred for new extensions?

Yet another example, extension name plurality. Do we want extension names to be singular, plural, or does it depend on the context? The current list of extensions are mixed. For example, we have extensions prefixed with `KHR_materials_` and  `KHR_texture_`. This is a weird inconsistency, why not `KHR_material_` or `KHR_textures_`? There is no difference in usage, I looked at both `KHR_materials_sheen` and `KHR_texture_basisu`, each instance of the extension defines properties for only one thing (a material or a texture). To be clear, I am not suggesting that we rename these existing extensions, but I am suggesting to create guidelines to be consistent for all new extensions. It's better to have a guideline that all extension authors can follow rather than having debates about conventions for each extension.

---

The conventions and best practices for glTF extensions would have the following goals:

* Ensure that the data is laid out in a consistent way between extensions, so that extension implementers are not surprised when implementing a new extension.

* Ensure that data in an extension is designed with a similar level of flexibility as other extensions or glTF itself.

[...]

Issue #2306, opened 2023-07-31
Subdivision surfaces via bezier patches


myaaaaaaaaa:
## Introduction

It's possible to take a model meant for Catmull-Clark subdivision and convert it into [a list of Bezier patches](https://en.wikipedia.org/wiki/B%C3%A9zier_surface#B%C3%A9zier_surfaces_in_computer_graphics).

[The proof-of-concept here](https://github.com/godotengine/godot/pull/79738) shows that it's possible to get results very close to Catmull-Clark this way, even at very low polycounts:

![1690833275](https://github.com/godotengine/godot/assets/103326468/8c9de893-7ff4-47b5-8fd1-b497978266fc)
![1690833137](https://github.com/godotengine/godot/assets/103326468/a01e60e0-7ce8-4c50-bb1d-ad07958b392c)
<sup>GREEN: Original lowpoly quad model ||| YELLOW: Bezier patch control points ||| RED: Bezier patch subdivision surface ||| BLUE: Catmull-Clark subdivision surface, for comparison ||| WHITE: Quad/Bezier patch borders</sup>

![1690833156](https://github.com/godotengine/godot/assets/103326468/63be45a2-1426-42fa-9fcf-b0c05cf72ffa)
<sup>Circle test. Note how Bezier patch subdivision surfaces (RED) need fewer points than Catmull-Clark subdivision surfaces (BLUE) to create rounder circles.<sup>

![1690833235](https://github.com/godotengine/godot/assets/103326468/370c28c2-6664-4a26-8ca0-937c2247e3b1)
<sup>Bevel test. Support loops work the same way for both surfaces.</sup>

## Rationale
Subdivision surfaces are a widely desired feature, see:

* #1362
* #1687
* #1783

The industry-standard algorithm for subdivision surfaces is the [Catmull-Clark method](https://en.wikipedia.org/wiki/Catmull%E2%80%93Clark_subdivision_surface), which has a number of downsides that makes it ill-suited for inclusion into glTF:

* Topology information is necessary since vertices must take a weighted average of nearby vertices. This information must somehow be stored in glTF or rebuilt by importers, adding complexity no matter which option is chosen.
* Supporting auxiliary data such as edge creases would be ideal, but this would require even more information to be stored in glTF and implemented in every renderer.
[...]

PR #2343, opened 2023-11-01
KITTYCAD_boundary_representation

alteous:


extensions/2.0/Vendor/KITTYCAD_boundary_representation/schema/glTF.KITTYCAD_boundary_representation.schema.json
alteous, line 1: This schema was automatically generated. It will likely require some tweaks to match the rest of the spec.

javagl:
The `KITTYCAD` prefix does not (yet) appear in https://github.com/KhronosGroup/glTF/blob/main/extensions/Prefixes.md . You might consider an issue/PR to add it.

Beyond that, there are some issues/discussions that may be related, e.g. https://github.com/KhronosGroup/glTF/issues/2306 or https://github.com/KhronosGroup/glTF/issues/1783 . It might be worthwhile checking these, to see whether there has been any previous discussion or similar proposals.

alteous:
@javagl, noted, the vendor prefix is tentative.

Bézier surfaces are a subset of the more general [NURBS](https://en.wikipedia.org/wiki/Non-uniform_rational_B-spline) curves and surfaces proposed in this extension.

#1783 mentions NURBS and B-reps but makes no proposal. This extension is intended for engineering use cases such as manufacturing and simulation.

alteous:
The ideas in this extension have been developed on a branch of the Rust glTF library: https://github.com/alteous/gltf/blob/brep/gltf-json/src/extensions/kittycad_boundary_representation.rs

This branch also contains the JSON schema generator: https://github.com/alteous/gltf/blob/brep/gltf-json/src/bin/generate_schema.rs

javagl:
> the vendor prefix is tentative.

I created a PR for adding it via https://github.com/KhronosGroup/glTF/pull/2345 , but should emphasize that there currently is no established policy for handling _renamings_ of prefixes. **Iff** there is **no** existing extension with such a prefix, then we could probably handle that, but the goal of the prefixes is to provide some sort of stability/reliability for the actual, final extension name.

Also have a look at the ["Naming" section](https://github.com/KhronosGroup/glTF/tree/main/extensions#naming):

> Names SHOULD be structured as <PREFIX>_<scope>_<feature>, where scope is an existing glTF concept (e.g. mesh, texture, image) and feature describes the functionality being added within that scope. This structure is recommended, but not required.

In this case, it might be something like `KITTYCAD_mesh_brep` or so (I haven't looked at the details of this PR yet).

The main reason for pointing to these issues is the encourage a form of "collaboration", with the goal of finding something that multiple people can _agree_ on. For example (_depending on the maturity of this PR and your own judgement!_) you could consider to `@mention` someone from these issues and point them to this PR - maybe they want to review it and chime in with their thoughts and ideas...

alteous:
> Iff there is no existing extension with such a prefix, then we could probably handle that, but the goal of the prefixes is to provide some sort of stability/reliability for the actual, final extension name.

That's fine. By 'tentative' I was thinking along the lines of "another party might be interested in this" and therefore it could be made `EXT`.

> In this case, it might be something like KITTYCAD_mesh_brep or so (I haven't looked at the details of this PR yet).

That's a good point. The `mesh` scope is related but feels inappropriate. What we're proposing essentially is a new 'solid' scope. So perhaps `X_solid_brep` would be a better name. This extension is related more closely to the proposed [`EXT_manifold`](https://github.com/KhronosGroup/glTF/pull/2286) extension. Is `manifold` is the name of the scope in that case?

javagl:
Leaving the option to change the prefix to `EXT` makes sense. After all, I assume that this PR is still work in progress, and possibly taking into account further feedback. (I cannot give that - I'm not technically involved in the topic - but I'm sure others might be interested in that)

(BTW: One could consider changing this to a 'Draft' _unless_ it is supposed to be 'Ready For Review'. But admittedly, the processes here are not perfectly sorted out. I once started trying to clarify some aspects of the extension development process, but ... the PR at https://github.com/KhronosGroup/glTF/pull/2225 is itself still a 'Draft' 😬 )

prideout:
This spec has several example JSON snippets for individual features, but I think it would be helpful to add a higher level JSON example near the top.  This would help reader who wish to quickly review the overall structure of a Brep.

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+
+```json
+"line": {
+  "start": [-1, -1, -1],
+  "end": [1, 1, 1],
+}
+```
+
+#### Circle
```
prideout, line 97: Would it be more flexible for a curve to be "circular arc" rather than a "circle"?

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
```diff
@@ -0,0 +1,250 @@
...
+  }
+}
+```
+
+#### Plane
+
+```json
+"plane": {
+  "normal": [0, 0, 1],
+  "point": [0, 0, 0]
```
prideout, line 150: This doesn't give the plane an orientation, so it's not obvious how to apply a texture to it.

prideout:
I love that someone is interested in opening up a modern standard for Brep representation!

As it now stands, this spec does not clearly leverage glTF's core strengths like its well-defined materials model, its animation model (skinning/morphing), and the node model (allows efficient mesh instancing and hierarchical transformations).

I think "playing nicely" with all those features are possible, but the spec should probably make it clear how it all works.

alteous:
> I love that someone is interested in opening up a modern standard for Brep representation!

It's great to have another party interested too, thanks for your review! Modernising B-reps is exactly what we're trying to achieve here. Our main use case is for manufacturing. We want to be able to ship an exact boundary representation, an optional tessellation, and manufacturing data in one bundle. The JSON + binary design of glTF allows this data to be stored more densely than existing formats and with potentially more diverse metadata.

> As it now stands, this spec does not clearly leverage glTF's core strengths like its well-defined materials model, its animation model (skinning/morphing), and the node model (allows efficient mesh instancing and hierarchical transformations).

That's a great point. The intention is to integrate well the materials and node model. I'll expand on the documentation there. Solids are meant be to instantiated in the same way as meshes, i.e., they don't appear in the scene unless they are referenced in the node hierarchy. It's up to the application whether they render the pre-tessellated mesh approximation or do something with the B-rep directly.

Animations targeting B-reps are expected to be limited to simple node transformations, i.e., translation, rotation, and scaling.

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
```diff
@@ -0,0 +1,250 @@
...
+  "circle":  {
+    "radius": 1,
+    "origin": [0, 0, 0],
+    "normal": [0, 0, 1],
+    "xbasis": [1, 0, 0]
+  }
+}
+```
+
+#### Torus (revolved circle)
```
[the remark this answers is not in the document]
alteous: Surfaces of revolution and extrusion are planned and in progress at the moment but haven't been added to this PR yet.

This particular section was intended to introduce a type of a parametric surface. Perhaps that could be made clearer.

Now that you mention it, I do wonder if it's worth having distinct types for cylinders and tori at all since they could be simply inferred as circle + extrusion/revolution.

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
alteous: It's a good point. I think this is a matter of naming. All curves have an optional subdomain so they are meant to be flexible in this way already, i.e., a circle curve with a subdomain would be an arc of the circle.

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+Curves are referenced by _edges_ to define the boundary between _faces_ in 3D space. Supplementary curves may be referenced by _loops_ to define the same boundary between _faces_ in 2D _surface_ parameter space.
+
+#### Base structure
+
+```json
+{
+  "type": "curveType",
+  "curveType": {
+    "curveSpecificValue": 1
```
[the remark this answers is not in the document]
alteous: This snippet is meant to convey the general structure of a curve entry. The structure is similar to that of cameras where there is a `type` field set to either `"perspective"` or `"orthographic"` and the camera-specific fields are put under `camera.perspective`/`camera.orthographic` respectively.

https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-camera

On a slightly unrelated note, this kind of sum type has been discussed in https://github.com/KhronosGroup/glTF/issues/2311.

prideout:
@alteous all your responses make sense to me.  My own interest is the architecture sector as opposed to manufacturing, but I think we'd both want the same things from a Brep format.

I look forward to seeing how this develops, let me know if there's anything I can do to help.

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
```diff
@@ -0,0 +1,422 @@
...
+  "numKnots": [6, 4],
+  "knotVector": [3, 3, 3, 4, 4, 4, 0, 0, 2, 2],
+  "order": [3, 2]
+}
+```
+
+## Orientation
+
+Many objects in boundary representation have two _orientation states_. Such objects are called _orientable_ objects. These are often given colloquial terms such as 'right/wrong side', 'up/down', 'in/out', 'forward/backward', et cetera. When referencing orientable objects, it is important to state which orientation of the object is desired. The orientation of a referenced object is described as being 'same-sense' or 'opposite-sense'. This extension utilizes the sign bit of floating point numbers to select the desired orientation. A positive sign selects the 'same-sense' and a negative sign selects the 'opposite-sense'.
+
```
alteous, line 388: In the working group meeting, it was raised that `-0` would likely be parsed as `0` by some JSON parsers. Also, it was raised that the IEEE754 specification says `-0 == 0` should evaluate as true. These are both good points, so I will come up with an alternative.

I have a few ideas:

1. Encode using an object `{ "index": 0, "reverse": true }`. This was actually the original proposal but since this use case is so common, this approach was deemed too verbose/bloated.
2. Use indices starting at 1. This avoids the tricky zeroth index but would be inconsistent with the core specification. I would like to avoid this one if possible.
3. Encode as a string instead, i.e,. `"-0"` instead of `-0`. This could work; however, it was raised this has its own set of problems. For example, glTF JSON permits additional integer representations such as exponential notation.
4. Move this data into the binary blob and encode as a one's complement integer.

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
```diff
@@ -0,0 +1,422 @@
...
+                    "start": 0,
+                    "end": 2,
+                    "curve": -2
+                }
+            ],
+            "curves": [
+                {
+                    "type": "line",
+                    "start": [-1, -1, 0],
+                    "end": [1, -1, 0]
```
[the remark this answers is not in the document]
alteous: Possibly. The idea here is that straight edges are a portion of a line curve. For example, an edge may span from vertex [-1, -1, -1] to vertex [1, 1, 1], but be associated with a line curve starting from [-2, -2, -2] in the direction [1, 1, 1].

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
```diff
@@ -0,0 +1,422 @@
...
+            ],
+            "loops": [
+                {
+                    "edges": [0, 1, -2]
+                }
+            ],
+            "edges": [
+                {
+                    "start": 0,
+                    "end": 1,
```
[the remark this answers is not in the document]
alteous: Edges typically join surfaces with differing surface equations. For example, consider the edge at the top of a closed cylinder which can be represented equivalently in three ways: a circle in 3D space, a circle on a planar surface in 2D space, or a line on a cylindrical surface in 2D space. The point of having edges is to allow all these curves to be provided: a 3D space curve is associated with the edge, and a 2D surface curves may be associated a with loop.

```
{
    "loops": [
        {
            "edges": [0, 1, 2, 3],
            "uvCurves": [0, 1, 2, 3]
        },
        {
            "edges": [0, 1, 2, 3],
            "uvCurves": [4, 5, 6, 7]
        }
    ],
    "edges": [
        {
            "curve": [4],
            ...
        }
    ]
}
```

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
```diff
@@ -0,0 +1,422 @@
...
+
+#### Example
+
+The interior hollow space of a tennis ball would be represented by an inner shell. The surface of the tennis ball would be represented by its outer shell.
+
+![tennis-ball-cross-section](https://content.instructables.com/FRV/UI1L/J2AW02WE/FRVUI1LJ2AW02WE.jpg?auto=webp&frame=1&width=1024&height=1024&fit=bounds&md=da5cf8a96b88acbea9eb082a7b378ae7)
+
+### Shell
+
+A _shell_ is a collection of _faces_ in 3D space which form a 'watertight' volume. A shell is represented by the `Shell` data structure.
```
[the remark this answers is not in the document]
alteous: That's a good point. I'm thinking of reworking many of the data structures in this direction. In particular, I'm considering removing the outer/inner loop distinction in faces, perhaps replacing it with extra metadata in the loop data structure instead. The same idea can be applied to shells.

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
alteous: A couple more ideas:

5. Encode using two integers per array, an index followed by a sign.

```json
{
    "edges": [
        0, 1,
        1, -1,
        2, 1,
        3, -1
    ]
}
```

6. Encode the sign using a separate paired array, which could be omitted if all items are same-sense:

```json
{
    "edges": [0, 1, 2, 3],
    "edgeOrientations": [1, -1, 1, -1]
}
```

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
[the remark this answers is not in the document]
alteous: Perhaps it could be either an indexed vertex or an explicit co-ordinate.

```json
{
    "curves": [
        {
            "type": "line",
            "line": {
                "start": 0,
                "end": 1
            }
        },
        {
            "type": "line",
            "line": {
                "start": [0.0, 0.0, 0.0],
                "end": [1.0, 1.0, 1.0]
            }
        }
    ]
}
```

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
alteous: I've reworked this as an integer/orientation pair. It looks like this:

Before

```json
{
    "loops": [0, -1],
    "surface": 0
}
```

After

```json
{
    "loops": [
        [0, 1],
        [1, -1]
    ],
    "surface": [0, 1]
}
```

`[index, 1]` could be implied by `[index]` but I'll keep it explicit for now.

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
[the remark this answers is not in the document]
alteous: Line curves have been simplified to a single form: an origin plus a normalised direction.

extensions/2.0/Vendor/KITTYCAD_boundary_representation/README.md
alteous: This has been amended now.

elalish:
This is interesting, thanks! I agree with @prideout that it would be nice if we could find a way to represent smoothly curved geometry within the existing data structures of glTF, since it would avoid duplication, which can inevitably result in conflicting data. I have some thoughts and progress on this, though it's a pretty different concept from NURBS, or even from subdivision surfaces - it's more a form of curved triangles. I have a short write-up here: https://elalish.blogspot.com/2022/03/smoothing-triangle-meshes.html - I have some improvements I want to make first, but I'm interested in attempting a glTF extension on top of my [EXT_mesh_manifold](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_mesh_manifold) to enable this. Of course, I'm not sure how helpful it'll be for engines that are already NURBS-based, but I'm curious if anyone else is interested in trying a new approach to B-reps?

One semantic nit: glTF meshes *are* boundary-representations. The difference between triangles and NURBS is first-order vs higher-order, and explicit edges vs implicit edges. I wrote up a taxonomy: https://elalish.blogspot.com/2022/03/solid-geometry-representations.html

PR #2347, opened 2023-11-16
KHR_materials_anisotropy should be exclusive vs unlit

zeux:
This exclusion note is present in every single material extension except for this one. I assume this is an omission since it's not clear why anisotropy specifically would be okay to mix with unlit and every other KHR_materials_ extension isn't.

Note that other material extensions also say it's not okay to mix them with `KHR_materials_pbrSpecularGlossiness`. I did not add this exclusion for now to reduce the scope of the change but let me know if that is preferred here as well.

If we don't end up adding this exclusion I feel like we should explicitly clarify the opposite.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
@@ -31,6 +31,7 @@ Written against the glTF 2.0 spec.
 
 ## Exclusions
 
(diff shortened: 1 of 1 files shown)


lexaknyazev:
Good catch, it's indeed an oversight. Please add the SpecGloss exclusion as well.

zeux:
Added `KHR_materials_pbrSpecularGlossiness` exclusion.

lexaknyazev (approved):

PR #2351, opened 2023-11-30
Fix typo: bufferView.byteLength

### LINKED ISSUE #2283: Typo in property name bufferView.length
In section [3.6.2.4. Data Alignment](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#data-alignment) you write:

> MUST be less than or equal to bufferView.length.

However, `length` is not an existing property of a `bufferView`. I assume you rather mean `bufferView.byteLength`?

emackey:
Fixes #2283.


aaronfranke (approved):

aaronfranke:
This fix is correct: https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/bufferView.schema.json#L19

PR #2366, opened 2024-02-29
Fix Typo: Remove duplicated "not"

takahirox:


javagl:
@lexaknyazev When things affect the core spec, then I'd hesitate to just review/merge anything without your approval (a similar case as in https://github.com/KhronosGroup/glTF/pull/2351 ). If you say that things like these (uncontroversial typos) are OK to be handled by anyone, then these PRs could be worked off more quickly.

emackey (approved):

aaronfranke (approved):

PR #2369, opened 2024-03-18
Add EXT_texture_astc extension

wasimabbas-arm:
Add an extension for native ASTC encoded ktx2 images. Extension only supports LDR, no HDR or 3D images supported yet.


extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,253 @@
...
+  - `supercompressionScheme` MUST be `0` (None).
+  - ASTC HDR vs LDR as well as block sizes can be determined from the vkFormat of the image
+  - DFD `colorModel` MUST be `KHR_DF_MODEL_ASTC`.
+  - DFD `channelId` MUST be `KHR_DF_CHANNEL_ASTC_DATA`.
+
+### Additional requirements
+
+Regardless of the format used, these additional restrictions apply for compatibility reasons:
+
+- Swizzling metadata (`KTXswizzle`) MUST be `rgba` or omitted.
```
wasimabbas-arm, line 172: TBC, where we need this meta data or not.

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds the ability to specify textures using KTX v2 images with (ASTC compression)[https://github.com/ARM-software/astc-encoder/blob/main/Docs/FormatOverview.md]. An implementation of this extension can use such images as an alternative to the PNG or JPEG images available in glTF 2.0 for more efficient asset transmission, reducing GPU memory footprint, higher texture throughput and faster execution.
+
+When this extension is used, it's allowed to use value `image/astc` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_astc` texture extension object.
```
janharaldfredriksen-arm, line 30: The example below uses mimeType `image/ktx2`. Is that intentional?

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+- **RGBA:** A texture that uses all four channels. Such textures MAY be encoded with the sRGB transfer function (not applied to Alpha).
+  > **Note:** RGBA textures from the core glTF 2.0 specification include:
+  > - `baseColorTexture` (non-opaque case)
+
+- **Red:** A texture that uses only red channel. All other channels are unused and their values are not sampled at runtime. Such textures MUST NOT be encoded with the sRGB transfer function.
+  > **Note:** Red textures from the core glTF 2.0 specification include:
+  > - `occlusionTexture` (standalone)
+
+- **Luminance-Alpha:** A texture that uses only Luminance and Alpha channels. To sample these textures use `.ga` swizzle in your shaders. Such textures MUST NOT be encoded with sRGB transfer function.
```
janharaldfredriksen-arm, line 159: Do we need Luminance-Alpha types here?
(I don't know what other glTF extensions do, but if Red-Green was sufficient that'd be better.)

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,253 @@
...
+  > **Note:** The core glTF 2.0 specification has no examples of luminance-alpha textures.
+
+### KTX header fields for ASTC payloads
+  - `supercompressionScheme` MUST be `0` (None).
+  - ASTC HDR vs LDR as well as block sizes can be determined from the vkFormat of the image
```
janharaldfredriksen-arm, line 164: nit: vkFormat => VkFormat
(in a few other places)

extensions/2.0/Khronos/KHR_texture_astc/schema/texture.KHR_texture_basisu.schema.json
```diff
@@ -0,0 +1,15 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_texture_astc glTF Texture Extension",
+    "type": "object",
+    "description": "glTF extension to specify textures using the KTX v2 images with astc compression.",
```
janharaldfredriksen-arm, line 5: nit: astc => ASTC

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Yes, the extension will only allow `image/ktx2`. `image/astc` is a typo however.

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds the ability to specify textures using KTX v2 images with (ASTC compression)[https://github.com/ARM-software/astc-encoder/blob/main/Docs/FormatOverview.md]. An implementation of this extension can use such images as an alternative to the PNG or JPEG images available in glTF 2.0 for more efficient asset transmission, reducing GPU memory footprint, higher texture throughput and faster execution.
```
MarkCallow, line 28: "using ASTC compressed images in KTX v2 containers" would be more accurate.

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+When this extension is used, it's allowed to use value `image/astc` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_astc` texture extension object.
+
+At runtime, engines can directly use the ktx images unlike EXT_texture_basisu which requires doing transcoding to other compressed formats. This also means you get support for all ASTC block formats and quality presets.
```
MarkCallow, line 32: s/doing//

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+## glTF Schema Updates
+
+The `KHR_texture_astc` extension is added to the `textures` object and specifies a `source` property that points to the index of the `image` which defines a reference to the KTX v2 image with ASTC compression.
```
MarkCallow, line 36: "KTX v2 file with the ASTC compressed image."

extensions/2.0/Khronos/KHR_texture_astc/README.md
MarkCallow, line 159: Why must it not be encoded with sRGB? It is quite normal to encode luminance in sRGB?

extensions/2.0/Khronos/KHR_texture_astc/README.md
MarkCallow: The field in the KTX v2 header is called "vkFormat". So here "value of vkFormat" would be correct. I haven't looked at the other instances of its use.

extensions/2.0/Khronos/KHR_texture_astc/README.md
MarkCallow: I don't think  glTF uses swizzle anywhere so probably not needed.

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,253 @@
...
+- Orientation metadata (`KTXorientation`) MUST be `rd` or omitted.
+- Color space information in the DFD MUST match the expected usage, namely:
+  - For textures with **color data** (e.g., base color maps),
+    - `colorPrimaries` MUST be `KHR_DF_PRIMARIES_BT709`;
+    - `transferFunction` MUST be `KHR_DF_TRANSFER_SRGB`.
+  - For textures with **non-color data** (e.g., normal maps),
+    - `colorPrimaries` MUST be `KHR_DF_PRIMARIES_UNSPECIFIED`;
+    - `transferFunction` MUST be `KHR_DF_TRANSFER_LINEAR`.
+- When a texture refers to a sampler with mipmap minification, the KTX image MUST contain a full mip pyramid.
+- When a texture referencing a KTX v2 image with ASTC compression is used for glTF 2.0 material maps (both color and non-color), the KTX v2 image MUST be of **2D** type as defined in the KTX v2 Specification, Section 4.1.
```
MarkCallow, line 182: "referencing a KTX v2 file with ASTC compressed image"

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,253 @@
...
+}
+```
+
+### JSON Schema
+
+[texture.KHR_texture_astc.schema.json](schema/texture.KHR_texture_astc.schema.json)
+
+## KTX v2 Images with Astc compression 
+
+To cover a broad range of use cases, this extension allows different ASTC block sizes as well as LDR, HDR and 3D images. These can be determined from the vkFormat of the KTX image.
```
MarkCallow, line 140: Given the fragmented nature of ASTC implementations you may need to have different extensions for LDR only, LDR+HDR and 3D.

MarkCallow (commented):
Looks good but I do think you'll need to address the fragmentation of ASTC implementations into LDR-only, LDR+HDR and LDR+HDR+3D possibly by having multiple extensions. I'm no glTF expert but perhaps you can just have different extension names in this single extension document. That's what the GL extensions did IIRC.

extensions/2.0/Khronos/KHR_texture_astc/README.md
abbaswasim: There is support for L+A in the encoder so it makes sense for gltf to have support too. Although my note below

`**Note:** The core glTF 2.0 specification has no examples of luminance-alpha textures.`

So maybe we remove it.

abbaswasim:
> Looks good but I do think you'll need to address the fragmentation of ASTC implementations into LDR-only, LDR+HDR and LDR+HDR+3D possibly by having multiple extensions. I'm no glTF expert but perhaps you can just have different extension names in this single extension document. That's what the GL extensions did IIRC.

I am not sure why we need to make the distinction? If all of these can be specified from the vkFormat then why have multiple extensions? 

We probably need to add a bit more details of what other texture types could be supported and the formats like RGBM or RGB9E5 etc.

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/94e429323ad4c901cc31a2d9deb56e18d718fe21

extensions/2.0/Khronos/KHR_texture_astc/schema/texture.KHR_texture_basisu.schema.json
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/94e429323ad4c901cc31a2d9deb56e18d718fe21

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/b7b6dd2397b7df15f740e40265650c74abb0cb77

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/b7b6dd2397b7df15f740e40265650c74abb0cb77

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/b7b6dd2397b7df15f740e40265650c74abb0cb77

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/b7b6dd2397b7df15f740e40265650c74abb0cb77

lexaknyazev (changes requested):

MarkCallow:
> I am not sure why we need to make the distinction? If all of these can be specified from the vkFormat then why have multiple extensions?

Because a glTF implementation says it supports the functionality by using it. If not supported it falls back to the alternative given in the glTF file. So if the implementation starts using the extension because it "supports ASTC" then finds it is an HDR image and the graphics API below it does not support HDR that's a problem. It may be too late to fall back.

@lexaknyazev any comment on this?

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,261 @@
...
+- **RGBA:** A texture that uses all four channels. Such textures MAY be encoded with the sRGB transfer function (not applied to Alpha).
+  > **Note:** RGBA textures from the core glTF 2.0 specification include:
+  > - `baseColorTexture` (non-opaque case)
+
+- **Red:** A texture that uses only red channel. All other channels are unused and their values are not sampled at runtime. Such textures MUST NOT be encoded with the sRGB transfer function.
+  > **Note:** Red textures from the core glTF 2.0 specification include:
+  > - `occlusionTexture` (standalone)
+
+- **Luminance-Alpha:** A texture that uses only Luminance and Alpha channels. To sample these textures use `.ga` swizzle in your shaders. Such textures MAY be encoded with sRGB transfer function.
+  > **Note:** The core glTF 2.0 specification has no examples of luminance-alpha textures.
```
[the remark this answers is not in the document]
wasimabbas-arm: @lexaknyazev Thanks for your contribution.

Can I just confirm you mean to remove the swizzle bits? or the whole section? I can remove the `.ga` swizzle requirements since it can be any of `.ra`, `.ga` or `.ba`. What I would like is to have some way to represent an `RG` texture type which is usually a Luminance-Alpha texture.

There is already precendence for this in the KHR_texture_basisu and I am tempted to just put that back in.

```
- **Red-Green:** A texture that uses only red and green channels. Blue and Alpha channels are unused and their values are not sampled at runtime. Such textures MUST NOT be encoded with sRGB transfer function.
  > **Note:** The core glTF 2.0 specification has no examples of red-green textures.
```

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+- Orientation metadata (`KTXorientation`) MUST be `rd` or omitted.
+- Color space information in the DFD MUST match the expected usage, namely:
+  - For textures with **color data** (e.g., base color maps),
+    - `colorPrimaries` MUST be `KHR_DF_PRIMARIES_BT709`;
+    - `transferFunction` MUST be `KHR_DF_TRANSFER_SRGB`.
+  - For textures with **non-color data** (e.g., normal maps),
+    - `colorPrimaries` MUST be `KHR_DF_PRIMARIES_UNSPECIFIED`;
+    - `transferFunction` MUST be `KHR_DF_TRANSFER_LINEAR`.
+- `pixelWidth` and `pixelHeight` MUST be multiples of 4.
```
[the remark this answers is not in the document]
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/3a34159967615d086a76dd033f6ba7d83f625e18

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,261 @@
...
+If the texture type is `normalTexture` the output will be a two component X+Y normal map stored as (RGB=X, A=Y). The Z component can be recovered programmatically in shader code by using the equation:
+
+```GLSL
+    nml.xy = texture(...).ga;              // Load in [0,1]
+    nml.xy = nml.xy * 2.0 - 1.0;           // Unpack to [-1,1]
+    nml.z = sqrt(1 - dot(nml.xy, nml.xy)); // Compute Z
+```
+
+### KTX header fields for ASTC payloads
+  - `supercompressionScheme` MUST be `0` (None).
```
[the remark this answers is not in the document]
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/3a34159967615d086a76dd033f6ba7d83f625e18

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds the ability to specify textures using (ASTC compressed images)[https://github.com/ARM-software/astc-encoder/blob/main/Docs/FormatOverview.md] in KTX v2 containers. An implementation of this extension can use such images as an alternative to the PNG or JPEG images available in glTF 2.0 for more efficient asset transmission, reducing GPU memory footprint, higher texture throughput and faster execution.
+
+When this extension is used, it's allowed to use value `image/ktx2` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_astc` texture extension object.
+
+At runtime, engines can directly use the ktx images unlike EXT_texture_basisu which requires transcoding to other compressed formats. This also means you get support for all ASTC block formats and quality presets.
```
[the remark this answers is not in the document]
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/3a34159967615d086a76dd033f6ba7d83f625e18

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: No sure what you mean by texture slots. If one reads the texture format and works out its an HDR image, the shader code for sampling that texture can be what's required. The hardware can deal with sampling this in any space.

You will always have special use of HDR images and special code/shader code to deal with it.

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,261 @@
...
+  > - `baseColorTexture` (non-opaque case)
+
+- **Red:** A texture that uses only red channel. All other channels are unused and their values are not sampled at runtime. Such textures MUST NOT be encoded with the sRGB transfer function.
+  > **Note:** Red textures from the core glTF 2.0 specification include:
+  > - `occlusionTexture` (standalone)
+
+- **Luminance-Alpha:** A texture that uses only Luminance and Alpha channels. To sample these textures use `.ga` swizzle in your shaders. Such textures MAY be encoded with sRGB transfer function.
+  > **Note:** The core glTF 2.0 specification has no examples of luminance-alpha textures.
+
+If the texture type is `normalTexture` the output will be a two component X+Y normal map stored as (RGB=X, A=Y). The Z component can be recovered programmatically in shader code by using the equation:
```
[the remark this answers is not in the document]
wasimabbas-arm: Won't this extension be that extension that adds support for two component normal maps? Similar argument to the HDR one, once you work out you have an ASTC `normalTexture` you will have to provide the right sampling code in your shaders.

extensions/2.0/Khronos/KHR_texture_astc/README.md
lexaknyazev: The logical model of glTF goes like this:

1. An `image` is just a resource reference.
2. A `texture` is an approximation of a 2D GPU texture object so it refers to an `image` (called `source` for historical reasons) and a sampler state (wrapping and filtering modes). To ensure ecosystem consistency, unextended glTF textures may only refer to JPEG/PNG images. A texture extension usually provides an alternative representation of a JPEG/PNG image source therefore that alternative source must not contain anything that cannot be represented as a regular JPEG/PNG image.
3. Many material properties, e.g., PBR base color, are texturable. This means that a property like `baseColorTexture` points to a texture object that provides ready-to-use per-texel values for that property. All such properties expect texel values in the normalized `[0, 1]` range and they are not aware of any texture encoding details.

All of this means that a texture object that refers to an image with HDR values cannot be used for providing texel data for glTF material properties.

extensions/2.0/Khronos/KHR_texture_astc/README.md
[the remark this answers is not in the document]
lexaknyazev: The Basis extension explicitly mentions Red and Red-Green textures because Basis ETC1S codec needs special handling during transcoding of such textures.

Since ASTC decoder always provides defined values for all four channels regardless of the compressed data, all of these subsections (RGB, RGBA, ...) are not needed for ASTC.

lexaknyazev:
> Because a glTF implementation says it supports the functionality by using it. If not supported it falls back to the alternative given in the glTF file. So if the implementation starts using the extension because it "supports ASTC" then finds it is an HDR image and the graphics API below it does not support HDR that's a problem. It may be too late to fall back.

If the goal of this extension is to provide a drop-in replacement for JPEG/PNG images in the existing glTF ecosystem, then only LDR payloads should be supported.

Besides, non-LDR textures cannot be used with the existing material properties in principle (see another comment above).

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Thanks for explaining this. 

> Many material properties, e.g., PBR base color, are texturable. 

I understand the glTF workflow and it makes perfect sense if you just think about using HDR images for PBR materials however many uses of HDR/float textures in graphics are not about storing color/PBR material properties like albedo/normal etc directly but rather things like environment maps or other image base lighting parameters. This is for that specific usecase.

> therefore that alternative source must not contain anything that cannot be represented as a regular JPEG/PNG image

And I don't think I am proposing that here. You can still have an alternative non-HDR JPEG/PNG image for your HDR texture target.

extensions/2.0/Khronos/KHR_texture_astc/README.md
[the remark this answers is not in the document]
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/8781d2296877799387f293c4ebc3da6bd0b9a254

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/8781d2296877799387f293c4ebc3da6bd0b9a254

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Contributors
+
+- Wasim Abbas [@wasimabbas-arm](https://github.com/wasimabbas-arm)
+- Jan-Harald Fredriksen [@janharaldfredriksen-arm](https://github.com/janharaldfredriksen-arm)
+- Peter Harris [@solidpixel](https://github.com/solidpixel)
+- Mark Callow [@MarkCallow](https://github.com/MarkCallow)
+- Alexey Knyazev [@lexaknyazev](https://github.com/lexaknyazev)
+
+Copyright 2023 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
```
lexaknyazev, line 16: Please replace this line with:
> Copyright 2024 The Khronos Group Inc.

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,240 @@
...
+        }
+    ]
+}
+```
+
+### JSON Schema
+
+[texture.KHR_texture_astc.schema.json](schema/texture.KHR_texture_astc.schema.json)
+
+## KTX v2 files with ASTC compression
```
lexaknyazev, line 139: Please update this section to reflect https://github.com/KhronosGroup/glTF/pull/2369#discussion_r1848786803.

wasimabbas-arm: > Please update this section to reflect [#2369 (comment)](https://github.com/KhronosGroup/glTF/pull/2369#discussion_r1848786803).

@lexaknyazev Apologies there is a typo in that comment. 

> Agree, I realised specifiying normal maps themselves are a problem at the moment in glTF. I will remove HDR references from the extension for now.

What I meant to say was 

> Agree, I realised specifiying ~normal~**environment** maps themselves are a problem at the moment in glTF. I will remove HDR references from the extension for now.

In https://github.com/KhronosGroup/glTF/commit/8781d2296877799387f293c4ebc3da6bd0b9a254 I have not removed mentions of HDR and 3D but mentioned that these are not supported instead.

What's left in that section is the normal map stuff, which I didn't remove yet because waiting for your answer on https://github.com/KhronosGroup/glTF/pull/2369#discussion_r1848253882

The 2-component Normal maps are an important ASTC feature. To get the most quality for your normal maps from high-compression you need this. There is some other normal map specific optimisations in the encoder as well that you don't want to lose by just considering it like any other RGB color texture.

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/cac347078219383aa212796fc29b346e4d3e39be

extensions/2.0/Khronos/KHR_texture_astc/README.md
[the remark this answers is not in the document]
lexaknyazev: Neither `image` nor `texture` glTF objects are aware of their usage as the same texture could technically be referenced from multiple material parameters (ignoring the fact that it does not make much sense for normal maps). Material shaders generally do not know anything extension-specific about texture encoding - they have a fixed interpretation of all channels.

Therefore, a concept of a "two-channel normal map" needs to be introduced as a material extension first (format-agnostic) and then that normal map could be provided by an ASTC texture.

This means that:
- This texture extension does not need to say anything about normal maps at all.
- A new material extension defining two-channel normal maps should be drafted someday. For compatibility reasons, such an extension should extend the `normalTextureInfo` objects by providing an alternative texture reference there.

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds the ability to specify textures using (ASTC compressed images)[https://github.com/ARM-software/astc-encoder/blob/main/Docs/FormatOverview.md] in KTX v2 containers. An implementation of this extension can use such images as an alternative to the PNG or JPEG images available in glTF 2.0 for more efficient asset transmission, reducing GPU memory footprint, higher texture throughput and faster execution.
+
+When this extension is used, it's allowed to use value `image/ktx2` for the `mimeType` property of images that are referenced by the `source` property of `KHR_texture_astc` texture extension object.
+
+At runtime, engines can directly use the ktx images unlike KHR_texture_basisu which requires transcoding to other compressed formats. This also means you get support for all ASTC block formats and quality presets.
```
lexaknyazev, line 33: We should not directly refer to Basis here. Suggested replacement:
> When running on supported platforms, engines can directly upload ASTC data without transcoding it to other compressed formats. This means that all ASTC block sizes and quality presets are supported unlike with other texture extensions that focus on portability of compressed texture payloads.

> Platforms that do not have hardware ASTC support would have to either use other texture sources or decompress ASTC data manually.

extensions/2.0/Khronos/KHR_texture_astc/schema/texture.KHR_texture_basisu.schema.json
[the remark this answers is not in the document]
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/b5717f9ffc6285701a87791f68f42d9d47712430

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,232 @@
...
+- When a texture referencing a KTX v2 file with ASTC compressed image is used for glTF 2.0 material maps (both color and non-color), the KTX v2 image MUST be of **2D** type as defined in the KTX v2 Specification, Section 4.1.
+- `KHR_DF_FLAG_ALPHA_PREMULTIPLIED` flag MUST NOT be set unless the material's specification requires premultiplied alpha.
+
+## Known Implementations
+
+Authoring:
+
+- [KTX Reference Software](https://github.com/KhronosGroup/KTX-Software/)
+
+Viewing:
```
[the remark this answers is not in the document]
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/b5717f9ffc6285701a87791f68f42d9d47712430

extensions/2.0/Khronos/KHR_texture_astc/README.md
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/b5717f9ffc6285701a87791f68f42d9d47712430

lexaknyazev (approved):

extensions/2.0/Khronos/KHR_texture_astc/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension adds the ability to specify textures using (ASTC compressed images)[https://github.com/ARM-software/astc-encoder/blob/main/Docs/FormatOverview.md] in KTX v2 containers. An implementation of this extension can use such images as an alternative to the PNG or JPEG images available in glTF 2.0 for more efficient asset transmission, reducing GPU memory footprint, higher texture throughput and faster execution.
```
lexaknyazev, line 29: Incorrect link syntax: should be `[title](link)`.

abbaswasim: Good catch, Fixed https://github.com/KhronosGroup/glTF/pull/2369/commits/b79c60a407e3971d3f5dac4b8a9cfbf93b427834

extensions/2.0/Vendor/EXT_texture_astc/schema/texture.KHR_texture_astc.schema.json
lexaknyazev, line 1: The file should be renamed as well.

extensions/2.0/Vendor/EXT_texture_astc/README.md
```diff
@@ -0,0 +1,236 @@
+<!--
+Copyright 2024 The Khronos Group Inc.
```
emackey: 2025

extensions/2.0/Vendor/EXT_texture_astc/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Contributors
+
+- Wasim Abbas [@wasimabbas-arm](https://github.com/wasimabbas-arm)
+- Jan-Harald Fredriksen [@janharaldfredriksen-arm](https://github.com/janharaldfredriksen-arm)
+- Peter Harris [@solidpixel](https://github.com/solidpixel)
+- Mark Callow [@MarkCallow](https://github.com/MarkCallow)
+- Alexey Knyazev [@lexaknyazev](https://github.com/lexaknyazev)
+
+Copyright 2024 The Khronos Group Inc.
```
emackey: 2025

extensions/2.0/Vendor/EXT_texture_astc/README.md
```diff
@@ -0,0 +1,236 @@
...
+
+[KTX File Format Specification, version 2](https://github.khronos.org/KTX-Specification/)
+
+[KTX Reference Software](https://github.com/KhronosGroup/KTX-Software/)
+
+[ASTC encoder](https://github.com/ARM-software/astc-encoder)
+
+## Appendix: Full Khronos Copyright Statement
+
+Copyright 2023 The Khronos Group Inc.
```
emackey: 2025

extensions/2.0/Vendor/EXT_texture_astc/README.md
```diff
@@ -0,0 +1,236 @@
...
+## Known Implementations
+
+Authoring:
+
+TBA
+
+
+Viewing:
+
+TBA
```
emackey: Do we want to merge this with remaining TBA messages? We could delete this section if the information is unknown.

emackey (commented):
Looks good.  Minor nitpicks here.

extensions/2.0/Vendor/EXT_texture_astc/README.md
wasimabbas-arm: I think better remove it. When we know of implementations we can add the section back in.

abbaswasim: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/d1145f8f08440ae80ac407156d328d987c746557

extensions/2.0/Vendor/EXT_texture_astc/README.md
abbaswasim: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/d1145f8f08440ae80ac407156d328d987c746557

extensions/2.0/Vendor/EXT_texture_astc/README.md
abbaswasim: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/d1145f8f08440ae80ac407156d328d987c746557

extensions/2.0/Vendor/EXT_texture_astc/README.md
abbaswasim: Fixed in https://github.com/KhronosGroup/glTF/pull/2369/commits/d1145f8f08440ae80ac407156d328d987c746557

emackey (approved):

MarkCallow:
Congratulations @wasimabbas-arm.

abbaswasim:
> Congratulations @wasimabbas-arm.

Thanks @MarkCallow and thank you for your contribution as well.

Issue #1986, opened 2021-05-28
Create a specification for primitive geometric shapes (boxes, spheres, etc)


aaronfranke:
I'm working on Godot Engine and I would like the ability to specify primitive geometric shapes in a GLTF file for the purpose of being read and used by a physics system and by other systems such as audio. 

I have seen that #1135 exists, and it has received some negative feedback. It's a huge proposal which includes things like physics materials, joint constraints, gravity, and more. Really the full extent of that proposal is overkill for what I want, so I am opening this proposal to hopefully get something simpler added to the specification.

Godot has a GLTF exporter, so I will try to be as specific as possible as to what the output currently looks like in the `.gltf` file and what the desired output would be if this specification was added.

Here is a screenshot of an example scene I am working with (`core.tscn` from the TPS demo):

![Screenshot from 2021-05-28 05-34-19](https://user-images.githubusercontent.com/1646875/119964867-a37fb080-bf77-11eb-8a15-942c72a9c2fc.png)

And here is that same screenshot but with the meshes hidden, so that only the primitive colliders are visible:

![Screenshot from 2021-05-28 05-34-25](https://user-images.githubusercontent.com/1646875/119964883-a67aa100-bf77-11eb-8b05-d0a41192a83e.png)

When exporting, the collision primitives easily visible in the second screenshot are not preserved. However, each of the collision shapes do get exported with a name, rotation, and translation, just with no shape defined. It looks like this in the `.gltf`:

```gltf
    {
      "extensions": {},
      "name": "MainFloor2",
      "rotation": [0, -0.078459, 0, 0.996917],
      "translation": [21.6304, -14.5, 3.42592]
    },
```

If allowed by the spec, the information about primitive shapes could look something like this (not sure if it belongs in "extensions", but I'm not sure where else it would go):

```gltf
    {
      "extensions": {
        "primitive_box": { "extents": [5.5, 2.1, 2] }
      },
      "name": "MainFloor2",
      "rotation": [0, -0.078459, 0, 0.996917],
      "translation": [21.6304, -14.5, 3.42592]
    },
```

Here is an overview of what the proposed API could look like.

* Box
  * Suggested name: `primitive_box`
  * Has `extents` member of type Vector3 (three numbers), which specify the "radius" of the box, so `extents` of `[1, 1, 1]` means a 2x2x2 cube.
[...]

Issue #1135, opened 2017-10-25
Create a specification for rigid body physics


jratcliff63367:
A number of people and companies have expressed an interest in having a GLTF extension which can describe rigid body physics.

A number of popular physics engines are widely used, each with a differing API and data serialization format.  While many concepts between them are the same, there also tends to be significant differences between them as well.

Past efforts to create a standardized physics data definition have ended up with something so watered down that it was of questionable value, or so highly specific to one physics engine, that it could not be easily used by others.

I see the goal of this effort to create a data spec which represents a 'union' of the features of a number of popular physics engines; such that those things which are in common between them use a shared DOM (data object model) and those properties which are unique to each can still be specified and, conceivably, ignored by other physics engines which do not need them.

Even just the concept of a 'physics engine' can be challenging, as the definition can be wide ranging in scope.  Some physics engines support soft bodies, particle systems, fluid dynamics, complex machines, and more.

While it would be desirable to support all features for all physics engines, that is probably unrealistic as an initial goal.

I would like to set the phase one goal to simply defining the basic components shared by any rigid body dynamics system; which is largely collision geometries, physical material properties, simulation settings, rigid body settings, collision filtering, and joints.

We should probably identify features that go beyond this, but leave them for future extensions efforts.  Future work should also address the specific needs of robots which include cameras, sensors, grippers, transmissions, and a lot complex machine data.

Typically a DOM for a physics engine is similar to the scene graph you would see for rendering data; there are meshes and geometries which can be instanced at different locations and scale and they can be linked together with joints.  There may need to be a way to specify a named binding between a physics representation and the visual representation it is associated with.  Likely a parallel scene graph which simply describes those bindings will be needed.

The basic components of any rigid body system can be described in the following outline:

* Simulation properties
  * Each physics engine defines certain global properties which control how that system performs.  Many of these properties tend to be unique for each different physics engine supported.  Includes things like gravity, solver iteration count, and other global settings.
* Data Resources (objects with large amounts of binary data)
  * Triangle meshes 
  * Convex Hulls
  * Heightfields
* Physics Materials
  * Defines physical properties of a material surface, friction, etc.
* Geometries (physical geometric representations for rigid bodies or parts of a compound rigid body)
  * Plane
  * Box
  * Sphere
  * Capsule
  * Cylinder
  * Triangle Mesh
  * Convex Hull
  * Heightfield
* Rigid Body (can be static, dynamic, or kinematic)
[...]

PR #2370, opened 2024-03-19
KHR_implicit_shapes extension Draft Proposal

eoineoineoin:
Draft of an extension which describes generic, non-renderable shapes suitable for collision detection.

## Known Implementations

[Blender importer/exporter](https://github.com/eoineoineoin/glTF_Physics_Blender_Exporter)
[Babylon.js importer](https://github.com/eoineoineoin/glTF_Physics_Babylon)
[Godot importer](https://github.com/eoineoineoin/glTF_Physics_Godot_Importer)

## Sample Assets

https://github.com/eoineoineoin/glTF_Physics/tree/master/samples


lexaknyazev (changes requested):

extensions/2.0/Khronos/KHR_collision_shapes/schema/glTF.KHR_collision_shapes.shape.convex.schema.json
```diff
@@ -0,0 +1,30 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_collision_shapes Convex Shape",
+    "type": "object",
+    "description": "Parameters describing a convex shape.",
```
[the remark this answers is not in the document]
eoineoineoin: Only concern here is that there's slightly different requirements for the `mode` of the `mesh.primitive` objects in the referenced mesh. I'd expect trimesh objects to require `TRIANGLES`, `TRIANGLE_STRIP` or `TRIANGLE_FAN`, though a convex hull could use any `mode`.

If we do want to combine them, would you agree that the `isConvex` flag should *not* be a mutable property?

extensions/2.0/Khronos/KHR_collision_shapes/schema/glTF.KHR_collision_shapes.shape.box.schema.json
```diff
@@ -0,0 +1,22 @@
...
+    "type": "object",
+    "description": "Parameters describing a box shape.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "size": {
+            "type": "array",
+            "description": "The extents of the box in each axis in local space.",
+            "items": {
+                "type": "number",
+                "minimum": 0.0
```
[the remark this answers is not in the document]
eoineoineoin: Yeah, though engines tend to support this, it's probably not a useful construct. I've converted properties to exclusiveMinimum where it makes sense; though left the radii properties on capsules/cylinders alone - consider a cone, which has one radius of 0.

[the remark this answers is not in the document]
lexaknyazev: Should it be disallowed to have both radii as zeros? Although it's not easily representable with schemas, the spec may have such normative language.

extensions/2.0/Khronos/KHR_collision_shapes/schema/glTF.KHR_collision_shapes.shape.convex.schema.json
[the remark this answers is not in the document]
lexaknyazev: Please add explicit trimesh reqs to the extension spec and to the `mesh` property schema description (also change "ID" to "index" there). The only current mention seems to be the word "triangulated" in the trimesh description, which is not strict enough.

Surely, the flag should be read only.

extensions/2.0/Khronos/KHR_collision_shapes/schema/glTF.KHR_collision_shapes.shape.box.schema.json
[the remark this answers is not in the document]
eoineoineoin: OK, I've added language clarifying degenerate shapes (https://github.com/KhronosGroup/glTF/pull/2370/commits/ef428bc842ca14bad7c64426aba7f389d54390f1) - one concern I have though, is how this requirement interacts with node scale; in particular, the third "Implementation Note" in section https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#transformations which allows nodes to be "optimized away" if they have zero scale.

Similarly, I'd _like_ if negative node scales were supported. i.e. a sphere attached to a node with scale [-1, -1, -1] should be identical to the sphere attached to a node with a scale of [1, 1, 1].

Have you any objection if we were to add text describing how to handle those cases?

extensions/2.0/Khronos/KHR_collision_shapes/schema/glTF.KHR_collision_shapes.shape.mesh.schema.json
```diff
@@ -0,0 +1,36 @@
...
+            "minItems": 1,
+            "items": {
+                "type": "number"
+            }
+        },
+        "extensions": { },
+        "extras": { }
+    },
+    "required": [
+        "convex",
```
lexaknyazev, line 33: Not a strong opinion but would it be reasonable to make the property not required with the `false` being the default value?

extensions/2.0/Khronos/KHR_collision_shapes/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Degenerate shapes are prohibited. A `sphere` must have a positive, non-zero radius. `box` shapes must have positive non-zero values for each component of `size`. `cylinder` and `capsule` shapes must have a positive, non-zero `height` and both `radiusTop` and `radiusBottom` should be positive; at least one of the radii should be non-zero. For `mesh` shapes whose `convex` property is `false`, the referenced mesh must contain at least one non-degenerate triangle primitive. For `mesh` shapes whose `convex` property is `true`, the referenced mesh must contain contain primitives with at least four non-coplanar vertices.
```
lexaknyazev, line 46: Please do not use backticks unless explicitly referring to properties. For example:
> A sphere must...

> Box shapes must...

EricGriffith:
1. It would be great to include example usage of the schemas in the spec.

2. **both** seems to be a typo here:
> As both the `mesh` shape

3. The `convex` property name is confusing. To me, it implies that the mesh is required to be convex. I prefer the previous version where convex and mesh were separate shapes. If momentum means it they must stay combined, please use a more descriptive name like `convexHull`.

4. The `skin` and `weights` properties should probably be disallowed when a mesh shape is used together with a node that uses the same mesh for rendering. Otherwise, every extension using these shapes will need to repeat the same normative text explaining the precedence of the two (node + shape) skin properties and the three (node + mesh + shape) weights properties.

extensions/2.0/Khronos/KHR_collision_shapes/schema/glTF.KHR_collision_shapes.shape.mesh.schema.json
```diff
@@ -0,0 +1,36 @@
...
+            "allOf": [ { "$ref": "glTFid.schema.json" } ]
+        },
+        "weights": {
+            "type": "array",
+            "description": "The weights of the instantiated morph target.",
+            "minItems": 1,
+            "items": {
+                "type": "number"
+            }
+        },
```
aaronfranke, line 29: I'm confused, why are skins and weights used for collision meshes? Where was the discussion on this? What is the use case?

aaronfranke (commented):
The contents of this PR do not match the repo in which the extension was developed: https://github.com/eoineoineoin/glTF_Physics/tree/master/extensions/2.0/Khronos/KHR_collision_shapes/schema

If changes are needed, the original repo should also be updated so that it stays in sync.

npolys:
note that in X3D

Collision for navigation can be a property on Grouping Node types

https://web3d.org/documents/specifications/19775-1/V4.0/Part01/components/navigation.html#Collision

and then there is a whole Rigid Body Physics Component for other object-object collisions
https://web3d.org/documents/specifications/19775-1/V4.0/Part01/components/rigidBodyPhysics.html
________________________________

extensions/2.0/Khronos/KHR_collision_shapes/schema/glTF.KHR_collision_shapes.shape.mesh.schema.json
eoineoineoin: This was discussed in several Khronos meetings on the interactivity spec. These parameters match functionality already in the glTF node. The usecase is to allow users to describe collision shapes which accurately match the render meshes.

lyuma: > collision shapes which accurately match the render meshes

I could see weights/skins being calculated once at load time, but since "interactivity" is brought up, it sounds like the intent is to permit changing vertices on any given frame. I cannot find a good specification for the intended behavior when vertices of a collision mesh are animated at runtime (e.g. on every frame). I see two big problems:

1. Firstly, it seems highly unperformant: both the need to regenerate optimized collision geometry (such as a convex hull or deduplicated trimesh vertices); and to run skinning on the CPU (some engines upload skinned mesh geometry to the GPU and do not store a copy on the CPU, so this may involve data transfer from the GPU on each frame). In the case of convex hulls, it might not be feasible to re-build the hull every frame in which it may have been animated.

2. In a physics sense, it seems to me problematic that the skinned mesh deformation is non-physical: if the change in shape causes a collision, does it exert a force? if so, what force? if not, will it be allowed to intersect with adjacent geometry from one frame to the next, or entirely phase through it?

As a concrete example, imagine a razor-thin mesh floor of an elevator, using a skin joint to move upwards, and a small marble resting on the floor of this elevator. Imagine an implementation performs CPU deformation by calling BakeMesh() every frame, effectively swapping out one mesh for another. If the joint node moves quickly enough, the floor of the elevator may move more in one physics frame than the radius of the marble, causing it to fall. This will be highly dependent on physics framerate, which could differ either due to hardware specs or by user agent.

Ultimately, I believe the likely outcome is the majority of conventional game engines will choose between ignoring support for skins/morphs for physics shapes (which means inconsistent support), or doing something unperformant (for example, calling BakeMesh() every frame). This sort of feature sounds cool, but it's a pie in the sky, and the likelihood of inconsistent support could scare users away from using the feature, which tells me that it belongs in a separate extension.

eoineoineoin: Note that `mesh` objects already have `weights` defined in the core glTF specification. Those weights may be animated, so this needs to be supported anyway. These properties merely the shapes consistent with how nodes can apply per-instance weights to a mesh, as well as using a skinned mesh.

As with many glTF features, it's possible to describe scenarios which are non-physical or may perform poorly, but at the end of the day, glTF content is made by a human who is ultimately making decisions about what that content is. The fact that a feature _may_ be misused is not a good reason to say "we don't support this feature which is in the core spec."

wallabyway:
@eoineoineoin - Thanks for bringing up 'implicit shapes'.

If I could pick two shapes that would help AEC, it would be cylinder and cube.  We could add a million other shapes, but I would compromise everything, just to get these two into a future spec somehow.

We pay a very high price for cylinders (HVAC, railings, cables, rebar, etc).  If the industry agreed on this, there would be a lot of precision gained, less data-loss.

Same for 'Cubes'.  If the industry agreed on a cube shape, rather than "I think these 16 triangles are a cube", adds a lot of meaning for AEC (design intent, manifold, volume, outline-style, improved z-fighting).

I know it's a very low bar, but you got to start somewhere and these two additions add an impressive amount of value IMHO.

javagl:
@wallabyway I've lost track of some of the discussion/developments here, and am (unfortuantely) not up to date. But a while ago, I also advocated for basically creating `KHR_shapes` and `KHR_collision` (using `KHR_shapes`) extensions. Somewhere around https://github.com/eoineoineoin/glTF_Physics/issues/1#issuecomment-1325503582 . The use-case would be for CAD/AEC (maybe going so far that these could even serve as a basis for some CSG at some point...). 

When "shapes" are defined here (non-renderable, as part of a "collision" extension), then someone **will** create a new extension that defines _the same shapes_, but "renderable". And in the worst case, one defines a "box" as (min, max), and the other one defines a "box" as (origin, size), and we all know what happens then...

Some further related discussion is in https://github.com/KhronosGroup/glTF/issues/1135 and around https://github.com/KhronosGroup/glTF/issues/1986#issuecomment-1023290360 , and many issues that are linked to/from these. (It's hard to wrap this up - that's why I lost track...).

eoineoineoin:
> @wallabyway I've lost track of some of the discussion/developments here, and am (unfortuantely) not up to date. But a while ago, I also advocated for basically creating `KHR_shapes` and `KHR_collision` (using `KHR_shapes`) extensions. Somewhere around [eoineoineoin/glTF_Physics#1 (comment)](https://github.com/eoineoineoin/glTF_Physics/issues/1#issuecomment-1325503582) . The use-case would be for CAD/AEC (maybe going so far that these could even serve as a basis for some CSG at some point...).
> 
> When "shapes" are defined here (non-renderable, as part of a "collision" extension), then someone **will** create a new extension that defines _the same shapes_, but "renderable". And in the worst case, one defines a "box" as (min, max), and the other one defines a "box" as (origin, size), and we all know what happens then...
> 
> Some further related discussion is in #1135 and around [#1986 (comment)](https://github.com/KhronosGroup/glTF/issues/1986#issuecomment-1023290360) , and many issues that are linked to/from these. (It's hard to wrap this up - that's why I lost track...).

Hey Marco,

Right, it does seem that there's interest in using the shapes for rendering, amongst other things. So, at the very least, there's a rename of the extension and associated reworking of documentation in the works, to ensure it's obviously usable for multiple different use-cases.

The "mesh" shape type is causing us some hassle, though, so I want to see if you have any thoughts here. For things like physics, selectability, UI, etc., it's useful to be able to describe a shape which is derived from a `glTF.mesh.` Though, for the "shape rendering" extension, this `shape.mesh` is redundant, since glTF meshes are already renderable via nodes. That leaves us with a couple of options:

1. Remove `shapes.mesh`, require "physics extension" and friends to reference `glTF.mesh` via some other mechanism
2. Leave as-is; "shape rendering extension" and friends would define how those `shapes.mesh `are handled

My concern is that with option 1, the options available in the `shapes.mesh` today will have to be re-defined by any extension which wants to use meshes. Similarly, with option 2, extensions might all require documentation to say how each shape type is handled (though, this might need to be the case anyway?) Any preferences? Other ideas?

lyuma:
I have been reading through the recent changes to use a collider or trigger "geometry" object.

I have some concern with the new version of the spec because it references "node", that implies referencing a node, which must be visible in core gltf. Is there a way to use meshes in physics colliders without rendering those meshes?

Related to this problem is a confusing paragraph here, so I would like some clarification on this:

> Collision shapes are parameterized in local space of the `node` they are associated with. If a shape is required to have an offset from the local space of the node the shape is associated with (for example a sphere _not_ centered at local origin or a rotated box,) a child node should be added with the desired offset applied, and the shape properties added to that child.

if I understand correctly, does this paragraph that the nodes and **child nodes** (a comment by Khronos members implies this actually means descendants) in the context of mesh geometry are taken in local space of the referenced `geometry` / `node`?

if so, can I ask if an acceptable way to implement hidden mesh collision shapes is to make a node whose scale is [0,0,0] (which hides the mesh as per the core gltf spec): since the collision shape children use a coordinate space relative to this node, child nodes will ignore the 0 scale. Does this make any sense?

if for some reason using [0,0,0] scale is not possible, the only alternatives I can think of are using a translation at 1e+99 to effectively hide the mesh, or relying on the node visbility extension (but this will not degrade well to older versions of Blender, for example)

HexaField:
@lyuma I can see some potential problems with your suggestions, most importantly with very large translations throwing off bounding volume calculations.

What if the spec allowed for referencing other scenes in the same document? This way, another scene would not be rendered (or optionally rendered, such as for rendering colliders in an editor) yet still accessible and loaded into memory under a separate hierarchy.

extensions/2.0/Khronos/KHR_implicit_shapes/README.md
```diff
@@ -0,0 +1,140 @@
...
+        ]
+    }
+}
+```
+
+To describe the geometry which represents the object, shapes must define at most one of the following properties:
+
+| |Type|Description|
+|-|-|-|
+|**sphere**|`object`|A sphere centered at the origin in local space.|
```
aaronfranke: Nitpick: Tables should be formatted.

extensions/2.0/Khronos/KHR_implicit_shapes/README.md
```diff
@@ -0,0 +1,140 @@
...
+Degenerate shapes are prohibited. A sphere must have a positive, non-zero radius. A box shape must have positive non-zero values for each component of `size`. The cylinder and capsule shapes must have a positive, non-zero `height`, both `radiusTop` and `radiusBottom` must be non-negative, and at least one of `radiusTop` and `radiusBottom` must be non-zero.
+
+### JSON Schema
+
+* glTF Document Extension: [glTF.KHR_implicit_shapes.schema.json](schema/glTF.KHR_implicit_shapes.schema.json)
+* Shape: [glTF.KHR_implicit_shapes.shape.schema.json](schema/glTF.KHR_implicit_shapes.shape.schema.json)
+* Sphere: [glTF.KHR_implicit_shapes.shape.sphere.schema.json](schema/glTF.KHR_implicit_shapes.shape.sphere.schema.json)
+* Box: [glTF.KHR_implicit_shapes.shape.box.schema.json](schema/glTF.KHR_implicit_shapes.shape.box.schema.json)
+* Cylinder: [glTF.KHR_implicit_shapes.shape.cylinder.schema.json](schema/glTF.KHR_implicit_shapes.shape.cylinder.schema.json)
+* Capsule: [glTF.KHR_implicit_shapes.shape.capsule.schema.json](schema/glTF.KHR_implicit_shapes.shape.capsule.schema.json)
```
aaronfranke: Nitpick: Khronos's preferred Markdown list style is using `-` symbols.

lexaknyazev (approved):

extensions/2.0/Khronos/KHR_implicit_shapes/README.md
```diff
@@ -0,0 +1,147 @@
...
+        ]
+    }
+}
+```
+
+To describe the geometry which represents the object, shapes must define at most one of the following properties:
+
+|              | Type     | Description                                                                                                                                                                                      |
+| ------------ | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| **plane**    | `object` | A plane centered at the origin in local space, with normal along the Y axis in local space, optionally with finite extents and may be double or single-sided.                                    |
```
aaronfranke:
```suggestion
| **plane**    | `object` | A plane centered at the origin in local space, with normal in the +Y direction in local space, optionally with finite extents and may be double or single-sided.                                    |
```

extensions/2.0/Khronos/KHR_implicit_shapes/README.md
```diff
@@ -0,0 +1,147 @@
...
+| **cylinder** | `object` | A cylinder centered at the origin and aligned along the Y axis in local space, with potentially different radii at each end. A cone is a special case of cylinder when one of the radii is zero. |
+| **capsule**  | `object` | A capsule (cylinder with hemispherical ends) centered at the origin and defined by two "capping" spheres with potentially different radii, aligned along the Y axis in local space.              |
+
+When the `type` of a shape refers to a shape type defined by this extension, the shape must not supply a value for a different shape type e.g. a shape whose `type` is "box" must not have a defined `sphere` property. Similarly, if `type` does not reference a shape type declared by `KHR_implicit_shapes`, an extension or other mechanism should supply an alternate shape definition.
+
+By default, planes have infinite extents along their tangent and bitangent vectors, specified as X and Z in local space, respectively. A plane may optionally provide `sizeX` and `sizeZ` properties to specify a finite size along these axes.
```
aaronfranke: This text is unclear, is the size "along these axes" referring to the radius or the diameter? Also, it should clarify +X and +Z for the tangent vectors.

eoineoineoin: I've updated this text but would be genuinely curious to hear if you knew of _any_ software which used the word "size" or "extents" to describe a half-extents.

extensions/2.0/Khronos/KHR_implicit_shapes/README.md
```diff
@@ -0,0 +1,147 @@
...
+            {
+                "sphere": { "radius": 0.1 },
+                "type": "sphere"
+            }
+        ]
+    }
+}
+```
+
+To describe the geometry which represents the object, shapes must define at most one of the following properties:
```
lexaknyazev, line 48: Consider requiring the corresponding object (so it's always available in the Object Model).

extensions/2.0/Khronos/KHR_implicit_shapes/README.md
```diff
@@ -0,0 +1,147 @@
...
+| **cylinder** | `object` | A cylinder centered at the origin and aligned along the Y axis in local space, with potentially different radii at each end. A cone is a special case of cylinder when one of the radii is zero. |
+| **capsule**  | `object` | A capsule (cylinder with hemispherical ends) centered at the origin and defined by two "capping" spheres with potentially different radii, aligned along the Y axis in local space.              |
+
+When the `type` of a shape refers to a shape type defined by this extension, the shape must not supply a value for a different shape type e.g. a shape whose `type` is "box" must not have a defined `sphere` property. Similarly, if `type` does not reference a shape type declared by `KHR_implicit_shapes`, an extension or other mechanism should supply an alternate shape definition.
+
+By default, planes have infinite extents along their tangent and bitangent vectors, specified as X and Z in local space, respectively. A plane may optionally provide `sizeX` and `sizeZ` properties to specify a finite total size along these axes.
```
lexaknyazev, line 60: As currently written, the language may imply that either both plane extents must be set or none.

extensions/2.0/Khronos/KHR_implicit_shapes/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+Degenerate shapes are prohibited. If provided, a plane must have positive non-zero `sizeX` and `sizeZ` extents. A sphere must have a positive, non-zero radius. A box shape must have positive non-zero values for each component of `size`. The cylinder and capsule shapes must have a positive, non-zero `height`, both `radiusTop` and `radiusBottom` must be non-negative, and at least one of `radiusTop` and `radiusBottom` must be non-zero.
```
lexaknyazev, line 62: Ditto about planes.

extensions/2.0/Khronos/KHR_implicit_shapes/README.md
eoineoineoin: Do you mean that even if a shape was utilizing the default values, you'd want to require an empty object? e.g.

```
"shapes" = [
{
     "type": "sphere",
     "sphere": {}
}
...
```

One complication on REQUIREing one of the objects that other extensions may add additional types and may not wish to have a fallback:

```
"shapes" = [
{
     "type": "teapot",
     "extensions": {
            "ext_teapot_shape" : {
                       ....
            }
    }
}
...
```

For those types, we wouldn't want to require an unrelated object in the shape.

lexaknyazev: Even when an object is syntactically empty it's not semantically empty because most shape properties have default values so, e.g., `"sphere": {}` actually means `"sphere": { "radius": 0.5 }`.

The extension should require an object matching the type so future extensions wouldn't require unrelated objects.

hybridherbst:
I think the "non-renderable" aspect could be made a bit more clear, at first I was confused on why there's no specification around UVs/tangents for these shapes.

(there's precedent for this kind of confusion creeping in with implicit shapes in OpenUSD, where it's not specified how UVs on implicit shapes work, and subsequently texturing a cube in one app and exporting it to another app as implicit shape is not interoperable at all, with people usually not being aware and then being confused.)

aaronfranke:
@hybridherbst A textured cube mesh should be represented as a mesh, not an implicit shape. An implicit shape might accompany such a cube for purposes like collision, but the intended visual representation is a mesh.

I believe the intention of renderability is for things like displaying a plain-colored outline in an editor, not runtime.

hybridherbst:
I just think that should be mentioned in the README for this extension to be less confusing. Reading just the extension made me wonder what happens when someone uses an implicit shape in for rendering – easily possible to do "accidentally wrong" via a custom extension. Only the first PR message here has the note "non-renderable".

javagl:
Throwback to [a comment from 2022](https://github.com/eoineoineoin/glTF_Physics/issues/1#issuecomment-1331001917) when the basics of what has become _this_ proposal have been sorted out:

> In the area of rendering, different questions arise - mainly that of tesselation or UV mapping.

That said: I think that the fact that the _current_ proposal here is focussing on the building block of the shapes themself makes sense, but think that there could, eventually, be an extension that defines a form of UV mapping for the `KHR_implicit_shapes` _only_ for the case that they are supposed to be used for (textured) rendering.

chriku:
Does this only require @eoineoineoin CLA approval or does anything else still require action? Anything one could do to progress this proposal? I'd really like to see #2424 become official and this seems to be a blocking dependency.

aaronfranke:
@chriku This PR is blocked because this will become a core glTF feature in glTF 2.1, rather than an extension.

chriku:
> @chriku This PR is blocked because this will become a core glTF feature in glTF 2.1, rather than an extension.

Thank you for the quick answer. This is both great news (long term) as well as horrible news (short term, at least until the spec has stabilized enough for a blender exporter. Can you make a VERY rough guess how long that would take naturally (like if it is in the days, years, centuries range)?)

aaronfranke:
@chriku Most likely, experimental support before the end of this year (such as in unmerged pull requests that you will have to explicitly grab and compile from source to use), and wide adoption over the course of 2027.

PR #2378, opened 2024-04-02
Additional fix to `GRIFFEL_bim_data` extension schema

### LINKED ISSUE #2373: GRIFFEL_bim_data.schema.json is an invalid json schema
`GRIFFEL_bim_data.schema.json` appears to be an invalid json shema.

The [`item`](
https://github.com/KhronosGroup/glTF/blob/ae1a90baf009203760304f06ed5645402de9dc09/extensions/2.0/Vendor/GRIFFEL_bim_data/schema/GRIFFEL_bim_data.schema.json#L20C18-L20C34)  property in ` properties/nodeProperties/items/properties/properties` should be `items`.

teddemunnik:
Fixes #2373

There was another instance of the `item` property inside of an array instance type, which should have been `items`. (`item` is not a known property in json schema draft 4)

### DIFF
--- a/extensions/2.0/Vendor/GRIFFEL_bim_data/schema/glTF.GRIFFEL_bim_data.schema.json
+++ b/extensions/2.0/Vendor/GRIFFEL_bim_data/schema/glTF.GRIFFEL_bim_data.schema.json
@@ -61,7 +61,7 @@
                     "properties": {
                         "type": "array",
                         "uniqueItems": true,
-                        "item": {
+                        "items": {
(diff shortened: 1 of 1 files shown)


javagl:
👍 

(I _did_ do full-text searches on the schemas, looking for further cases, but ... it did slip through, apparently. If something like this comes up again, I'd probably create a local copy of the meta-schema, and _disallow_ additional properties for the validation. What should an "additional property" in a JSON schema be, after all?)

Issue #2049, opened 2021-09-29
Consistent JSON schema version for core and extensions


javagl:
The JSON Schema version of the core specification has been updated from `draft-04` to `draft/2020-12` (via https://github.com/KhronosGroup/glTF/pull/2031 and https://github.com/KhronosGroup/glTF/pull/2034).

This raises some questions about the consistency of the specification. The core specification now uses `draft/2020-12`, whereas the extensions usually use `draft-04`. Without any guidelines or further actions, this could cause considerable fragmentation and inconsistencies.

Will the existing extension specifications be updated to also use `draft/2020-12`? (Who will do that, and how?)

Is there any guideline for the schema versions that extension specifications should use? (The section at https://github.com/KhronosGroup/glTF/tree/main/extensions#creating-extensions does not say anything).

lexaknyazev:
Yes, the extension specs need to be upgraded:
- to the new JSON schema version;
- to use BCP 14 normative language;
- to use AsciiDoc tooling (long-term).

javagl:
I'm trying to get a bit up to date with the latest schema, and I hope that the following is closely enough related to this topic that it is OK to bring that up here:

---

Section https://json-schema.org/understanding-json-schema/structuring.html#id says

> However, using a relative reference when setting a base URI can be problematic. For example, we couldn’t use this schema as an anonymous schema because there would be no Retrieval URI and you can’t resolve a relative reference against nothing. For this and other reasons, it’s recommended that you always use an absolute URI when declaring a base URI with $id.

Right now, the `$id` are relative. But actually inserting the whole `https://raw.githubusercontent.com/KhronosGroup/glTF/main/specification/2.0/schema/` prefix there looks strange at the first glance...

---

Similarly, section https://json-schema.org/understanding-json-schema/structuring.html#ref points out

> When using $ref in an anonymous schema, relative references may not be resolvable. 

This can be a problem for the extensions (even though they are not "anonymous" - they only refer to a different context than the one that is implied by their base URL). Specifically, I wondered whether extensions that currently contain relative references like

    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],

should actually contain absolute references. Again, imagining the full path there ...

    "allOf": [ { "$ref": "https://raw.githubusercontent.com/KhronosGroup/glTF/main/specification/2.0/schema/glTFProperty.schema.json" } ],

[...]

PR #2379, opened 2024-04-02
Update KHR_light_punctual schema from json schema draft 04 to json schema draft 2020-12

teddemunnik:
As discussed in #2049 this PR does the update just for the `KHR_lights_punctual` extension.

### DIFF
--- a/extensions/2.0/Khronos/KHR_lights_punctual/schema/light.schema.json
+++ b/extensions/2.0/Khronos/KHR_lights_punctual/schema/light.schema.json
@@ -1,5 +1,6 @@
 {
-    "$schema": "http://json-schema.org/draft-04/schema",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "light.schema.json",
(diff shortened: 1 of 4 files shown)


extensions/2.0/Khronos/KHR_lights_punctual/schema/light.schema.json
```diff
@@ -51,8 +52,7 @@
         "range": {
             "type": "number",
             "description": "A distance cutoff at which the light's intensity may be considered to have reached zero.",
-            "minimum": 0,
-            "exclusiveMinimum": true
+            "exclusiveMinimum": 0
```
aaronfranke, line 55: Nitpick: Since this value is not restricted to integers, I would prefer `0.0` instead of just `0`, to make it clear to readers. This would also be consistent with [other values](https://github.com/KhronosGroup/glTF/pull/2380/files#diff-fe74ec687a590b4d07790b3deed9e0c93dd941cc960506fdfa67602fdb4de120R23).

Same for `light.spot.schema.json`.

PR #2383, opened 2024-04-05
Fix type of CESIUM_primitive_outline's `indices` property.

kring:
When I wrote the schema for this extension, I incorrectly used `integer` as the type for the `indices` property, which is meant to refer to an accessor in the glTF. It should instead be `"allOf": [ { "$ref" : "glTFid.schema.json" } ]`.


lexaknyazev (approved):

javagl (approved):

javagl:
(The same is still TODO for other extensions, but this is tracked in https://github.com/CesiumGS/3d-tiles/issues/751 )

PR #2391, opened 2024-04-26
Framebuffer light intensity mapping extension - draft spec

rsahlin:
One way of declaring an extension that will provide a light scale factor to be used in shading to produce valid display output.

### DIFF
--- a/extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
+++ b/extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
@@ -1,0 +1,134 @@
+# KHR_framebuffer_value_mapping
+
+## Contributors
+
+Rickard Sahlin, <mailto:rickard.sahlin@inter.ikea.com>  
+
+Copyright (C) 2024 The Khronos Group Inc. All Rights Reserved. glTF is a trademark of The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.01 spec.
+
+## Exclusions
+
+## Termonilogy
+
+This extension uses the following terms:
+
+Framebuffer : One or more memory buffers used by the target platform to store pixel values prior to being output to display.  
+framebuffer_max_value : The maximum value that can be stored in the framebuffer, for an 8 bit unorm buffer this is 1.0.  
+
+
+## Overview
+
+This extension is intended for implementations that targets a light emitting display with the goal of outputting interactive framerates in a physically correct manner.    
+The purpose is to define how light calculation output values are mapped to framebuffer values.  
+
+There are two main goals for this extension:  
+  
+1: Provide a method to linearly scale RGB values to the framebuffer format of the target device - called lightScaleFactor.  
+  
+2: Define how to limit values prior to writing them to framebuffer when the scaled value exceeds the maximum range of the framebuffer - called RGBThreshold.  
+  
+  
+## Calculation of the lightScaleFactor  
+
+The light scale factor is calculated from the extension "maxLight" value.  
+This value defines the light intensity (lm/m2) value that will be mapped to the maximum value of the framebuffer.  
+ 
+  
+float lightScaleFactor = framebuffer_max_value / maxLight  
+ 
+Each calculated RGB pixel value shall be scaled by the lightScaleFactor at some point prior to being output to display.  
+Exactly how this is done is an implementation detail, the normal usecase would be to scale the value in the fragment shader.   
+
+## Limit of RGB pixel values before output to framebuffer  
+    
+If, after applying the lightScaleFactor, RGB values exceed the maximum value of the framebuffer then the RGB triplet must be scaled.  
+The scaling shall be done using the max component value from the RGB triplet, ensuring that color is retained.  
+  
+pseudocode for scaling.  
+vec3 scaledRGB = RGB / max(R, max(G,B));  
+  
+
+## Defining an asset to use KHR_framebuffer_value_mapping
+  
+The extension is put in the scene declaration.  
+All uses of the scene, for instance in a usecase with glTFX, shall take the extension into account.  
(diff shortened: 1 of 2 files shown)


extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
```diff
@@ -0,0 +1,134 @@
...
+2: Define how to limit values prior to writing them to framebuffer when the scaled value exceeds the maximum range of the framebuffer - called RGBThreshold.  
+  
+  
+## Calculation of the lightScaleFactor  
+
+The light scale factor is calculated from the extension "maxLight" value.  
+This value defines the light intensity (lm/m2) value that will be mapped to the maximum value of the framebuffer.  
+ 
+  
+float lightScaleFactor = framebuffer_max_value / maxLight
```
elalish, line 46: Common industry parlance for `lightScaleFactor` is exposure.

extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
```diff
@@ -0,0 +1,134 @@
...
+Each calculated RGB pixel value shall be scaled by the lightScaleFactor at some point prior to being output to display.  
+Exactly how this is done is an implementation detail, the normal usecase would be to scale the value in the fragment shader.   
+
+## Limit of RGB pixel values before output to framebuffer  
+    
+If, after applying the lightScaleFactor, RGB values exceed the maximum value of the framebuffer then the RGB triplet must be scaled.  
+The scaling shall be done using the max component value from the RGB triplet, ensuring that color is retained.  
+  
+pseudocode for scaling.  
+vec3 scaledRGB = RGB / max(R, max(G,B));
```
elalish, line 57: Common industry parlance for a function like this that maps out of bounds RGB values to a framebuffer's limits is a tone mapping function.

extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
```diff
@@ -0,0 +1,134 @@
...
+## Limit of RGB pixel values before output to framebuffer  
+    
+If, after applying the lightScaleFactor, RGB values exceed the maximum value of the framebuffer then the RGB triplet must be scaled.  
+The scaling shall be done using the max component value from the RGB triplet, ensuring that color is retained.  
+  
+pseudocode for scaling.  
+vec3 scaledRGB = RGB / max(R, max(G,B));  
+  
+
+## Defining an asset to use KHR_framebuffer_value_mapping
```
elalish, line 60: Can you add example renders comparing your tone mapping function to others, e.g. Linear, ACES, PBR Neutral and demonstrating the pros and cons compared to each? One model that might make for an interesting comparison is my 3D Macbeth color chart [model](https://github.com/google/model-viewer/blob/5476def9c5bf3d65f04524ce1cc2e6bc849751fb/packages/shared-assets/models/MacbethBalls.glb).

elalish (changes requested):

extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
rsahlin: This is not exposure.
There is nothing being exposed - instead this is a linear scale factor so that light intensity is limited inside the desired range. 
Industry parlance for parameters that provide this function is scale or factor.

extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
rsahlin: No, this is not a tonemapping function - it is RGB  scale or normalization.

extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
rsahlin: I can certainly include a couple of images of how glTFs will look using this extension with different maxLight values.

extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
elalish: Mathematically, what you have defined is exactly equivalent to what all current PBR renderers call `exposure`. In general they do not deal with aperture, shutter, or ISO either, but their simplified combination, known as exposure, which is exactly the factor you define.

extensions/2.0/Khronos/KHR_framebuffer_value_mapping/images/macbeth_bright.jpg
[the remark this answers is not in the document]
rsahlin: I have no idea what you mean by `wildly different colors`
All spheres have the sama chroma and hue - the matte spheres are simply darker due to more scattering of transmitted light.
This is what matters in PBR, the hue/chroma must be the same but the brightness may change
Compare to a glossy or matte paint.

extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
rsahlin: Nope - this is a mathematical function to limit the precision of a 3 component color value prior to being stored in a buffer.
Tonemapping is the method of applying a 'color tone' to an image.

extensions/2.0/Khronos/KHR_framebuffer_value_mapping/README.md
rsahlin: As I have stated before, you need to understand the goal of this extension.
This extension takes care of limiting the precision of light intensity values prior to being stored in a framebuffer that has less precision.

Exposure is such a diluted term that it could mean more or less anyting, the usage in PBR renderers vary wildly - some has a proper camera with aperture, shutter and ISO - some will just apply a final cludge factor to get the wanted light intensity.
Looking at the available PBR tools and renderers it's far from general usage.

As I also have stated before - you clearly want/need the type of exposure value that is ascociated with a camera - make sure to propose such an extension.
This is not that extension (you are looking for)

PR #2397, opened 2024-05-07
KHR_accessor_float64 and KHR_accessor_float16 Draft Proposals

lexaknyazev:
Split from #2395.

As discussed in the Interactivity DTSG, it's important for the animation pointer data to be able to represent the same Object Model property values that could be set via behavior graphs.

### DIFF
--- a/extensions/2.0/Khronos/KHR_accessor_float64/README.md
+++ b/extensions/2.0/Khronos/KHR_accessor_float64/README.md
@@ -42,8 +42,9 @@ When the extension is supported, the following data type is valid for accessors 
 Such accessors **MUST** adhere to the following additional rules:
 
 - their `normalized` property **MUST NOT** be set to true;
-- their effective byte offsets **MUST** be aligned to 8-byte boundaries;
-- when referring to data in a GLB-stored buffer, the start of the buffer **MUST** be aligned to 8-byte boundaries by adding necessary padding bytes to the JSON chunk as defined in [Section 4.4.3.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#structured-json-content).
+- their effective byte offsets, as defined in [Section 3.6.2.4](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#data-alignment), **MUST** be aligned to 8-byte boundaries.
+
+When referring to data in a GLB-stored buffer, the start of the buffer **SHOULD** be aligned to an 8-byte boundary by adding necessary trailing padding bytes to the JSON chunk as defined in [Section 4.4.3.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#structured-json-content).
 
 ## Extending Accessor Usage
 
--- a/extensions/2.0/Khronos/KHR_accessor_float64/README.md
+++ b/extensions/2.0/Khronos/KHR_accessor_float64/README.md
@@ -1,0 +1,129 @@
+<!--
+Copyright 2024 The Khronos Group Inc.
+SPDX-License-Identifier: LicenseRef-KhronosSpecCopyright
+-->
+
+# KHR\_accessor\_float64
+
+## Contributors
+
+- Alexey Knyazev, [@lexaknyazev](https://github.com/lexaknyazev)
+- Sean, [@spnda](https://github.com/spnda)
+
+Copyright 2024 The Khronos Group Inc.
+See [Appendix](#appendix-full-khronos-copyright-statement) for full Khronos Copyright Statement.
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 Specification.
+
+## Overview
+
+The base glTF 2.0 Specification supports only one floating-point component type - IEEE-754 _binary32_ (also known as single-precision). Although this is sufficient for most applications, there are use cases that benefit from increased range and precision provided by the 64-bit floating-point data type (also known as double-precision).
+
+This extension expands the set of recognized accessor component types with a new enum value `5130` that represents IEEE-754 _binary64_.
+
+This extension allows usage of such accessors in all places where the single-precision floating-point component type is accepted by the base glTF 2.0 Specification or extensions.
+
+Because this extension does not provide a way to specify alternative data representations, assets that use this extension **MUST** specify it in the `extensionsRequired` array.
+
+## Extending Accessors
+
+When the extension is supported, the following data type is valid for accessors in addition to the types defined in [Section 3.6.2.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#accessor-data-types).
+
+| `componentType` | Data Type | Signed | Bits |
+|-----------------|-----------|--------|------|
+| `5130`          | _double_  | Signed | 64   |
+
+Such accessors **MUST** adhere to the following additional rules:
+
+- their `normalized` property **MUST NOT** be set to true;
+- their effective byte offsets **MUST** be aligned to 8-byte boundaries;
+- when referring to data in a GLB-stored buffer, the start of the buffer **MUST** be aligned to 8-byte boundaries by adding necessary padding bytes to the JSON chunk as defined in [Section 4.4.3.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#structured-json-content).
+
+## Extending Accessor Usage
+
+When the extension is supported, it is valid to use _double_ component type accessors in **all** places where _float_ component type accessors are valid.
+
+The following sections cover only the base glTF 2.0 Specification and ratified extensions.
+
+### Mesh Attributes
+
+When the extension is supported, _double_ component type accessors are valid for all mesh and morph target attributes that support _float_ component type accessors as defined in [Section 3.7.2.1](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#meshes-overview) and [Section 3.7.2.2](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#morph-targets) respectively.
+
+#### Interaction with `EXT_mesh_gpu_instancing`
+
+When the extension is supported, _double_ component type accessors are valid for all instancing semantics defined in the `EXT_mesh_gpu_instancing` extension.
+
+### Skins
+
+When the extension is supported, _double_ component type accessors are valid for inverse-bind matrices defined in [Section 3.7.3.1](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#skins-overview).
+
+### Animation Samplers
+
+When the extension is supported, _double_ component type accessors are valid for animation sampler inputs and outputs defined in [Section 3.11](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#animations).
+
+Interpolation of double-precision keyframe data **MUST** be performed with double-precision arithmetic.
+
+#### Interaction with `KHR_animation_pointer`
+
+When the extension is supported, _double_ component type accessors are valid for keyframe data of all Object Model properties of `float*` data types.
+
+## Compatibility
+
+Since not all graphics environments support double-precision floating-point data, implementations **MAY** employ different lossy strategies to render such assets, e.g., converting mesh attribute data to single-precision values when uploading data to GPUs.
+
+When a lossy conversion is involved, implementations **SHOULD** warn their users that the asset's rendering can be incorrect.
+
+Given transmission size and portability limitations, asset creators **SHOULD** use _double_ component type accessors only when absolutely necessary.
+
+## Appendix: Full Khronos Copyright Statement
+
+Copyright 2024 The Khronos Group Inc.
+
+This Specification is protected by copyright laws and contains material proprietary
+to Khronos. Except as described by these terms, it or any components
+may not be reproduced, republished, distributed, transmitted, displayed, broadcast
+or otherwise exploited in any manner without the express prior written permission
+of Khronos.
+
+Khronos grants a conditional copyright license to use and reproduce the unmodified
+Specification for any purpose, without fee or royalty, EXCEPT no licenses to any patent,
+trademark or other intellectual property rights are granted under these terms.
+
+Khronos makes no, and expressly disclaims any, representations or warranties,
+express or implied, regarding this Specification, including, without limitation:
+merchantability, fitness for a particular purpose, non-infringement of any
+intellectual property, correctness, accuracy, completeness, timeliness, and
+reliability. Under no circumstances will Khronos, or any of its Promoters,
+Contributors or Members, or their respective partners, officers, directors,
+employees, agents or representatives be liable for any damages, whether direct,
+indirect, special or consequential damages for lost revenues, lost profits, or
+otherwise, arising from or in connection with these materials.
+
+This specification has been created under the Khronos Intellectual Property Rights
+Policy, which is Attachment A of the Khronos Group Membership Agreement available at
+https://www.khronos.org/files/member_agreement.pdf. Khronos grants a conditional
+copyright license to use and reproduce the unmodified specification for any purpose,
+without fee or royalty, EXCEPT no licenses to any patent, trademark or other
+intellectual property rights are granted under these terms. Parties desiring to
+implement the specification and make use of Khronos trademarks in relation to that
+implementation, and receive reciprocal patent license protection under the Khronos
+IP Policy must become Adopters and confirm the implementation as conformant under
+the process defined by Khronos for this specification;
+see https://www.khronos.org/conformance/adopters/file-format-adopter-program.
+
+Where this Specification identifies specific sections of external references, only those
+specifically identified sections define normative functionality. The Khronos Intellectual
+Property Rights Policy excludes external references to materials and associated enabling
+technology not created by Khronos from the Scope of this Specification, and any licenses
+that may be required to implement such referenced materials and associated technologies
+must be obtained separately and may involve royalty payments.
+
+Khronos® is a registered trademark, and glTF™ is a trademark of The Khronos Group Inc. All
+other product names, trademarks, and/or company names are used solely for identification
+and belong to their respective owners.
--- a/extensions/2.0/Khronos/KHR_accessor_float16/README.md
+++ b/extensions/2.0/Khronos/KHR_accessor_float16/README.md
@@ -3,7 +3,7 @@ Copyright 2025 The Khronos Group Inc.
 SPDX-License-Identifier: LicenseRef-KhronosSpecCopyright
 -->
 
-# KHR\_accessor\_float64
+# KHR\_accessor\_float16
 
 ## Contributors
 
@@ -23,9 +23,9 @@ Written against the glTF 2.0 Specification.
 
 ## Overview
 
-The base glTF 2.0 Specification supports only one floating-point component type—IEEE-754 _binary32_ (also known as single-precision). Although this is sufficient for most applications, there are use cases that benefit from increased range and precision provided by the 64-bit floating-point data type (also known as double-precision).
+The base glTF 2.0 Specification supports only one floating-point component type—IEEE-754 _binary32_ (also known as single-precision).
 
-This extension expands the set of recognized accessor component types with a new enum value `5130` that represents IEEE-754 _binary64_.
+This extension expands the set of recognized accessor component types with a new enum value `5131` that represents IEEE-754 _binary16_.
 
 This extension allows usage of such accessors in all places where the single-precision floating-point component type is accepted by the base glTF 2.0 Specification or extensions.
 
@@ -37,47 +37,41 @@ When the extension is supported, the following data type is valid for accessors 
 
 | `componentType` | Data Type | Signed | Bits |
 |-----------------|-----------|--------|------|
-| `5130`          | _double_  | Signed |  64  |
+| `5131`          |  _half_   | Signed |  16  |
 
 The `normalized` property of such accessors **MUST NOT** be set to true.
 
-When a _double_ accessor refers to a buffer view that refers to a GLB-stored buffer, the start of the buffer **SHOULD** be aligned to an 8-byte boundary by adding necessary trailing padding bytes to the previous chunks as defined in the [GLB File Format Specification](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#chunks).
-
 ## Extending Accessor Usage
 
-When the extension is supported, it is valid to use _double_ component type accessors in **all** places where _float_ component type accessors are valid.
+When the extension is supported, it is valid to use _half_ component type accessors in **all** places where _float_ component type accessors are valid.
 
 The following sections are not exhaustive and cover only the base glTF 2.0 Specification and some ratified extensions.
 
 ### Mesh Attributes
 
-When the extension is supported, _double_ component type accessors are valid for all mesh and morph target attributes that support _float_ component type accessors as defined in the [Meshes](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#meshes-overview) and [Morph Targets](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#morph-targets) sections of the glTF 2.0 Specification respectively.
+When the extension is supported, _half_ component type accessors are valid for all mesh and morph target attributes that support _float_ component type accessors as defined in the [Meshes](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#meshes-overview) and [Morph Targets](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#morph-targets) sections of the glTF 2.0 Specification respectively.
 
 #### Interaction with `EXT_mesh_gpu_instancing`
 
-When the extension is supported, _double_ component type accessors are valid for all instanced attribute semantics defined in the `EXT_mesh_gpu_instancing` extension.
+When the extension is supported, _half_ component type accessors are valid for all instanced attribute semantics defined in the `EXT_mesh_gpu_instancing` extension.
 
 ### Skins
 
-When the extension is supported, _double_ component type accessors are valid for inverse-bind matrices defined in the [Skins](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#skins-overview) section of the glTF 2.0 Specification.
+When the extension is supported, _half_ component type accessors are valid for inverse-bind matrices defined in the [Skins](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#skins-overview) section of the glTF 2.0 Specification.
 
 ### Animation Samplers
 
-When the extension is supported, _double_ component type accessors are valid for animation sampler inputs and outputs defined in the [Animations](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#animations) section of the glTF 2.0 Specification.
+When the extension is supported, _half_ component type accessors are valid for animation sampler inputs and outputs defined in the [Animations](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#animations) section of the glTF 2.0 Specification.
 
-Interpolation of double-precision keyframe data **MUST** be performed with double-precision arithmetic.
+Use of _half_ component type accessors animation data storage **MUST NOT** affect the implementation precision used for animation interpolation.
 
 #### Interaction with `KHR_animation_pointer`
 
-When the extension is supported, _double_ component type accessors are valid for keyframe data of all Object Model properties of `float*` data types.
+When the extension is supported, _half_ component type accessors are valid for keyframe data of all Object Model properties of `float*` data types.
 
 ## Compatibility
 
-Since not all graphics environments support double-precision floating-point data, implementations **MAY** reject assets using double-precision data for mesh attributes or any other purpose that would require native hardware support.
-
-If an implementation converts double-precision data to single-precision due to platform limitations, it **SHOULD** warn users that the asset's rendering can be incorrect.
-
-Given transmission size and portability implications, _double_ component type accessors **SHOULD NOT** be used unless absolutely necessary.
+Since not all programming languages directly support working with half-precision floating-point data, implementations **MAY** need to convert such data to higher precision representations, i.e., single or double, when operating on individual values. This is usually not needed when uploading half-precision data to GPUs.
 
 ## Appendix: Full Khronos Copyright Statement
 


extensions/2.0/Khronos/KHR_accessor_float64/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+| `componentType` | Data Type | Signed | Bits |
+|-----------------|-----------|--------|------|
+| `5130`          | _double_  | Signed | 64   |
+
+Such accessors **MUST** adhere to the following additional rules:
+
+- their `normalized` property **MUST NOT** be set to true;
+- their effective byte offsets **MUST** be aligned to 8-byte boundaries;
+- when referring to data in a GLB-stored buffer, the start of the buffer **MUST** be aligned to 8-byte boundaries by adding necessary padding bytes to the JSON chunk as defined in [Section 4.4.3.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#structured-json-content).
```
javagl, line 46: Is it really necessary to require that alignment for the _buffer_? 

I thought that only the start of the accessor (accessor.byteOffset + bufferView.byteOffset) needs to be aligned accordingly, as of https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#data-alignment . As it is written now, it will have implications on the 'chunk' alignment of https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#chunks-overview 

(Maybe I need a refresher for some implications, for things like the JavaScript `Float64Array`, and what _exactly_ the alignment requirements refer to in this case. But I thought that having a `Float64Array` with an offset of 16 inside an `ArrayBuffer` would be OK, even if the glTF buffer formally starts at offset 12 of this `ArrayBuffer`, for example)

lexaknyazev: If the whole GLB is received as an `ArrayBuffer`, apps won't be able to create a `Float64Array` from it without the 8-byte buffer alignment requirement.

```js
// Uncaught RangeError: start offset of Float64Array should be a multiple of 8
new Float64Array(new ArrayBuffer(16), 4)
```

Without this requirement, apps would have to make an extra copy. This sounds worse than adding four trailing spaces to JSON during export if needed.

lexaknyazev: After revisiting this one more time I think that this requirement could be demoted to suggestion, i.e., saying SHOULD instead of MUST.

The reasoning is that some pipeline tools that produce GLB files may be able to passthrough assets using this extension but won't be updated any time soon to specifically handle the 8-byte alignment.

javagl: My thoughts had been about the point is that this `Float64Array` will not be created for the whole buffer, but only for the _accessor_ that is _part_ of the buffer. So...
```
accessor = new Float64Array(glbArrayBuffer, 4 + 4)
                                            ^   ^offset of BV/accessor in buffer
                                            |offset of BIN buffer in GLB 
```
would be valid. 

But I can see that the alignment requirement has to be phrased in terms of the `buffer` (i.e. ~"the `BIN` chunk"). That's the only way to ensure that 'everything works', regardless of whether _the whole GLB_ is stored in one `ArrayBuffer`, or whether only the BIN chunk is stored in one `ArrayBuffer`.

---

I think it could be tricky to support this for _exporters_: Usually, they'll do 
```
glbData.fillWith(jsonData);
glbData.padTo(4);
glbData.fillWith(bufferData);
```
But now they'll have to do
```
glbData.fillWith(jsonData);
if (bufferData.containsAccessorWith64bitValues()) {
    glbData.padTo(8);
} else {
    glbData.padTo(4);
}
glbData.fillWith(bufferData);
```
where that `containsAccessorWith64bitValues` could be difficult to implement. 

Exporters might be tempted to just use the "lazy" solution of _always_ aligning to 8, to be safe. But that might raise questions in view of https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#glb-stored-buffer which says

> The byte length of the BIN chunk MAY be up to 3 bytes bigger than JSON-defined buffer.byteLength value to satisfy GLB padding requirements.

and where the BIN chunk may become up to 7 bytes bigger than the `buffer.byteLength` when this "lazy" approach is used. I think the validator would then complain about `BUFFER_GLB_CHUNK_TOO_BIG`.

(Note: Nothing of this is "critical". I'm just trying to think through corner cases and implications...)

lexaknyazev: The BIN chunk trailing bytes are unrelated to accessor alignment. They are needed only to ensure 4-byte alignment of hypothetical custom chunks present in GLB afterwards.

javagl: Exporters will have to ensure the 8-byte alignment of the start of the `BIN` chunk (by padding the JSON chunk). Strictly speaking, they _only_ have to do this when the glTF uses this extension. But I think that they will _always_ do it. And that's OK. 

(They'll probably figure out that they should _not_ align _all_ chunks starts/ends to 8 bytes (ignoring what the alignment requirements for a 'custom' chunk could be), so maybe that's not so much of an issue...)

lexaknyazev: Relaxed the GLB-related language to not invalidate all existing tools that passthrough unknown extensions.

Please review the rest of the extension text.

javagl: The remaining text looks fine for me (I'd probably need more time (and maybe try to _implement_ it...) to find _potential_ pitfalls, but the description itself _looks_ straightforward - let's see whether others have further comments here)

extensions/2.0/Khronos/KHR_accessor_float64/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+Because this extension does not provide a way to specify alternative data representations, assets that use this extension **MUST** specify it in the `extensionsRequired` array.
+
+## Extending Accessors
+
+When the extension is supported, the following data type is valid for accessors in addition to the types defined in [Section 3.6.2.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#accessor-data-types).
+
+| `componentType` | Data Type | Signed | Bits |
+|-----------------|-----------|--------|------|
+| `5130`          | _double_  | Signed | 64   |
```
emackey, line 40: Out of curiosity, does `5130` come from anywhere in WebGL, or is it unique to glTF?

emackey (approved):

extensions/2.0/Khronos/KHR_accessor_float64/README.md
lexaknyazev: It comes from desktop OpenGL (like all other component type enums originally).

emackey: Of course. Thanks!

javagl: Not sure what "THE" official source is, but https://registry.khronos.org/OpenGL/api/GL/glcorearb.h says
`#define GL_DOUBLE                         0x140A`
(and it's found in many other places via https://github.com/search?q=org%3AKhronosGroup+GL_DOUBLE&type=code

(I'm usually using https://javagl.github.io/GLConstantsTranslator/GLConstantsTranslator.html to check stuff like that, but... I don't even remember what header file I originally used for that 8 years ago - it may be incomplete...)

bghgary:
How does this extension interact with mesh compression extensions like Draco or Meshopt?

donmccurdy:
For Draco, here's an (old) quote...

- https://github.com/google/draco/issues/360

... so it doesn't sound like f64 is supported. Draco's "edgebreaker" codec will change vertex count and order, so compressing only a subset of a mesh primitive's attributes must be avoided. The "sequential" codec is safer, but in practice it [can change the vertex count too](https://github.com/google/draco/issues/720#issuecomment-886135249). I think it would be necessary today to avoid using Draco for any mesh primitive containing f64 attributes, but perhaps that's something that can be added to the Draco library without changes to the extension.

***

For Meshopt, I don't think there's support for f64 in the library today, but vertex count and order does not change, so excluding specific vertex attributes from compression is trivial. That's probably what I'd do in glTF Transform, just skip compression for these attributes. I assume that adding f64 support to EXT_meshopt_compression is technically possible, but not trivial: 

- https://github.com/zeux/meshoptimizer/issues/2#issuecomment-806359065

***

Not to say that KHR_accessor_float64 should prohibit these compression methods. But there are technical obstacles.

emackey:
In theory, if one is using f64 for precision, one should not switch to quantized, right?  That's less precision than f32.

donmccurdy:
I tend to think of the component type's precision only as an upper bound on quantization precision. It's common to quantize `POSITION` accessors to 12 or 14 bits, stored in 16-bit integer accessors, for example. HTTP-level compression will then reduce the transferred size. The same approach should work with f64, with accessors quantized to 32–64 bits. See [gltfpack's `-vpf` flag](https://meshoptimizer.org/gltf/#options), for an example with f32.

lexaknyazev:
> accessors quantized to 32–64 bits

Note that there are no 32- or 64-bit normalized data types.

@bghgary, @emackey, @javagl
OK to move the draft to Release Candidate?

bghgary:
> OK to move the draft to Release Candidate?

Should we add sample models? We usually want to include sample models if the intention is to get implementations so that people can test.

javagl:
I agree that sample models are important (and I'd probably see whether I can quickly create one of these "minimal test models", like the `Triangle` storing everything with `double` or so). But I don't know whether they are _required_ for this to go from 'draft' to 'release candidate'.

bghgary:
If RC means we want implementations (which I believe is the intention), then whoever implements the spec will need something to test. If we don't provide the sample, the implementors will have to create something instead which will make it less likely for the implementation to happen.

javagl:
Don already covered the topic of generating test models via https://github.com/donmccurdy/glTF-Transform/pull/1417

In the meantime, if it matters, here's the `AnimatedTriangle` sample asset (embedded), using `double/float64` values for the `POSITION` data and the rotation animation key frames/quaternions:

```json
{
  "extensionsUsed" : [ "KHR_accessor_float64" ],
  "extensionsRequired" : [ "KHR_accessor_float64" ],
  "accessors" : [ {
    "bufferView" : 0,
    "byteOffset" : 0,
    "componentType" : 5123,
    "count" : 3,
    "type" : "SCALAR",
    "max" : [ 2 ],
    "min" : [ 0 ]
  }, {
    "bufferView" : 1,
    "byteOffset" : 0,
    "componentType" : 5130,
    "count" : 3,
    "type" : "VEC3",
    "max" : [ 1.0, 1.0, 0.0 ],
    "min" : [ 0.0, 0.0, 0.0 ]
  }, {
    "bufferView" : 2,
    "byteOffset" : 0,
    "componentType" : 5130,
    "count" : 5,
    "type" : "SCALAR",
    "max" : [ 1.0 ],
    "min" : [ 0.0 ]
  }, {
    "bufferView" : 2,
    "byteOffset" : 40,
    "componentType" : 5130,
    "count" : 5,
    "type" : "VEC4",
    "max" : [ 0.0, 0.0, 1.0, 1.0 ],
    "min" : [ 0.0, 0.0, 0.0, -0.7070000171661377 ]
  } ],
  "animations" : [ {
    "channels" : [ {
      "sampler" : 0,
      "target" : {
        "node" : 0,
        "path" : "rotation"
      }
    } ],
    "samplers" : [ {
      "input" : 2,
      "interpolation" : "LINEAR",
      "output" : 3
    } ]
  } ],
  "asset" : {
    "generator" : "JglTF from https://github.com/javagl/JglTF",
    "version" : "2.0"
  },
  "buffers" : [ {
    "uri" : "data:application/gltf-buffer;base64,AAABAAIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADwPwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAPA/AAAAAAAAAAAAAAAAAAAAAAAAAAAAANA/AAAAAAAA4D8AAAAAAADoPwAAAAAAAPA/AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA8D8AAAAAAAAAAAAAAAAAAAAAAAAAgL6f5j8AAACAvp/mPwAAAAAAAAAAAAAAAAAAAAAAAAAAAADwPwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIC+n+Y/AAAAgL6f5r8AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADwPw==",
    "byteLength" : 280
  } ],
  "bufferViews" : [ {
    "buffer" : 0,
    "byteOffset" : 0,
    "byteLength" : 6,
    "target" : 34963
  }, {
    "buffer" : 0,
    "byteOffset" : 8,
    "byteLength" : 72,
    "target" : 34962
  }, {
    "buffer" : 0,
    "byteOffset" : 80,
    "byteLength" : 200
  } ],
  "meshes" : [ {
    "primitives" : [ {
      "attributes" : {
        "POSITION" : 1
      },
      "indices" : 0,
      "mode" : 4
    } ]
  } ],
  "nodes" : [ {
    "mesh" : 0,
    "rotation" : [ 0.0, 0.0, 0.0, 1.0 ]
  } ],
  "scene" : 0,
  "scenes" : [ {
    "nodes" : [ 0 ]
  } ]
}
```

---

<sub>Generator code: [GeneratedAnimatedTriangleFloat64.zip](https://github.com/KhronosGroup/glTF/files/15363092/GeneratedAnimatedTriangleFloat64.zip)</sub>

donmccurdy:
Here's a 64-byte MorphStressTest:

[MorphStressTest.zip](https://github.com/KhronosGroup/glTF/files/15365710/MorphStressTest.zip)

I just picked this arbitrarily, as a scene containing both mesh data and animation. Converting other samples would be easy enough, happy to add whichever are wanted. The gist of the script (after other changes to support f64 in gltf transform) is ...

```javascript
import { NodeIO } from '@gltf-transform/core';
import { KHRONOS_EXTENSIONS, KHRAccessorFloat64 } from '@gltf-transform/extensions';
import { dequantize } from '@gltf-transform/functions';

const io = new NodeIO().registerExtensions(KHRONOS_EXTENSIONS)
const document = await io.read('sample.glb');

function f64(options = {}) {

  return (document) => {
    document.createExtension(KHRAccessorFloat64).setRequired(true);

    for (const accessor of document.getRoot().listAccessors()) {
      if (accessor.getComponentType() === 5126 /* FLOAT */) {
        accessor.setArray(new Float64Array(accessor.getArray()));
      }
    }
  }

}

await document.transform(dequantize(), f64());

await io.write('sample64.glb', document);
```

So it's a direct conversion from f32 to f64, the additional precision is not required to render the asset correctly.

Further changes will be required in glTF Transform to fully support f64 data, so I wouldn't expect features like mesh simplification or compression to work with f64 yet. See:

- https://github.com/donmccurdy/glTF-Transform/pull/1417

zeux:
> I assume that adding f64 support to EXT_meshopt_compression is technically possible, but not trivial

If filters aren’t used then it should just work except maybe some spec addendums if the strides are restricted currently. For translation type vectors filters are typically omitted or the exponential is used; it isn’t hard to make it work but I doubt it’s super useful because that implies a precision tradeoff and f64 implies the opposite.

Without filters but with attribute codecs on f64 you will see some meaningful compression iff the values have less entropy than the format allows: for example I would expect reasonable results if f64 encodes f32 values, and minimal compression (10%?) for random values in 0..1 interval.

zeux:
The coupling between animation usage and mesh/instance usage in this extension is a little unfortunate.

WebGL, WebGL2, WebGPU don't support double precision buffers or use of doubles in shaders. On desktop you do get some level of support through native APIs although it's conditional on hardware support and fairly sparsely represented in the API (eg in D3D12 there's no way to load 64-bit floats from a buffer directly, so you have to synthesize it from 2 32-bit halves manually).

Overall I would expect that the 99.99% path for a renderer to support this extension for accessors that require GPU access is to decode f64->f32 on load. This is obviously possible, and for some loaders this will not be that hard because they don't work with buffer views directly and assume a "accessor -> GPU buffer" conversion model anyhow, but that does make life rather more difficult for renderers that map bufferViews to GPU directly, as you need to analyze accessors that refer to that.

On the flip side, choosing not to support this extension means you can't support f64 property animation, which might be undesirable. This extension could also help for some cases where you need eg instance positions to be f64 and you're using mesh_gpu_instancing extension purely as a compression technique (rendering instances one by one on CPU) or have special careful shader plumbing to actually support this via a GPU path...

I almost wish that this extension would at least not apply to mesh attributes but maybe there's some reasons to favor that.

lexaknyazev:
> that does make life rather more difficult for renderers that map bufferViews to GPU directly, as you need to analyze accessors that refer to that.

The sparse accessors feature from the base glTF 2.0 spec already requires implementations to be able to read/write individual attribute values.

NorbertNopper-Huawei:
@lexaknyazev We are in interest of having an EXT_accessor_float16 (or a KHR_) extension.
We are also aware of this proposal for other data types: https://github.com/KhronosGroup/glTF/pull/2395

As @spnda worked on both extensions, can we somehow align what potential next steps could be?

From our side, our main interest is in having a half-float extension and I think it would make sense to have several additional data types extensions, however to be discussed at once to avoid fragmentation.

emackey (approved):
I agree with earlier concerns that several implementations may want to pick this up for the sake of 64-bit animation keyframes, but not be willing to take 64-bit vertex attributes at the same time. But I guess it's easy enough for an implementation to just report an error saying they don't support the latter when found.  The draft looks good to me, and some sample models might be a good idea before moving to Release Candidate status.

NorbertNopper-Huawei:
> I agree with earlier concerns that several implementations may want to pick this up for the sake of 64-bit animation keyframes, but not be willing to take 64-bit vertex attributes at the same time. But I guess it's easy enough for an implementation to just report an error saying they don't support the latter when found. The draft looks good to me, and some sample models might be a good idea before moving to Release Candidate status.

I assume, there are no objections. Will share the pull request internally. Huawei is mainly interested in the fp16 one.

NorbertNopper-Huawei:
One more detail: Both extensions are still in "Draft" @emackey maybe we review this first?

lexaknyazev:
> Both extensions are still in "Draft"

We need sample assets and at least one implementation to update the status.

NorbertNopper-Huawei:
Who is expected to create these assets?

javagl:
There are some tests shown in and linked from https://github.com/KhronosGroup/glTF/pull/2397#issuecomment-2118802415 above. But based on the points mentioned above, it could make sense to create models that _explicitly_ use these new types in specific contexts (e.g. "_only_ for animation but _not_ for vertex attributes", or "_both_, animation _and_ vertex attributes").

I have not yet looked at the updated state here (i.e. my review is still pending). During my first pass, there wasn't too much to say, and I assume that not sooo much has changed since that last pass, but I'll try to allocate some time to review the latest changes.

NorbertNopper-Huawei:
> There are some tests shown in and linked from [#2397 (comment)](https://github.com/KhronosGroup/glTF/pull/2397#issuecomment-2118802415) above. But based on the points mentioned above, it could make sense to create models that _explicitly_ use these new types in specific contexts (e.g. "_only_ for animation but _not_ for vertex attributes", or "_both_, animation _and_ vertex attributes").
> 
> I have not yet looked at the updated state here (i.e. my review is still pending). During my first pass, there wasn't too much to say, and I assume that not sooo much has changed since that last pass, but I'll try to allocate some time to review the latest changes.

To play god's advocate, you also create an example for `half-float/float16`?

extensions/2.0/Khronos/KHR_accessor_float64/README.md
```diff
@@ -33,40 +33,37 @@ Because this extension does not provide a way to specify alternative data repres
...
+| `5130`          | _double_  | Signed |  64  |
 
-Such accessors **MUST** adhere to the following additional rules:
+The `normalized` property of such accessors **MUST NOT** be set to true.
 
-- their `normalized` property **MUST NOT** be set to true;
-- their effective byte offsets, as defined in [Section 3.6.2.4](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#data-alignment), **MUST** be aligned to 8-byte boundaries.
-
-When referring to data in a GLB-stored buffer, the start of the buffer **SHOULD** be aligned to an 8-byte boundary by adding necessary trailing padding bytes to the JSON chunk as defined in [Section 4.4.3.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#structured-json-content).
+When a _double_ accessor refers to a buffer view that refers to a GLB-stored buffer, the start of the buffer **SHOULD** be aligned to an 8-byte boundary by adding necessary trailing padding bytes to the previous chunks as defined in the [GLB File Format Specification](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#chunks).
```
javagl, line 44: This seems to be the most significant change from the normative perspective: Changing the alignment requirement from MUST to SHOULD. It's hard to foresee implications for implementors in all environments/languages. I hope that it does not imply that some JavaScript implementation will have to do something like
`if (aligned(data, 8)) use(data); else use(copyTheWholeThingToBeAligned(data));`

extensions/2.0/Khronos/KHR_accessor_float64/README.md
```diff
@@ -76,15 +73,15 @@ When the extension is supported, _double_ component type accessors are valid for
 
 ## Compatibility
 
-Since not all graphics environments support double-precision floating-point data, implementations **MAY** employ different lossy strategies to render such assets, e.g., converting mesh attribute data to single-precision values when uploading data to GPUs.
+Since not all graphics environments support double-precision floating-point data, implementations **MAY** reject assets using double-precision data for mesh attributes or any other purpose that would require native hardware support.
```
javagl, line 76: While I see the point of this change (in view of the previous discussion and the distinction between e.g. animation data and actual (GPU) vertex attributes), it raises the question of which conditions have to be met for an implementation to claim that it *"supports"* this extension. 

This may not necessarily be in the scope of the spec text. But could it make sense to say something like

- implementations that want to support this extension MUST accept if everywhere
- BUT if they need to convert to single-precision (for hardware support reasons), then they MUST warn the user

? (Details TBD)

Otherwise, there is no way to know (for a content creator) whether the extension can be used at all, because it might work in all cases, and then an asset happens to use it for an obscure morphed `TEXCOORD_42`, and this suddenly breaks...

javagl (commented):
I only looked at the changes in the `float64` part from the previous state, and assume that `float16` is largely copy+paste. **Iff** there is something specific to review in `float16`, drop me a note.

javagl:
@HuaweiNopper It probably would make sense to have the same basic test models for both extensions. I can try to create a `float16` version (although this is usually not supported on the CPU, specifically not in Java, I can probably just manually encode some `float` into a `short` and shove it into a `ShortBuffer`...)

extensions/2.0/Khronos/KHR_accessor_float64/README.md
lexaknyazev: That change happened long time ago, see https://github.com/KhronosGroup/glTF/pull/2397#discussion_r1594373340.

The practical implications are as follows:
1. If an implementation keeps the original GLB file and only creates various views, then the BIN chunk may need to be aligned (it must in JavaScript). If it's not aligned, creating an `f64` view may fail (it will in JS).
2. If an implementation drops the original GLB file and keeps only the BIN chunk, e.g., by creating a copy, then chunk alignment is irrelevant.

Since the choice between 1 and 2 is implementation-specific, we can't assume anything here. On the other hand, it seems overly strict to require exporters to adjust alignment based on the extension usage.

It's important to have sample assets covering this issue.

lexaknyazev:
> although this is usually not supported on the CPU, specifically not in Java

Java has helpers for half-float operations since JDK 20.

extensions/2.0/Khronos/KHR_accessor_float64/README.md
lexaknyazev: > there is no way to know (for a content creator) whether the extension can be used at all

That's technically true but I'd say it falls into the same bucket as various other platform-specific limits, like max texture dimensions or the amount of attributes/vertices/indices/meshes in general.

Implementations may choose to reject the asset with double-precision attributes instead of converting mesh data because it's unknown whether such conversions are safe for a given asset. The spec even says that double-precision accessors should no be used unless absolutely necessary, which implies that authors should know exactly what they are doing.

> because it might work in all cases

That's unlikely in practice. GPU support (both hardware and API) for double-precision data and/or operations is far from pervasive. The main reason to allow double-precision mesh data accessors is to not block various DCC interchange and custom workflows that are less concerned with portability.

lexaknyazev:
> **Iff** there is something specific to review in `float16`, drop me a note.

The `f16` sections about animation samplers and compatibility are quite different than in `f64`.

extensions/2.0/Khronos/KHR_accessor_float64/README.md
javagl: I had seen that comment, but not yet the commit and actual change. 

Given that some details of how implementations _accomplish_ different degrees and forms of "support" for this extension are exactly that, _implementation-specific_, the spec should not constrain the options for content creators. Even if implementatons have to do specific alignment checks and perform a copy for proper alignment, they still can do this.

> It's important to have sample assets covering this issue.

Agreed. (I can probably create such assets as well (some details in the comment below), but ... it will need some tweaks. Right now, I'm trying to make sure that the data **is** aligned properly, and this includes automatisms to apply padding based on the accessor component type size - but I can probably disable that for such tests)

extensions/2.0/Khronos/KHR_accessor_float64/README.md
javagl: These are still different categories. There are always "_upper_ limits". Nobody would expect a web-based rendered to swallow a 1000000x1000000 pixel PNG texture. The point is that clients may implement the necessary `64`->`32` conversion for _some_ cases, and there is no reasonable way to communicate which ones this are. I don't know what the exact criteria are for this being "safe for a given asset" (maybe _that_ should be clarified?). Right now, clients could claim that they support the extension, but reject _all_ assets containing it, claiming that the conversion is never "safe"... (Playing devil's advocate, as usual...). 

I agree that this is so much implementation-dependent that the spec should not be overly restrictive. But I think that it could make sense to impose _some_ requirements for clients that claim to support it. The _conversion_ of the data itself should be fairly trivial to implement. Requiring a warning for the cases where the implementor deems it to be "not safe" sounds reasonable. (No _really_ strong opinion, though...)

NorbertNopper-Huawei:
> @HuaweiNopper It probably would make sense to have the same basic test models for both extensions. I can try to create a `float16` version (although this is usually not supported on the CPU, specifically not in Java, I can probably just manually encode some `float` into a `short` and shove it into a `ShortBuffer`...)

Since C++23, half floats are also available in the C++ standard: https://en.cppreference.com/w/cpp/types/floating-point.html

javagl:
> Java has helpers for half-float operations since JDK 20.

I didn't have that on the radar. (I'm still mainly using Java 8 👴 )

---

> The f16 sections about animation samplers and compatibility are quite different than in f64.

I had another look at this. The main differences seem to be 

- the claim that the interpolation itself must be unaffected by the `float16` type, 
- no aligment requirement issues
- the "Compatibility" is a bit simpler for the `float16` case 
 
(The latter two somewhat related to the inlined comments)

No specific comments about that.

---

About test models: I created some flavors of the `AnimatedTriangle` model. The file name suffix indicates whether it uses the new accessor for the `POSITION`, for the animation accessors, or for both. (No file suffix are the baselines that do not really use the extensions)

[AnimatedTriangle-accessor_floatN-2025-10-23.zip](https://github.com/user-attachments/files/23100224/AnimatedTriangle-accessor_floatN-2025-10-23.zip)

From the inlined comments:

> It's important to have sample assets covering this issue.

As mentioned there, my tooling currently aims at getting the alignment _right_. For example, the one where `float64` is used for the animations (but not the positions):

Khronos float 64 alignment

The animation data starts at byte 48 and not 44. But maybe I can create dedicated tests for this with the data starting at byte 44.

extensions/2.0/Khronos/KHR_accessor_float64/README.md
lexaknyazev: I think the current spec language covers that well enough from the engines' point of view:
> or any other purpose that would require native hardware support

Double-precision floats are nearly universally supported by CPUs but barely supported by GPUs. This implies that using `f64` for animations should be fine but mesh data using it would be less portable. An engine that claims to support the extension but rejects all assets with it can do so only when running on a platform without any `f64` support (neither GPU nor CPU) and that's probably the right thing to do.

> safe for a given asset

This is very content-specific. I could think of three cases:
1. Double-precision attributes are used by accident, i.e., actual data is perfectly representable with single-precision values. In this case, it's likely a tooling issue that may be eventually caught by the validator.
2. Double-precision attributes are used to exactly preserve original values in some DCC workflow but they are not required in practice, i.e., the values still fit in the `f32` range and precision differences are not noticeable. If such assets are used for wide distribution, it's an authoring issue.
3. Double-precision attributes are used intentionally because either the intended values do not fit into `f32` range or the increased precision is actually required, e.g., different vertex positions would be equal after converting them to `f32`.

The first case is safe, the second is maybe safe, and the third is not safe. I don't think engines could distinguish between these.

javagl: That clarifies the "safe". I think that the spec text may leave a bit too many degrees of freedom, but that may also be adequate given the variety of platforms and possible usages of the extension

spnda:
> > @HuaweiNopper It probably would make sense to have the same basic test models for both extensions. I can try to create a `float16` version (although this is usually not supported on the CPU, specifically not in Java, I can probably just manually encode some `float` into a `short` and shove it into a `ShortBuffer`...)
> 
> Since C++23, half floats are also available in the C++ standard: https://en.cppreference.com/w/cpp/types/floating-point.html

Since the `<stdfloat>` header is optional it is barely supported by any compiler, even less support IEEE754 types that are not supported by the target hardware. GCC supports 32-bit and 64-bit since GCC 13, and MSVC supports 32-bit and 64-bit since 17.7. As far as I know nobody supports 16-bit anywhere, but I might be wrong.

I think your best bet is just to rely on x86/armv8 intrinsics for specific instruction set extensions that add support for binary16 data types, if you even want hardware accelerated processing over software floats.

NorbertNopper-Huawei:
> > > @HuaweiNopper It probably would make sense to have the same basic test models for both extensions. I can try to create a `float16` version (although this is usually not supported on the CPU, specifically not in Java, I can probably just manually encode some `float` into a `short` and shove it into a `ShortBuffer`...)
> > 
> > 
> > Since C++23, half floats are also available in the C++ standard: https://en.cppreference.com/w/cpp/types/floating-point.html
> 
> Since the `<stdfloat>` header is optional it is barely supported by any compiler, even less support IEEE754 types that are not supported by the target hardware. GCC supports 32-bit and 64-bit since GCC 13, and MSVC supports 32-bit and 64-bit since 17.7. As far as I know nobody supports 16-bit anywhere, but I might be wrong.
> 
> I think your best bet is just to rely on x86/armv8 intrinsics for specific instruction set extensions that add support for binary16 data types, if you even want hardware accelerated processing over software floats.

It is more about creating the examples. Furthermore, in most cases the data can be passed 1:1 to the GPU.

If needed to be processed on the CPU, e.g. GCC supports float16 since GCC 13:
https://gcc.gnu.org/projects/cxx-status.html#cxx23

spnda:
> It is more about creating the examples. Furthermore, in most cases the data can be passed 1:1 to the GPU.

I understand that this is just for examples, I was just trying to say that it might not be feasible as an example since the support isn't quite there yet so some might have issues running it or testing it. Yes, a GPU does support this but a lot animation processing often still happens on the CPU in many renderers.

> If needed to be processed on the CPU, e.g. GCC supports float16 since GCC 13: https://gcc.gnu.org/projects/cxx-status.html#cxx23

I was wrong about it outright not supporting it, yes. However, I checked and it can effectively only use it as a storage data type, since the x87 instructions don't support it. All processing converts them to binary32 first, then calls the instruction on those, and then converts the resulting binary64 float back to binary16: https://godbolt.org/z/EbvsTKh1f. So it is actually kinda usable to my surprise, but brings no real gain for most platforms.

NorbertNopper-Huawei:
> > It is more about creating the examples. Furthermore, in most cases the data can be passed 1:1 to the GPU.
> 
> I understand that this is just for examples, I was just trying to say that it might not be feasible as an example since the support isn't quite there yet so some might have issues running it or testing it. Yes, a GPU does support this but a lot animation processing often still happens on the CPU in many renderers.
> 
> > If needed to be processed on the CPU, e.g. GCC supports float16 since GCC 13: https://gcc.gnu.org/projects/cxx-status.html#cxx23
> 
> I was wrong about it outright not supporting it, yes. However, I checked and it can effectively only use it as a storage data type, since the x87 instructions don't support it. All processing converts them to binary32 first, then calls the instruction on those, and then converts the resulting binary64 float back to binary16: https://godbolt.org/z/EbvsTKh1f. So it is actually kinda usable to my surprise, but brings no real gain for most platforms.

We mainly request the `KHR_accessor_float16` extension to save the overall size of the glTF file and pass directly to the GPU e.g. using GL_HALF_FLOAT in an OpenGL context. Our purpose is not for e.g. interactivity and mainly for data size reduction.

lexaknyazev:
> brings no real gain for most platforms

It reduces storage (and potentially memory usage) so it's not that bad even for animations.

> We mainly request the `KHR_accessor_float16` extension to save the overall size of the glTF file

Do you have any specific use case (that could be shared) showing `f16` benefits over 16-bit quantization? FWIW, `unorm16` provides generally better precision than `float16` at the same storage cost. The only drawback is that the actual mesh size would have to be handled via node transforms.

spnda:
> We mainly request the `KHR_accessor_float16` extension to save the overall size of the glTF file and pass directly to the GPU

> It reduces storage (and potentially memory usage) so it's not that bad even for animations.

Yes, that's why I wrote the spec to begin with 😂.

My point about it being "no real gain" is the fact that using `std::float16_t` on the CPU side brings no gain over just using `uint16_t`, or even just plain `std::byte`, and passing that unchanged to the GPU. And if you do want to use it on the CPU just do the conversion yourself (GCC also just calls `__extendhfsf2`) and work with the traditional floating point types. I am not questioning the validity of these extensions one bit.

NorbertNopper-Huawei:
> Do you have any specific use case

In the Standards team we are not deeply involved in the uses cases. It is more like we want to  have `VK_FORMAT_R16_SFLOAT` or `GL_HALF_FLOAT` in glTF.

NorbertNopper-Huawei:
@lexaknyazev I am currently in discussion with my colleagues.
Is it possible that we can have two pull requests for `KHR_accessor_float16 `and `KHR_accessor_float64`?
We are mainly interested on the `KHR_accessor_float16` extension at the moment.

lexaknyazev:
> Is it possible that we can have two pull requests

What's the point in that? We'll ratify them together anyway.

NorbertNopper-Huawei:
> > Is it possible that we can have two pull requests
> 
> What's the point in that? We'll ratify them together anyway.

Why are the ratified at the same time?

On a first sight, the extensions look similar. However, they are causing dependencies on the hardware and maybe the 3D Formats group decides to have e.g. a `KHR_accessor_float16` or `EXT_accessor_float64` extension for some reasons.

zeux:
I’m wondering what the status of these extensions is and is release candidate/ratification expected any time soon? Primarily interested in the float16 accessor extension out of these two.

NorbertNopper-Huawei:
> I’m wondering what the status of these extensions is and is release candidate/ratification expected any time soon? Primarily interested in the float16 accessor extension out of these two.

From Huawei side, float16 also has the higher priority.

Issue #2401, opened 2024-05-20
Visibility function glyph not rendered in PDF


emackey:
In section B.3.2, following #2386 and #2400, there is a glyph that renders in HTML but not in PDF.

Here's a screenshot of Specification.html:

![good screenshot](https://github.com/KhronosGroup/glTF/assets/1235080/20244b47-4255-4933-9ac2-beaa0ea87ef8)

And here's Specification.pdf:

![image](https://github.com/KhronosGroup/glTF/assets/1235080/e1062335-890b-48eb-adf6-bd1664320bcb)

emackey:
@oddhack Do you happen to know any quick fix for the missing glyph here?  Asciidoc source is:

https://github.com/KhronosGroup/glTF/blob/6014140040039d0bc25547fa1255569a1e31c6d1/specification/2.0/Specification.adoc?plain=1#L3061

oddhack:
It is not trivial. The font used by asciidoctor-pdf when rendering math is (intentionally) incomplete  - see https://github.com/asciidoctor/asciidoctor-pdf/issues/1832 and related PR https://github.com/asciidoctor/asciidoctor-pdf/pull/1864, wherein I got them to add the special characters needed by Vulkan to the distributed font. That took several months (some of it taken up by a long digression into licensing issues, which are resolved now), and in the meantime I was building my own fallback font with the additional characters included for use in the Vulkan build.

You could open an issue about this on their repo - it is a very small change to the build script once you known the needed Unicode value - but TBH you are probably better off just using a different character that's already in the PDF font. Maybe Greek nu (ν)? I infer the problem arose when noticing that 'V' was being used for two different things in that equation so the actual LHS is probably pretty arbitrary.

emackey:
@oddhack Thanks as always for the insights into the documentation mechanics here, it's much appreciated!

PR #2403, opened 2024-05-22
Fix missing glyph

emackey:
This is a simple search-and-replace of `\mathcal{V}` with `\nu` as a fix for the PDF render issue in #2401.

@bsdorra @proog128 Any objection to changing the letter like this?


emackey:
Here's a screenshot of the rendered PDF on this branch, it's much better than in #2401.  The rendered HTML looks correct as well.

screenshot

proog128:
Looks good!

PR #2405, opened 2024-05-23
EXT_mesh_gpu_instancing: Suggest extensionRequired usage

### LINKED ISSUE #2402: EXT_mesh_gpu_instancing: used vs required?
The spec for EXT_mesh_gpu_instancing doesn't mention the expected usage pattern: there are some extensions that modify the appearance of the scene in a way that can be insignificant if unsupported -- for example, for a lot of the new material extensions a renderer that doesn't support them will just render a slightly more basic version of the object, so I'd expect most material extensions to be used- but not required-.

Some extensions like Meshopt or Basis allow required- style usage with no fallback if desired, and optionally support a fallback for renderers that don't support the extension (with associated penalties on load time / size / memory impact) in which case the extension would be used- but not required-.

It's not clear to me what the expectations are for EXT_mesh_gpu_instancing. If extension is not supported, the rendered result will omit all instanced meshes and instead render a single instance of the basic mesh with a wrong transform. There isn't really a useful way that I can see (please correct me otherwise) to specify a fallback.

However, the instanced example here https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/SimpleInstancing doesn't list the extension as required. Any renderer or tool that doesn't support the extension will render a single cube at origin.

Should the extension spec recommend that scenes that use this extension put it in `requiredExtensions`? Should the model above be updated accordingly? Or am I missing some subtlety here?

zeux:
Using this extension without specifying it as required is guaranteed to lead to significant differences in appearance between renderers that support it and renderers that don't due to how it is specified.

Fixes #2402.

### DIFF
--- a/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md
+++ b/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md
@@ -57,6 +57,8 @@ All attribute accessors in a given node **must** have the same `count`.
 
 > **Implementation Note:** When instancing is used on the node, the non-instanced version of the mesh should not be rendered.
 
+Because the extension does not provide a way to specify a non-instanced fallback, files that use the extension should specify it in `extensionsRequired` array.
+
 ## Transformation Order
 
 When using instancing, the instance transform matrix is constructed by multiplying translation (if present), rotation (if present), and scale in the same order as they are for nodes:


extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md
```diff
@@ -57,6 +57,8 @@ All attribute accessors in a given node **must** have the same `count`.
 
 > **Implementation Note:** When instancing is used on the node, the non-instanced version of the mesh should not be rendered.
 
+Because the extension does not provide a way to specify a non-instanced fallback, files that use the extension should specify it in `extensionsRequired` array.
```
lexaknyazev, line 60: Please use formatting for "should" here.

bghgary:
There are two sample models in this folder. Should we update it with `extensionsRequired`?

https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/samples/teapots_galore/teapots_galore.gltf#L286-L288

https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/samples/teapots_galore_id/teapots_galore_id.gltf#L320-L322

zeux:
There’s an instancing example (SimpleInstancing) so I don’t know if we need a second one; not sure what the policy is wrt having examples in this repository. I plan to submit a PR for SimpleInstancing if this gets merged but I didn’t realize there are examples in this repository as well…

extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md
zeux: Updated, thanks!

bghgary:
@zeux Sorry, my question isn't directed at you specifically, though you are welcome to comment also of course. Maybe @lexaknyazev and others can chime in.

> not sure what the policy is wrt having examples in this repository

I'm not sure either. I think this might be the only extension with sample assets inside the extension folder itself.

lexaknyazev:
That's 100% an oversight. Samples should be in their own repo.

javagl:
I agree that the models should not be here. 
(Moving them to `glTF-Sample-Assets` will require some work (just to make clear that it's not CTRL-X/V), and ... I think that only the one _with_ the `_ID` should be enough...)

zeux:
Please let me know if any further changes are needed on this PR (or if there are no plans to merge it).

PR #2409, opened 2024-06-10
Update anisotropy extension spec

### LINKED ISSUE #2407: Clarify anisotropy usage
The extension says:
> All meshes with materials that use anisotropy **SHOULD** supply `TANGENT` vectors as a mesh attribute. If `TANGENT` vectors are not supplied for such a mesh, the mesh **MUST** supply a `normalTexture`, and tangents are computed according to rules in the [Meshes section](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#meshes) of the glTF specification. Likewise when `TANGENT` vectors are not supplied for a mesh, the mesh **MUST NOT** supply different texture coordinates on the `normalTexture` and `anisotropyTexture`.

This section should have more details, specifically answering the following questions:
1. It's clear that two sets of tangent vectors are unwanted hence the requirement to not use two separate texture coordinate sets for normal and anisotropy textures. But why cannot anisotropy texture coordinates be used for tangent space derivation when a normal texture is not present?
2. What rules apply to the tangent space (and why) when an anisotropy texture is not present? Another section of the extension spec defines that texture sampling returns a constant value in this case; it clearly does not have any texture coordinates.

lexaknyazev:
Fixes #2407.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
@@ -64,18 +64,24 @@ Sample values:
 
 |                         | Type     | Description               | Required           |
 | ----------------------- | -------- | ------------------------- | ------------------ |
-| **anisotropyStrength**  | `number` | The anisotropy strength. When anisotropyTexture is present, this value is multiplied by the blue channel. | No, default: `0.0` |
-| **anisotropyRotation**  | `number` | The rotation of the anisotropy in tangent, bitangent space, measured in radians counter-clockwise from the tangent. When anisotropyTexture is present, anisotropyRotation provides additional rotation to the vectors in the texture. | No, default: `0.0` |
-| **anisotropyTexture**   | [`textureInfo`](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | The anisotropy texture. Red and green channels represent the anisotropy direction in [-1, 1] tangent, bitangent space, to be rotated by anisotropyRotation. The blue channel contains strength as [0, 1] to be multiplied by anisotropyStrength. | No |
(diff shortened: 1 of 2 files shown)


extensions/2.0/Khronos/KHR_materials_anisotropy/README.md
```diff
@@ -64,18 +64,24 @@ Sample values:
...
+The strength parameter is a dimensionless number in the range $[0, 1]$ and increases the roughness along a chosen direction. The default direction aligns with the tangent to the mesh as described in the glTF 2.0 specification, [Meshes section](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#meshes).
 
-All meshes with materials that use anisotropy **SHOULD** supply `TANGENT` vectors as a mesh attribute.  If `TANGENT` vectors are not supplied for such a mesh, the mesh **MUST** supply a `normalTexture`, and tangents are computed according to rules in the [Meshes section](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#meshes) of the glTF specification. Likewise when `TANGENT` vectors are not supplied for a mesh, the mesh **MUST NOT** supply different texture coordinates on the `normalTexture` and `anisotropyTexture`.
+A mesh primitive using an anisotropy material **MUST** have a defined tangent space, i.e., it **MUST** have `NORMAL` and `TANGENT` attributes or its base material **MUST** have a normal texture. When the mesh primitive does not have `NORMAL` or `TANGENT` vectors, they are computed as defined in the glTF 2.0 specification.
```
emackey, line 76: This seems good.  One detail the previous text offered is that users "**SHOULD**" prefer `TANGENT` over the normal texture.  Can that intention be preserved here?

emackey: Oh nevermind, it's below, I didn't read far enough.

emackey (approved):

PR #2410, opened 2024-06-12
KHR_node_visibility Draft Proposal

lexaknyazev:
As discussed in the Interactivity DTSG, there should be a way to control node visibility through `KHR_animation_pointer` and/or `KHR_interactivity` workflows.

### DIFF
--- a/extensions/2.0/Khronos/KHR_node_visibility/README.md
+++ b/extensions/2.0/Khronos/KHR_node_visibility/README.md
@@ -1,0 +1,63 @@
+# KHR_node_visibility
+
+## Contributors
+
+- Dwight Rodgers, Adobe
+- Peter Martin, Adobe
+- Emmett Lalish, Google
+- Alexey Knyazev, Independent
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension allows glTF animations and interactivity to control visibility of node hierarchies. It is intended for use in conjunction with `KHR_animation_pointer` and/or `KHR_interactivity` extensions.
+
+## Extending Nodes
+
+The `KHR_node_visibility` extension object is added to the objects within the `nodes` array. The extension object contains a single boolean `hidden` property. This value is mutable through JSON pointers as defined in the glTF 2.0 Asset Object Model and controls visibility of the node that contains it and all its children nodes recursively. A value of `true` causes all nodes below in the hierarchy to be omitted from display, even any nodes below that have a value of `false`.
+
+|            |   Type    |              Description              |       Required      |
+|------------|-----------|---------------------------------------|---------------------|
+| **hidden** | `boolean` | Specifies whether the node is hidden. | No, default: `false`|
+
+In other words, a node is visible if and only if its own `hidden` property is `false` and all its parents are visible. This allows a single change of a `hidden` property at a high level of the hierarchy to hide or show complex (multi-node) objects.
+
+In the following example, both nodes (and therefore their meshes) are initially hidden.
+
+```json
+{
+    "nodes": [
+        {
+            "children": [1],
+            "mesh": 0,
+            "extensions": {
+                "KHR_node_visibility": {
+                    "hidden": true
+                }
+            },
+        },
+        {
+            "mesh": 1
+        }
+    ]
+}
+```
+
+## Extending glTF 2.0 Asset Object Model
+
+The following pointer template represents the mutable property defined by this extension.
+
+| Pointer                                           |  Type  |
+|---------------------------------------------------|--------|
+| `/nodes/{}/extensions/KHR_node_visibility/hidden` | `bool` |
+
+## JSON Schema
+
+- [node.KHR_node_visibility.schema.json](schema/node.KHR_node_visibility.schema.json)
--- a/extensions/2.0/Khronos/KHR_node_visibility/schema/node.KHR_node_visibility.schema.json
+++ b/extensions/2.0/Khronos/KHR_node_visibility/schema/node.KHR_node_visibility.schema.json
@@ -1,0 +1,17 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_node_visibility glTF Node Extension",
+    "type": "object",
+    "description": "glTF extension that defines node's visibility.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "hidden": {
+            "type": "boolean",
+            "description": "Specifies whether the node is hidden.",
+            "default": false,
+            "gltf_detailedDescription": "Specifies whether the node is hidden. A value of true means that the node and all its children are hidden."
+        },
+        "extensions": { },
+        "extras": { }
+    }
+}


javagl (commented):
I'm lacking some context from the `KHR_animation_pointer`/`KHR_interactivity`, but this looks pretty straightforward.

Maybe `hidden` should be `required`, but when `default`ing to `false`, this may not be strictly necessary.

lexaknyazev:
> Maybe `hidden` should be `required`, but when `default`ing to `false`, this may not be strictly necessary.

Having a default value is exactly why it's not required. So to make node's visibility controllable, an empty extension object is all that's needed.

extensions/2.0/Khronos/KHR_node_visibility/README.md
```diff
@@ -0,0 +1,63 @@
+# KHR_node_visibility
```
emackey, line 1: Nitpick: We often escape underscores when outside of code blocks, such as this title string here.  In markdown, underscores are used to denote _italics_, so the above string may be rendered by some systems with _node_ in italics.  At some point GitHub markdown became smart enough to avoid mid-word italics like this, so the problem isn't always visible anymore.  But for future-proofing and compatibility with other markdown systems, I still recommend following what we've done in other extensions:

`# KHR\_node\_visibility`

emackey:
This sort of "Inverted Boolean" is a design pitfall that I usually lobby against, for the sake of non-programmer users who are sometimes confused by double-negatives.

Please take a moment and really think through the implications of changing this to parameter `visible`, default `true`, instead of parameter `hidden`, default `false`.  Think about the Boolean flag in isolation, where its value may appear in code or data separated from the input parameter name that appears here.  Normally a value of `true` appearing in code or data means _YES_, do it, show it, use it, go for it.  Normally a value of `false` means no, don't go, don't show, get that thing out of here.

Currently this draft's `hidden` parameter is asking users to accept a value of `true` as indicating "Don't show that now", and a value of `false` as "Yes, put it in there please."  That's backwards, and causes confusion wherever the value appears independently from the parameter being manipulated.

Apologies if this is just a pet peeve of mine.  I've been opposed to this kind of Boolean use since the late 90's when some software I work on sported a checkbox on the UI that looked like this:

- [ ] Don't overwrite the file

There was a similarly named `DontOverwriteFile` Boolean in the corresponding API.  How many users got confused by turning the checkbox off thinking it would keep their file safe from being overwritten, we'll never know, but we know a few were quite vocal to tech support about it.  I've been opposed to inverted Boolean design ever since.

javagl:
@emackey That was also the first thing that I thought. And I fully agree that this is an antipattern: Nobody wants things like `boolean doNotIgnoreNonPositiveExclusions` 🤪 . But I (have to) _assume_ that this was an intentional and conscious choice, most likely related to a form of backward compatibility for systems that do not know the extension. But I'd also be curious about the exact reasoning for *not* using a `visible` with `default:true`. In which cases could this cause trouble or have undesired effects?

lexaknyazev:
Consider the following:
- All nodes are visible in unextended glTF.
- An empty extension object must not change the default visibility.
- Subject to engine design, it may be an open question whether "hiding a node (tree)" is a positive or negative action.

Out of all possible combinations of the property direction (positive/negative) and the default value (true/false), only two options remain:
- `hidden` with default `false`
- `visible` with default `true`

An optional boolean JSON property has three states:
- `undefined`
- `true`
- `false`

If `undefined` and `true` are supposed to be semantically the same, then distinguishing between them may require special error-prone handling not needed for other glTF boolean properties found in the core spec. For example in JS:
```js
let ext1 = {visible: false}
let ext2 = {visible: true}
let ext3 = {};

// Naive check
if (ext1.visible) {} // ok
if (ext2.visible) {} // ok
if (ext3.visible) {} // wrong

// Inverted naive check
if (!ext1.visible) {} // ok
if (!ext2.visible) {} // ok
if (!ext3.visible) {} // wrong

// Accurate check
if (ext1.visible !== false) {} // ok
if (ext2.visible !== false) {} // ok
if (ext3.visible !== false) {} // ok

//

let ext4 = {hidden: false}
let ext5 = {hidden: true}
let ext6 = {};

// Naive check
if (ext4.hidden) {} // ok
if (ext5.hidden) {} // ok
if (ext6.hidden) {} // ok

// Inverted naive check
if (!ext4.hidden) {} // ok
if (!ext5.hidden) {} // ok
if (!ext6.hidden) {} // ok
```

The same logic applies to other languages that may not even have a concept of an undefined boolean. DCC tools do not have to present the flag literally so the double-negative UI could be easily avoided in user-facing tools.

emackey:
Yes.  This is the one I'm advocating for here:

* Property name `visible` with default value `true`

Yes, JavaScript and JSON have some interesting wrinkles when distinguishing between `undefined`, `null`, `false`, `0`, and the empty string.  My asking for a default value of `true` will feed into this problem, but it's well-known by JS developers (and I have experience with how these issues were routinely handled from my time on the early Cesium team and other WebGL projects).  We do already have a case where `undefined` defaults to infinity (attenuation distance if I recall), and another case where `undefined` defaults to `0.5` (which happens in `alphaMask` I believe).  So, this should not place any undue burden on developers.  It will be immediately apparent if some client application defaults visibility to `false` by accident.

>  it may be an open question whether "hiding a node (tree)" is a positive or negative action.

In your code samples, sure.  But I'm talking simple human understanding here.  I'm talking about people trying to learn foreign design details of unfamiliar standards.  There's no wiggle room here to say that the default behavior of displaying ordinary glTF nodes is somehow a negative action, and that hiding them from view is positive.  It's clearly inverted.  The cleaner design is to use `true` as the default state here for displaying visible nodes.

javagl:
> JavaScript ... _interesting wrinkles_

That's an _interesting_ euphemism for "irrecoverably broken clusterf..." 🙂 
(There are probably hundreds and thousands of `if (!something) throw "Something is undefined!!!111");` out there...)

---

I already suspected that the reasoning was in some form of "backward compatibility". And this indeed is the reason here - namely, via the statement

> An empty extension object must not change the default visibility.

which causes a problem (only) _in combination with_ the case of 

`let ext3 = {};` 
`if (!ext3.visible) {} // wrong`

Now... one could argue about that. One could call this "an implementation error". And something like this is less likely to happen for _experienced_ developers who know the ... wrinkles... of JS. However, there's certainly that trade-off between the more intuitive approach of `visible` with `default:true`, which is tailored for _humans_, and the less error-prone approach of `hidden` with `default:false`, which is tailored for _developers_. And I do understand the reasoning for both sides. 

If we chose to use the `visible`/`default:true` approach, then it might be possible to alleviate the problem by emphasizing the special case in the spec text. Maybe as an "Implementation Note" containing the relevant part of the example code shown above. But I don't have a really strong opinion here.

dwrodger:
The important thing from a functionality perspective is that it must be possible to hide a subtree of the nodes with a single value change, since we don’t provide constructs for iterating an entire tree, it would inefficient to do so, and it could lose important state at lower levels (such as where a diorama is to be made visible and the animation within the diorama involves flipping the visibility state of some diorama elements – we must be able to toggle on the visibility of the diorama as a whole while allowing the animation track to have unfettered control of visibility within the diorama).

So, if you have “visible” with default “true”, then the logic for whether any node should display must be “display node if and only if node and all parents have visible=true (including by default)”.

This means that a node X may be hidden despite having “visible=true”. To me, it feels slightly more intuitive to reason that “node X is visible even though hidden=false, because its parent is hidden” than it is to reason that “node X is invisible even though visible=true, because its parent is invisible”, but I don’t have a strong opinion here. Mostly I want to point out that either way people will have to learn the design detail of how the effect of the property propagates in the hierarchy, which is the most important thing.

Dwight

From: Ed Mackey ***@***.***>
Date: Thursday, June 13, 2024 at 8:49 AM
To: KhronosGroup/glTF ***@***.***>
Cc: Dwight Rodgers ***@***.***>, Review requested ***@***.***>
Subject: Re: [KhronosGroup/glTF] KHR_node_visibility Draft Proposal (PR #2410)

EXTERNAL: Use caution when clicking on links or opening attachments.



Yes. This is the one I'm advocating for here:

  *   Property name visible with default value true

Yes, JavaScript and JSON have some interesting wrinkles when distinguishing between undefined, null, false, 0, and the empty string. My asking for a default value of true will feed into this problem, but it's well-known by JS developers (and I have experience with how these issues were routinely handled from my time on the early Cesium team and other WebGL projects). We do already have a case where undefined defaults to infinity (attenuation distance if I recall), and another case where undefined defaults to 0.5 (which happens in alphaMask I believe). So, this should not place any undue burden on developers. It will be immediately apparent if some client application defaults visibility to false by accident.

it may be an open question whether "hiding a node (tree)" is a positive or negative action.

In your code samples, sure. But I'm talking simple human understanding here. I'm talking about people trying to learn foreign design details of unfamiliar standards. There's no wiggle room here to say that the default behavior of displaying ordinary glTF nodes is somehow a negative action, and that hiding them from view is positive. It's clearly inverted. The cleaner design is to use true as the default state here for displaying visible nodes.

emackey:
All good points.  And @lexaknyazev does show the correct code for checking this:

```js
// Accurate check
if (ext1.visible !== false) {} // ok
if (ext2.visible !== false) {} // ok
if (ext3.visible !== false) {} // ok
```

Yes, implementers will still need to pay attention to the details and the hierarchy.  But we can easily supply test asset(s) that will make it very plain to see if an entire node has gone missing or inappropriately shown.

The folks I'm concerned about are the adopters: The users, the artists, the designers.  They use this spec too, and they can't be kept in line with a simple unit test like programmers can be.  And this is an instance of a known design antipattern for them that seems easy to avoid.

I don't want to stand in the way of progress here, so if the consensus is to dismiss my objection, I can accept that.  But so far in this thread I have only seen, pardon my expression, programmer reasoning here, coder's logic, which is not always along the same lines as how the artists and creators like to think.

lexaknyazev:
Agreed within the Interactivity DTSG to use the "positive" property name provided that test assets for this extension will include multi-node objects relying on the implicit (default) property value.

javagl (commented):
I don't have a strong opinion for either `hidden` or `visible`. The trade off is that the latter is more "intuitive", and the former avoids that possible `if (!node.visible) shouldDefaultToTrue();` bug. (Maybe there could be an 'Implementation Note' about that - but that may not be too important). So the new approach is fine for me.

emackey (approved):

javagl:
A quick attempt to create a test asset:

[KHR_node_visibility test 0001 - 2024-06-26.zip](https://github.com/user-attachments/files/15993099/KHR_node_visibility.test.0001.-.2024-06-26.zip)

- When the extension is not supported, then it shows "The extension is not supported" (and nothing else).
- When `visible:undefined` defaults to `true`, then it shows "Does handle defaults properly" (and if it doesn't, it shows "Does not handle defaults properly").
- When the constraint that _a node is visible if itself and all its ancestors are visible_ is not met, then it shows "Does not properly handle hierarchy" (otherwise, it does not display this)

The behavior is "emulated" in BabylonJS here:

Khronos KHR visibility test

There's probably room for improvement, but maybe someone wants to give it a spin...

aaronfranke:
For content creation tools exporting glTF files, is it recommended that files using `KHR_node_visibility` put this in `extensionsRequired`? It would make sense considering that the model does not render correctly without it, and it is likely that the asset author explicitly needed it, but I wanted to ask here.

lexaknyazev:
> For content creation tools exporting glTF files, is it recommended that files using `KHR_node_visibility` put this in `extensionsRequired`?

That should depend on the intended usage. Couple examples:
- Visibility flag is actively used (toggled) by the animation pointer and/or interactivity extensions. The extension(s) should likely be required.
- Visibility flag is used just to temporarily hide something during DCC interchange. The extension should not be required and likely should not be included in the final asset at all.

aaronfranke:
I have implemented this in Godot Engine here: https://github.com/godotengine/godot/pull/93722. Both import and export are supported, so Godot Engine can be used for both authoring and consuming glTF assets with `KHR_node_visibility`. We decided to make the handling of visibility configurable on export, so that users can choose what they prefer.

Test file: [cube_visibility_example.zip](https://github.com/user-attachments/files/16037124/cube_visibility_example.zip)

extensions/2.0/Khronos/KHR_node_visibility/schema/node.KHR_node_visibility.schema.json
```diff
@@ -0,0 +1,17 @@
...
+    "title": "KHR_node_visibility glTF Node Extension",
+    "type": "object",
+    "description": "glTF extension that defines node's visibility.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "visible": {
+            "type": "boolean",
+            "description": "Specifies whether the node is visible.",
+            "default": true,
+            "gltf_detailedDescription": "Specifies whether the node is visible. A value of false means that the node and all its children are hidden."
```
aaronfranke, line 12: Nitpick

```suggestion
            "gltf_detailedDescription": "Specifies whether the node is visible. A value of false means that the node and all its descendants are hidden."
```

extensions/2.0/Khronos/KHR_node_visibility/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Extending Nodes
+
+The `KHR_node_visibility` extension object is added to the objects within the `nodes` array. The extension object contains a single boolean `visible` property. This value is mutable through JSON pointers as defined in the glTF 2.0 Asset Object Model and controls visibility of the node that contains it and all its children nodes recursively. A value of `false` causes all nodes below in the hierarchy to be omitted from display, even any nodes below that have a value of `true`.
+
+|             |   Type    |               Description              |      Required      |
+|-------------|-----------|----------------------------------------|--------------------|
+| **visible** | `boolean` | Specifies whether the node is visible. | No, default: `true`|
+
+In other words, a node is visible if and only if its own `visible` property is `true` and all its parents are visible. This allows a single change of a `visible` property at a high level of the hierarchy to hide or show complex (multi-node) objects.
```
aaronfranke, line 30:
```suggestion
In other words, a node is visible if and only if its own `visible` property is `true` and all its ancestors are visible. This allows a single change of a `visible` property at a high level of the hierarchy to hide or show complex (multi-node) objects.
```

aaronfranke (approved):

extensions/2.0/Khronos/KHR_node_visibility/schema/node.KHR_node_visibility.schema.json
javagl: I also considered to bring that up, but... it transitively applies to all relevant nodes, so this does not change the _meaning_ on a technical level. (One could argue which of them is easier to understand, conceptually, but I don't have a strong opinion there...)

extensions/2.0/Khronos/KHR_node_visibility/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Overview
+
+This extension allows glTF animations and interactivity to control visibility of node hierarchies. It is intended for use in conjunction with `KHR_animation_pointer` and/or `KHR_interactivity` extensions.
+
+## Extending Nodes
+
+The `KHR_node_visibility` extension object is added to the objects within the `nodes` array. The extension object contains a single boolean `visible` property. This value is mutable through JSON pointers as defined in the glTF 2.0 Asset Object Model and controls visibility of the node that contains it and all its children nodes recursively. A value of `false` causes all nodes below in the hierarchy to be omitted from display, even any nodes below that have a value of `true`.
+
+|             |   Type    |               Description              |      Required      |
```
aaronfranke, line 26: The titles here are center-aligned, but they are left-aligned in most files in this repo.

extensions/2.0/Khronos/KHR_node_visibility/README.md
```diff
@@ -0,0 +1,63 @@
...
+|-------------|-----------|----------------------------------------|--------------------|
+| **visible** | `boolean` | Specifies whether the node is visible. | No, default: `true`|
```
aaronfranke, line 28:
```suggestion
| **visible** | `boolean` | Specifies whether the node is visible. | No, default: `true` |
```

extensions/2.0/Khronos/KHR_node_visibility/README.md
lexaknyazev: The extension uses the same terms as the core spec.

DennisSmolek:
I have a question regarding this proposal, mostly with how other upcoming extensions would/could use it:

The current system deals with mesh/object visibility within the scene and assumes it will be a part of the systems scenegraph.

ThreeJS for example has `.visible` that [keeps the mesh in the graph](https://threejs.org/docs/#api/en/core/Object3D.visible) and its children as well. Thus in memory.

GLTFLoader (I believe) will also generate materials for these objects if not referenced.

The problem is, not all nodes will ever be visible and shouldn't be treated as such. For example, Colliders for physics systems. Convex Hulls, Trimeshes, even plain shapes from what I understand are being tracked as nodes.
These "invisible" items would still be processed as meshes, especially as some extensions propose putting them as children of parents to keep things like world space intact.

I'm looking at LOD as well potentially using visibility/availability flags on nodes/assets.

If a node/asset exists for other purposes (physics, interaction, other processing, lod, etc) but never renders, and still exists in the node tree, I think it may be helpful to mark the asset as such, and there's a difference between not CURRENTLY visible, and NEVER will be visible. No materials will be generated, UUIDS, etc etc.

I can't think of a good term, but `isReference` or something would work.

Without a flag will have to do checks/process them out some other way.


Sidenote re: falsy, I agree the `hidden: true` setup way better.

```ts
if (!node.hidden) resultingMesh.visible = false;
``` 
Is a better setup than being `visible: true` as it wouldn't break existing gltfs/loaders and keeps json (ever so slightly) smaller.

aaronfranke:
@DennisSmolek Nodes that will never be visible will not be affected by this property. A physics shape is not expected to be visible at runtime. Whether the visible property is true or false will not change that. What gives you the impression that they "would still be processed as meshes"? A physics shape is not the same as a visible mesh.

DennisSmolek:
> @DennisSmolek Nodes that will never be visible will not be affected by this property. A physics shape is not expected to be visible at runtime. Whether the visible property is true or false will not change that. What gives you the impression that they "would still be processed as meshes"? A physics shape is not the same as a visible mesh.

Consider:
```js
{
    "children": [1],
    "mesh": 1,
    "extensions": {
        "KHR_node_visibility": {
            "visible": false
        }
    },
}
```
        
Lets say a extension uses this node to hold a collider shape. A Trimesh or a convexHull. It is a mesh. It has mesh data. 
      
Currently (From what I understand), when threeJS process the node in `GLTFLoader` it would load it as a regular mesh and put it into the active scenegraph, albeit with the `visible: false`. property. 
It will get a material, a world position Matrix, it's buffer copied,  etc etc.
If this had a prop like `nonRenderable` this would be mitigated.
      
I'm going to be trying to get the physics extension working with three so saw some mention of this.
      
It's fine, I just have to add more logical processing for the meshes to check if they are LODs, Physics, etc.

aaronfranke:
@DennisSmolek Indeed, that is a concern with the recent changes to the KHR physics extension. However, note that this is not a problem with the old version of the KHR physics extension, or with the OMI physics extension. Document-level:

```json
{
    "extensions": {
        "OMI_physics_shape": {
            "shapes": [
                {
                    "type": "trimesh",
                    "trimesh": { "mesh": 1 }
                }
            ]
        }
    }
}
```

On a node:

```json
{
    "children": [1],
    "extensions": {
        "KHR_node_visibility": {
            "visible": false
        },
        "OMI_physics_body": {
            "collider": {
                "shape": 0
            }
        }
    }
}
```

The same structure [used to be possible in KHR_collision_shapes before it was refactored into KHR_implicit_shapes](https://github.com/eoineoineoin/glTF_Physics/commit/24e219fa4032f529ef2598c664305837facb8c15).

pfcDorn:
I ran into an issue with selectable nodes, where visibility is set to false. Should theses invisible nodes still be selectable? I saw other implementations (like Godot and BabylonJS), where only the visibility will be set on the nodes, but the nodes are still enabled (e.g. for physics). That would mean invisible nodes could block raycasts for other selectable nodes which are visible.  So does the term "visible" means pure visibility of a mesh or a general "enabled" state of the node?

aaronfranke:
@pfcDorn Visible is strictly for render visibility (meshes/particles/etc), it is not a general disable nodes feature.

As for selectability, if the application allows for selecting things outside of clicking on the visible mesh, then that should not be impacted by invisibility. Also if an editor has a scene hierarchy editor, the invisible nodes should still be shown in there.

extensions/2.0/Khronos/KHR_node_visibility/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension allows glTF animations and interactivity to control visibility of node hierarchies. It is intended for use in conjunction with `KHR_animation_pointer` and/or `KHR_interactivity` extensions.
```
bghgary: Do we need to update the [Object Model](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/ObjectModel.adoc) with this PR?

aaronfranke: The extension itself defines what object model pointers are available.

The `ObjectModel.adoc` file is only for the core spec and extensions created before the object model existed.

lexaknyazev: We'll update the object model file for reference after this is merged (just for convenience).

bghgary: I agree it's for convenience. It's very easy to miss adding this to the object model in implementations like we did in Babylon.js, though it's less of a problem with the sample assets.

bghgary:
> A quick attempt to create a test asset:
> 
> [KHR_node_visibility test 0001 - 2024-06-26.zip](https://github.com/user-attachments/files/15993099/KHR_node_visibility.test.0001.-.2024-06-26.zip)
> 
> * When the extension is not supported, then it shows "The extension is not supported" (and nothing else).
> * When `visible:undefined` defaults to `true`, then it shows "Does handle defaults properly" (and if it doesn't, it shows "Does not handle defaults properly").
> * When the constraint that _a node is visible if itself and all its ancestors are visible_ is not met, then it shows "Does not properly handle hierarchy" (otherwise, it does not display this)
> 
> The behavior is "emulated" in BabylonJS here:
> 
> Khronos KHR visibility test [ Khronos KHR visibility test ](https://private-user-images.githubusercontent.com/5597569/343269027-e8b66cad-bce4-4199-b9b1-b4b1526c3e75.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjE1ODYzMTUsIm5iZiI6MTc2MTU4NjAxNSwicGF0aCI6Ii81NTk3NTY5LzM0MzI2OTAyNy1lOGI2NmNhZC1iY2U0LTQxOTktYjliMS1iNGIxNTI2YzNlNzUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MTAyNyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTEwMjdUMTcyNjU1WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YzgyZjc1OTUzNzZiZTRiMmNmNTc0YTE4NjgxODEwNWNmOWRjYTUzMTgxMmM4ZDczYmE0NTBjOGU0Yzg2Y2E4NiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.ivMh_J3NOVM1LVd_zjjqx8kcEFYR9MAX_p2mRhlQUec) [ ](https://private-user-images.githubusercontent.com/5597569/343269027-e8b66cad-bce4-4199-b9b1-b4b1526c3e75.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjE1ODYzMTUsIm5iZiI6MTc2MTU4NjAxNSwicGF0aCI6Ii81NTk3NTY5LzM0MzI2OTAyNy1lOGI2NmNhZC1iY2U0LTQxOTktYjliMS1iNGIxNTI2YzNlNzUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MTAyNyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTEwMjdUMTcyNjU1WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YzgyZjc1OTUzNzZiZTRiMmNmNTc0YTE4NjgxODEwNWNmOWRjYTUzMTgxMmM4ZDczYmE0NTBjOGU0Yzg2Y2E4NiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.ivMh_J3NOVM1LVd_zjjqx8kcEFYR9MAX_p2mRhlQUec)
> 
> There's probably room for improvement, but maybe someone wants to give it a spin...

Hi @javagl, we are using this asset to test in Babylon.js. I think it would be great to make this test available in the glTF sample assets. There were a couple of small things we talked about in the WG to improve this asset:
1. Add green check marks / red Xs to the textures to make it obvious whether it is working or not.
1. Add one more level to the parent hierarchy to test ancestors and not just direct parents.

Is this something you would be interested in changing?

javagl:
@bghgary 

> Is this something you would be interested in changing?

Can do (I might not be able to do this in the next few days, but will try to carve out some time)

extensions/2.0/Khronos/KHR_node_visibility/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+Release Candidate
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension allows to control visibility of node hierarchies. It is intended for use in conjunction with `KHR_animation_pointer` and/or interactivity extensions.
```
aaronfranke: Should we also note that it can be used by itself? Because it can be.

extensions/2.0/Khronos/KHR_node_visibility/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Extending Nodes
+
+The `KHR_node_visibility` extension object is added to the objects within the `nodes` array. The extension object contains a single boolean `visible` property. This value is mutable through JSON pointers as defined in the glTF 2.0 Asset Object Model and controls visibility of the node that contains it and all its children nodes recursively. A value of `false` causes all nodes below in the hierarchy to be omitted from display, even any nodes below that have a value of `true`.
+
+When a node is not visible, all its visual features including but not limited to meshes, light sources (e.g., attached with `KHR_lights_punctual`), point clouds, particles, billboards, volumetric effects, etc., are not rendered. Visibility affects neither cameras, nor node's interactivity features such as selectability or hoverability.
```
aaronfranke, line 26: Nitpick: Prefer English words over Latin initialisms.

```suggestion
When a node is not visible, all its visual features including but not limited to meshes, light sources (such as attached with `KHR_lights_punctual`), point clouds, particles, billboards, volumetric effects, and so on, are not rendered. Visibility affects neither cameras, nor node's interactivity features such as selectability or hoverability.
```

javagl:
@bghgary An updated test is attached here.

> Add one more level to the parent hierarchy to test ancestors and not just direct parents.

I wasn't entirely sure about the exact shape for this. The previous asset had the two cases
``` 
- node (visible:true)
  - child with mesh (visible:false)
```
and
``` 
- node (visible:false)
  - child with mesh (visible:true)
```
with the mesh showing the message "Does not properly handle hierarchy", and being expected to **not** be displayed (because either the node containing it or its parent had 'visible:false').

I now just inserted another node in-between
``` 
- node (visible:false)
  - child (visble:false)
    - grandchild with mesh (visible:true)
```
and
``` 
- node (visible:true)
  - child (visible:true)
    - grandchild with mesh (visible:false)
```
But I wondered whether there should also be "true-false-true" and "false-true-false" chains (or even more of the 8 possible combinations)...? 
(I'd probably have to think more closely about what _exactly_ is supposed to be tested there in a rendering engine...)

[KHR_node_visibility test 0002 - 2025-10-30.zip](https://github.com/user-attachments/files/23222079/KHR_node_visibility.test.0002.-.2025-10-30.zip)

It's only the "test 0002", so we can iterate on that 😁

extensions/2.0/Khronos/KHR_node_visibility/README.md
lexaknyazev: The base glTF spec, extensions, and other current Khronos specs use Latin abbreviations a lot. We could eventually make a style decision to stop using them though.

javagl:
@bghgary Depending on whether there already are specific points about that "0002" version that should be fixed, just drop me a note. Otherwise, I can allocate some time to create a "cleaned up" version of that and create a PR into glTF-Sample-Assets.

bghgary:
Thanks @javagl!

> But I wondered whether there should also be "true-false-true" and "false-true-false" chains (or even more of the 8 possible combinations)...?

I think testing all 8 combinations is overkill. I would just test "false-true-true" to make sure "parent of a parent" is working. @lexaknyazev You made this suggestion during the WG call. Do you have any thoughts?

> Otherwise, I can allocate some time to create a "cleaned up" version of that and create a PR into glTF-Sample-Assets.

Since this PR is already merged, I think it would be good to open a new PR in glTF-Sample-Assets with the current state (even without any cleanup) so that we can continue the discussion there?

javagl:
@bghgary Opened https://github.com/KhronosGroup/glTF-Sample-Assets/pull/244 with the current state from here.

Whether the hierarchy test cases should _only_ cover the `false-true-true` case can be discussed there.

bghgary:
Thanks @javagl!

PR #2421, opened 2024-07-02
KHR audio framework design proposal

cashah:
During the recent 3D Formats working group meeting, we reviewed the [proposal](https://docs.google.com/presentation/d/1IrrQaE-jHyzOtFRabjtLAzeP5UOirFOEj8FRADAceqk/edit#slide=id.g27310d50ada_0_104) to define the KHR audio glTF specification using an audio graph framework. The purpose of the current document is to delve deeper into that proposal, offering a comprehensive design of the KHR audio graph. This includes a detailed description of each node object within the graph along with functionality, the specific properties associated with it, and how it interacts with other nodes in the graph. The document is structured to facilitate clear understanding and to solicit feedback on the proposed design. Based on the feedback we will update and finalize the design before it is formally schematized into KHR core audio spec, extensions, animations, and interactivity.

### DIFF
--- a/extensions/2.0/Khronos/KHR_audio_graph/README.md
+++ b/extensions/2.0/Khronos/KHR_audio_graph/README.md
@@ -1,0 +1,1279 @@
+<!-----
+
+
+
+Conversion time: 1.967 seconds.
+
+
+Using this Markdown file:
+
+1. Paste this output into your source file.
+2. See the notes and action items below regarding this conversion run.
+3. Check the rendered output (headings, lists, code blocks, tables) for proper
+   formatting and use a linkchecker before you publish this page.
+
+Conversion notes:
+
+* Docs to Markdown version 1.0β36
+* Fri Jun 14 2024 17:25:52 GMT-0700 (PDT)
+* Source doc: KHR Audio Graph Design
+* Tables are currently converted to HTML tables.
+----->
+
+
+
+## KHR Audio Graph Design
+
+
+
+
+## **Contributors**
+
+
+
+* Chintan Shah, Meta
+* Alexey Medvedev, Meta
+
+
+## Context
+
+During the recent Khronos 3D formats working group meeting held on 5/29, we reviewed the [proposal to define the KHR audio glTF specification using an audio graph framework](https://docs.google.com/presentation/d/1IrrQaE-jHyzOtFRabjtLAzeP5UOirFOEj8FRADAceqk/edit?usp=sharing). The purpose of this document is to delve deeper into that proposal, offering a comprehensive design of the KHR audio graph. This includes a detailed description of each node object within the graph along with functionality, the specific properties associated with it, and how it interacts with other nodes in the graph. The document is structured to facilitate clear understanding and to solicit feedback on the proposed design. Based on the feedback we will update and finalize the design before it is formally schematized into KHR core audio spec, extensions, animations, and interactivity.
+
+
+## 1. Introduction
+
+This document provides a detailed design for managing audio routing, mixing, and processing in various applications for desktop, mobile, and wearable devices. The core idea involves an audio graph consisting of multiple interconnected audio node objects to create the final audio output. This design seeks to incorporate the audio capabilities found in modern web, game, and XR engines as well as processing, mixing, and filtering functions available in audio production softwares. Although this system is designed with diverse use cases in mind, it might not cover every specialized feature found in state-of-the-art audio tools. In such instances, users are encouraged to develop custom extensions. Nevertheless, the proposed system will support many complex audio applications by default and has been designed to facilitate future expansion with more sophisticated features.
+
+
+## 2. Features
+
+The core specification should support these primary features:
+
+
+
+* Graph based audio routing for simple or complex mixing and processing architectures.
+* Processing of audio data stored in memory buffer or accessed via file paths.
+* Capturing audio metadata such as encoding properties.
+* Audio playback functionalities including playing, stopping, pausing, looping, and controlling playback speed.
+* Spatialized audio supporting a wide range of 3D applications and immersive environments with 6DoF source/listener capabilities, panning models (equal power, HRTF), distance attenuation, and sound cones.
+* Basic audio signal processing to control gain, delay, and pitch.
+* Flexible handling of channels in an audio stream, allowing splitting, merging, up-mixing, or down-mixing.
+* Audio mixing, reverb, and filtering with a set of low-order audio filters.
+* Animation control and dynamic update of node properties.
+
+
+## 3. Graph based audio processing
+
+Audio nodes are the building blocks of an audio graph for rendering audio to the audio hardware. Graph based audio routing allows arbitrary connections between different audio node objects. An audio graph can be represented by audio sources, the audio destination/sink, and intermediate processing nodes. Each node can have inputs and/or outputs. A source node has no inputs and a single output. A destination or sink node has one input and no outputs. In the simplest case, a single source can be routed directly to the output.
+
+One or more intermediate processing nodes such as filters can be placed between the source and destination nodes. Most processing nodes will have one input and one output. Each type of audio node differs in the details of how it processes or synthesizes audio. But, in general, an audio node will process its inputs (if it has any), and generate audio for its outputs (if it has any). An output may connect to one or more audio node inputs, thus fan-out is supported. An input (except source and sink) may be connected to one or more audio node outputs, thus fan-in is supported. Each input and output has one or more channels. The exact number of channels depends on the details of the specific audio node.
+
+
+## 4. Audio source
+
+
+### 4.1 Source node (0 input / 1 output)
+
+Audio sources reference audio data and define playback properties for it. An audio source node has no inputs and exactly one output, which has the same number of channels as indicated in encoding properties.
+
+
+<table>
+  <tr>
+   <td>
+   </td>
+   <td><strong>Type</strong>
+   </td>
+   <td><strong>Description</strong>
+   </td>
+   <td><strong>Required</strong>
+   </td>
+   <td><strong>Notes</strong>
+   </td>
+  </tr>
+  <tr>
+   <td>data
+   </td>
+   <td>object
+   </td>
+   <td>audio or oscillator
+<p>
+See 4.2 Audio data.
+<p>
+See 4.3 Oscillator data.
+   </td>
+   <td>
+   </td>
+   <td>Only selective source node properties apply with oscillator data.
+   </td>
+  </tr>
+  <tr>
+   <td>priority
+   </td>
+   <td>integer
+   </td>
+   <td>Determines the priority of this audio source among all the ones that coexist in the scene (0 = most important, 256 = least important, default = 128).
+   </td>
+   <td>
+   </td>
+   <td>Need to  persist and propagate priority with downstream processing.
+   </td>
+  </tr>
+  <tr>
+   <td>state
+   </td>
+   <td>string
+   </td>
+   <td>Playback state (paused, playing, stopped).
+   </td>
+   <td>
+   </td>
+   <td>
+   </td>
+  </tr>
+  <tr>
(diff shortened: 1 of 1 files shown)


rudybear (approved):

extensions/2.0/Khronos/KHR_audio_graph/README.md
```diff
@@ -0,0 +1,1279 @@
...
+   </td>
+   <td><strong>Notes</strong>
+   </td>
+  </tr>
+  <tr>
+   <td>type
+   </td>
+   <td>string
+   </td>
+   <td>Shape in which emitter emits audio (cone, omnidirectional, custom).
```
docEdub, line 655: Why is `custom` included here? Is there a way of defining a custom shape? How would it be implemented?

docEdub, line 655: Does `omnidirectional` need to be included when it can be achieved with a 360-degree cone angle?

extensions/2.0/Khronos/KHR_audio_graph/README.md
```diff
@@ -0,0 +1,1279 @@
...
+   <td>
+   </td>
+   <td>
+   </td>
+  </tr>
+</table>
+
+
+
+### 4.3 Oscillator data
```
docEdub, line 303: Has band-limiting been considered for audio rate oscillator implementations? Is it assumed? Or does it need to be explicitly required? I would think it should be explicitly required if audio quality is a priority.

extensions/2.0/Khronos/KHR_audio_graph/README.md
```diff
@@ -0,0 +1,1279 @@
...
+### 6.7 Audio mixer node (N inputs / 1 output)
+
+Use the Audio mixer node to combine the output from multiple audio sources. Number of channels should be the same across all inputs.
+
+[no properties for this node]
+
+
+### 6.8 Filter node (1 input / 1 output)
+
+Use the Audio Mixer node to combine the output from multiple audio sources. A filter node always has exactly one input and one output.
```
docEdub, line 999: This looks to be partially copied from the previous item by mistake.

extensions/2.0/Khronos/KHR_audio_graph/README.md
```diff
@@ -0,0 +1,1279 @@
...
+   <td>
+   </td>
+   <td>
+   </td>
+  </tr>
+</table>
+
+
+
+### 6.9 Reverb node (1 input / 1 output)
```
docEdub, line 1079: Reverb implementations vary widely. Has any thought been given to consistency across implementations for this node? Web-based implementations would have to create their own reverb nodes since there is no WebAudio reverb node, yet. Would a `Convolver` node make more sense?

extensions/2.0/Khronos/KHR_audio_graph/README.md
```diff
@@ -0,0 +1,1279 @@
...
+----->
+
+
+
+## KHR Audio Graph Design
+
+
+
+
+## **Contributors**
```
aaronfranke, line 30: Headings shouldn't have any additional formatting symbols.
```suggestion
## Contributors
```

extensions/2.0/Khronos/KHR_audio_graph/README.md
```diff
@@ -0,0 +1,1279 @@
...
+
+
+
+* Chintan Shah, Meta
+* Alexey Medvedev, Meta
```
aaronfranke, line 35: The Khronos preferred Markdown style is to use `-` for unordered lists:
```suggestion
- Chintan Shah, Meta
- Alexey Medvedev, Meta
```

extensions/2.0/Khronos/KHR_audio_graph/README.md
```diff
@@ -0,0 +1,1279 @@
...
+
+Conversion time: 1.967 seconds.
+
+
+Using this Markdown file:
+
+1. Paste this output into your source file.
+2. See the notes and action items below regarding this conversion run.
+3. Check the rendered output (headings, lists, code blocks, tables) for proper
+   formatting and use a linkchecker before you publish this page.
```
aaronfranke, line 13: Is this leftover from something? It should be removed.

extensions/2.0/Khronos/KHR_audio_graph/README.md
```diff
@@ -0,0 +1,1279 @@
...
+
+
+
+* Chintan Shah, Meta
+* Alexey Medvedev, Meta
+
+
+## Context
+
+During the recent Khronos 3D formats working group meeting held on 5/29, we reviewed the [proposal to define the KHR audio glTF specification using an audio graph framework](https://docs.google.com/presentation/d/1IrrQaE-jHyzOtFRabjtLAzeP5UOirFOEj8FRADAceqk/edit?usp=sharing). The purpose of this document is to delve deeper into that proposal, offering a comprehensive design of the KHR audio graph. This includes a detailed description of each node object within the graph along with functionality, the specific properties associated with it, and how it interacts with other nodes in the graph. The document is structured to facilitate clear understanding and to solicit feedback on the proposed design. Based on the feedback we will update and finalize the design before it is formally schematized into KHR core audio spec, extensions, animations, and interactivity.
```
aaronfranke, line 40: Please include the year and use ISO 8601 to make this unambiguous to future readers.

```suggestion
During the recent Khronos 3D formats working group meeting held on 2024-05-29, we reviewed the [proposal to define the KHR audio glTF specification using an audio graph framework](https://docs.google.com/presentation/d/1IrrQaE-jHyzOtFRabjtLAzeP5UOirFOEj8FRADAceqk/edit?usp=sharing). The purpose of this document is to delve deeper into that proposal, offering a comprehensive design of the KHR audio graph. This includes a detailed description of each node object within the graph along with functionality, the specific properties associated with it, and how it interacts with other nodes in the graph. The document is structured to facilitate clear understanding and to solicit feedback on the proposed design. Based on the feedback we will update and finalize the design before it is formally schematized into KHR core audio spec, extensions, animations, and interactivity.
```

aaronfranke (changes requested):
There is a lot missing from this PR. Is KHR_audio_graph implemented anywhere? Are there any sample assets? Where are the JSON schemas?

Please define the relationship between KHR_audio_graph and KHR_audio_emitter. An audio graph is too complicated for most scenes and engines, which often only need to define basic emission. The Google Slides document linked in KHR_audio_graph mentions Unity DSPGraph, which is an extra package, not part of the base Unity. I wrote a longer list of specific discussion points here: https://github.com/KhronosGroup/glTF/pull/2137#issuecomment-2197388138

extensions/2.0/Khronos/KHR_audio_graph/README.md
aaronfranke: Note: This is how [KHR_audio_emitter](https://github.com/KhronosGroup/glTF/pull/2137) works.

aaronfranke: It would be implemented by another glTF extension, extending this extension.

extensions/2.0/Khronos/KHR_audio_graph/README.md
cashah: Good callout, extended support for convolution reverb in addition to algorithmic reverb.

extensions/2.0/Khronos/KHR_audio_graph/README.md
cashah: Keeping it explicit to make it runtime friendly.

extensions/2.0/Khronos/KHR_audio_graph/README.md
cashah: Made that explicit by including a detune parameter. The implementation can then use this with frequency to calculate the desired oscillator frequency.

extensions/2.0/Khronos/KHR_audio_graph/README.md
cashah: Updated.

extensions/2.0/Khronos/KHR_audio_graph/README.md
docEdub: Ok, thanks, but I don't think adding a `detune` parameter has anything to do with making band-limiting explicit. If it's assumed that implementations will apply band-limiting to avoid aliasing, then that's fine, but I suspect many implementations will take the easier route of skipping band-limiting if it's not required, which will reduce audio quality.

It may be that this concern is outside the scope of the standard, but if there a lot of poor-quality implementations then it may cause creators to avoid using it, so we might want to make band-limiting required.

extensions/2.0/Khronos/KHR_audio_graph/README.md
docEdub: Ok, thanks, but this doesn't address my concern about the algorithmic reverb sounding different across implementations. I don't think sound designers are going to use a reverb if it doesn't sound the same everywhere, so how do we get the algorithmic reverb to sound the same on all implementations? Can a specific reverb algorithm be required by the standard, like maybe a basic plate reverb?

cashah: Got it. Unfortunately, there is no universal standard for reverb implementation, which can vary significantly in core business logic and its complexity. Many implementations remain unpublished. Therefore, the objective is not to restrict reverb to a particular implementation but to rather parameterize it sufficiently to accommodate most implementations. Requiring a specific reverb algorithm as a standard would, in fact, diminish the value for users. That said, one can always leverage convolution reverb with a specific impulse response.

extensions/2.0/Khronos/KHR_audio_graph/README.md
cashah: There are various practical methods an implementation might use to prevent aliasing. While the ideal discrete-time digital audio signal is clearly defined mathematically, each implementation must balance the trade-off between computational cost and fidelity to this ideal. It is anticipated that implementations will strive towards this ideal, but it is also reasonable to consider lower-quality, more cost-effective approaches for less powerful hardware. We can include more guidance/recommendation around this.

extensions/2.0/Khronos/KHR_audio_graph/README.md
docEdub: > Requiring a specific reverb algorithm as a standard would, in fact, diminish the value for users.

Maybe, but I think users won't create graphs with algorithmic reverbs because they won't know how they'll sound on all implementations. Instead, they'll bake the exact reverb they want directly into the audio sources. They'll do this with any audio nodes that vary significantly across implementations.

cashah: >because they won't know how they'll sound on all implementations 

That's a valid point, and it extends beyond just different implementations to the same implementation being used across various device types, such as mobiles, desktops, and HMDs. Each device type, and indeed each variant within those types, can potentially alter the sound characteristics, making it challenging for users to predict how the reverb will sound without testing it on each device themselves. Moreover, developers often face a trade-off between computational cost and fidelity in their implementations. This balance is crucial to ensure that the reverb not only meets quality expectations but also remains computationally feasible and customizable across diverse hardware platforms.

andybak:
Is there any documentation or records of discussions that covers the "why" rather than the "what"?

i.e the case for _why_ this belongs in GLTF, discussions on challenges related to implementation and adoption and any exploration of the costs vs the benefits.

I have no doubt this conversations happened - are they published anywhere?

Igor-dvr:
> Is there any documentation or records of discussions that covers the "why" rather than the "what"?
> 
> i.e the case for _why_ this belongs in GLTF, discussions on challenges related to implementation and adoption and any exploration of the costs vs the benefits.
> 
> I have no doubt this conversations happened - are they published anywhere?

There is a video https://youtu.be/5hXk_NKv3FE?t=1353 that mentions audio extensions including 
KHR_audio_emitter and  KHR_audio_graph

PR #2422, opened 2024-07-03
KHR_node_selectability Draft Proposal

lexaknyazev:
As discussed in the Interactivity DTSG.

### DIFF
--- a/extensions/2.0/Khronos/KHR_node_selectability/README.md
+++ b/extensions/2.0/Khronos/KHR_node_selectability/README.md
@@ -1,0 +1,107 @@
+# KHR\_node\_selectability
+
+## Contributors
+
+- Dwight Rodgers, Adobe
+- Peter Martin, Adobe
+- Emmett Lalish, Google
+- Alexey Knyazev, Independent
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension allows glTF animations and interactivity to control selectability of node hierarchies. It is intended for use in conjunction with `KHR_animation_pointer` and/or `KHR_interactivity` extensions, although it may be used independently.
+
+## Concepts
+
+### Selecting / Select action
+
+The concept of “selecting” in this extension refers to a user interaction with an object in the scene. Typical examples of a “selection” action would be: a user clicking a mouse button while the mouse is somewhere over the 2D render of the scene, a user tapping with a finger on a point in the 2D render of the scene, a user using a six degree of freedom controller to point in a 3D scene and then pressing a trigger or other action button to indicate their intent to select, or a user using finger and hand gestures.
+
+### Selection Ray
+
+In all cases, we assume that a “selection” indicates that the user wishes to interact with a node in the scene that is along a specific ray, for example along the ray from the camera through the pixel and out into the scene, or along the ray from the controller out to the scene.
+
+This ray will be called the “selection ray”.
+
+### Selectability
+
+In this section of this specification we will use the term “object” to refer to the geometry of a specific node (not including its subtree), whether defined by a mesh on that node or by any future extensions.
+
+It is usually assumed that the user wishes to interact with the first object that intersects the selection ray. However, this may not always be the case. For example, a mostly transparent object representing and atmospheric effect like mist may be intended only for appearance, and a selection ray ought to skip this object and terminate at the first object thereafter.
+
+The concept of “selectability” defines whether the selection ray should terminate at a given object. A selection action should be interpreted as applying to the first “selectable” object along the ray, with all earlier “unselectable” objects being skipped, regardless of the opacity, visibility or appearance of the objects.
+
+## Extending Nodes
+
+The `KHR_node_selectability` extension object is added to the objects within the `nodes` array. The extension object contains a single boolean `selectable` property. This value is mutable through JSON pointers as defined in the glTF 2.0 Asset Object Model and controls “selectability” of the node that contains it and all its children nodes recursively. A value of `false` causes all nodes below in the hierarchy to be omitted from “selection”, even any nodes below that have a value of `true`.
+
+|                | Type      | Description                               | Required            |
+|----------------|-----------|-------------------------------------------|---------------------|
+| **selectable** | `boolean` | Specifies whether the node is selectable. | No, default: `true` |
+
+In other words, a node is “selectable” if and only if there exists no node including or above it in the hierarchy where such node has a `selectable` property with value `false`. This allows a single change of a `selectable` property at a high level of the hierarchy to enable or disable the selection of complex (multi-node) objects.
+
+In the following example, both nodes (and therefore their meshes) are initially unselectable.
+
+```json
+{
+    "nodes": [
+        {
+            "children": [1],
+            "mesh": 0,
+            "extensions": {
+                "KHR_node_selectability": {
+                    "selectable": false
+                }
+            },
+        },
+        {
+            "mesh": 1
+        }
+    ]
+}
+```
+
+## Extending glTF 2.0 Asset Object Model
+
+The following pointer template represents the mutable property defined by this extension.
+
+| Pointer                                                  | Type   |
+|----------------------------------------------------------|--------|
+| `/nodes/{}/extensions/KHR_node_selectability/selectable` | `bool` |
+
+## Interaction with KHR_interactivity
+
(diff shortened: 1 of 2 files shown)


javagl (commented):
Looks straightforward to me. No real objections.

I cannot say much about the connection to `KHR_interactivity` on a low technical level. 

On a high level, I'd mention possible alternatives for the output values: When there is a `selectionRayOrigin`, one could consider some `selectionPointDistance` (i.e. the distance along the ray), or a `selectionRayDirection`, or a `selectionLine` or so. But there are some ways of deriving one from the other...

- `selectionLine = selectionPoint - selectionRayOrigin`
- `selectionRayDirection = normalize(selectionLine)`
- `selectionPoint = selectionRayOrigin + selectionLine`
- `selectionPointDistance = distance(selectionRayOrigin, selectionPoint)`
- `selectionLine = selectionPointDistance * selectionRayDirection`
- `selectionPoint = selectionRayOrigin + selectionPointDistance * selectionRayDirection`
- ...

Which one is "the best"? I don't know. Others might even want to throw in some "picked triangle" and the barycentric coordinates of the hit point in there...

In the past, I did model the 'result of a picking operation' as `result = { ray, distance, pickedObject }` with `ray = { origin, direction }`, but others may have different views on that.

---

A detail that _might_ already be covered by some general "event handling" specification part of `KHR_interactivity` (I still have to finish reading that one)...: 

> This interactivity event node is activated when a “select” event occurs on a glTF node nodeIndex or on any node in its subtree subject to the following propagation rule: the lowest node in the tree receives the select event first, and the event bubbles up the tree until a glTF node with an associated event/onSelect behavior graph node with its stopPropagation configuration value set to true is found.

When `stopPropagation` is false, then the event bubbles up further, and that single operation could cause _multiple_ elements to be selected with a single operation - is that correct?

lexaknyazev:
> But there are some ways of deriving one from the other...

Sure. Two points seem to be the most atomic and fundamental values in this interconnected system. Various math nodes could be used to trivially derive everything else.

> that single operation could cause _multiple_ elements to be selected with a single operation

Yes if they have event "listeners", i.e., `event/onSelect` nodes with the corresponding `nodeIndex` configuration values.

extensions/2.0/Khronos/KHR_node_selectability/README.md
```diff
@@ -0,0 +1,107 @@
...
+Unless used in conjunction with `KHR_interactivity`, the result of a “select” action on a given object is left up to the implementation.
+
+When used in conjunction with `KHR_interactivity`, a new interactivity event is defined as follows:
+
+|                         |                  | |
+|-------------------------|------------------|-|
+| **Type**                | `event/onSelect` | Select event |
+| **Configuration**       | `int nodeIndex`        | Index of a node |
+|                         | `bool stopPropagation` | Whether to allow parents of this node to also receive the event |
+| **Output values**       | `int selectedNodeIndex`  | Index of the actual node that was selected |
```
aaronfranke, line 92: What is the distinction between the configuration `int nodeIndex` and the output value `int selectedNodeIndex`? Wouldn't they just be set to the same value?

The description below defines `selectedNodeIndex` is initialized to -1, so is the distinction that `selectedNodeIndex` can either be -1 or set to `nodeIndex`?

dwrodger: nodeIndex is in the configuration, meaning that when you put an onSelect node into your node graph, you choose a nodeIndex to listen to. selectedNodeIndex is an output, meaning that it is a value that is set at runtime when the onSelect event occurs. If you set your nodeIndex to a node that has no children, then selectedNodeIndex will always be set to nodeIndex. If you set your nodeIndex to be a node that has children, then selectedNodeIndex may be any node (that can accept select events) within the subtree of nodeIndex, depending on which specific node was selected.

aaronfranke: Ah, I see, that makes sense. So it allows selections to "bubble up", reaching an event listening for an ancestor, but still point to a specific child. Thanks!

extensions/2.0/Khronos/KHR_node_selectability/README.md
```diff
@@ -0,0 +1,107 @@
...
+|                         | `float3 selectionPoint`   | Position of intersection of the select ray with the geometry of the selected node in global space |
+|                         | `float3 selectionRayOrigin` | Position of the origin of the select ray in global space |
```
emackey, line 94: In an earlier section, this was defined as the "selection ray", not the "select ray".

```suggestion
|                         | `float3 selectionPoint`   | Position of intersection of the selection ray with the geometry of the selected node in global space |
|                         | `float3 selectionRayOrigin` | Position of the origin of the selection ray in global space |
```

emackey:
Just a general question here from someone who hasn't been involved in the interactivity discussions: Do we need to assume a "selection ray" in this extension?  I can easily imagine VR/XR scenarios where I could tap a "selection point" in space, or drag out a selection box of some kind, or use some not-as-yet-invented controller to indicate selection of a node by more futuristic means.  Must it always be a ray?

extensions/2.0/Khronos/KHR_node_selectability/README.md
emackey: So, maybe rename `nodeIndex` to `listenerNodeIndex`?

lexaknyazev: That's implied by the value being an _event_ configuration. Let's clarify the description instead.

extensions/2.0/Khronos/KHR_node_selectability/README.md
lexaknyazev: Good catch!

lexaknyazev:
> tap a "selection point" in space, or drag out a selection box of some kind

This sounds like new, more advanced events.

---

We could probably allow returning NaN if the exact coordinates cannot be provided for some reason. @dwrodger WDYT?

dwrodger:
> > tap a "selection point" in space, or drag out a selection box of some kind
> 
> This sounds like new, more advanced events.
> 
> We could probably allow returning NaN if the exact coordinates cannot be provided for some reason. @dwrodger WDYT?

Yes, I think that NaN in the case that the ray can't be defined sounds fine. It may also be appropriate to update the language where it defines selection and what it means for an object to be invisible to selection. That language could just say that implementations that use systems other than ray-based selection are free to interpret "invisible to selection" in whatever way makes sense for their selection mechanics.

emackey (approved):

javagl (commented):
I don't have substantial comments beyond what I already wrote in the first review. If the review was about the latest changes: I'm lacking some technical context there (e.g. about the 'controllers'). From a _structural_ perspective, I could only brainstorm whether the absence of a point/origin should be modelled with NaN, or whether that should be differentiated into a "selection event" (without these properties) and a "picking event" (with these properties).

hybridherbst:
Maybe as additional input here, in WebXR each controller has a "ray pose" and a "grip pose" and both are very much needed for different use cases. 

For example, visuals for controller models are usually aligned with the "grip pose", and subsequently the action of "dragging an object" often is based on the "grip pose". The "ray pose" or "aim pose" is needed for clicking on things.

These two so far have survived a number of novel interaction mechanisms, including things like Apple Vision Pro where suddenly pointers are transient (they exist only temporarily while the user is interacting) and have very different "ray pose" (ray from the eye to where the user is looking) and "grip pose" (point and orientation for where in space the user has started the hand gesture for the selection).

hybridherbst:
Reading a bit more in the spec, I wonder about this section:
> In the case of multiple-controller systems, the controllerIndex output value MUST be set to the index of the controller that generated the event; in single-controller systems, this output value MUST be set to zero.

Maybe it should read "the unique ID of the controller that has generated the event" instead? The index of a controller can change; for example, in touch-based systems usually each new touch has a new pointer ID (see https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerId). Or in WebXR, where users can connect and disconnect new controllers arbitrarily (e.g. switch from controllers to hands and back) or have transient pointers as well.

8vama:
>In other words, a node is “selectable” if and only if there exists no node including or above it in the hierarchy where such node has a selectable property with value false.

Are nodes by default selectable, even if the node doesn't have the extension? It was ambiguous in the spec. 

From a performance point of view, it would be great if this is opt-in only and default to false.

PR #2451, opened 2024-10-26
Add link to Chinese version of overview

javagl:
Follow-up from https://github.com/javagl/gltfOverview/pull/11

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ Please provide spec feedback by submitting [issues](https://github.com/KhronosGr
     <a href="https://www.khronos.org/developers/reference-cards/">Khronos reference cards</a>
   </sup><br>
   <sup>
-    <a href=https://github.com/randall2835/gltfOverviewJapanese/releases/download/v0.1/gltfOverview2.0-Japanese.png>Japanese translation</a> by <a href=https://github.com/randall2835/gltfOverviewJapanese>Takuto Takahashi</a>
+    <a href=https://github.com/randall2835/gltfOverviewJapanese/releases/download/v0.1/gltfOverview2.0-Japanese.png>Japanese translation</a> by <a href=https://github.com/randall2835/gltfOverviewJapanese>Takuto Takahashi</a><br>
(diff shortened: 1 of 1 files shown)


emackey:
(This may be out-of-scope for this PR, but I'm going to ask anyway...)

The Japanese and Chinese translation links here lead to direct downloads.  It might be more friendly to link to the GitHub listing, or some kind of preview page where one could observe the material without being committed to save it in a local file and open from there.  Would anyone object to changing these links to be something that would allow users to preview in their browsers as opposed to initiating a download?

javagl:
Yeah, the "official" overview [GitHub releases](https://github.com/javagl/gltfOverview/releases/tag/v2.0.0d) do include the "...small.png" that could be suitable for that (and is copied here to be shown in the README, with additional compression). These small ones don't exist for the Japanese and Chinese versions. 

Even if the links here pointed to "...small.png" versions, people might be confused when clicking them ("Why so small?"). Maybe letting the links point directly to the "releases" at https://github.com/randall2835/gltfOverviewJapanese/releases/tag/v0.1 and https://github.com/leeyeel/gltfOverview-Chinese/releases/tag/v2.0.0d could be better. Otherwise: Do you have a preference where the "...small.png" versions could reside, and how to make clear that these are only _previews_ for the real ones? (And where to include the PDFs - for zooming and/or printing, these are actually more important than the large PNGs...)

javagl:
(Completely unrelated note to myself: https://github.com/KhronosGroup/glTF?tab=readme-ov-file#for-developers links to glTF-Sample-Models - this has been archived for a while, and should point to glTF-Sample-Assets. I'll fix this, but not now, to avoid concurrent changes to the README)

emackey:
+1 for the "Middle ground" option you mentioned.  It seems good to me.

javagl:
@emackey Updated the links to point to the releases, and added the respective `| PDF |` link. 

Direct preview:

https://github.com/javagl/glTF/tree/add-chinese-overview-links?tab=readme-ov-file#overview

PR #2453, opened 2024-10-31
KHR_materials_volume_scatter (replaces #1928)

bsdorra:
The extension in this PR defines volumetric scattering and serves as an replacement/alternative to #1928

## Changes wrt. #1928
* Naming
Choosing "KHR_materials_volume_scatter" over "KHR_materials_subsurface" clarifies that this extension addresses general volumetric scattering and that it functions only in combination with a real volume (by using KHR_materials_volume). The term "subsurface" scattering is often linked with dense scattering scenarios, such as wax or skin, which can be misleading - dense scattering is just one specific configuration of this extension (also see section on scattering + diffuse transmission).

* Scatter albedo replaces scatter coefficient
In #1928, scatter color and distance are used to specify the scattering coefficient directly. This PR proposes using multi-scatter albedo instead, aligning more closely with approaches in [OpenPBR (subsurface)](https://academysoftwarefoundation.github.io/OpenPBR/#model/basesubstrate/subsurface) and [Enterprise PBR](https://dassaultsystemes-technology.github.io/EnterprisePBRShadingModel/spec-2025x.md.html#components/volume). Using scatter albedo also better harmonizes with the definition and naming of attenuation parameters in KHR_materials_volume.

* Adding scatter anisotropy

## Discussion
There has already been some discussion in the PBR TSG on whether to use single-scatter albedo or multi-scatter albedo. Multi-scatter albedo, being a derived parameter, requires conversion to single-scatter albedo. The PR currently suggests the conversion method proposed by [Kulla & Conty](https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides_v2.pdf). What happens if potential future alternative conversions prove more effective? Comments suggested that it is possible to keep the normative part of the multi-scatter albedo abstract and declare the actual equation non-normative.

Currently, volume parameters are not allowed to be texturable. OpenPBR does permit this. As @proog128 notes, should we also decide to allow this at some point, for texturable scatter albedo, the conversion would need to be computed per pixel within the shader.


DennisSmolek:
I personally don't care what you call it, but MaterialX standardSurface, Disney BSDF, Blender Principled, OpenPBR etc. All call it subsurface scattering.

Trying to build bridges between systems works easiest (both mentally and technically) when systems align.

If they are fundamentally different then I totally get it, but if they align, it may be operationally easier to call it the same thing..

bsdorra:
There's a conceptual difference. The mentioned material models define subsurface scattering as a surface based effect, i.g. a BSSRDF. Some of them additionally define volume scattering (e.g. the transmission_scatter properties in the translucent base of OpenPBR). However, in reality, subsurface scattering is simply dense volume scattering, making separate parameter sets somewhat redundant. 

This proposal aims to generalize and eliminate redundancy by introducing just a single parameter set that defines the physical behavior of a volumetric medium using scatter and attenuation properties (KHR_materials_volume). The name KHR_materials_volume_scatter should emphasize that it is a pure volume feature, that depends on KHR_materials_volume.

Renderers may not support actual volume scattering or may have optimized techniques for dense scattering. [This section](https://github.com/KhronosGroup/glTF/blob/e17468db6fd9ae3ce73504a9f317bd853af01a30/extensions/2.0/Khronos/KHR_materials_volume_scatter/README.md#scattering--diffuse-transmission) explains how combining this extension with KHR_materials_diffuse_transmission can help signal a dense (subsurface) scattering scenario.

extensions/2.0/Khronos/KHR_materials_volume_scatter/README.md
```diff
@@ -0,0 +1,237 @@
...
+This cue allows renderers to use diffusion approximation rather than random walk subsurface scattering, producing near-ground-truth results for dense materials with greater efficiency, essential for both real-time and offline rendering. [Christensen and Burley (2015)](#ChristensenBurley2015) demonstrate the mapping of physical parameters to diffusion-based reflectance profiles, while [Jimenez et al. (2015)](#Jimenez2015) introduce a real-time method using a separable kernel for reflectance profile rendering.
+
+<figure style="text-align:center">
+<img src="./figures/transmission-translucency.png"/>
+<figcaption><em>Comparison of combining denes volume scattering with either transmission or diffuse transmission. Left: Rough transmission and scattering. Middle: Diffuse transmission and scattering. Right: Diffuse transmission without scattering using a thin-walled material. Colors have been adjusted manually in order to account for differences in distances and to minimize the impact of energy loss from the rough microfacet BTDF.</em></figcaption>
+</figure>
+
+## Schema
+
+- [glTF.KHR_materials_subsurface.schema.json](schema/glTF.KHR_materials_subsurface.schema.json)
```
emackey: The schema filename here should begin with `material.*` instead of `glTF.*` now. This is a relatively recent change applied to other material schemas in this repo.

extensions/2.0/Khronos/KHR_materials_volume_scatter/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+* [Christensen, P. and B. Burley (2015): Approximate Reflectance Profiles for Efficient Subsurface Scattering](https://graphics.pixar.com/library/ApproxBSSRDF/paper.pdf)<a name="ChristensenBurley2015"></a>
+* [Louis G. Henyey and Jesse L. Greenstein: Diffuse radiation in the galaxy. Astrophysical Journal 93, 70–83, 1941.](https://ui.adsabs.harvard.edu/abs/1941ApJ....93...70H/abstract)<a name="HenyeyGreenstein"></a>
+* [Jimenez J., K. Zsolnai, A. Jarabo, C. Freude, T. Auzinger, X.-C. Wu, J. Pahlen, M. Wimmer and D. Gutierrez (2015): Separable Subsurface Scattering](http://www.iryoku.com/separable-sss/)<a name="Jimenez2015"></a>
+* [Kulla C., Conty A. (2017): Revisiting Physically Based Shading at Imageworks](https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides_v2.pdf)<a name="KullaConty2017"></a>
+
+
+## Appendix: Full Khronos Copyright Statement
+
+Copyright 2018-2020 The Khronos Group Inc.
```
emackey: 2025

extensions/2.0/Khronos/KHR_materials_volume_scatter/schema/glTF.KHR_materials_volume_scatter.schema.json
```diff
@@ -0,0 +1,28 @@
...
+            "default": [ 0.0, 0.0, 0.0 ],
+            "minItems": 3,
+            "maxItems": 3
+        },
+        "scatterAnisotropy": {
+            "type": "number",
+            "minimum": -1.0,
+            "maximum": 1.0,
+            "default": 0.0,
+            "description": "The anisotropy of scatter events."
```
emackey: Can we add a description (either to this directly, or with the `gltf_detailedDescription` pattern [used elsewhere](https://github.com/KhronosGroup/glTF/blob/90139f8f126c222a4b684219f226f85004ad45fc/extensions/2.0/Khronos/KHR_materials_volume/schema/material.KHR_materials_volume.schema.json#L25)?

It could say something like "The anisotropy of scatter events. Range is (-1, 1).  1 indicates forward scattering, -1 indicates backscattering, and the default is zero."

Did I get that backwards?  We should spell it out here.

PR #2460, opened 2024-11-24
Removed attributes that are already available through composition

MennoVink:
The `extensions` and `extras` attributes are not needed when glTFProperty has already been composited into the type.
Additionally `name` is not needed when compositing glTFChildOfRootProperty instead.

### DIFF
--- a/specification/2.0/schema/accessor.schema.json
+++ b/specification/2.0/schema/accessor.schema.json
@@ -124,10 +124,7 @@
         "sparse": {
             "allOf": [ { "$ref": "accessor.sparse.schema.json" } ],
             "description": "Sparse storage of elements that deviate from their initialization value."
-        },
-        "name": { },
-        "extensions": { },
-        "extras": { }
+        }
     },
     "dependencies": {
         "byteOffset": [ "bufferView" ]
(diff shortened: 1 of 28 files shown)


bghgary:
Maybe check if this change works with glTF Tools for vscode, just in case.

javagl:
^ This, but also check whether the output generated by https://github.com/CesiumGS/wetzel is still the same after the change (as this output is, in fact, a _normative_ part of the spec...)

aaronfranke (approved):
As far as I can tell, it's completely redundant to have these properties repeated in every schema. We should get rid of them.

MennoVink:
Signed though i dont think this is moving anywhere.
Personally i have no interest in checking these unrelated external tools. I just needed the json to be a valid json schema according to the json specs.

javagl:
@lexaknyazev Any strong opinion on this? I think that both (having and not having these properties) is "valid", but I'm not sure about possible "best practices" here. (Just wondering if the PR could be closed, or whether checking possible differences of the `wetzel` output is worthwhile before merging)

lexaknyazev:
Although those properties are technically redundant, the schema files are not invalid because there is no contradiction there. For example, a JSON object that validates against `accessor.schema.json` must also be valid as per `glTFChildOfRootProperty.schema.json` because the former depends on the latter.

This redundancy exists mostly for historical reasons: very early versions of schema files disallowed additional properties altogether so we had to add the properties explicitly to use composition.

As of today, I think it's still useful to list them explicitly even if as a reminder for implementors.

PR #2463, opened 2024-12-16
Fix K-bits range for meshopt octahedral filter

lilleyse:
Looking at [`meshopt_encodeFilterOct`](https://github.com/zeux/meshoptimizer/blob/20b26f98746a439600b20065a69ab0c66c4cca5c/src/vertexfilter.cpp#L877) as a reference, the bit range should be `1 <= K <= 16`, not `4 <= K <= 16`.

@zeux does this change look correct?

### DIFF
--- a/extensions/2.0/Vendor/EXT_meshopt_compression/README.md
+++ b/extensions/2.0/Vendor/EXT_meshopt_compression/README.md
@@ -441,7 +441,7 @@ Octahedral filter allows to encode unit length 3D vectors (normals/tangents) usi
 
 This filter is only valid if `byteStride` is 4 or 8. When `byteStride` is 4, then the input and output of this filter are four 8-bit components, and when `byteStride` is 8, the input and output of this filter are four 16-bit signed components.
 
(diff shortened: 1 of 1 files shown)


zeux:
While 1 is the smallest possible K from the encoding perspective, unsure if the spec should state that as the bounds - a 1-bit signed normalized integer has two values, -1 and 0, so that's not super useful for normal encoding :) My recollection for why the spec stated K=4 is the minimum was that I felt that this is the lowest value assets could be reasonably encoded in (at a great quality loss!), but maybe this is too conservative. When normals are used for specular reflections, Suzanne from glTF-Sample-Assets looks broken on K=3:

K=4
image

K=3
image

... but BrainStem model is perhaps okay for some contexts with K=3:

image

... or K=2

image

A 2-bit signed normalized integer can represent values -1, 0, 1 in [-1, 1] range. So if we want to update the spec then minimum K should probably start with 2; K=1 is not useful.

emackey:
@lilleyse What's the use case here, is it 3D Tiles with unlit and hence no normals are being used?  What's the next step here?

zeux:
FWIW I've updated this for the new extension proposal (#2517).

lexaknyazev:
@lilleyse Could you please change 1 to 2 as discussed above so this PR could be merged?

lilleyse:
@lexaknyazev done

lexaknyazev (approved):

PR #2478, opened 2025-03-05
EXT_mesh_primitive_restart

pmconne:
This extension permits the prohibition against maximal index values in index buffers to be selectively relaxed for groups of primitives, while providing a trivial fallback for implementations that don't support primitive restart. Discussed in #1142.


extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Overview
+
+"Primitive restart" is a feature of the input assembly stage that restarts the current primitive when the vertex index value is the maximum possible value for a given index buffer type. For example, the line strip primitive usually produces one continuous connected series of line segments, but with primitive restart enabled, a maximal vertex index value (e.g., 65535 for unsigned 16-bit integer indices) indicates the beginning of a new line string disconnected from those preceding it. Primitive restart can be useful for batching multiple line strips, line loops, triangle strips, or triangle fans into a single draw call. Alternatively, batching can be achieved by decomposing the primitives into lines or triangles, but this may introduce many redundant vertices, increasing the amount of data required to describe the geometry.
+
+glTF 2.0 explicitly prohibits index buffers from containing maximal index values because support for primitive restart [varies](https://github.com/KhronosGroup/glTF/issues/1142#issuecomment-433717774) amongst graphics APIs. Per [section 3.7.2.1](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#meshes-overview) of the spec,
```
lexaknyazev, line 24: Please remove the issue comment link as that information is already inaccurate, e.g., Vulkan support for primitive restart has changed since then.

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
```diff
@@ -0,0 +1,142 @@
...
+    ],
+```
+
+By default, this mesh draws two separate line strip primitives, each using its own `indices` accessor. The `BENTLEY_primitive_restart` extension specifies that both primitives can be replaced with a single one using a combined `indices` accessor containing primitive restart values.
+
+## glTF Schema Updates
+
+The `BENTLEY_primitive_restart` extension is applied to a mesh. Its `primitiveGroups` property is a list of groups of primitives that can be replaced with a single primitive using primitive restart. Each group is described by a list of indices into `mesh.primitives`, along with the ID of the accessor that supplies the vertex indices for the replacement primitive.
+
+Every primitive in a group **MUST** have identical property values (e.g., attributes, material, mode, etc), with the exception of `indices`.
```
lexaknyazev, line 138: Several more restrictions need to be mentioned here to ensure consistency and compatibility. I suggest putting all of them in a subsection (feel free to adjust language).

- A primitive index **MUST NOT** appear more than once across all primitive groups, i.e., two primitive groups cannot contain the same primitive index.
- The primitives referred by the extension **MUST** use one of the following topology types (represented by the `mode` property): 2 (line loop), 3 (line strip), 5 (triangle strip), or 6 (triangle fan). Other topology types are not allowed.
- The primitives referred by the extension **MUST** have their `indices` properties defined, i.e., they **MUST** use indexed geometry.
- Accessors referred by the extension's `indices` properties **MUST** be valid index accessors as per the base glTF 2.0 specification, i.e., their types **MUST** be scalar, their component types **MUST** be any of the unsigned integer types, and their buffer views (if defined) **MUST NOT** be used for any other purpose except vertex indices.

pmconne: Thanks for your feedback. Done.

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+The `BENTLEY_primitive_restart` extension is applied to a mesh. Its `primitiveGroups` property is a list of groups of primitives that can be replaced with a single primitive using primitive restart. Each group is described by a list of indices into `mesh.primitives`, along with the ID of the accessor that supplies the vertex indices for the replacement primitive.
+
+## Constraints
+
+A given primitive index **MUST NOT** appear in more than one primitive group.
+
+Each primitive in each group **MUST** use one of the following topology types, as specified by the `mode` property: 2 (line loop), 3 (line strip), 5 (triangle strip), or 6 (triangle fan). No other topology types are permitted.
+
+All primitives in a given group **MUST** have identical property values (e.g., attributes, material, mode, etc), with the exception of `indices`.
```
lexaknyazev, line 144: Although it's technically implied here, it's still worth explicitly mentioning that if any primitive of a group has a `KHR_materials_variants` extension object, then all other primitives of the same group **MUST** have that extension with identical content.

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
```diff
@@ -0,0 +1,152 @@
+<!--
+Copyright 2015-2021 The Khronos Group Inc.
```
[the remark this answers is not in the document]
pmconne: @pjcozzi any guidance on copyright and license here? The extension template includes this placeholder, but only one vendor extension actually includes it.

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+By default, this mesh draws two separate line strip primitives, each using its own `indices` accessor. The `BENTLEY_primitive_restart` extension specifies that both primitives can be replaced with a single one using a combined `indices` accessor containing primitive restart values.
+
+## glTF Schema Updates
+
+The `BENTLEY_primitive_restart` extension is applied to a mesh. Its `primitiveGroups` property is a list of groups of primitives that can be replaced with a single primitive using primitive restart. Each group is described by a list of indices into `mesh.primitives`, along with the ID of the accessor that supplies the vertex indices for the replacement primitive.
+
+## Constraints
+
+A given primitive index **MUST NOT** appear in more than one primitive group.
```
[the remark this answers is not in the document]
pmconne: Added.

extensions/2.0/Vendor/BENTLEY_primitive_restart/schema/primitiveGroup.schema.json
```diff
@@ -0,0 +1,40 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "$id": "primitiveGroup",
+    "title": "Primitive Group",
+    "type": "object",
+    "description": "A group of primitives that can be drawn together using a single index buffer containing restart values",
+    "allOf": [
+        {
+            "$ref": "glTFChildOfRootProperty.schema.json"
```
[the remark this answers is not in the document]
pmconne: I misread [this guidance](https://github.com/KhronosGroup/glTF/tree/main/extensions#creating-extensions). Updated.

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
pmconne: Clarified - thanks.

extensions/2.0/Vendor/BENTLEY_primitive_restart/schema/primitiveGroup.schema.json
```diff
@@ -0,0 +1,40 @@
...
+    "properties": {
+        "primitives": {
+            "type": "array",
+            "items": {
+                "$ref": "glTFid.schema.json"
+            },
+            "uniqueItems": true,
+            "minItems": 1,
+            "description": "The indices of the primitives that will be combined and drawn using a single set of indices",
+            "gltf_detailedDescription": "The indices of the primitives that will be combined and drawn using a single set of indices. All properties except for `indices` (e.g., `material`, `mode`, and `attributes` will be obtained from the first primitive in the list, and all other primitives in the list **MUST** have identical values for those properties. The `mode` must be TRIANGLE_FAN, TRIANGLE_STRIP, LINE_LOOP, or LINE_STRIP."
```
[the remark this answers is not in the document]
pmconne: Sorry about that - fixed.

extensions/2.0/Vendor/BENTLEY_primitive_restart/schema/mesh.BENTLEY_primitive_restart.schema.json
```diff
@@ -0,0 +1,26 @@
...
+    "type": "object",
+    "description": "glTF extension enabling the use of primitive restart values in index buffers",
+    "allOf": [
+        {
+            "$ref": "glTFProperty.schema.json"
+        }
+    ],
+    "properties": {
+        "primitiveGroups": {
+            "type": "array",
```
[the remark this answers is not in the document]
pmconne: Added.

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
```diff
@@ -0,0 +1,154 @@
...
+    ],
+```
+
+By default, this mesh draws two separate line strip primitives, each using its own `indices` accessor. The `BENTLEY_primitive_restart` extension specifies that both primitives can be replaced with a single one using a combined `indices` accessor containing primitive restart values.
+
+## glTF Schema Updates
+
+The `BENTLEY_primitive_restart` extension is applied to a mesh. Its `primitiveGroups` property is a list of groups of primitives that can be replaced with a single primitive using primitive restart. Each group is described by a list of indices into `mesh.primitives`, along with the index of the accessor that supplies the vertex indices for the replacement primitive.
+
+## Constraints
```
lexaknyazev, line 138: There is a high-level issue with all these constraints.

Let's say an asset that violates them (but perfectly valid otherwise) is loaded in an engine that does not support the extension. In this case, the asset would render fine.

But if the same asset is loaded in an engine that does support the extension, its rendering becomes undefined because the base spec does not generally prescribe what to do with invalid content.

Since this extension is fundamentally a performance hint and does not have any user-observable effects, it could have a built-in fallback to avoid undefined behavior by saying that if these constraints are violated, the extension becomes implicitly disabled and the affected primitive group (or the whole mesh) is rendered as usual.

pmconne: Added language to this effect.

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+The extension is subject to the following constraints. Violation of any constraint renders the entire extension invalid, in which case the engine should ignore the extension and render `mesh.primitives` instead.
```
lexaknyazev, line 140: In general, glTF specifications define asset semantics rather than engine operations so normative language is applied to the former.

Consider changing
> the engine should ignore the extension and render `mesh.primitives` instead

to

> the extension **MUST** be ignored and the `mesh.primitives` objects are rendered as defined in the glTF 2.0 specification

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+- A given primitive index **MUST NOT** appear in more than one primitive group.
+
+- Each primitive index in a primitive group **MUST NOT** appear more than once in that group.
+
+- Each primitive in each group **MUST** use one of the following topology types, as specified by the `mode` property: 2 (line loop), 3 (line strip), 5 (triangle strip), or 6 (triangle fan). No other topology types are permitted.
+
+- All primitives in a given group **MUST** have identical property values (e.g., attributes, material, mode, etc), with the exception of `indices`. This includes the `extensions` property - e.g., if any primitive in a group has a `KHR_materials_variants` extension object, then all other primitives in the same group **MUST** have that extension with identical content.
+
```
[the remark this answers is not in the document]
pmconne: Thanks for pointing that out. Added.

lexaknyazev:
@emackey @javagl
What do you think about making this an `EXT` extension? It's seems quite trivial to support it in viewers.

Also since it extends `mesh` objects, should it be called `*_mesh_primitive_restart`?

javagl (commented):
<sup>Disclosure: I'm an independent contributor, but also a freelancer with a connection to Bentley. I do not speak on behalf of either side here. I also have not yet reviewed this extension proposal on a technical level (only causally watching the comments and discussions).</sup>

**Only** about the naming:

Whether the prefix is `BENTEY` or `EXT` is up do decide for the person who proposes the extension. 

(Although some of the extension development process is about to be refined, the section at https://github.com/KhronosGroup/glTF/tree/main/extensions#naming will likely still be kept in a similar form. One addition here _might_ be that the `EXT` prefix suggests some broad, general applicability or that the extension is not bound to any form of ~"vendor-specific functionality" - but that's very vague...)

The naming section also says

> Names SHOULD be structured as  `<PREFIX>_<scope>_<feature>`, where _scope_ is an existing glTF concept (e.g. mesh, texture, image) 

It is not entirely clear whether that only refers to _top-level elements_ (somehow related to the point of [`gltfProperty` vs. `glTFChildOfRootProperty` above](https://github.com/KhronosGroup/glTF/pull/2478#discussion_r1983982172)), and whether any "sub-element" should be part of the name.

One could make a case to always _include_ the "top-level" element name. Be it at least for the case that someone defines an extension that goes into an `animation.sampler`, and we want to make sure that it is called `EXT_animation_sampler_example`, and not just `EXT_sampler_example`, which would cause an ambiguity with the top-level `sampler`. 

So I'd be leaning towards `_mesh_primitive_restart` here.

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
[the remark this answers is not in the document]
lexaknyazev: It's a part of REUSE compliance. All files must have such comments (if some don't we should fix that).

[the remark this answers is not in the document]
lexaknyazev: A vendor-prefixed extension should be owned by the vendor. Khronos usually owns extensions with `KHR` and `EXT` prefixes.

pmconne:
> Whether the prefix is BENTEY or EXT is up do decide for the person who proposes the extension.

If there's no objection, I will push a change to rename the extension to `EXT_mesh_primitive_restart` and assign the copyright to Khronos.

bghgary:
Should we just name it KHR_ directly? We have already ratified some EXT_ extensions and wasn't able to rename to KHR_ due to compatibility. Would we ever consider ratifying this extension?

pjcozzi:
>>Should we just name it KHR_ directly?
>
>We'd have to get an agreement within 3DF WG first and commit to providing Sample Assets at least.

@lexaknyazev @bghgary any update on this?  Or any additional feedback for @pmconne?

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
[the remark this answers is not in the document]
pmconne: Changed prefix to `EXT` leaving Khronos as the copyright holder (2025) pending discussion about whether or not this might be ratified.

extensions/2.0/Vendor/EXT_mesh_primitive_restart/schema/EXT_mesh_primitive_restart.schema.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "$id": "EXT_mesh_primitive_restart",
+  "title": "EXT_mesh_primitive_restart glTF extension",
```
aaronfranke, line 4:
```suggestion
  "title": "EXT_mesh_primitive_restart glTF Mesh Primitive Extension",
```

Also, this schema file should use 4-space indentation to match other schema files in this repo.

aaronfranke (commented):
The feature makes sense as an optimization when this is supported, and adds minimal data, while keeping the same visual appearance. However, I am curious, what prevents importers from doing this implicitly instead of checking for this explicit extension? I guess it would be complicated to check for mostly-identical primitives of the right topology type not used by morph targets.

extensions/2.0/Vendor/EXT_mesh_primitive_restart/schema/EXT_mesh_primitive_restart.schema.json
pmconne: Thanks, addressed.

extensions/2.0/Vendor/EXT_mesh_primitive_restart/README.md
```diff
@@ -0,0 +1,159 @@
...
+                    "attributes": {
+                        "POSITION": 0
+                    },
+                    "indices": 0,
+                    "material": 0,
+                    "mode": 3
+                },
+                {
+                    "attributes": {
+                        "POSITION": 0
```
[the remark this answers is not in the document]
pmconne: It should actually be 3. Thanks for spotting that.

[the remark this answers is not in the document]
emackey: Yes, position 3, of course.  Also the primitive above should use position 3 as well.

extensions/2.0/Vendor/EXT_mesh_primitive_restart/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+The mesh looks like this:
+
+```json
+    "meshes": [
+        {
+            "primitives": [
+                {
+                    "attributes": {
+                        "POSITION": 0
```
emackey, line 100: `"POSITION": 3` here as well.

bghgary:
Did we discuss whether this should be KHR or EXT? Should there be a sample model that uses it?

extensions/2.0/Vendor/EXT_mesh_primitive_restart/schema/primitiveGroup.schema.json
```diff
@@ -0,0 +1,40 @@
...
+        },
+        "name": {},
+        "extensions": {},
+        "extras": {}
+    },
+    "required": [
+        "primitives",
+        "indices"
+    ]
+}
```
aaronfranke, line 40: Nitpick: The file should have a trailing `\n` to be a POSIX-compliant text file. GitHub displays a ⛔ symbol on such files.

extensions/2.0/Vendor/EXT_mesh_primitive_restart/schema/primitiveGroup.schema.json
```diff
@@ -0,0 +1,40 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
```
aaronfranke, line 2: Nitpick: Upgrade to JSON schema draft 2020-12 to match glTF schemas.

extensions/2.0/Vendor/EXT_mesh_primitive_restart/README.md
```diff
@@ -0,0 +1,159 @@
...
+            ],
+            "min": [
+                -0.5,
+                -0.5,
+                0.0
+            ],
+            "name": "Positions accessor"
+        }
+    ],
+
```
aaronfranke, line 89:
```suggestion
```

extensions/2.0/Vendor/EXT_mesh_primitive_restart/schema/primitiveGroup.schema.json
pmconne: Done.

pmconne:
What can I do to help get this over the finish line?

extensions/2.0/Vendor/BENTLEY_primitive_restart/README.md
pmconne: @lexaknyazev *requiring* clients to implement fallback amounts to requiring them to exhaustively validate all of the constraints. Some of that may be trivial (e.g., is `primitiveGroups` empty), but not all of it (e.g., do all primitives in a group have exactly the same property values recursively). Fallback in case of malformed encoding is nice to have, but the onus should be on the encoders to conform to the spec.

lexaknyazev: Then please replace "MUST" with "SHOULD" on line 139.

pmconne:
We have [implemented](https://github.com/iTwin/itwinjs-core/pull/8312) this extension. Please let me know what if anything is required before this can merge.

extensions/2.0/Vendor/EXT_mesh_primitive_restart/README.md
```diff
@@ -0,0 +1,162 @@
...
+# EXT_mesh_primitive_restart
+
+## Contributors
+
+* Paul Connelly, Bentley Systems, [@pmconne](https://github.com/pmconne)
+
+## Status
+
+Draft
```
lexaknyazev: This should be changed to "Complete" before merging.

extensions/2.0/Vendor/EXT_mesh_primitive_restart/schema/primitiveGroup.schema.json
```diff
@@ -0,0 +1,40 @@
...
+        "indices": {
+            "allOf": [
+                {
+                    "$ref": "glTFid.schema.json"
+                }
+            ],
+            "description": "The index of the accessor that contains the vertex indices",
+            "gltf_detailedDescription": "The index of the accessor that contains the vertex indices. The accessor **MUST** have `SCALAR` type and an unsigned integer component type. The indices are permitted to include the maximal index value for the component type, indicating the start of a new primitive."
+        },
+        "name": {},
```
lexaknyazev: Is it intentional to have `name` here? As per the base spec, primitives do not have names, only meshes do.

pmconne: Removed.

extensions/2.0/Vendor/EXT_mesh_primitive_restart/README.md
pmconne: Completed.

weegeekps:
@lexaknyazev Are there any other changes you'd like before we merge?

lexaknyazev:
@weegeekps Please confirm the copyright info. If it's fine, feel free to merge.

aaronfranke (approved):

lexaknyazev:
@weegeekps All checked-in files must have SPDX copyright headers. The question was who is the owner in this particular case.

weegeekps:
@lexaknyazev My mistake. I've opened a new PR to correct this: https://github.com/KhronosGroup/glTF/pull/2533

PR #2479, opened 2025-03-14
EXT_mesh_primitive_edge_visibility

pmconne:
The `EXT_mesh_primitive_edge_visibility` extension augments a triangle mesh primitive with sufficient information to enable engines to produce non-photorealistic visualizations of 3D objects with visible edges. The edge visibility is encoded in a highly compact form to avoid excessively bloating the glTF asset. Adapted for glTF from https://github.com/iTwin/itwinjs-core/pull/5581. We intend to add support for the extension to the [CesiumJS](https://github.com/CesiumGS/cesium/) engine.


extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+</figure>
+
+The hard edges could be encoded explicitly into the glTF asset as additional primitives to be drawn together with the triangle mesh. However, this will generally produce a "stippling" effect where the edges and triangles collide in the depth buffer, as shown in Figure 4.
+
+<figure>
+<img src="./figures/depth-fighting.png"/>
+<figcaption><em><b>Figure 4</b> Edges drawn as separate primitives</em></figcaption>
+</figure>
+
+The [CESIUM_primitive_outline](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/CESIUM_primitive_outline) extension provides an additional index buffer describing each hard edge as a line segment (a pair of indices into the triangle mesh's list of vertices). The extension leaves the details of how to render the edges up to the engine, requiring only that they render without depth-fighting. This approach satisfies the use case for which it was intended - displaying the edges of boxy, low-resolution buildings - but suffers some limitations:
```
pjcozzi, line 58: It would not need to be near-term, but is there potential that this extension could replace `CESIUM_primitive_outline` in the future?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+- 1: Silhouette edge - the edge should be drawn only in silhouette (i.e., when separating a front-facing triangle from a back-facing triangle).
+- 2: Hard edge - the edge should always be drawn.
+- 3: Hard edge encoded in `primitives` - a representation of the edge is included in the extension's `primitives` property. The edge should always be drawn, either by drawing the `primitives` array or by drawing it like an edge with visibility value `2` (never both).
+
+The extension's `visibility` property specifies the index of an accessor of `SCALAR` type and component type 5121 (unsigned byte) that encodes as a bitfield the visibility of every edge of every triangle in the mesh. The ordering of triangles and vertices is as described by [Section 3.7.2.1](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#meshes-overview) of the glTF 2.0 specification. For each triangle `(v0, v1, v2)`, the bitfield encodes three visibility values for the edges `(v0:v1, v1:v2, v2:v0)` in that order. Therefore, the accessor's `count` **MUST** be `6 * N / 8` rounded up to the nearest whole number, where `N` is the number of triangles in the primitive.
```
pjcozzi, line 81: It's an edge case (no pun intended), but should the spec define the value of bits when there are extra/unused bits at the end of the bitstream?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+The `primitives` property **MUST** be defined *if and only if* at least one edge is encoded with visibility value `3` in `visibility`.
+
+### Silhouette Normals
+
+The extension's `silhouetteNormals` property specifies the index of an accessor of `SCALAR` type and component type 5123 (unsigned short) providing normal vectors used to determine the visibility of silhouette edges at display time. For each edge encoded as a silhouette (visibility value `1`) in `visibility`, the silhouette normals buffer provides the two outward-facing normal vectors of the pair of triangles sharing the edge. Each normal vector is compressed into 16 bits using the "oct16" encoding described [here](https://jcgt.org/published/0003/02/01/). The ordering of the normal vector pairs corresponds to the ordering of the edges in `visibility`; that is, the first pair of normals corresponds to the first edge encoded with visibility `1`, the second pair to the second occurrence of visibility `1`; and so on. The accessor's `count` **MUST** be twice the number of edges encoded with visibility value `1`.
```
pjcozzi, line 126: I don't know if I have a strong opinion but others may ask about the oct16 encoding and, if instead, this should be uncompressed and then able to use meshopt, etc.  Would that fit into the ecosystem better and have similar characteristics?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+Engines should render a silhouette edge unless both adjacent triangles are front-facing or both are back-facing, as determined by their normal vectors.
+
+The `silhouetteNormals` property **MUST** be defined *if and only if* at least one edge is encoded with visibility value `1` in `visibility`.
+
+### Extensions
+
+The `BENTLEY_primitive_restart` extension *may* be applied to the `EXT_mesh_primitive_edge_visibility` extension object, in which case it applies to the `primitives` property in the same way that it would apply to `mesh.primitives`.
```
pjcozzi, line 134: Same comment as above.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+## JSON Schema
+
+[./schema/primitive.EXT_mesh_primitive_edge_visibility.schema.json](primitive.EXT_mesh_primitive_edge_visibility.schema.json)
+
+## Implementation Notes
```
pjcozzi, line 140: Excellent to see these references!

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+* Paul Connelly, Bentley Systems, [@pmconne](https://github.com/pmconne)
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec. Depends on `BENTLEY_primitive_restart` if primitive restart is used.
```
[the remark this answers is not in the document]
pmconne: Yes.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
pmconne: Yes, the hard edge visibility encoded by that extension is a subset of that encoded by this extension.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
pmconne: Yeah, I went back and forth on it and decided to leave as-is initially pending feedback. I note that while oct-encoded normals get some use in 3D Tiles 1.0, the closest thing in glTF is KHR_mesh_quantization (min 1 byte per component, plus padding for byte stride).

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
pjcozzi: @lilleyse if you agree, do you want to submit an issue to the `3d-tiles` repo for future consideration to replace `CESIUM_primitive_outline` with this extension over time?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
pmconne: Added requirement that unused bits are zero-initialized.

kring (commented):
Thanks for opening this @pmconne! I have a few thoughts below.

Overall it looks good! I can't see any particular reason it couldn't be implemented in Unreal Engine or Unity.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/schema/primitive.BENTLEY_edge_visibility.schema.json
```diff
@@ -0,0 +1,45 @@
...
+            ],
+            "description": "The index of the accessor encoding the pair of outward-facing normals for the two faces sharing each silhouette edge (visibility 1) encoded in `visibility`.",
+            "gltf_detailedDescription": "This property **MUST** be defined if and only if `visibility` encodes at least one edge with visibility value `1`."
+        },
+        "primitives": {
+            "type": "array",
+            "description": "An array of primitives defining geometry that renders all of the edges encoded as `3` in `visibility`.",
+            "gltf_detailedDescription": "This property **MUST** be defined if and only if `visibility` encodes at least one edge with visibility value `3`",
+            "items": {
+                "$ref": "mesh.primitive.schema.json"
```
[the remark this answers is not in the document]
javagl: I haven't looked at all this on a technical level, but looking at this point in isolation, I think that (it could pose some challenges for code generators, due to the "recursion" through an extension, and) it could be worthwhile to point out that this mesh primitive may not _itself_ contain _another_ instance of the extension 😬

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+The extension's `visibility` property specifies the index of an accessor of `SCALAR` type and component type 5121 (unsigned byte) that encodes as a bitfield the visibility of every edge of every triangle in the mesh. The ordering of triangles and vertices is as described by [Section 3.7.2.1](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#meshes-overview) of the glTF 2.0 specification. For each triangle `(v0, v1, v2)`, the bitfield encodes three visibility values for the edges `(v0:v1, v1:v2, v2:v0)` in that order. Therefore, the accessor's `count` **MUST** be `6 * N / 8` rounded up to the nearest whole number, where `N` is the number of triangles in the primitive. Any unused bits in the last byte **MUST** be set to zero.
+
+The visibility of a given edge shared by a given pair of triangles **MUST** be encoded as a non-zero value no more than once. All other occurrences of the same edge in the bitfield **MUST** be encoded as zero, to prevent engines from either producing redundant geometry for shared edges or having to manually detect and account for such redundancies.
```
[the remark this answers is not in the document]
pmconne: That technique won't work for silhouettes, nor can it provide consistent edge widths (an important related requirement for AEC) since it can only draw *inside* the triangles. But I don't want to dictate specific implementations. I have changed the meaning of visibility value `3` to indicate a repeated hard edge.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+Engines **MUST** render all edges according to their specified visibility values. The edges **MUST** draw in front of their corresponding triangles with no depth-fighting. Engines may choose a material with which to draw the edges - e.g., they could apply the primitive's material to the edges, or apply some uniform material to the edges of all primitives in the scene, or some other consistent option. If engines choose to render the edges specified by the `primitives` property, then they **MUST** render those primitives according to the glTF 2.0 specification (e.g., using the specified material) and they **MUST NOT** also produce their own rendering of the edges with visibility value `3`.
```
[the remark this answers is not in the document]
pmconne: The extension now specifies the default material and permits alternate material(s) to be defined.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+Binary │ 00│ 01│ 00│ 10│ │ 00│ 00│ 00│ 10│
+       └───┴───┴───┴───┘ └───┴───┴───┴───┘   
+Decimal              18                 2
+```
+
+### Edge Primitives
+
+The extension's `primitives` property optionally provides an array of primitives representing all of the hard edges encoded with visibility value `3` in the `visibility` property. Engines should render these primitives directly, in which case they **MUST** do so without depth-fighting and **MUST NOT** also produce their own rendering of these edges. The primary use case is for edges that should be drawn using a different material than the corresponding surface - for example, a red outline drawn around a filled green shape.
+
+The `primitives` property **MUST** be defined *if and only if* at least one edge is encoded with visibility value `3` in `visibility`.
```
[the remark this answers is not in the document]
pmconne: I have replaced `primitives` with `lineStrings`, and those edges are no longer encoded in `visibility`. Please review the revised spec.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+- [GLSL](https://jcgt.org/published/0003/02/01/) (Section 3.1)
+- [JavaScript](https://github.com/CesiumGS/cesium/blob/dd9c25b652460e8cd77ecd29fbbf7cc44d8b62a2/packages/engine/Source/Core/AttributeCompression.js#L77)
+- [C++](https://github.com/iTwin/imodel-native/blob/4f0a5008dba61124b490a7506a6cd82ce1b8f148/iModelCore/iModelPlatform/PublicAPI/DgnPlatform/Render.h#L2012)
+
+The [pull request](https://github.com/iTwin/itwinjs-core/pull/5581) that informed the design of this extension provides [an iterator](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/common/imdl/CompactEdges.ts#L42) over the `visibility` and `silhouetteNormals` buffers.
+
+iTwin.js [implements](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/internal/render/webgl/glsl/Edge.ts#L107) conditional display of silhouette edges. It draws edges in a separate pass from surfaces to [ensure](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/internal/render/webgl/glsl/FeatureSymbology.ts#L426) the mitigate z-fighting.
```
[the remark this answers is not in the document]
pmconne: Yes, we invested a lot of effort into ensuring the edges consistently display in front of their surfaces (and with consistent pixel width). I think we might want to rough out an implementation in CesiumJS before writing up implementation guidance here.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/schema/primitive.BENTLEY_edge_visibility.schema.json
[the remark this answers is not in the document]
pmconne: I have removed this quirk.

pmconne:
@kring @javagl @pjcozzi thank you for your feedback. I've revised the spec accordingly.

pjcozzi:
@kring @javagl @lexaknyazev or others - do you have any additional feedback for @pmconne on this extension?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/schema/primitive.BENTLEY_edge_visibility.schema.json
```diff
@@ -0,0 +1,45 @@
+{
```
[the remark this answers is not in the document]
javagl: Bump @pmconne (triggered by the request for comments below).

kring:
@pmconne I read through the spec again and I think this looks great! No further suggestions from me.

I would like to see a complete implementation of this extension in CesiumJS or iTwinJS, though, if there isn't one already. It would be useful both as proof that it can be implemented sensibly in WebGL (2?), and for my own interest in the rendering techniques involved.

pmconne:
> I would like to see a complete implementation of this extension in CesiumJS or iTwinJS

Most of the rendering pieces are already in place in iTwin.js. I think the quickest path to a complete implementation would be for me to update the iModel -> 3D Tiles pipeline to encode the edges according to the spec and update iTwin.js decode accordingly.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/schema/primitive.BENTLEY_edge_visibility.schema.json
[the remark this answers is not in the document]
pmconne: Renamed, thank you for the reminder.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
lilleyse: Yeah, makes sense to eventually replace `CESIUM_primitive_outline`. I opened https://github.com/CesiumGS/3d-tiles/issues/807.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,221 @@
...
+## JSON Schema
+
+- [primitive.EXT_mesh_primitive_edge_visibility.schema.json](./schema/primitive.EXT_mesh_primitive_edge_visibility.schema.json)
+- [lineString.schema.json](./schema/lineString.schema.json)
+
+## Implementation Notes
+
+The [pull request](https://github.com/iTwin/itwinjs-core/pull/5581) that informed the design of this extension provides [an iterator](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/common/imdl/CompactEdges.ts#L42) over the `visibility` buffer.
+
+iTwin.js [implements](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/internal/render/webgl/glsl/Edge.ts#L107) conditional display of silhouette edges. It also draws edges in a separate pass from surfaces to [ensure](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/internal/render/webgl/glsl/FeatureSymbology.ts#L426) the mitigate z-fighting.
```
lilleyse, line 221: Thanks for the links. I think we can do this in CesiumJS with a separate pass after the `OPAQUE` pass.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/schema/primitive.EXT_mesh_primitive_edge_visibility.schema.json
```diff
@@ -0,0 +1,51 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "EXT_mesh_primitive_edge_visibility glTF extension",
```
aaronfranke, line 3:
```suggestion
    "title": "EXT_mesh_primitive_edge_visibility glTF Mesh Primitive Extension",
```

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,221 @@
...
+#### Examples
+
+Consider the following simple triangle fan with triangles described by the index list `[0,1,2, 0,2,3, 0,3,4, 0,4,5]`:
+```
+  2______3______4
+  |＼    |    ／|
+  |  ＼  |  ／  |
+  |____＼|／____|
+  1      0      5
+```
```
aaronfranke, line 160: The winding order in this example indicates that the triangles are all facing away from view.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,221 @@
...
+            "indices": 5,
+            "material": 1
+        }
+    ]
+}
+```
+
+##### Two materials
+
+Assume that the top edge is to be drawn using the same material as the vertical edges, and the bottom edge is to be drawn using a different material. Then there would be two index buffers - `[2,3,4]` and `[1,0,5]` - and the `lineStrings` property would look something like this:
```
aaronfranke, line 180: In this case, why wouldn't the top edge just be included together with the vertical edges in the visibility indices like `[0,2,0, 0,2,2, 3,2,0, 0,2,0]`? Doesn't that produce the same visual effect?

aaronfranke (commented):
How does this extension define the ends of each edge? Do they just end? Are they beveled/rounded? How do they connect to other edges?

Untitled

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
pmconne: Adjusted to CCW - thanks.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
pmconne: I have reverted this change - the extension now (once again) expects a pair of normals to be supplied for each silhouette edge. Every client will need them to render the silhouettes correctly. Computing them on the client side is onerous, especially because the client must infer winding order and iterate triangles (non-trivial if using fans or strips, especially if primitive restart is also used). Computing the normals from the encoded positions may result in loss of precision, especially if the positions are quantized.

This reopens the question of in what format the normals should be encoded. Models with silhouette edges have **a lot** of silhouette edges, so it's important to keep them small. oct16 is the smallest (2 bytes per normal). octahedral is twice as large (with one byte wasted to meet alignment requirements). Float is 8 times as large (wasting four bytes for alignment). Compression can help reduce the size of the transmitted asset but the client still must allocate memory for the uncompressed data.

pmconne:
@kring said:

> I would like to see a complete implementation of this extension in CesiumJS or iTwinJS, though, if there isn't one already. It would be useful both as proof that it can be implemented sensibly in WebGL (2?), and for my own interest in the rendering techniques involved.

We have [implemented](https://github.com/iTwin/itwinjs-core/pull/8366) this extension in iTwin.js.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
lexaknyazev: Whatever the chosen format is, it should be defined in the extension text. A 30-page PDF could only be put in the "References" section of the document.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+The extension's `silhouetteNormals` property specifies the index of an accessor of `SCALAR` type and component type 5123 (unsigned short) providing normal vectors used to determine the visibility of silhouette edges at display time. For each edge encoded as a silhouette (visibility value `1`) in `visibility`, the silhouette normals buffer provides the two outward-facing normal vectors of the pair of triangles sharing the edge. Each normal vector is compressed into 16 bits using an "oct16" encoding. The ordering of the normal vector pairs corresponds to the ordering of the edges in `visibility`; that is, the first pair of normals corresponds to the first edge encoded with visibility `1`, the second pair to the second occurrence of visibility `1`; and so on. The accessor's `count` **MUST** be twice the number of edges encoded with visibility value `1`.
+
+### oct16 encoding
+
+Octahedral normal encoding projects a normal vector onto an octahedron and folds it in half, mapping to two signed normalized components in [-1, 1]. The "oct16" form of this encoding maps each component to the integer range [0, 255], permitting a normal vector to be encoded into two bytes.
```
lexaknyazev: Is there any reason not to use VEC2 accessor type with normalized BYTE component type?

It would enable implicit conversions from quantized values using the standard glTF rules. Also that component type is generally considered more accurate as it preserves zeros.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
lilleyse: I'd like to echo @pjcozzi's comments above - is there a need to define oct16 encoding in this extension when oct encoding is already available in meshopt?

pmconne: I tried to explain that in the second paragraph of my previous reply. The amount of data raises concerns about client memory use, not just glTF transmission size.

A sphere is the worst case - 2 of every 3 edges in a sphere is a silhouette, though because we encode each shared silhouette only once, it works out to number of encoded silhouettes = number of triangles. Float normals require 16x as much memory as oct16 (`KHR_mesh_quantization` only applies to vertex attributes so not relevant here AFAICT). That's an extra 21 bytes per vertex, versus an extra 1.3 bytes per vertex using oct16.

I understand wanting to separate concerns (i.e., let this extension worry about encoding edges, and extensions like meshopt worry about minimizing data size), but that's a big difference.

lilleyse: You should be able to use `EXT_meshopt_compression` for non vertex attribute data. This extension would just have to say which `componentType`'s are allowed for silhouette normals.

```jsons
// accessor
{
  "bufferView": 4,
  "componentType": 5120,
  "normalized": true,
  "count": 24,
  "type": "VEC3"
}

// bufferView
{
  "buffer": 1,
  "byteOffset": 456,
  "byteLength": 96,
  "byteStride": 4,
  "target": 34962,
  "extensions": {
    "EXT_meshopt_compression": {
      "buffer": 0,
      "byteOffset": 280,
      "byteLength": 61,
      "byteStride": 4,
      "count": 24,
      "mode": "ATTRIBUTES",
      "filter": "OCTAHEDRAL"
    }
  }
}
```

The `MeshoptDecoder` API would decode to 3-bytes per normal. Not as good as oct16, but still acceptable for GPUs that prefer 4-byte alignment.

pmconne: OK, I'll do that.

pmconne: Updated spec and [implementation](https://github.com/iTwin/itwinjs-core/pull/8416).

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
pmconne: That would make sense. However I updated the spec with regards to the normal encoding based on feedback from @lilleyse and @pjcozzi.

weegeekps:
@lexaknyazev Any other changes required preventing us from getting this approved and merged? Like `EXT_primitive_restart`, this is intended to just be an open multi-vendor extension.

markschlosseratbentley:
The spec has been updated to now list CesiumJS as a known implementation in addition to iTwin.js. These implementations have the caveat that they do not implement the full capability of this specification, namely materials. This support is forthcoming.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+3D modeling and computer-aided drafting environments like SketchUp, MicroStation, and Revit provide non-photorealistic visualizations that render 3D objects with their edges visible. The edges can improve the readability of complex models and convey semantics of the underlying topology. The `EXT_mesh_primitive_edge_visibility` extension augments a triangle mesh primitive with sufficient information to enable engines to produce such visualizations. The information is encoded in a compact form to minimize the impact on the size of the glTF asset.
```
lexaknyazev: There should probably be some language about 3rd-party trademarks being used solely for identification etc, or preferably don't mention them at all.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,214 @@
...
+Assume that the vertical edges `0:1` and `2:3` and the shared diagonal edge `0:2` are all hard edges, and the horizontal edges `3:0` and `1:2` are hidden edges. Then the corresponding visibility values would be `[2,0,2, 2,2,0]`. However, we must ensure that the visibility of the shared hard edge is encoded only once as `2`, with the other occurrence encoded as `3`, resulting in visibility values `[2,0,2, 3,2,0]`. Encoding the edge visibility produces the sequence of bytes `[226, 2]`, as illustrated below.
+
+<figure>
+<img src="./figures/shared-hard-edge-encoding.svg"/>
+<figcaption><em><b>Figure 9</b> Sequence of bytes with an encoded shared hard edge</em></figcaption>
+</figure>
+
+### Material
+
+The extension's `material` property optionally specifies the index of a glTF material with which the edges are to be drawn. If unspecified, the edges should be drawn using the same material as the triangle mesh primitive.
```
lexaknyazev: - Would the lack of a defined material make the edges effectively invisible assuming that triangles are rendered?
- What happens If the referenced material requires surface normals or full tangent space?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The extension's `lineStrings` property provides an alternate representation for some or all of the hard edges that would otherwise be encoded in `visibility`. It is an array in which each entry is an object with the following properties:
+- `indices` (required): the index of an accessor of type `SCALAR` and component type `UNSIGNED_BYTE` (5121), `UNSIGNED_SHORT` (5123), or `UNSIGNED_INT` (5125) encoding one or more line strings as indices into the triangle mesh's vertex attribute array. Each line string is encoded as a series of two or more indices. Multiple line strings can be encoded by inserting a "primitive restart" value (that is, the maximal possible value for the accessor's component type, like 255 for `UNSIGNED_BYTE`) between each line string. The `indices` buffer **MUST NOT** begin or end with a primitive restart value, and **MUST NOT** contain consecutive primitive restart values.
```
lexaknyazev: > The `indices` buffer

should likely say "accessor"

---

Given that these values directly refer to vertices, can they define edges not existing in the original mesh topology?

---

Can multiple entries of the `lineStrings` array refer to the same edges?

---

Do the type-maximum values always cause an edge restart? For example, if the accessor uses unsigned bytes and the mesh contains more than 255 vertices, does `255` still cause a restart?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The `silhouetteNormals` property **MUST** be defined *if and only if* at least one edge is encoded with visibility value `1` in `visibility`.
+
+## Rendering requirements
+
+- Engines **MUST** render all edges according to their specified visibility values, though some engines may permit the user to toggle the display of edges on and off.
+- The edges **MUST** draw in front of their corresponding triangles with no depth-fighting.
+- The edges **MUST** draw using the materials specified by the extension by default, though some engines may provide options for the user to override those materials.
```
lexaknyazev: Shouldn't this say "**MUST** be drawn"?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,214 @@
...
+- Each silhouette edge **MUST** be rendered unless both adjacent triangles are front-facing or both are back-facing, as determined by their normal vectors.
```
lexaknyazev: Is there any interaction with double-sided materials?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/schema/lineString.schema.json
```diff
@@ -0,0 +1,38 @@
...
+        "material": {
+            "allOf": [
+                {
+                    "$ref": "glTFid.schema.json"
+                }
+            ],
+            "description": "The index of a glTF material with which the edges are to be drawn.",
+            "gltf_detailedDescription": "The index of a glTF material with which the edges are to be drawn. If omitted, the edges are drawn using the same material as the edges encoded by the extension's `visibility` field, or else that of the triangle mesh primitive."
+        },
+        "name": {},
```
lexaknyazev: Is `name` an intended property here?

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The extension's `lineStrings` property provides an alternate representation for some or all of the hard edges that would otherwise be encoded in `visibility`. It is an array in which each entry is an object with the following properties:
+- `indices` (required): the index of an accessor of type `SCALAR` and component type `UNSIGNED_BYTE` (5121), `UNSIGNED_SHORT` (5123), or `UNSIGNED_INT` (5125) encoding one or more line strings as indices into the triangle mesh's vertex attribute array. Each line string is encoded as a series of two or more indices. Multiple line strings can be encoded by inserting a "primitive restart" value (that is, the maximal possible value for the accessor's component type, like 255 for `UNSIGNED_BYTE`) between each line string. The `indices` buffer **MUST NOT** begin or end with a primitive restart value, and **MUST NOT** contain consecutive primitive restart values.
+- `material` (optional): the index of the material with which the edges encoded by `indices` are to be drawn. If omitted, the extension's own `material` property is used, if defined; otherwise, the material of the triangle mesh primitive is used.
```
lexaknyazev: What happens If the referenced material requires surface normals or full tangent space (likely the same as above)?

lexaknyazev:
In addition to the review comments please optimize all PNG images with
```
$ oxipng --zopfli --strip all -o max *.png
```

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/schema/lineString.schema.json
danielzhong: Removed, thanks!

danielzhong:
@lexaknyazev Thanks! I’ve gone through and updated all the PR feedback.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,218 @@
...
+- `indices` (required): the index of an accessor of type `SCALAR` and component type `UNSIGNED_BYTE` (5121), `UNSIGNED_SHORT` (5123), or `UNSIGNED_INT` (5125) encoding one or more line strings as indices into the triangle mesh's vertex attribute array. Each line string is encoded as a series of two or more indices. Multiple line strings can be encoded by inserting a primitive restart value between each line string. The primitive restart value is always the maximal possible value for the accessor's component type, such as 255 for `UNSIGNED_BYTE`, and **MUST** be interpreted as a restart marker rather than as a vertex index. Therefore, the accessor referenced by `indices` **MUST NOT** begin or end with a primitive restart value, **MUST NOT** contain consecutive primitive restart values, and **MUST NOT** contain any vertex index equal to the primitive restart value. Writers **MUST** choose a component type large enough that all referenced vertex indices are strictly less than the primitive restart value.
```
lexaknyazev: > Each line string is encoded as a series of two or more indices.

Could this format be more precise? Specifically, the validity of:
- repeated non-restart values (probably implicitly invalid because of the "edge of the triangle mesh primitive" language below)
- repeated edges, e.g., `1 2 1 2`;
- incomplete pairs, e.g., `1 2 255 3`.

danielzhong: Added, thanks!

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,220 @@
+<!--
+Copyright 2025 Bentley Systems, Incorporated
```
weegeekps:
```suggestion
Copyright 2026 Bentley Systems, Incorporated
```

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
```diff
@@ -0,0 +1,220 @@
...
+- [iTwin.js](https://github.com/iTwin/itwinjs-core/pull/8366)
+- [CesiumJS](https://github.com/CesiumGS/cesium/pull/12859)
+
+## Implementation Notes
+
+The [pull request](https://github.com/iTwin/itwinjs-core/pull/5581) that informed the design of this extension provides [an iterator](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/common/imdl/CompactEdges.ts#L42) over the `visibility` buffer.
+
+iTwin.js [implements](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/internal/render/webgl/glsl/Edge.ts#L107) conditional display of silhouette edges. It also draws edges in a separate pass from surfaces to [mitigate z-fighting](https://github.com/iTwin/itwinjs-core/blob/03b760e1e91bde5221aa7370ea45c52f966e3368/core/frontend/src/internal/render/webgl/glsl/FeatureSymbology.ts#L426).
+
+Assets containing silhouette edges tend to contain *a lot* of them. Using component type 5120 (byte) to encode the silhouette normal vectors is recommended to minimize their impact on the size of the glTF asset. Using `EXT_meshopt_compression` can further reduce the transmission size.
```
weegeekps, line 220: Add a `License` section after this line and add this text so it's understood that it's free to use.

> This work is licensed under CC BY 4.0. To view a copy of this license, visit https://creativecommons.org/licenses/by/4.0/

weegeekps (commented):
I'd like to get this merged today. @markschlosseratbentley, you've got two changes that are needed and then we should be good to go since this is a Vendor extension.

extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md
markschlosseratbentley: Added.

weegeekps (approved):

PR #2481, opened 2025-03-19
KHR_materials_diffuse_roughness

MiiBond:
As part of Adobe's effort to support the OpenPBR feature set, I'm proposing an extension to support base diffuse roughness in glTF.


extensions/2.0/Vendor/EXT_materials_diffuse_roughness/schema/glTF.EXT_materials_diffuse_roughness.schema.json
```diff
@@ -0,0 +1,24 @@
...
+            "description": "The base, diffuse layer's roughness.",
+            "default": 1.0,
+            "minimum": 0.0,
+            "maximum": 1.0,
+            "gltf_detailedDescription": "The base, diffuse layer's roughness."
+        },
+        "diffuseRoughnessTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "A greyscale texture, representing the diffuse layer's roughness.",
+            "gltf_detailedDescription": "A greyscale texture, representing the diffuse layer's roughness. The value will modulate the diffuseRoughnessFactor."
```
lexaknyazev, line 19: Instead of "greyscale" (which is ambiguous and not used in glTF specs) please specify an exact channel (likely red) from which the values are sourced.

This should also be mentioned in README for better visibility.

bghgary: Example:
- https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_transmission#properties
- https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_transmission/schema/material.KHR_materials_transmission.schema.json#L18

bghgary: We should decide which channel though. Does it make sense to pack this channel with other textures?

javagl:
A cross-link to https://github.com/KhronosGroup/glTF/issues/2482 : I think the schema file should be called `material.EXT_materials_diffuse_roughness.schema.json`.

bsdorra:
To be consistent with [OpenPBR]([OpenPBR](https://academysoftwarefoundation.github.io/OpenPBR/index.html#model/thin-walledcase)), this extension should also affect [KHR_materials_diffuse_transmission](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_diffuse_transmission/README.md), I believe.

extensions/2.0/Vendor/KHR_materials_diffuse_roughness/README.md
lexaknyazev, line 1: Please update the directory name to `Khronos`.

dfattal:
### `diffuseRoughnessFactor` default disagrees between the README and the schema

README:

> |**diffuseRoughnessFactor** | `number` | The diffuse roughness factor. | No, default: `0.0` |

`schema/material.KHR_materials_diffuse_roughness.schema.json`:

```json
"diffuseRoughnessFactor": {
    "type": "number",
    "description": "The base, diffuse layer's roughness.",
    "default": 1.0,
```

We implemented `0.0`, on backwards-compatibility grounds: glTF's base diffuse BRDF is Lambertian, `0.0` is the value that reproduces it, and a default of `1.0` restyles every material in an asset that adds this extension for one of them. But the argument matters less than the disagreement — worth reconciling before two implementations pick differently and assets start depending on one of them.

(For what it's worth the example in *Extending Materials* sets `diffuseRoughnessFactor` explicitly, so it doesn't disambiguate.)

---
David Fattal — implementing for the [DisplayXR project](https://github.com/DisplayXR) ([model viewer](https://github.com/DisplayXR/displayxr-demo-modelviewer)). This extension is implemented there alongside `KHR_materials_coat` and `KHR_materials_fuzz`; happy to contribute conformance assets.

PR #2488, opened 2025-05-12
EXT_materials_specular_openpbr

MiiBond:
Here's an extremely rough draft of an OpenPBR-compatible specular extension.
Based on @lexaknyazev's advice, I've made this a sub extension of `KHR_materials_specular` as they share the same parameters.
A few issues so far:
1. I'm not sure how to specify that this needs to be defined under the `KHR_materials_specular` in the schema.
2. The `specularFactor` is basically the same as `specular_weight` in OpenPBR. However, for metals, this creates a bit of an issue. In OpenPBR, decreasing `specular_weight` to zero creates a black colour at F82 while, in glTF, we'd expect the factor being 0 to effectively turn off the extension.

Feedback would be very appreciated.


extensions/2.0/Vendor/EXT_materials_specular_openpbr/README.md
```diff
@@ -0,0 +1,145 @@
...
+## Exclusions
+
+- This extension must not be used on a material that also uses `KHR_materials_pbrSpecularGlossiness`.
+- This extension must not be used on a material that also uses `KHR_materials_unlit`.
+
+## Overview
+
+Specular reflectance in glTF as defined by `KHR_materials_specular`  behaves a bit different than in OpenPBR. This extension defines an interpretation of the specular parameters that is compatible with OpenPBR's specular reflectance for both dielectrics and metals.
+
+This extension adds no additional parameters. Its presence merely changes how the parameters of `KHR_materials_specular` are used.
```
lexaknyazev, line 25: PBR Material extensions are generally designed to be "as if always defined" with their default state not affecting the material appearance. To follow this convention, consider adding a boolean property (optional, false by default) that enables the new semantics.

extensions/2.0/Vendor/EXT_materials_specular_openpbr/README.md
```diff
@@ -0,0 +1,145 @@
+# EXT\_materials\_specular\_openpbr
```
lexaknyazev, line 1: It would be safer to name the extension based on the actual interpretation difference rather than using a reference.

MiiBond: I was expecting that issue to be raised but I'm not sure of a good name.
`EXT_materials_specular_edge_color`?
`EXT_materials_specular_f82`? - the F82 really only refers to the metallic model so probably isn't ideal.

bghgary (commented):
I don't know the math to be able to review that part, but the wording and explanation seems fine to me.

extensions/2.0/Vendor/EXT_materials_specular_openpbr/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+In OpenPBR, metals use the [F82 reflectance model](https://academysoftwarefoundation.github.io/OpenPBR/index.html#model/basesubstrate/metal) (see also the Adobe Standard Material [technical specifications](https://helpx.adobe.com/content/dam/substance-3d/general-knowledge/asm/Adobe%20Standard%20Material%20-%20Technical%20Documentation%20-%20May2023.pdf)).  This allows the specular color to be applied at grazing angles to simulate metals with complex IOR's. Many real-world metals exhibit dips in brightness or sublte hue changes at grazing angles.
+
+Here are three real-world metals and one made up, "Unobtainium", to demonstrate the edge-coloring.
+![](figures/openpbrMetals.png)
+
+For real-world metals, the effect can be hard to see. To make it more visible, here are the results of rendering in a white furnace.
+![](figures/openpbrMetalsFurnace.png)
+
+This model adds an extra term to the Schlick approximation to accomplish this. If `μ` is the cosine of the incident angle and `μ¯` is `cos(82) ~= 1/7`. `F(μ¯)` is the specular color at F82:
```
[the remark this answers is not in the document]
MiiBond: Oh, that's excellent. Thank you.

extensions/2.0/Vendor/EXT_materials_specular_openpbr/README.md
MiiBond: I've added a single boolean value to the extension to enable it.

extensions/2.0/Vendor/EXT_materials_specular_openpbr/README.md
MiiBond: I've changed the name to `EXT_materials_specular_edge_color` for now.

extensions/2.0/Vendor/EXT_materials_specular_edge_color/README.md
```diff
@@ -0,0 +1,154 @@
...
+*This section is non-normative.*
+
+[Appendix B](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#appendix-b-brdf-implementation) defines the function `fresnel_mix`. In this extension, apply this function to the dielectric lobe only and we add two additional arguments called `weight` and `specular_color`.
+
+```
+function fresnel_mix(specular_color, ior, weight, base, layer) {
+  f0 = ((1-ior)/(1+ior))^2
+  f0 = min(f0, float3(1.0))
+  fr = f0 + (1 - f0)*(1 - abs(VdotH))^5
+  return (1 - weight * fr) * base + weight * fr * specular_color * layer
```
bsdorra, line 122: I think the base weight needs to consider the `specular_color` for reasons of energy preservation. E.g. low specular_color will lead to low specular contribution that is not compensated by the base layer. Also specular color values > 1.0 need to be handled somehow. Haven't fully thought this through but maybe sth. like 

```
specular_weight =  weight * fr * min(specular_color, float3(1.0))
base_weight = 1 -  max_value(specular_weight)
return base_weight * base + specular_weight * layer
```

extensions/2.0/Vendor/EXT_materials_specular_edge_color/README.md
```diff
@@ -0,0 +1,154 @@
...
+    const b_denominator_reciprocal = 1.0 / b_denominator;
+    vec3 b = b_numerator * b_denominator_reciprocal; // analogous to "a" in the "Fresnel Equations Considered Harmful" slides
+    cos_theta = max(roughness, NdotV);
+    one_minus_cos_theta = 1.0 - cos_theta;
+    vec3 offset_from_F0 = (white_minus_F0 - b * cos_theta * one_minus_cos_theta) * pow(one_minus_cos_theta, 5.0);
+    return clamp(F0 + offset_from_F0, 0.0, 1.0);
+}
+```
+All that remains is a simple lerp between the dielectric and metallic lobes based on the metallic value.
+
```
bsdorra, line 147: Maybe small code chunk that shows the lerp for clarity?

extensions/2.0/Vendor/EXT_materials_specular_edge_color/README.md
MiiBond: My understanding is that, in OpenPBR, `specular_color` does not effect the base weight.
From the specs:

> The specular_color parameter modulates the Fresnel factor of Fdielectric, but only for the initial reflection of light incident from above, while the light transmitted from above or below (or reflected from below) is assumed to be unaffected. This is technically unphysical if altered from the default white color (as real dielectrics have a Fresnel factor dependent only on the index of refraction), but can be useful in practice to artificially tint the specular highlight without disturbing other aspects of the light transport, i.e. the reflection due to scattering from the internal medium, or the reflection from below, or the transmission from above or below.

extensions/2.0/Vendor/EXT_materials_specular_edge_color/README.md
```diff
@@ -0,0 +1,154 @@
...
+![](figures/edgeColorDielectricWhite.png)
+
+Here is the same comparison for colored specular, increasing from [0,0,0] to [1,0,0] (bottom).
+
+#### KHR_materials_specular
+![](figures/defaultDielectricColor.png)
+#### EXT_materials_specular_edge_color
+![](figures/edgeColorDielectricColor.png)
+
+The specular color factor is allowed to be set to values greater than [1, 1, 1]. Thus, the reflection amount can go beyond what is determined by the index of refraction (IOR). To still ensure energy conservation, the product of specular color factor, specular color texture, and f0 reflectance from IOR is clamped to 1. Please refer to [Implementation](#Implementation) for an example on where to place the clamping operation.
```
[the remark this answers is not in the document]
MiiBond: Yeah, the needed logic here is a bit more complex. I've tried to clarify.

Basically, OpenPBR allows specular_weight > 1.0 but not specular_color. Assuming that we want the data to remain the same for this sub-extension, we need to apply specularColorFactors > 1 to the weighting while keeping the color itself as a multiply on the specular lobe. I've updated the implementation code as well. Does it make sense?

extensions/2.0/Vendor/EXT_materials_specular_edge_color/README.md
MiiBond: I think I've got something that works in the Implementation section now.
I definitely wasn't handling the > 1.0 case correctly.

PR #2494, opened 2025-06-11
EXT_implicit_ellipsoid_region Proposal

j9liu:
This extension proposal builds on top of **KHR_implicit_shapes** (#2370) to enable the definition of implicit ellipsoid-based regions. This is useful in geospatial applications where regions are defined in longitude / latitude / height coordinates relative to an ellipsoid model, e.g., the WGS84 ellipsoid.
 
This enables the rendering of ellipsoid-conforming voxels in **EXT_primitive_voxels** (#2496), allowing larger regions of data to follow the ellipsoid's surface more closely and improve the real-world accuracy of the data over uniform box voxels.


aaronfranke:
The `EXT` prefix is reserved for multi-vendor extensions. Which vendors besides Cesium have worked on this?

javagl:
Multiple vendors working on a specific extension is not (or no longer) a _strict_ requirement for the `EXT_` prefix. The informal meaning of that prefix is rather something like the vendor ~"not claiming the (naming) rights". 
(Most of the open extension PRs are using the `EXT_` prefix, FWIW. The exceptions are mainly the `KHR_` ones, which is not much more than the "Khronos vendor prefix" nowadays)

j9liu:
Closing for now. Will reopen later to target glTF 2.1 where shapes are core. https://github.com/KhronosGroup/glTF/issues/2588

PR #2495, opened 2025-06-11
EXT_implicit_cylinder_region Proposal

j9liu:
This extension proposal builds on top of **KHR_implicit_shapes** (#2370) to enable the definition of implicit cylinder-based regions. Although **KHR_implicit_shapes** already provides a definition for cylinders, this extension provides granular control over advanced properties, such as the inner radius and angular bounds of the cylinder.
 
This is most useful for **EXT_primitive_voxels** (#2496) to enable rendering of cylinder-conforming voxels. Some categories of real-world data may be collected by cylindrical sensors, in which case **EXT_implicit_cylinder_region** offers a better fit for representing the data (over uniform box voxels, for example).

PR #2496, opened 2025-06-11
EXT_primitive_voxels Proposal

j9liu:
This extension would enable volumetric data to be stored as voxels in glTF. Through this extension, a glTF mesh primitive could define a uniform voxel grid with a specified shape and dimensions. The primitive's attributes are then inferred to fill that voxel grid, representing continuous data within that volume.
 
Some notable use cases for storing and visualizing voxel data include geothermal temperature, weather, and ocean salinity. Support for **EXT_primitive_voxels** has been implemented in CesiumJS, with future support incoming in Cesium for Unreal. (https://github.com/CesiumGS/cesium-unreal/pull/1685).
 
### Notes

This relies on **KHR_implicit_shapes** (#2370) to define the shape of the voxel volume, with support for box voxels as well as **EXT_implicit_ellipsoid_region** (#2494) and **EXT_implicit_cylinder_region** (#2495).
 
This extension also enables the definition of `noData` values to represent empty voxels.
 
A significant advantage of supporting voxels in glTF is accessing compression methods like **EXT_meshopt_compression** for improved storage. Some initial results from testing on a user's dataset:

- original - **250 MB**
- lossless meshopt - **135.3 MB**
- 16-bit quantization + meshopt - **93.8 MB**
- 8-bit quantization + meshopt - **38.1 MB**
- 16-bit exponential filter + meshopt - **64.6 MB**
- 8-bit exponential filter + meshopt - **26.3 MB**


weegeekps:
@lexaknyazev I know you are busy, but would you be able to take a pass at this if you have some time? While the features that use this are currently available for select customers in Cesium at the moment, we are looking to release them fully to everyone sometime later this year. We would eventually like this to become ratified if possible, so if there are any sweeping changes we need to make, it would be very appreciated to know those so we can fix them prior to that release.

PR #2516, opened 2025-08-04
Update read-only object model pointers

lexaknyazev:
This update ensures that applications can query asset's nodes structure and material references.


mattmacf98 (approved):

DRx3D (approved):
LGTM

PR #2520, opened 2025-08-19
KHR_materials_clearcoat_color extension

MiiBond:
This extension adds the ability to control the coloured transmission of the clear coat layer.
It's designed to provide the same functionality as the coat_color parameter of [OpenPBR.](https://academysoftwarefoundation.github.io/OpenPBR/index.html#model/coat/view-dependentabsorption)


extensions/2.0/Vendor/EXT_materials_clearcoat_color/README.md
```diff
@@ -0,0 +1,131 @@
...
+* Requires the `KHR_materials_clearcoat` extension to also be defined on the material
+
+## Exclusions
+
+* This extension must not be used on a material that also uses `KHR_materials_pbrSpecularGlossiness`.
+* This extension must not be used on a material that also uses `KHR_materials_unlit`.
+
+## Overview
+
+This extension defines a way to add colored absorption to the clear coating provided by the `KHR_materials_clearcoat` extension. This allows materials to simulate the appearance of colored varnishes or lacquers.
```
bghgary: nit: for consistency

```suggestion
This extension defines a way to add colored absorption to the clearcoating provided by the `KHR_materials_clearcoat` extension. This allows materials to simulate the appearance of colored varnishes or lacquers.
```

bghgary: Maybe provide an example of a real object that shows the varnish or lacquer so that people can visually identify with the concept?

extensions/2.0/Vendor/EXT_materials_clearcoat_color/README.md
```diff
@@ -0,0 +1,131 @@
...
+        }
+    ]
+}
+```
+
+### Clearcoat Color Parameters
+
+All implementations should use the same calculations for the input parameters. Implementations of the BRDF itself can vary based on device performance and resource constraints. See [Appendix B](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#appendix-b-brdf-implementation) for more details on the BRDF calculations.
+
+As with the core glTF 2.0 spec, all coloured textures are assumed to be sRGB and all coloured factors are assumed to be linear.
```
bghgary: nit: There are inconsistent spelling of color/colour in the doc. I would suggest using the same ones to be consistent.
```suggestion
As with the core glTF 2.0 spec, all colored textures are assumed to be sRGB and all colored factors are assumed to be linear.
```

extensions/2.0/Vendor/EXT_materials_clearcoat_color/README.md
```diff
@@ -0,0 +1,131 @@
...
+|**clearcoatColorTexture**   | [`textureInfo`](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo)  | The clearcoat layer's color texture.  | No                               |
+
+As with other factor/texture combinations, these two values are multiplied before being used in calculations. If not defined, the defaults represent a white transmission color and, therefore, no absorption.
+```
+clearcoat_transmission_color = clearcoatColorFactor * clearcoatColorTexture
+```
+
+## Physical Interpretation
+
+The clearcoat is a thin, dielectric layer that specularly-transmits light through it but, by default, doesn't absorb light. This extension allows us to specify the observed color tint of light passing through the coating from the underlying layer. The `clearcoat_transmission_color` represents the **square of transmittance of the coating**. This is because the observed color tint of the underlying material results after light has passed through the coating twice (in incoming and outgoing directions), making the final color more easily artist-directable.
```
bghgary: Just for understanding, does the squared value also have impact on texture precision if 8-bit? Similar to gamma space?

extensions/2.0/Vendor/EXT_materials_clearcoat_color/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+### View-Dependent Absorption
+
+Light rays travel different distances through the clearcoat layer depending on the angle of incidence. At grazing angles (when viewing the surface nearly edge-on), light travels a longer distance through the clearcoat medium, resulting in stronger absorption and more pronounced color saturation. Conversely, at normal incidence (when viewing the surface straight-on), light travels the shortest path through the clearcoat, producing less absorption. The `clearcoat_transmission_color` represents the color observed at normal incidence.
+
+Note that this view-dependent absorption is more pronouced when the ratio of IOR between the coating and surounding medium (usually air) is low. This is because, at higher IOR's, light impacting the coating at grazing angles gets refracted into the coating more strongly, shortening the path length through the coat and decreasing the absorption.
```
bghgary:
```suggestion
Note that this view-dependent absorption is more pronounced when the ratio of IOR between the coating and surrounding medium (usually air) is low. This is because, at higher IOR's, light impacting the coating at grazing angles gets refracted into the coating more strongly, shortening the path length through the coat and decreasing the absorption.
```

extensions/2.0/Vendor/EXT_materials_clearcoat_color/README.md
MiiBond: No more than any other coloured texture. The texture itself will be stored as sRGB, just like any coloured texture and then linearized before using the value in calculations. The colour of the texture represents the final, observed look of the tinting. For the sake of path tracers, we specify that what is actually happening in the coat is the light ray passing through the coat twice, once on the way in and once on the way out. Therefore, the tint colour represents the square of the transmittance.

PR #2522, opened 2025-09-03
KHR_billboard Draft Proposal

weegeekps:
Billboards have been [discussed within the context of glTF for some time now](https://github.com/KhronosGroup/glTF-External-Reference/blob/main/explainers/billboard.md) and we've found a growing need for being able to define them within a glTF for 3D Tiles and other geospatial use cases. This draft proposal is for `KHR_billboard`, an extension based on [the Billboard explainer](https://github.com/KhronosGroup/glTF-External-Reference/blob/main/explainers/billboard.md) that @UX3D-haertl wrote last year.

Human-readable link to [KHR_billboard](https://github.com/CesiumGS/glTF/tree/billboard-draft/extensions/2.0/Khronos/KHR_billboard/README.md).

Billboards are viewport-aligned meshes or images that have many uses including:

 - Spatially-oriented labels.
 - Displaying foliage, trees, clouds with 2D sprites.
 - Animated visual effects and particles.
 - Distant low level-of-detail sprites.

Most CGI tools and game engines support some sort of Billboard functionality. While this proposal is based on the excellent explainer written by @UX3D-haertl, I have done some further analysis and looked into how the popular engines support billboards. Given that, I made some small adjustments to how rotation axis is defined and attempted to make it clearer how the parent-child relationship should work with children of a billboard node.

In the full spirit of the [explainer](https://github.com/KhronosGroup/glTF-External-Reference/blob/main/explainers/billboard.md) the content of the billboard is handled by the associated mesh or children of the billboard node. This allows a variety of content to be handled by the billboard.

_Note: We are also investigating an extension for defining labels for use with billboards similar to what we do in CesiumJS and we'll share that soon._

### DIFF
--- a/extensions/2.0/Khronos/KHR_billboard/README.md
+++ b/extensions/2.0/Khronos/KHR_billboard/README.md
@@ -58,7 +58,7 @@ The content of a billboard is defined by the meshes and children of the Node. Th
                 "scaleWithDistance" : false,
                 "viewDirection": [0, 0, 1],
                 "up": [0, 1, 0],
-                "rotationAxis": "All",
+                "rotationAxis": ["X", "Y", "Z"],
                 "overlay": false
             }
         }
--- a/extensions/2.0/Khronos/KHR_billboard/README.md
+++ b/extensions/2.0/Khronos/KHR_billboard/README.md
@@ -72,7 +72,7 @@ When the `KHR_billboard` extension is defined on a node, the node's global rotat
 
 ### Child Nodes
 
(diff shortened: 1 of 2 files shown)


andybak:
This might be a bit of a stretch for this specific extension but I'd like to throw it into the mix.

Open Brush (née Tilt Brush) has custom shaders for various brush types. Some of these simulate typical particle effects using camera-facing quads. We aim to provide shaders for as many platforms as possible but some simply don't support custom shaders so we want to supply a "best" effort GLTF export that leverages native GLTF features and extensions to get as close as possible for as many brushes as possible.

So specifically some meshes exported from Open Brush will have a sequence of triangle pairs forming billboards. These are used for smoke, snow, stars dot trails a few others similar effects.

PR #2525, opened 2025-09-12
EXT_lights_area

MiiBond:
This is almost identical to the previously-proposed KHR area lights extension except that it excludes "sphere" as a valid shape in favour of only flat lights. (See https://github.com/KhronosGroup/glTF/pull/1948).

### DIFF
--- a/extensions/2.0/Vendor/EXT_lights_area/README.md
+++ b/extensions/2.0/Vendor/EXT_lights_area/README.md
@@ -33,18 +33,16 @@ The following example defines both a rectangular and a disk area light:
                 "color": [1.0, 1.0, 1.0],
                 "intensity": 1000.0,
                 "type": "rect",
+                "size": 1.0,
                 "rect": {
-                    "width": 2.0,
-                    "height": 1.0
+                    "aspect": 2.0
                 }
             },
             {
                 "color": [1.0, 0.9, 0.8],
                 "intensity": 1500.0,
                 "type": "disk",
-                "disk": {
-                    "radius": 1.2
-                }
+                "size": 1.2
             }
         ]
     }
--- a/extensions/2.0/Vendor/EXT_lights_area/README.md
+++ b/extensions/2.0/Vendor/EXT_lights_area/README.md
@@ -17,7 +17,7 @@ Written against the glTF 2.0 spec.
 This extension defines area lights with physically accurate intensity values for use with glTF 2.0.
 
 Area lights are flat surfaces that uniformly emit light from one side (the side facing in the -Z direction in local space).
-This extension defines two area light shapes: `rect` (rectangle) and `disk` (circle). These lights are referenced by nodes and inherit the transform of that node, allowing them to be placed in a scene.
+This extension defines two area light types: `rect` (rectangle) and `disk` (circle). These lights are referenced by nodes and inherit the transform of that node, allowing them to be placed in a scene.
 
 ## Defining Area Lights
 
@@ -86,7 +86,7 @@ All light types share the common set of properties listed below.
 
 ### Rectangle Lights
 
-When a light's `type` is `rect`, the `rect` property on the light is required. Its properties (below) are optional.
+When a light's `type` is `rect`, the `rect` property contains the following optional properties.
 
 | Property | Type | Description | Required | Default |
 |:---------|:-----|:------------|:---------|:--------|
--- a/extensions/2.0/Vendor/EXT_lights_area/README.md
+++ b/extensions/2.0/Vendor/EXT_lights_area/README.md
@@ -1,0 +1,140 @@
+# EXT_lights_area
+
+## Contributors
+
+* Mike Bond, Adobe, [@miibond](https://github.com/MiiBond)
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
+
+This extension defines area lights with physically accurate intensity values for use with glTF 2.0.
+
+Area lights are flat surfaces that uniformly emit light from one side (the side facing in the -Z direction in local space).
+This extension defines two area light shapes: `rect` (rectangle) and `disk` (circle). These lights are referenced by nodes and inherit the transform of that node, allowing them to be placed in a scene.
+
+## Defining Area Lights
+
+As with KHR_lights_punctual, area lights are defined as a `lights` array inside the `EXT_lights_area` extension at the root of the document.
+
+The following example defines both a rectangular and a disk area light:
+
+```json
+"extensions": {
+    "EXT_lights_area": {
+        "lights": [
+            {
+                "color": [1.0, 1.0, 1.0],
+                "intensity": 1000.0,
+                "shape": "rect",
+                "width": 2.0,
+                "height": 1.0
+            },
+            {
+                "color": [1.0, 0.9, 0.8],
+                "intensity": 1500.0,
+                "shape": "disk",
+                "radius": 1.2
+            }
+        ]
+    }
+}
+```
+
+## Adding Light Instances to Nodes
+
+Lights must be attached to a node by defining the `extensions.EXT_lights_area` property and, within that, an index into the `lights` array using the `light` property.
+
+```json
+"nodes": [
+    {
+        "extensions": {
+            "EXT_lights_area": {
+                "light": 0
+            }
+        }
+    }
+]
+```
+
+## Behaviour
+
+The light will inherit the transform of the node that it is attached to. Area lights are centered at the origin in local space:
+
+* **Rectangular lights**: extend from -width/2 to +width/2 along the local X-axis, and from -height/2 to +height/2 along the Y-axis
+* **Disk lights**: extend in a circle of the specified radius 
+around the origin in the local XY-plane
+
+Area lights emit light uniformly from one side of the surface (the side facing in the -Z direction in local space). The light emission follows Lambert's cosine law.
+
+**Lambert's Cosine Law**: This law states that the radiant intensity (or luminous intensity) observed from a perfectly diffusing surface is directly proportional to the cosine of the angle between the direction of the incident light and the surface normal. In practical terms, this means that an area light appears brightest when viewed straight-on (perpendicular to the surface) and dimmer as the viewing angle increases, creating realistic soft lighting that mimics real-world diffuse light sources like LED panels or softboxes.
+
+## Properties
+
+### Light Properties
+
+| Property | Type | Description | Required | Default |
+|:---------|:-----|:------------|:---------|:--------|
+| `color` | `number[3]` | RGB color of the light in linear space. Each component should be in the range [0, 1]. | No | `[1.0, 1.0, 1.0]` |
+| `intensity` | `number` | The luminance of the light surface in nits (cd/m²). Must be a positive value. | No | `1000.0` |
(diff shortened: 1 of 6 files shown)


extensions/2.0/Vendor/EXT_lights_area/README.md
```diff
@@ -0,0 +1,140 @@
...
+The light inherits the full transform of its parent node, including:
+
+* Position (translation)
+* Orientation (rotation) 
+* Scale
+
+If the node is scaled:
+
+* For rectangular lights: the width and height are scaled accordingly
+* For disk lights: the radius is scaled by the average of the X and Y scale factors
```
lexaknyazev: How exactly does this work? A node may inherit multiple transform matrices affecting all axes.

extensions/2.0/Vendor/EXT_lights_area/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+* `luminance` is in nits (cd/m²)
+* `area` is the surface area of the light in square meters:
+  * For rectangular lights: `area = width × height`
+  * For disk lights: `area = π × radius²`
+* The result is in lumens
+
+### Transform Inheritance
+
+The light inherits the full transform of its parent node, including:
```
lexaknyazev: The "transform of its parent node" may be misleading as "parent node" usually refers to a node containing the current node.

In this case, it should be "transform of the node instantiating the light".

extensions/2.0/Vendor/EXT_lights_area/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+* Position (translation)
+* Orientation (rotation) 
+* Scale
```
lexaknyazev: How would negative scale values work?

extensions/2.0/Vendor/EXT_lights_area/README.md
```diff
@@ -0,0 +1,140 @@
...
+### Light Properties
+
+| Property | Type | Description | Required | Default |
+|:---------|:-----|:------------|:---------|:--------|
+| `color` | `number[3]` | RGB color of the light in linear space. Each component should be in the range [0, 1]. | No | `[1.0, 1.0, 1.0]` |
+| `intensity` | `number` | The luminance of the light surface in nits (cd/m²). Must be a positive value. | No | `1000.0` |
+| `shape` | `string` | The shape of the area light. Must be either `"rect"` or `"disk"`. | Yes | - |
+| `width` | `number` | Width of the rectangular light in meters. Must be a positive value. Required for `rect` lights, ignored for `disk` lights. | Conditional | `1.0` |
+| `height` | `number` | Height of the rectangular light in meters. Must be a positive value. Required for `rect` lights, ignored for `disk` lights. | Conditional | `1.0` |
+| `radius` | `number` | Radius of the disk light in meters. Must be a positive value. Required for `disk` lights, ignored for `rect` lights. | Conditional | `1.0` |
```
lexaknyazev: For consistency, it might be better to follow the punctual lights schema with shape-specific properties located in nested objects.

Alternatively:
- There might be a common property, e.g., `size`. The rectangular shape may then have a `rect.aspectRatio` property while the `disk` shape wouldn't need anything else.
- All these properties could be assumed as always being 1.0 and the actual size would always be controlled by node transforms.

extensions/2.0/Vendor/EXT_lights_area/schema/light.area.schema.json
```diff
@@ -0,0 +1,60 @@
...
+                "type": "number",
+                "minimum": 0.0,
+                "maximum": 1.0
+            },
+            "minItems": 3,
+            "maxItems": 3,
+            "description": "RGB color of the light in linear space. Each component should be in the range [0, 1].",
+            "default": [1.0, 1.0, 1.0]
+        },
+        "luminance": {
```
lexaknyazev: This property name does not match the examples in the spec.

extensions/2.0/Vendor/EXT_lights_area/schema/light.area.schema.json
```diff
@@ -0,0 +1,60 @@
...
+            "type": "number",
+            "description": "The luminance of the light surface in nits (cd/m²). Must be a positive value.",
+            "default": 1000.0,
+            "minimum": 0.0,
+            "exclusiveMinimum": true
```
lexaknyazev: Why is zero not allowed here?

extensions/2.0/Vendor/EXT_lights_area/README.md
MiiBond: Great call. I'll make that change.

extensions/2.0/Vendor/EXT_lights_area/schema/light.area.schema.json
MiiBond: Mistake

extensions/2.0/Vendor/EXT_lights_area/schema/light.area.schema.json
MiiBond: Fixed. Thanks.

extensions/2.0/Vendor/EXT_lights_area/README.md
MiiBond: I've updated the readme to make the scaling a lot clearer. To avoid non-uniform and negative scaling, I've specified that the absolute value of the largest component of the node's world scale is used... I'm not sure I like that but I don't have a better idea.

extensions/2.0/Vendor/EXT_lights_area/README.md
MiiBond: I've rewritten this part now.

extensions/2.0/Vendor/EXT_lights_area/README.md
MiiBond: I'm disallowing them now by always taking the absolute value.

extensions/2.0/Vendor/EXT_lights_area/README.md
```diff
@@ -0,0 +1,134 @@
...
+When a light's `type` is `rect`, the `rect` property on the light is required. Its properties (below) are optional.
+
+| Property | Type | Description | Required | Default |
+|:---------|:-----|:------------|:---------|:--------|
+| `aspect` | `number` | Defines the relative size of the width of the rectangle compared to the height. | No | `1.0` |
+
+The rectangle area light's shape extends from -width/2 to +width/2 along the local X-axis, and from -height/2 to +height/2 along the Y-axis where width and height are defined as follows:
+
+`width = world scale * size * aspect`
+`height = world scale * size`
```
bghgary: I'm guessing you want this on two lines.

```suggestion
`width = world scale * size * aspect` \
`height = world scale * size`
```

extensions/2.0/Vendor/EXT_lights_area/README.md
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+    }
+]
+```
+
+The light will inherit the transform of the node in the following manner:
+1. The light's centre is defined as the node's world location.
+2. The light's direction is defined as the 3-vector `(0.0, 0.0, -1.0)` and the rotation of the node orients the light accordingly. That is, an untransformed light points down the -Z axis.
+3. The light's size is defined as its "size" property multiplied by the absolute value of the largest component of the node's world scale. i.e. the scaling is always uniform and positive.
```
bghgary: nit:
```suggestion
3. The light's size is defined as its "size" property multiplied by the absolute value of the largest component of the node's world scale, i.e., the scaling is always uniform and positive.
```

extensions/2.0/Vendor/EXT_lights_area/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Overview
```
bghgary, line 15: Would it help to add a couple of pictures with an example of a rect and a disk in the real world for illustration purposes?

bghgary (commented):
Small comments. Do we need to discuss whether this extension should be EXT or KHR?

siliconvoodoo:
I'm going to have some remarks to make regarding unit, beam angle and transformation.

## Spread

Blender cycles renderer adds the possibility to setup an area light emission distribution. Named "Beam shape's Spread". Alike to a BRDF a flat emitter usually respects a lambert law, thus has a "180 degrees emission distribution with a cosine lobe".

But more directional emitters may exist, like a TN LCD screen doesn't emit on all angles. This can happen with polarization filters, a LED array or a VCSEL Array, or some optic construct. No idea if the actual reality of them is relevant, but perhaps their convenience as an artistic gimmick may?

Another thing to consider is the implementation, not necessarily all engines will desire to support that property, perhaps it should be optional? But it would make it super sweet that export -> import full cycle preserves that property in any software including blender.

image

As one can see, it's not immediately trivial because of some echoing problem that may happen during shader implementation, I personally fixed them sprinkling some `min` and `saturate`:

image

See the violet line how it shows up again on the right of the X=0 axis? If one doesn't account for that, area lights can cause secondary lobe energy on their sides, it's funny 🤣 .

## Units

This is in connection with the unit, the choice of Nits and not Lux makes it that the beam angle becomes irrelevant to the intensity received on the lit surfaces. If that proposal changed in favor of Lux, it would behave like the blender effect, enlarging the beam diminishes intensity on the surface: (blender uses `Watt` that's why)

image

1 to 180 light array, purely for illustration purposes:

image

I think the `KHR_lights_punctual` spot light using `lm/sr` is coherent with the choice of Nit in this case (Nit is $\frac{\mathrm{lm}}{\mathrm{m}^2 \cdot \mathrm{sr}}$  )

## Transform

For the scale:
> world scale` is the largest component of the absolute value of the node's scale.

I'm guessing this is in synch with the behaviour of Unity.
The problem with allowing the usual transform matrix to transform the area lights is that with a few combinations of non-uniform-scale and rotate in the hierarchical parent nodes, shear becomes possible. A sheared disk is an ellipse, a sheared rect light is a parallelogram light. Which are both hard to support in shaders, and most implementers would probably overlook that possibility resulting in broken cases.
Using the `max(|x|, |y|, |z|)` somehow feels like a guard against odd transforms but at the same time meeting in the middle with the scale request. Unity forgoes the scale matrix at the last node-level of the scene graph.

image

In my implementation to reproduce Unity behavior I did something in the likes of:
```
let worldmatrix = node->multiply_hierarchy_tm();
vec3 scale;
decompose_matrix(worldmatrix, &scale);
lightmatrix = vec3(rect.x, rect.y, 1) * (1 / scale) * worldmatrix;
```

the `1/scale` part is optional when detecting a Unity origin data. (gltFast exporter with some arrangements).
I wrote a report document looking like this as a memo after that:

image

If the optional part is off, then sheared area lights (prallelogram lights) are totally possible. I just chose to accept them and adjust my shaders to support them, this happens as a difference if I disable the Unity-flag in the importer:

Left is full transform, Right is Unity:
image

As you can see, even the direction (-z) become slanted with respect to the base light plane, which makes them difficult to handle. Not impossible, but not trivial.

I'd recommend a spec that simplifies this in the spirit of Unity. But doesn't make it too stupidly hard to understand in respect of the parent hierarchy of transforms either. So perhaps your max method + one word about what scale you are talking about. The current node final-result scale, meaning _the scale as the result of extracting the scale vector from the final child matrix of the light node after multiplication with all the parents_.  or something

PR #2530, opened 2025-09-25
KHR_materials_clearcoat_ior

MiiBond:
This adds an IOR uniform to the coat layer.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_clearcoat_ior/schema/material.KHR_materials_clearcoat_ior.schema.json
+++ b/extensions/2.0/Khronos/KHR_materials_clearcoat_ior/schema/material.KHR_materials_clearcoat_ior.schema.json
@@ -1,0 +1,26 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_materials_clearcoat_ior glTF Material Extension",
+    "type": "object",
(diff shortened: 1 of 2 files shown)


extensions/2.0/Khronos/KHR_materials_clearcoat_ior/schema/material.KHR_materials_clearcoat_ior.schema.json
```diff
@@ -0,0 +1,26 @@
...
+    "description": "glTF extension that defines the index of refraction of the clearcoat layer.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "clearcoatIor": {
+            "type": "number",
+            "description": "The index of refraction of the clearcoat layer.",
+            "default": 1.5,
+            "oneOf": [
+                {
+                    "minimum": 0.0,
```
MiiBond, line 14: This is how the base layer IOR extension is defined but, to be honest, I don't quite understand this. What's with the minimum: 0 and maximum 0?

lexaknyazev: The base IOR extension allows `(ior == 0.0) || (ior >= 1.0)`. The zero is a special case to support legacy content, see the extension spec for details.

PR #2531, opened 2025-09-29
KHR_gaussian_splatting_compression_spz_2

weegeekps:
This is a draft extension proposal for using [Niantic Spatial's SPZ compression library](https://github.com/nianticlabs/spz) for compression 3D Gaussian splatting data using the base [KHR_gaussian_splatting](https://github.com/KhronosGroup/glTF/pull/2490) extension with glTF. This extension targets version 2 of SPZ.

As we get closer to done with the base extension for 3D Gaussian splatting, [KHR_gaussian_splatting](https://github.com/KhronosGroup/glTF/pull/2490), it has made more and more sense to break out the `KHR_gaussian_splatting_compression_spz_2` extension for SPZ compression of 3DGS data. This allows them each to independently go through the ratification process and neither will hold up the other.

For previous history and conversations regarding the development of this extension, please look at the PR for [KHR_gaussian_splatting before this linked comment](https://github.com/KhronosGroup/glTF/pull/2490#issuecomment-3348956097).

### DIFF
--- a/extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/README.md
+++ b/extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/README.md
@@ -1,0 +1,165 @@
+# KHR\_gaussian\_splatting\_compression\_spz\_2
+
+## Contributors
+
+- Jason Sobotka, Cesium
+- Renaud Keriven, Cesium
+- Adam Morris, Cesium
+- Sean Lilley, Cesium
+- Projit Bandyopadhyay, Niantic Spatial
+- Daniel Knoblauch, Niantic Spatial
+- Ronald Poirrier, Esri
+- Jean-Philippe Pons, Esri
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+Depends on the `KHR_gaussian_splatting` extension specification for attribute definitions.
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Compressing 3D Gaussian splats using SPZ](#compressing-3d-gaussian-splats-using-spz)
+  - [Geometry Type](#geometry-type)
+  - [Schema Example](#schema-example)
+  - [Extension Properties](#extension-properties)
+  - [Accessors](#accessors)
+- [Conformance](#conformance)
+- [Schema](#schema)
+- [Known Implementations](#known-implementations)
+- [Resources](#resources)
+
+## Overview
+
+This extension defines support for compressing 3D Gaussian splats stored within in glTF using [v2 of the SPZ compression format](https://github.com/nianticlabs/spz/tree/v2.0.0) for efficient streaming and storage.
+
+SPZ is a compression format from Niantic Spatial designed for Gaussian splats. Currently, it is open sourced under the MIT license. The SPZ format is primarily used in the Niantic Spatial Scaniverse app. It was purpose built for Gaussian splats and offers a balance of high compression with minimal visual fidelity loss, and allows for storing some or all of the 3D Gaussian's spherical harmonics.
+
+At rest, the 3D Gaussian splats are stored within the SPZ compression format. Upon being decompressed, values are mapped to the glTF attributes defined by `KHR_gaussian_splatting` for rendering. The [conformance](#conformance) section defines what an implementation must do when encountering this extension, and how the extension interacts with the base specification.
+
+## Compressing 3D Gaussian splats using SPZ
+
+To use this extension, it must be defined in the `extensions` property in a `KHR_gaussian_splatting` extension definition. Any mesh primitive using `KHR_gaussian_splatting` that is using this extension will use the SPZ payload to retreive the values for `POSITION`, `COLOR_0`, `KHR_gaussian_splatting:SCALE`, `KHR_gaussian_splatting:ROTATION`, and all `KHR_gaussian_splatting:SH_DEGREE_ℓ_COEF_n` attributes. The [attribute mapping](#attribute-mapping) section defines how the SPZ data is mapped to these attributes.
+
+The extension must then be listed in `extensionsUsed` alongside `KHR_gaussian_splatting`.
+
+```json
+  "extensionsUsed" : [
+    "KHR_gaussian_splatting",
+    "KHR_gaussian_splatting_compression_spz"
+  ]
+```
+
+### Inheritance from base extension
+
+As this extension extends the base extension, all components of the base extension are inherited. Any modifications to the base definition are explicitly defined in the sections below.
+
+### Schema Example
+
+Example SPZ extension shown below. This extension only affects any `primitive` nodes containting Gaussian splat data. A `bufferView` is provided by the extension which points to where the SPZ blob is stored.
+
+```json
+  "meshes": [{
+      "primitives": [{
+          "attributes": {
+            "POSITION": 0,
+            "COLOR_0": 1,
+            "KHR_gaussian_splatting:SCALE": 2,
+            "KHR_gaussian_splatting:ROTATION": 3,
+            "KHR_gaussian_splatting:SH_DEGREE_1_COEF_0": 4,
+            "KHR_gaussian_splatting:SH_DEGREE_1_COEF_1": 5,
+            "KHR_gaussian_splatting:SH_DEGREE_1_COEF_2": 6
+          },
+          "material": 0,
+          "mode": 0,
+          "extensions": {
+            "KHR_gaussian_splatting": {
+              "extensions": {
+                "KHR_gaussian_splatting_compression_spz": {
+                  "bufferView": 0,
+                }
+              }
+            }
+          }
+        }]
+    }],
+  "buffers": [{
+      "uri": "0.bin",
+      "byteLength": 9753142
+    }],
+  "bufferViews": [{
+      "buffer": 0,
+      "byteLength": 9753142
+    }],
+```
+
+### Extension Properties
+
+#### bufferView
+
+This property points to the bufferView containing the Gaussian splat data compressed with SPZ.
+
+### Attribute Mapping
+
+Data may be used directly from SPZ or may be mapped to the placeholder attributes this extension provides. When mapping to attributes, the data from SPZ is mapped from the `GaussianCloudData` struct in SPZ to the following attributes in glTF. Several fields require additional conversion to make them glTF and renderer ready.
+
+| SPZ `GaussianCloudData` field | glTF Attribute | Required Conversion |
+| --- | --- |
+| `positions` | `POSITION` | |
+| `colors` | `COLOR_0` RGB components | Compute _0.5 + 0.282095 * x_ to get the color between 0 and 1. |
+| `alphas` | `COLOR_0` A component | Compute the sigmoid to get the alpha between 0 and 1. (Sigmoid formula: _1 / (1 + e^-x)_) |
+| `scales` | `KHR_gaussian_splatting:SCALE` | Compute the base-e exponential of each scale value. (e.g. _e^x_ or `std::exp(x)`) |
+| `rotations` | `KHR_gaussian_splatting:ROTATION` | |
+| `sh` index 0 to 2 | `KHR_gaussian_splatting:SH_DEGREE_1_COEF_0` | |
+| `sh` index 3 to 5 | `KHR_gaussian_splatting:SH_DEGREE_1_COEF_2` | |
+| `sh` index 6 to 8 | `KHR_gaussian_splatting:SH_DEGREE_1_COEF_3` | |
+| `sh` index 9 to 11 | `KHR_gaussian_splatting:SH_DEGREE_2_COEF_0` | |
+| `sh` index 12 to 14 | `KHR_gaussian_splatting:SH_DEGREE_2_COEF_1` | |
+| `sh` index 15 to 17 | `KHR_gaussian_splatting:SH_DEGREE_2_COEF_2` | |
+| `sh` index 18 to 20 | `KHR_gaussian_splatting:SH_DEGREE_2_COEF_3` | |
+| `sh` index 21 to 23 | `KHR_gaussian_splatting:SH_DEGREE_2_COEF_4` | |
+| `sh` index 24 to 26 | `KHR_gaussian_splatting:SH_DEGREE_3_COEF_0` | |
+| `sh` index 27 to 29 | `KHR_gaussian_splatting:SH_DEGREE_3_COEF_1` | |
+| `sh` index 30 to 32 | `KHR_gaussian_splatting:SH_DEGREE_3_COEF_2` | |
+| `sh` index 33 to 35 | `KHR_gaussian_splatting:SH_DEGREE_3_COEF_3` | |
+| `sh` index 36 to 38 | `KHR_gaussian_splatting:SH_DEGREE_3_COEF_4` | |
+| `sh` index 39 to 41 | `KHR_gaussian_splatting:SH_DEGREE_3_COEF_5` | |
+| `sh` index 42 to 44 | `KHR_gaussian_splatting:SH_DEGREE_3_COEF_6` | |
+
+### Accessors
+
+Accessor requirements are modified from the base `KHR_gaussian_splatting` extension with the following adjustments to definition:
+
+ - Accessor `type` is defined for the resulting type after decompression and dequantization has occurred.
+ - The accessor `count` must match the number of points in the compressed SPZ data.
+
+## Conformance
+
+The recommended process for handling SPZ compression is as follows:
+
+- If the loader does not support `KHR_gaussian_splatting_compression_spz` and `accessor.bufferView` is undefined, other extensions may be providing the data.
+- If the loader does not support `KHR_gaussian_splatting_compression_spz` and `accessor.bufferView` is defined, accessor data should be sourced as usual.
+- If the loader does support `KHR_gaussian_splatting_compression_spz` then the loader must process `KHR_gaussian_splatting_compression_spz` data first. The loader must get the data from `KHR_gaussian_splatting_compression_spz`'s `bufferView` extension property.
+
+This allows for a graceful fallback when an implementation does not support this extension.
+  
+When compressing or decompressing the SPZ data to be stored within the glTF, you must specify a Left-Up-Front (`LUF`) coordinate system in the SPZ `PackOptions` or `UnpackOptions` within the SPZ library. This ensures that the data is compressed and decompressed appropriately for glTF.
+
+## Schema
+
+[SPZ Compression Schema](./schema/mesh.primitive.KHR_gaussian_splatting_compression_spz.schema.json)
+
+## Known Implementations
+
+This is currently implemented within [3D Tiles and CesiumJS as an experimental feature](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html?classFilter=3D).
+
+## Resources
+
+[https://github.com/nianticlabs/spz](https://github.com/nianticlabs/spz)
+
+[https://github.com/drumath2237/spz-loader/tree/main](https://github.com/drumath2237/spz-loader/tree/main)
--- a/extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/README.md
+++ b/extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/README.md
@@ -80,8 +80,8 @@ Example SPZ extension shown below. This extension only affects any `primitive` n
           "extensions": {
             "KHR_gaussian_splatting": {
               "extensions": {
-                "KHR_gaussian_splatting_compression_spz": {
-                  "bufferView": 0,
+                "KHR_gaussian_splatting_compression_spz_2": {
+                  "bufferView": 0
                 }
               }
             }
--- a/extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/schema/mesh.primitive.KHR_gaussian_splatting_compression_spz_2.schema.json
+++ b/extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/schema/mesh.primitive.KHR_gaussian_splatting_compression_spz_2.schema.json
@@ -1,0 +1,16 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "KHR_gaussian_splatting_compression_spz_2 glTF Mesh Primitive Extension",
+    "type": "object",
+    "description": "Compressed data for SPZ v2 primitive.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "bufferView": {
+            "allOf": [ { "$ref" : "glTFid.schema.json" } ],
+            "description": "The index of the bufferView."
+        },
+        "extensions": { },
+        "extras": { }
+    },
+    "required": [ "bufferView" ]
+}


extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/README.md
```diff
@@ -0,0 +1,165 @@
...
+            "KHR_gaussian_splatting:SH_DEGREE_1_COEF_2": 6
+          },
+          "material": 0,
+          "mode": 0,
+          "extensions": {
+            "KHR_gaussian_splatting": {
+              "extensions": {
+                "KHR_gaussian_splatting_compression_spz": {
+                  "bufferView": 0,
+                }
```
lilleyse:
```suggestion
                "KHR_gaussian_splatting_compression_spz_2": {
                  "bufferView": 0
                }
```

weegeekps: Nice catch. Thank you.

extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/README.md
```diff
@@ -0,0 +1,165 @@
...
+#### bufferView
+
+This property points to the bufferView containing the Gaussian splat data compressed with SPZ.
+
+### Attribute Mapping
+
+Data may be used directly from SPZ or may be mapped to the placeholder attributes this extension provides. When mapping to attributes, the data from SPZ is mapped from the `GaussianCloudData` struct in SPZ to the following attributes in glTF. Several fields require additional conversion to make them glTF and renderer ready.
+
+| SPZ `GaussianCloudData` field | glTF Attribute | Required Conversion |
+| --- | --- |
```
javagl, line 112: The table formatting is broken

javagl:
Not sure whether it's on the radar (I know that there currently is some turmoils around this topic...), but the wording and description around `COLOR_0` may have to be updated.

_This_ extension casually mentions `COLOR_0`, "as-if" it was a required attribute in the base extension. But in the base extension, it is described as an optional fallback. (I think that this means that there will be "fewer conversions" necessary, so in some way, it may actually _simplify_ things, but the details should be reviewed carefully)

weegeekps:
@javagl I have updates in progress that bring this in line with the current base spec including the changes around color spaces. I plan on pushing those changes as soon as that work is fully settled and we have consensus.

extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+The recommended process for handling SPZ compression is as follows:
+
+- If the loader does not support `KHR_gaussian_splatting_compression_spz` and `accessor.bufferView` is undefined, other extensions may be providing the data.
+- If the loader does not support `KHR_gaussian_splatting_compression_spz` and `accessor.bufferView` is defined, accessor data should be sourced as usual.
+- If the loader does support `KHR_gaussian_splatting_compression_spz` then the loader must process `KHR_gaussian_splatting_compression_spz` data first. The loader must get the data from `KHR_gaussian_splatting_compression_spz`'s `bufferView` extension property.
+
+This allows for a graceful fallback when an implementation does not support this extension.
+  
+When compressing or decompressing the SPZ data to be stored within the glTF, you must specify a Left-Up-Front (`LUF`) coordinate system in the SPZ `PackOptions` or `UnpackOptions` within the SPZ library. This ensures that the data is compressed and decompressed appropriately for glTF.
```
larsdal-hexagon, line 151: Hi, I've been looking into how KHR_gaussian_splatting_compression_spz_2 is implemented in Cesium, and there the spz data is stored in LUF coordinates, rather than the standard RUB, with no conversion on load.
Is the intention here for data in the spz files to be stored in the standard RUB coordinates and converted to/from LUF on read/write, or for the spz data to be stored in LUF coordinates like in the current Cesium implementation?
My interpretation of the spec would indicate the former.

javagl: I think that this is one point that (has been brought up occasionally and) may have to be clarified. The comment at https://github.com/CesiumGS/cesium/issues/12682#issuecomment-3039079463 refers to an _older_ state of the implementation, and it seems like the _specific_ rotations that this comment is linking to are no longer applied. But it might be that some rotations are now done elsewhere.

(An aside: The rotation was originally done by the SPZ-loader library. And this is using the [spz `convertCoordinates`](https://github.com/nianticlabs/spz/blob/ef094fd1a96ca6ff414d72d7904ee4f4f6d97be9/src/cc/splat-types.h#L134) function under the hood. And I'm about 95% sure that this implementation is wrong. But I haven't confirmed that yet...)

javagl:
Coming back to the `COLOR_0` handling mentioned above:

According to the current base extension specification, the `COLOR_0` attribute is only a "fallback". It can not replace the `SH_DEGREE_0_COEF_0`+`OPACITY` attributes from the base extension. One task here is to align this requirement with the SPZ compression extension. The SPZ library/format itself always stores separate color and alpha, so that could be _relatively_ easy to align with the glTF world. It mainly means that these values have to be converted according to their domain (e.g. the color into the SH coefficients).

weegeekps:
Just so folks are aware: I have a lot of updates in queue for this extension, but I am waiting to see the final editorial draft of the base extension.

> It mainly means that these values have to be converted according to their domain (e.g. the color into the SH coefficients).

Most all of the SPZ loaders immediately apply `0.282` and the `0.5` offset to the `colors` result and return an RGBA with opacity encoded in. As a clarifying point, I'm saying "most", but the truth is all of the loaders I've seen do this. There _may_ be a loader out there that doesn't that I haven't seen, but the popular ones including the Babylon.js loader, spz-loader library, and others all perform this calculation immediately.

As far as I can tell, this is very much in the "spirit" of SPZ. I don't believe we should require language to the SPZ extension to require implementations to change. I believe this is also why it's called "colors" and not DC or SH0. (DC with spherical harmonics does not mean diffuse component although that is a common point of confusion.) This is one of the reasons we put this statement in the spec:

> Extensions extending this extension may define alternative lighting methods, have specific requirements for handling compression, or define different spherical harmonics handling.

javagl:
> Most all of the SPZ loaders immediately apply 0.282

And I'm not even sure that it is "correct" to do this, because it does not anticipate the "colorScale" in SPZ. See the "Colors" point at the bottom of the first comment of https://github.com/nianticlabs/spz/issues/42

But the main reason for "reviving" this PR by adding that comment:

We talked about some of that in the 3D Formats call. And one point that we talked about was that the `COLOR_0` can **not** replace the `SH_DEGREE_0_COEF_0+OPACITY`. That wasn't self-evident for me. I thought that it _could_ be reasonable to see that `COLOR_0` as an ~"alternative way" of encoding that data. But the consensus was that content creators should **not** mix the "real" attributes and that "fallback" attribute. 

That reduces ambiguities and simplifies the loaders that _do_ support the extension. It has the drawback that any glTF asset that wants to offer the fallback path has to store this as _additional_ (!) data. If it was non-exclusive, people could opt in for _always_ representing `SH0+OPACITY` as `COLOR_0`. But I don't have a strong opinion.

(And I know one "creator" of glTF-SPZ splat file that _does_ always write the `COLOR_0` and _omit_ the opacity. You probably know where this could be tracked)

weegeekps:
> And I'm not even sure that it is "correct" to do this, because it does not anticipate the "colorScale" in SPZ.

Under the covers, SPZ packs these as RGB for stability. While doing so and to support values greater than 1.0, they use a scale of `0.15`. When unpacking, it reverses this calculation leaving you with the DC components with the intent of you applying the expected scale of 0.282 in your codebase.

> And one point that we talked about was that the COLOR_0 can not replace the SH_DEGREE_0_COEF_0+OPACITY.

On the surface I think this is a reasonable take, but when I think about it more it feels like we've pushed past the fine line we've been treading. There is a separation of concerns matter here that I think we have to account for; we've made it upfront in the base extension that other extensions, such as compression extensions, may have slightly different ways of providing the data. This rule violates that policy and to me it seems like we're telling compression extensions that if you don't fit this mold correctly, you can't build off of this base extension. _That seems very wrong to me._

There are some maths that we must specify for consistent rendering of 3DGS from glTF. The `kernel` property specifies those. We should not mandate that a particular compression format cannot provide the variables for those maths in a different manner than the base extension would. At that point we're mandating implementation details that should be left to the implementers.

gugu23456789:
> 这是一个基于 glTF 的基础 [KHR_gaussian_splatting](https://github.com/KhronosGroup/glTF/pull/2490) 扩展，使用 [Niantic Spatial 的 SPZ 压缩](https://github.com/nianticlabs/spz)库进行压缩 3D 高斯散布数据的草稿扩展提案。该扩展面向SPZ第2版。
> 
> 随着我们接近完成3D高斯喷溅的基础扩展，[KHR_gaussian_splatting](https://github.com/KhronosGroup/glTF/pull/2490)，拆分用于3DGS数据的SPZ压缩扩展越来越有意义。这使得双方各自独立完成批准程序，双方都不会拖延对方。`KHR_gaussian_splatting_compression_spz_2`
> 
> 关于此扩展开发的历史和讨论，请查看[本链接评论前](https://github.com/KhronosGroup/glTF/pull/2490#issuecomment-3348956097)的PR KHR_gaussian_splatting。




Hi @Weegeekps and Khronos glTF community,

This is a fascinating and much-needed proposal. We’ve been closely following the development of the KHR_gaussian_splatting and KHR_gaussian_splatting_compression_spz_2 extensions.

As the draft approaches a critical stage, we wanted to share that a complete, production-ready, and community-vetted open-source toolchain implementing this specification already exists and is actively used.

We are the maintainers of the SPZ Open-Source Ecosystem (spz-ecosystem), and we have built two core, interoperable projects that form a full pipeline for SPZ v2:

**spz_gatekeeper - The Format Compliance Engine**
*   **L2-Only Legal Validator**: A pure C++17/WASM tool that performs rigorous, non-destructive validation of SPZ headers, flags, and TLV trailer extensions. It defines a clear contract for what constitutes a compliant SPZ derivative.
*   **Governance & Registry**: Features a split ExtensionSpecRegistry / ExtensionValidatorRegistry system, providing a framework for registering, documenting, and validating vendor extensions (starting with 0xADBE0002).
*   **Production-Grade**: Includes a fully functional WebAssembly UI for in-browser validation, comprehensive CI/CD, and a dual-audit model (browser_lightweight_wasm_audit + local_cli_spz_artifact_audit).
*   Repository: [spz_gatekeeper](https://github.com/spz-ecosystem/spz_gatekeeper)

**spz2glb - The High-Performance Conversion Hub**
*   **Lossless, Spec-Compliant Packaging**: A converter that embeds SPZ streams byte-for-byte into GLB containers using the KHR_gaussian_splatting_compression_spz_2 extension. It features a built-in 3-layer verification suite (structure, losslessness, decode-consistency).
*   **Web-Optimized & Performant**: Offers a high-performance WebAssembly build, enabling sub-second, client-side conversion in the browser with no data upload. Our benchmarks show ~506ms for a 25MB SPZ file.
*   **Ecosystem Integration**: Our output is already being used experimentally by community projects like [Real3DViewer](https://github.com/real3d-com/Real3dViewer), validating the end-to-end workflow from SPZ to rendered content.
*   Repository: [spz2glb](https://github.com/spz-ecosystem/spz2glb)

**Why This Matters for the Spec:**
*   **Provides a Production Reference**: We offer a neutral, community-driven implementation that can serve as a concrete, testable reference for spec discussions and compatibility testing.
*   **Solves Real Tooling Gaps**: Our tools address the immediate needs for validation, conversion, and developer onboarding in the SPZ format.
*   **Independent & Safe Foundation**: The projects include clear legal and political separation statements, ensuring they are a clean, vendor-neutral foundation for the ecosystem.

We are fully aligned with the goal of a robust, interoperable standard. If helpful, we are ready to:
*   Run compatibility tests using our toolchain against various draft iterations.
*   Provide detailed feedback from our implementation experience.
*   Collaborate on fostering a broader open-source ecosystem around this standard.

The repositories are under active development (currently at v2.0.1), and we welcome any feedback or collaboration.

Best regards,
The spz-ecosystem maintainers.

extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/README.md
weegeekps: Hi, @larsdal-hexagon. If you're using 3D Tiles 1.1, use `CoordinateSystem::UNSPECIFIED` and either train the splats in the LUF coordinate system or convert them into the LUF coordinate system prior to packing.

To be clear, this is **only** for _3D Tiles 1.1_. If you're doing pure glTF, unpack using the correct `CoordinateSystem::LUF` value.

gugu23456789:
Hi @weegeekps,

I'm the maintainer of the neutral SPZ ecosystem tooling:

· spz_gatekeeper – an open-source SPZ compliance validator (currently actively aligning with Niantic's latest v3.0.0 / SPZ v4 release).
· spz2glb – a converter implemented against the current SPZ_2 draft (this PR).

To better understand the intended direction of this draft and ensure consistency between the specification and the actual evolving ecosystem, could you kindly clarify:

Will this SPZ_2 draft be updated to match the SPZ v4 reference implementation (Zstd compression, new extension layout, etc.), or is it expected to remain anchored to the older SPZ v2 specification?

As an early external implementer, I'd be happy to contribute testing or feedback to help keep the specification interoperable with the wider ecosystem.

Thank you for your guidance!

extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+| SPZ `GaussianCloudData` field | glTF Attribute | Required Conversion |
+| --- | --- |
+| `positions` | `POSITION` | |
+| `colors` | `COLOR_0` RGB components | Compute _0.5 + 0.282095 * x_ to get the color between 0 and 1. |
+| `alphas` | `COLOR_0` A component | Compute the sigmoid to get the alpha between 0 and 1. (Sigmoid formula: _1 / (1 + e^-x)_) |
+| `scales` | `KHR_gaussian_splatting:SCALE` | Compute the base-e exponential of each scale value. (e.g. _e^x_ or `std::exp(x)`) |
+| `rotations` | `KHR_gaussian_splatting:ROTATION` | |
+| `sh` index 0 to 2 | `KHR_gaussian_splatting:SH_DEGREE_1_COEF_0` | |
+| `sh` index 3 to 5 | `KHR_gaussian_splatting:SH_DEGREE_1_COEF_2` | |
```
ntandesri, line 119: It should be `KHR_gaussian_splatting:SH_DEGREE_1_COEF_1` and `KHR_gaussian_splatting:SH_DEGREE_1_COEF_2` (instead of 2 & 3).

weegeekps:
Hi @gugu23456789 (and anyone else watching), please stay tuned. We are working on this and will have a clarification posted as soon as we can.

gugu23456789:
@weegeekps and working group members — thank you for your patience on this PR. We fully understand that the working group's current focus is on finalizing the base extension's editorial draft, and this message is in no way meant to rush any timeline. We would simply like to share some engineering observations from our reference implementations of the current draft, as input for the community and working group to consider. These points reflect our own implementation experience; we may well have misunderstandings, and we welcome corrections.

1. Relationship between the draft and SPZ v4 In our implementation we noticed that the draft's naming and description still anchor to SPZ v2, while Niantic's reference implementation has since moved to SPZ v4 (Zstd compression, 32-byte header, ILV header-zone extension layout); our own converter already produces v4-based output, so we would be curious to learn whether the working group plans to have the draft explicitly cover v4, and whether our reading is correct.

2. GLB packaging approach We currently package the original SPZ stream byte-identically into the GLB (SPZ stored as-is in buffer → bufferView → spz_2.bufferView , so GLB size ≈ SPZ size); in practice this preserves the "lossless" semantics the draft describes well, and we would like to ask whether the draft intends to allow or encourage re-compression, or keep it strictly pass-through.

3. JSON fields Besides the draft's bufferView , our current GLB JSON also carries three metadata fields — spzVersion , compression , coordinateSystem — which are non-standard since the draft schema only defines bufferView ; we are unsure whether they should be standardized or better placed in extras , and would appreciate the community's thoughts.

4. How extension fields are added We add fields by extending fastgltf's GaussianSplatSpzCompression struct ( bufferView + the three metadata fields) and serializing them in the JSON writer; the mechanism itself works well, but a canonical field set in the schema would likely improve interoperability.

5. Handling of the ILV 003 extension We parse the Adobe coordinate system extension ( 0xADBE0003 ) from the SPZ v4 ILV header zone, treat it as descriptive metadata rather than an instruction (consistent with the "descriptor, not instruction" consensus in the base PR discussion), validate its value range [0, 16], and record it in the GLB metadata; we would like to ask how this coordinate system information should be reconciled with the draft's LUF requirement on the glTF side — or whether our understanding is off.

6. High-performance WASM and queue mode Both projects ship usable high-performance WASM builds — spz2glb (memory pooling + device-aware memory management) and spz_gatekeeper (zero-copy pointer inspection + serial batch queue) — enabling in-browser SPZ→GLB conversion and compliance validation; this may indicate the extension has practical potential for web-scale use, though there may well be gaps we haven't considered.

7. Benchmark dataset (experimental early work) Our companion spz-anime-text2scene-bench dataset (6/10 scenes, ~587 MB, CC BY 4.0) is currently an experimental early-stage effort that we ourselves treat as a "to-be-improved baseline"; we welcome community feedback and plan to expand coverage and align formats as the specification stabilizes.

Thank you again for the working group's patience. We are happy to contribute testing and feedback as a reference implementation, and we look forward to the clarification whenever the working group is ready to post it.

javagl:
Disclaimer: I don't have the authority to answer these questions. Until Adam chimes in, I'll share my thoughts and interpretations. These should be considered as purely subjective.

---

> Relationship between the draft and SPZ v4 

This is related to the first point that was brought up in the [discussion of the original PR](https://github.com/KhronosGroup/glTF/pull/2490#issuecomment-2902604707) (even before splitting out the base extension). From what I see, there are roughly three options:

- Leave this extension to be specific for SPZ v2
- Introduce new extensions, `...spz_3`, `...spz_4`....
- Generalize this extension to cover other SPZ versions

All this is revolving around aspects of stability and backward compatibility. Of course, when SPZ versions 5, 6, 7 ... are published, then this should not imply `...spz_5`, `...spz_6`, `...spz_7` extensions. When the current extension is generalized, then it _still_ has to be limited to a particular set of supported SPZ versions (as supporting the extension _now_ cannot mean that it will also work with _future_ SPZ versions).

And to be clear: SPZ v4 is a new file format. It is completely unrelated to SPZ v2 (except for trivial details, like quantizing a float into a byte). It could have been published by a different company, with a different name, and nobody would even think of asking about any form of "compatibility" between SPZ v2 and this new file format.

---

> GLB packaging approach We currently package the original SPZ stream byte-identically into the GLB

This sounds reasonable. Compressing data that is already compressed rarely makes sense. (In addition to the complexity that could result from adding another "layer" of compression here, both for the specification and for the implementations)

--- 

> JSON fields Besides the draft's bufferView , our current GLB JSON also carries three metadata fields — spzVersion , compression , coordinateSystem

The `spzVersion` is related to the first point. It still raises the question about the _possible_ values for this field (I assume 2, 3, 4, but not 1 or 5...)
The `compression` is something that I'd consider to be an inherent part of the SPZ payload.
The `coordinateSystem` seems to be related to [a comment above](https://github.com/KhronosGroup/glTF/pull/2531#pullrequestreview-3896032564), and https://github.com/KhronosGroup/glTF/issues/2613 (even for the base extension). glTF does (fortunately) define a single coordinate system, and I'd be leaning towards using this coordinate system for splat data as well (regardless of the encoding). **Iff** SPZ allows for different coordinate systems, then this should be made absolutely clear.

Related: https://github.com/nianticlabs/spz/issues/14 , https://github.com/nianticlabs/spz/issues/84 , https://github.com/nianticlabs/spz/issues/42 ... (these questions don't just go away - they'll bubble up again and again...)

---

> How extension fields are added We add fields by extending fastgltf's GaussianSplatSpzCompression struct ( bufferView + the three metadata fields) and serializing them in the JSON writer; the mechanism itself works well, but a canonical field set in the schema would likely improve interoperability.

This sounds like an implementation detail. In the previous point, you mentioned that you stored some ~"(arbitrary) additional fields", and of course, these may not be supported by some serialization layer, unless they are officially made part of the specification and its JSON schema.

---

> Handling of the ILV 003 extension 

This sounds like something that came in with SPZ v4, and is not (and cannot be) handled in this extension unless it is explicitly generalized to handle SPZ v4 as well. (And it _looks_ like it's a good example why claiming support for ~"different versions" can be difficult. If such a new format can add ~"arbitrary metadata", it may be necessary to be very explicit about which parts or flavors of this metadata are anticipated to make sense in the context of glTF).

---

> High-performance WASM and queue mode 

I think this is related to the specific library that is used for loading the SPZ data. And it may unfold into two (fairly separate) questions: 

1. Which versions of SPZ does that library support? (For example, https://github.com/drumath2237/spz-loader is used in some projects, but only supports SPZ v2...)
2. Which _technology_ is that library using? (Referring to WASM, zero-copy or whatnot...)

---

> Benchmark dataset (experimental early work) 

You seem to refer to https://github.com/spz-ecosystem/spz-anime-text2scene-bench (aber ich habe mir das noch nicht näher angeschaut, weil ich dazu Google Translate benutzen müßte). In general, the questions about data sets also fall into two categories:

1. Compliance tests. These can usually be "small" data sets with specific features. I started creating some of these at https://github.com/KhronosGroup/glTF/issues/2562 
2. Benchmark data sets. Of course, these should be based on real-world example data sets. But I think that for _some_ forms of analysis, it can also be very useful to generate _artificial_ test data. I've been doing that in some very narrow contexts as well. But usually, there are several "dimensions" along which this test data set generation can happen. A very shallow one is whether the goal of these benchmarks is to analyze the performance of a renderer, or the compression ratios that can be achieved with a certain compression method (and the nature and structure of the data sets will be vastly different depending on the exact goal)

gugu23456789:
@javagl  — thank you very much for the detailed and candid response, and for the related issues you linked. We cross-checked several points against both of our projects ([spz_gatekeeper](https://github.com/spz-ecosystem/spz_gatekeeper) and [spz2glb](https://github.com/spz-ecosystem/spz2glb)), and would like to add the following details.

**1. How we support multiple SPZ versions**

Both projects dispatch on the compression magic, rather than assuming a single version:

- **spz2glb** first detects the magic — gzip (`0x1F8B`) routes to the SPZ v2/v3 header parser, Zstd (`0x28B52FFD`) routes to the v4 32-byte header parser. Rotation data is handled per version (`version >= 3` selects the 4-byte smallest-three quaternion layout; v2 uses the 3-byte layout). The v4-only ILV header zone is only read on the Zstd path.
- **spz_gatekeeper** applies a version policy: `version < 1` → error, `version 1..4` → normal validation, `version > 4` → warning but continue (via the `kKnownMaxVersion` constant). v4 uses its own header parsing path. Compression mapping: v3 = gzip, v4 = Zstd.

**2. How we validate across versions**

- **spz2glb** ships a built-in five-layer verification: L1 GLB structure & KHR extension contract → L2 binary losslessness (byte-identical to the original SPZ) → L3 decoding consistency → L4 GLB↔SPZ metadata consistency → L5 ILV extension integrity (TLV structure + 003 value range [0, 16]).
- **spz_gatekeeper** is the format-compliance engine: it parses the ILV layout `[u32 type][u32 length][value]` and **skips unknown types via `length`** (no false positives). Semantic validation is only applied to extensions registered in the extension registry (currently `0xADBE0002` / `0xADBE0003`) — which is precisely the "boundary" you mentioned: registered = semantic commitment, unregistered = structurally valid but no semantic promise.

**3. How we handle the 003 extension**

Both projects treat `0xADBE0003` (Adobe Coordinate System) consistently: as a **descriptor, not an instruction** (no coordinate transformation — that is left to the renderer's `coordinateConverter()`). L5 validation enforces the value range [0, 16]. The ILV header zone is located via `tocByteOffset` in the SPZ v4 header, whose field layout is aligned with the upstream Niantic `NgspFileHeader` (32-byte) spec — identical parsing in both projects.

**4. The GLB structure we produce**

The GLB JSON carries the full extension chain: outer `KHR_gaussian_splatting {kernel, colorSpace, sortingMethod, projection}`, nested `KHR_gaussian_splatting_compression_spz_2 {bufferView, spzVersion, compression, coordinateSystem}` — `bufferView` points directly at the **original SPZ stream** (byte-identical embedding; GLB size ≈ SPZ size), and `spzVersion` explicitly declares the version.

**5. WASM + CLI dual-end collaboration**

Both projects provide in-browser **WASM modules** — spz2glb for fast batch v3/v4 packaging with per-file JSON conversion reports; spz_gatekeeper for drag-and-drop validation with audit reports — and the reports are **consumable by each project's CLI**: spz_gatekeeper's browser report can be exported as `browser_to_cli_handoff` and re-checked by the CLI (`compat-check --handoff --json`, final verdict on the CLI side); spz2glb's web queue mirror produces the same JSON report structure as the CLI. Both projects are **WASM + CLI dual-end, sharing one JSON contract**.

**6. Benchmark dataset (in response to your compliance-vs-benchmark split, with clarifications)**

The dataset actually started as an attempt to address [spz issue #87](https://github.com/nianticlabs/spz/issues/87) (canonical samples per format version). It is an **experimental composition; we do not claim it is correct**:

- **Composition**: per scene, 9 files — `{raw/v3.spz, raw/v4.spz, ply, glb, assets/360°, prompt.txt, 3×JSON}`. JSON files use version-suffixed names: `{scene}_v3-info.json`, `{scene}_v4-info.json`, `{scene}_splats-info.json`.
- **GLB note**: the GLB files currently in the dataset were packaged by an **early version of spz2glb**, covering **v3 SPZ only**, and the extension fields may be incomplete. Since `KHR_gaussian_splatting_compression_spz_2` is still a **Draft**, once it advances to Release Candidate (RC) or ratification, we will re-align spz2glb against the then-current draft and re-package the GLB files accordingly.
- **Data source**: splat files generated by **Tencent's open-sourced World Model 2.0 (HY-World 2.0)** — model weights and code are fully open-sourced on [GitHub](https://github.com/Tencent-Hunyuan/HY-World-2.0) and [Hugging Face](https://huggingface.co/Tencent-Hunyuan/HY-World-2.0), so the community can reproduce locally; generation is also available through the Hunyuan3D online platform ([CN site](https://3d.hunyuan.tencent.com/) / [international site](https://3d.hunyuanglobal.com/)). Users may choose either path. The prompt→scene flow is easy to reproduce at low cost, which suits rapid multi-version conversion validation.
- **Positioning**: early work at v0.0.1a, offered as a reference template for the "canonical samples" discussion in #87.

**7. Community accessibility (i18n)**

One more note: the two WASM UIs and the dataset documentation are currently mostly in Chinese, which is not ideal for international collaboration. We plan to add multi-language (i18n) support across all three SPZ projects — spz_gatekeeper, spz2glb, and spz-anime-text2scene-bench — to make them more accessible to the wider community.

Thank you again for the discussion. We are happy to share more implementation details or test data if that would be useful.

javagl:
@gugu23456789 Thanks for the infos. The infos go beyond what is immediately relevant for the `spz_2` glTF extension, but maybe the approach for handling different SPZ versions can help to inform whether _this_ extension should remain fixed for 'v2', or whether it should be generalized to also handle other SPZ versions. 
(Again, given that SPZ v4 is an entirely new file format, one could make a strong case to _not_ try and cover both v2 and v4 with the same glTF extension, but all that is not my call).

(An aside: For a moment, I was confused by that last link, [`https://github.com/KhronosGroup/glTF/issues/87`](https://github.com/KhronosGroup/glTF/issues/87), but that was just the automatic conversion that GitHub did after entering `#87` with the intention to link to [`https://github.com/nianticlabs/spz/issues/87`](https://github.com/nianticlabs/spz/issues/87) 🙂 )

Regarding the "canonical examples": I agree that this is important. I've tried to cover this _to some extent_, with "JSplat". It contains an [examples project](https://github.com/javagl/JSplat/tree/master/jsplat-examples/src/main/java/de/javagl/jsplat/examples) with some related functions. For example, it has functions to create that "Unit Cube" (shown in the [main README](https://github.com/javagl/JSplat/blob/master/README.md)) purely programmatically, and then write it out in different formats ( GSPLAT, PLY_ASCII, PLY_BINARY_LE, PLY_BINARY_BE, SPZ, GLTF, GLTF_SPZ, SOG ). But it does **not** yet handle newer SPZ versions, and there has been a bit of back and forth while I've been tracking the ongoing specifications for SPZ, glTF, and _this_ extension - it may need an update when there are new decisions here, regarding the handling of the SPZ versions...

gugu23456789:
@javagl — thank you for the follow-up, and for the exploration and community contributions behind JSplat. As before, these are just notes from our own implementation experience; we may well be wrong on some points, and we welcome corrections.

**1. Design background: why registry-driven**

We noticed that the upstream Niantic library uses a `tryParseExtension` switch-case, hardcoding extension handling type by type. As SPZ versions (v2/v3/v4) and vendors (Adobe, Niantic, ...) keep producing new extensions,such a structure can become harder to maintain as the number of extensions grows over time. The gatekeeper's role is a neutral format-compliance validator (not a storage or writer), so we moved "extension management" into a governance layer that is independent of version parsing: a two-layer registry — one layer declares contracts, the other performs semantics.

**2. The two-layer registry**

- **Spec layer `ExtensionSpecRegistry`**: each extension registers an 11-field contract — `type / vendor_id / extension_id / vendor_name / extension_name / category / status / spec_url / short_description / min_spz_version / requires_has_extensions_flag` — registered per type, not bound to a specific SPZ version.
- **Validator layer `ExtensionValidatorRegistry`**: runtime payload validators keyed by type, auto-registered at static init via `RegisterValidator<T>`; the validator interface has only 3 virtual methods (Validate/GetName/GetType) — we only validate, never write, which differs from the upstream `SpzExtensionBase` with its 8 virtual methods.
- **Parsing**: ILV records are walked as `[u32 type][u32 length][value]`, and unknown types are skipped via `length` — structurally valid by construction, no false positives.

**3. `min_spz_version`: making "generalization" operable**

Version-dependent differences (gzip/zstd header parsing, quat layout) are handled by the core parser, so extension contracts are decoupled from versions — the registry itself expresses "which SPZ versions this extension applies to" through `min_spz_version`. Current registrations: 0xADBE0002 is `stable` without a version restriction; 0xADBE0003 is `draft` with `min_spz_version=4`. Should spz_2 ever be generalized, this kind of structure could directly host cross-version contract management.

**4. The 4-state matrix: an explicit boundary for semantic commitment**

Registration × validator = 4 combinations, each with its own issue code (A/B/C/D, no gaps):
- spec ✓ + validator ✓ → full semantic commitment (`L2_EXT_VALIDATION` only on validation failure)
- spec ✓ + validator ✗ → contract declared, validation pending (`L2_EXT_REGISTERED_NO_VALIDATOR`)
- spec ✗ + validator ✓ → validator exists but undeclared (inconsistent, `L2_EXT_UNREGISTERED_VALIDATOR`)
- spec ✗ + validator ✗ → structurally valid, no semantic promise (`L2_EXT_UNKNOWN`)

This is our engineering answer to "claiming support for different versions is difficult": no overclaiming (unregistered = no semantic promise), no blind spots (all four states covered), and observable status (queryable from both CLI and WASM).

**5. WASM capabilities**

The WASM module shares the same registry/report vocabulary as the CLI. Browser-side capabilities include:

- **Zero-copy inspection**: `inspectSpzPtr(ptr, size, strict)` reads SPZ bytes directly from WASM memory (avoiding JS↔WASM copies); `inspectSpz`/`inspectSpzText` are the regular entry points.
- **What the current checks cover**:
  - Format layer: magic, version validity (invalid / newer-than-known-max → error/warning respectively), SH degree range, reserved-zero, file sizes (too small / truncated / base payload)
  - Compression layer: gzip decompression failure; ZSTD decompression failure on the v4 path
  - v4-specific: 32-byte header parsing, ILV extension-zone parsing, TOC parsing
  - Extension layer: declared-but-no-data, ILV parsing (error in strict mode, downgraded to warning in non-strict), validator semantics, the three state codes of the 4-state matrix, undeclared trailer
  - strict/non-strict runs through all of the above
- **Governance exports**: `listRegisteredExtensions()` / `describeExtension(type)` / `getCompatibilityBoard()` — the registry contracts and the maturity board are browsable inside the browser.
- **JSON report contents**:
  - `inspectSpz` returns top-level `ok` (true when no errors) / `strict`, plus the full GateReport JSON (`spz_l2`: flags / trailer_size / ilv_records[{type,length,offset}]; `issues`: code / severity / message / location)
  - Both audit modes share a 12-field verdict schema: `audit_profile / policy_name / policy_version / policy_mode / audit_mode / verdict / final_verdict / release_ready / summary / budgets / issues / next_action` (budgets items carry per-item `status`: within_budget / over_budget)
  - Handoff: the browser can export `browser_to_cli_handoff`, which the CLI merges via `compat-check --handoff --json` — the final verdict always lives on the CLI side
- **Memory engineering** : device-aware SmartMemoryManager, a serial BatchQueue (max 2), 64KB chunked writes into a reserved buffer, and C++-side HotObjectPool/BumpAllocator

6. Canonical samples and our synthetic fixture

To be honest about the canonical samples: our current synthetic fixture is still quite early-stage. gen-fixture produces a minimal SPZ with a placeholder payload and a single ILV record — a minimal-sample tool aimed at extension-validator development and compliance smoke tests within the gatekeeper itself. It does not yet cover the diversity of real scenes.

We have also learned from projects like JSplat that programmatically generated scene samples (e.g., a unit-cube scene read and written across multiple formats) are closer to a "reproducible sample" target than a placeholder fixture. We are considering borrowing that idea to grow our synthetic samples over time. Building compliance test data sets is a long-term community effort, and we will keep following the relevant discussions (such as #2562 ).

These are our current practices and reflections. Thank you again for the discussion — we are happy to share more implementation details if that would be useful.

javagl:
Much of this information could probably be part of some "Design Philosopy" or "Project Structure" section in one of the 'Documentation'-README files of your projects (and maybe it already _is_ part of this documentation - maybe even just in a different language...). And as far as I understood this, much of this is about the handling of the 'extensions' that have been introduced in SPZ v4. (In addition to the overarching question about how to handle v2-vs-v4 to begin with, but it sounds like this boils down to the inspection of a few magic bytes). 

I think that SPZ was originally chosen to be the compression method for splats in glTF because (it was around and gained some attention and support at this time 🤷‍♂️ and) it is a relatively simple file format. 

But now I have to say: It **was** a relatively simple file format. In version v2, there wasn't much magic behind all this: Quantize a few floats, wrap them into ZIP, and you're done. 
(I wrote that [JSpz](https://github.com/javagl/JSpz) library in an afternoon - some of the unspecified parts have been 'ported' from the original implementation, but if they _had_ been specified, the implementation would likely have taken _less_ time, because I wouldn't have had to reverse engineer the existing code...)

From what I've seen about v4, and your description of the 'extension' mechanism, it sounds like this added a _considerable_ complexity. And I could imagine that these things could raise some pretty fundamental questions. For example (just _very_ roughly): When an SPZv4 extension defines a certain coordinate system, what does this mean for glTF? Is this extension "allowed" in glTF? Where do coordinate conversions take place then...?

Again: All these are subjective observations (and maybe I have a wrong perception of things). 

---

About the "canonical samples": There certainly are pros and cons for different approaches, depending on what _exactly_ should be tested. 

For example, when you want to compare the "compression ratio" of two file formats, then you can pick a real-world data set, compress it using formats A and B, and see that A achieves a compression of 60% and B achieves a compression of 70%. 

What is the insight here? Maybe B uses fewer quantization bits, and produces less accurate results (i.e. maybe the rendered output will have lower visual quality?) How to quantify the "lossiness" of the compression (beyond PSNR)? Maybe A has some command-line option for the compression strength, and you could set this parameter to achieve a compression of 80%? And maybe, if you had chosen a _different_ real-world data set, the results would be the exact opposite? 

Gaussian Splat compression is a complex topic, and an active research topic, FWIW. One can start at https://w-m.github.io/3dgs-compression-survey/ and probably spend _years_ with (and write a PhD thesis about) that. 

---

BTW: While some aspects may be relevant (e.g. how to handle different SPZ versions), much of this discussion is "off-topic" for **this** pull request. Before sharing more thoughts (and maybe distracting even further), I'll probably wait to see how your _original_ questions are supposed to be addressed.

gugu23456789:
@javagl — thank you for the follow-up, and for naming the fundamental question that sits exactly at the junction of our two tools: "what does an SPZ v4 extension (e.g. a coordinate system) mean for glTF, is it allowed, and where do conversions happen?" Let me share how we split that responsibility in our implementations.

spz_gatekeeper — the semantic-commitment side. As a neutral format-compliance engine, it separates "structurally valid" from "semantically compliant" through the two-layer registry ( ExtensionSpecRegistry + ExtensionValidatorRegistry ) and the 4-state matrix (spec × validator): a registered extension carries a semantic commitment; an unregistered one is only structurally valid with no promise. In short, it defines what an extension can be understood to mean .

spz2glb — the data-transport side. It follows a strict pass-through principle: it only reads extension metadata from the SPZ ILV records and packs it into the GLB extensions field as-is — no coordinate transforms, no semantic interpretation; when there is no ILV extension, the corresponding metadata fields are simply omitted. In short, it defines how extension data can be carried .

On your question of whether an extension is "allowed" in glTF: we treat permissibility as a spec-side decision (a working-group boundary call), and transportability + observability as the implementation's responsibility . The tools do not preempt the spec on permissibility — they faithfully carry whatever is declared and make its status queryable via the registry / reports.

From the start, the two tools were designed as complementary, interoperable components of a single end-to-end toolchain: spz_gatekeeper confirms "the input is compliant", spz2glb completes "the output is standard". We hope this engineering experience can serve as a humble reference anchor for the working group's discussion.

On engineering feasibility, our test data may be useful as reference:

- Test coverage : the toolchain is validated on two sources of SPZ files — high-fidelity .spz (v3) and .ply exported from Tencent HY-World (web-generated) , and .ply → .spz v4 produced by Niantic SPZ Converter . The two classes differ in geometric detail and encoding path, covering the main version branches of the current SPZ ecosystem.
- spz2glb packaging performance : for 25+ MB SPZ files from both sources, lossless SPZ→GLB packaging is stable at 200+ ms with a memory peak around 50+ MB (~2× input size) — suggesting the draft's lossless packaging approach is engineering-efficient, with no theory-to-practice gap.
- spz_gatekeeper compliance validation : its current mainline is WASM security hardening (performance optimization is a follow-up). Current compliance checks — header integrity, ILV extension parsing, ZSTD/gzip decompression paths — are stable in the 800+ ms range (observed values of the current WASM implementation); we expect them to shrink after the performance work.
- Draft feasibility : together the two tools show with measured data that the SPZ_2 draft's "compressed-stream pass-through into GLB" scheme is not only self-consistent in theory but also landable in engineering — the end-to-end flow (compliance validation + lossless packaging) completes in about 1 second , which is practically meaningful for browser-side real-time distribution.
We understand the current focus is the spz_2 extension itself; the above is engineering background for reference. We are happy to narrow the scope if needed, and we will adjust our implementation per the community consensus. Thank you again for the discussion.

PR #2533, opened 2025-10-06
Adding a copyright SPDX back to the mesh primitive restart extension

weegeekps:
Correcting this. Our intent is to release this extension openly for use by others.


weegeekps:
@lexaknyazev or @javagl, do either of you have any time to approve this?

weegeekps:
Thanks, @javagl. It's correct that this only addresses the one comment for this extension. We probably need to do some clean up and add SPDX identifiers to older extensions in another PR. However, I'm unsure what is correct for those extensions.

PR #2538, opened 2025-10-29
Add KHR_materials_fuzz

MiiBond:
This extension provides an OpenPBR-compatible fuzz/sheen extension.
The main differences are:
1. The fuzz sits on top of the coat (sheen sat below it)
2. The fuzz has a dedicated weighting factor, whereas sheen only had color. A consequence of this is that fuzz can be darker than the surface below it. This allows effects like black soot.

### DIFF
--- a/extensions/2.0/Khronos/KHR_materials_fuzz/README.md
+++ b/extensions/2.0/Khronos/KHR_materials_fuzz/README.md
@@ -1,0 +1,126 @@
+<!--
+SPDX-License-Identifier: LicenseRef-KhronosSpecCopyright
+-->
+
+# KHR_materials_fuzz
+
+## Contributors
+
+- Mike Bond, Adobe [@MiiBond](https://github.com/MiiBond)
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
+## Exclusions
+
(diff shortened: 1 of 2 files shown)


dfattal:
### `fuzzRoughnessFactor` default disagrees between the README and the schema

README:

> |**fuzzRoughnessFactor** | `number` | The fuzz roughness | No, default: `0.5` |

`schema/material.KHR_materials_fuzz.schema.json`:

```json
"fuzzRoughnessFactor": {
    "type": "number",
    "default": 0.0,
```

`fuzzFactor` (`0.0`) and `fuzzColorFactor` (`[1.0, 1.0, 1.0]`) agree between the two; this is the only one that doesn't.

We took the schema's `0.0` purely to be consistent with the machine-readable source, and we don't have a strong view on which is correct — `0.5` is arguably the more useful default for a fuzz lobe, since a perfectly smooth fuzz is an odd thing to ask for. Flagging it rather than proposing a value.

---
David Fattal — implementing for the [DisplayXR project](https://github.com/DisplayXR) ([model viewer](https://github.com/DisplayXR/displayxr-demo-modelviewer)). This extension is implemented there alongside `KHR_materials_coat` and `KHR_materials_diffuse_roughness`; happy to contribute conformance assets.

PR #2539, opened 2025-11-04
KHR_gaussian_splatting_wide_gamut_color

weegeekps:
As mentioned in the [base KHR_gaussian_splatting extension pull request](https://github.com/KhronosGroup/glTF/pull/2490#discussion_r2487892627), this extension adds support for several wide-gamut colors and their associated encoding methods to `KHR_gaussian_splatting`.

Some initial open questions:
 - We still need to decide on how we want to [handle the questions about which function to use at runtime for blending/composition with PQ and HLG](https://github.com/KhronosGroup/glTF/pull/2490#discussion_r2479229868).
 - What additional language around tone mapping do we need?

### DIFF
--- a/extensions/2.0/Khronos/KHR_gaussian_splatting_wide_gamut_color/README.md
+++ b/extensions/2.0/Khronos/KHR_gaussian_splatting_wide_gamut_color/README.md
@@ -1,0 +1,86 @@
+<!--
+Copyright 2015-2025 The Khronos Group Inc.
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# KHR_gaussian_splatting_wide_gamut_color
+
+## Contributors
+
+- Alexey Knyazev, Khronos
+- Adam Morris, Cesium
+- Zehui Lin, Huawei
+- Michael Nikelsky, Autodesk
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec. Extends KHR_gaussian_splatting.
+
+## Overview
+
+This extension adds wide-gamut color support to the KHR_gaussian_splatting extension. It introduces several additional wide-gamut color space options for the `colorSpace` property. These additional color spaces enable more accurate color representation for high dynamic range (HDR) content and wide-gamut displays when using Gaussian splatting in glTF assets.
+
+## Additional Color Spaces
+
+| Color Space | Description |
+| --- | --- |
+| BT.2020-ITU | BT.2020-ITU color space. |
+| BT.2020-linear | BT.2020 linear color space. |
+| BT.2100-PQ | BT.2100 PQ (Perceptual Quantizer) color space. |
+| BT.2100-HLG | BT.2100 HLG (Hybrid Log-Gamma) color space. |
+| Display-P3 | Display P3 color space. |
+
+## Extending the base KHR_gaussian_splatting extension
+
+The base `KHR_gaussian_splatting` extension is required for this extension to be used. The new `colorSpace` values can be used in place of the existing ones defined in the base extension.
+
+To extend the base extension, include both extensions in the glTF asset's `extensionsUsed` array:
+
+```json
+{
+  "extensionsUsed": [
+    "KHR_gaussian_splatting",
+    "KHR_gaussian_splatting_wide_gamut_color"
+  ]
+}
+```
+
+Then add the `KHR_gaussian_splatting_wide_gamut_color` extension to the `extensions` property of the `KHR_gaussian_splatting` object:
+
+```json
+{
+  "meshes": [
+    {
+      "primitives": [
+        {
+          // snip...
+          "extensions": {
+            "KHR_gaussian_splatting": {
+              "colorSpace": "Display-P3",
+              "extensions": {
+                "KHR_gaussian_splatting_wide_gamut_color": {}
+              }
+            }
+          }
+        }
+      ]
+    }
+  ]
+}
+```
+
+There are no additional properties introduced by this extension; it solely extends the `colorSpace` options of the base extension.
+
+## Known Implementations
+
+- TODO: List of known implementations, with links to each if available.
+
+## Resources
+
+- BT.2020: https://www.itu.int/rec/R-REC-BT.2020
+- BT.2100: https://www.itu.int/rec/r-rec-bt.2100
+- Display P3: https://www.color.org/chardata/rgb/DisplayP3.xalter


NorbertNopper-Huawei:
In general, I like splitting up the extension. However, to avoid future fragmentation and/or duplication, I recommend to define it more generic, as defining a color encoding is required not just for gaussian splatting.

extensions/2.0/Khronos/KHR_gaussian_splatting_wide_gamut_color/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec. Extends KHR_gaussian_splatting.
+
+## Overview
+
+This extension adds wide-gamut color support to the KHR_gaussian_splatting extension. It introduces several additional wide-gamut color space options for the `colorSpace` property. These additional color spaces enable more accurate color representation for high dynamic range (HDR) content and wide-gamut displays when using Gaussian splatting in glTF assets.
```
lexaknyazev, line 25: WCG does not imply HDR. Could we be more specific about the intended use cases here?

extensions/2.0/Khronos/KHR_gaussian_splatting_wide_gamut_color/README.md
```diff
@@ -0,0 +1,86 @@
+<!--
+Copyright 2015-2025 The Khronos Group Inc.
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# KHR_gaussian_splatting_wide_gamut_color
```
lexaknyazev, line 6: Saying "wide color gamut" (here and throughout) would be more standard than "wide-gamut color", which is often used as an adjective.

extensions/2.0/Khronos/KHR_gaussian_splatting_wide_gamut_color/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Additional Color Spaces
+
+| Color Space | Description |
+| --- | --- |
+| BT.2020-ITU | BT.2020-ITU color space. |
+| BT.2020-linear | BT.2020 linear color space. |
+| BT.2100-PQ | BT.2100 PQ (Perceptual Quantizer) color space. |
+| BT.2100-HLG | BT.2100 HLG (Hybrid Log-Gamma) color space. |
+| Display-P3 | Display P3 color space. |
```
lexaknyazev, line 35: glTF strings are case-sensitive. Could we align lower/upper case usage here? Not a very strong opinion but I'd prefer making these enums all-lowercase (or at least capitalizing Linear).

NorbertNopper-Huawei:
As this has been discussed.
There are folks, who want to use RAW image data.
E.g. this paper is describing it:
https://aryan-garg.github.io/hdrsplat/

extensions/2.0/Khronos/KHR_gaussian_splatting_wide_gamut_color/README.md
ZehuiLin-Huawei: To include HDR and WCG in this extension, naming as color_space should be better.

extensions/2.0/Khronos/KHR_gaussian_splatting_wide_gamut_color/README.md
ZehuiLin-Huawei, line 35: It would be clearer if we explicitly state what color primiaries and transfer characteristics are being used for each color space. For example, describing BT.2020-PQ as "BT.2020 color primaires and SMPTE ST2084 Perceptual Quantizer (PQ) transfer function".

PQ is defined for HDR, while BT.2020 defines SDR and WCG. BT.2100 extends BT.2020 to HDR, but using the same color primaires.

extensions/2.0/Khronos/KHR_gaussian_splatting_wide_gamut_color/README.md
ZehuiLin-Huawei, line 25: It would be helpful to add informative language somewhere explaining that the splats are intended for alpha blending in the specified linear or non-linear color space. This behavior differs from traditional meshes where people should first convert colors to linear color space before performing any other color operations.

extensions/2.0/Khronos/KHR_gaussian_splatting_wide_gamut_color/README.md
```diff
@@ -0,0 +1,86 @@
...
+    {
+      "primitives": [
+        {
+          // snip...
+          "extensions": {
+            "KHR_gaussian_splatting": {
+              "colorSpace": "Display-P3",
+              "extensions": {
+                "KHR_gaussian_splatting_wide_gamut_color": {}
+              }
```
ZehuiLin-Huawei, line 66: While gaussian splatting extension is in `primitives` level, I feel the `colorSpace` property should be put in `meshes` level, as multiple gaussian splatting "primitives" in a "mesh" probably share the same color space.

NorbertNopper-Huawei: It has to be on inside the primitive, as e.g. another primitive in the same mesh is e.g. common, triangle based glTF asset.

ZehuiLin-Huawei: Okay. If we are going to add color space for other kinds of primitives, then in `meshes` would be better.

doug-walker:
Hi, I work for Autodesk and am serving as the chief architect on the [OpenColorIO](https://opencolorio.org/) project, which is a color management system that is used in applications from Autodesk, Adobe, Foundry, Unreal, SideFX, V-Ray, and many more. We have a collaboration going with senior representatives from the OpenUSD and MaterialX projects to establish good color interop across the Academy Software Foundation (ASWF) projects and OpenUSD. I was wondering if we could interest you in adopting our color space naming?

We have published a recommendation for [scene-referred color spaces](https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/01_TextureAssetColorSpaces/TextureAssetColorSpaces.md#summary-table--overview-of-the-recommendations) used in computer graphics. And we have another recommendation for display-referred color spaces that is [in progress here](https://docs.google.com/document/d/1MmBG4a3Dr6S6EO781WjK-xZW7QdHpuo-zd7wtMvG1Rs/edit?usp=sharing). We are providing open source OpenColorIO implementations of all of the color spaces to try and make them very clearly defined.

The key item is a compact name that we are now calling the "[color interop ID](https://docs.google.com/document/d/1T94lYbis9uCskL_ZEMxGBF2JryLfZnjxlEoNgRHZzBE/edit?usp=sharing)", and that is the string I would propose you adopt. This has recently been added to OpenEXR as a standard attribute and there is work being done to implement that in OpenImageIO, Blender, and other software. 

The splats schema that is in progress for OpenUSD [will use the OpenUSD ColorSpaceAPI](https://github.com/PixarAnimationStudios/OpenUSD/pull/3716#discussion_r2208711490), which utilizes the interop IDs and these color space definitions.

Sharing common naming and precise color space definitions would help promote good interop between all of these projects.

NorbertNopper-Huawei:
> Hi, I work for Autodesk and am serving as the chief architect on the [OpenColorIO](https://opencolorio.org/) project, which is a color management system that is used in applications from Autodesk, Adobe, Foundry, Unreal, SideFX, V-Ray, and many more. We have a collaboration going with senior representatives from the OpenUSD and MaterialX projects to establish good color interop across the Academy Software Foundation (ASWF) projects and OpenUSD. I was wondering if we could interest you in adopting our color space naming?
> 
> We have published a recommendation for [scene-referred color spaces](https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/01_TextureAssetColorSpaces/TextureAssetColorSpaces.md#summary-table--overview-of-the-recommendations) used in computer graphics. And we have another recommendation for display-referred color spaces that is [in progress here](https://docs.google.com/document/d/1MmBG4a3Dr6S6EO781WjK-xZW7QdHpuo-zd7wtMvG1Rs/edit?usp=sharing). We are providing open source OpenColorIO implementations of all of the color spaces to try and make them very clearly defined.
> 
> The key item is a compact name that we are now calling the "[color interop ID](https://docs.google.com/document/d/1T94lYbis9uCskL_ZEMxGBF2JryLfZnjxlEoNgRHZzBE/edit?usp=sharing)", and that is the string I would propose you adopt. This has recently been added to OpenEXR as a standard attribute and there is work being done to implement that in OpenImageIO, Blender, and other software.
> 
> The splats schema that is in progress for OpenUSD [will use the OpenUSD ColorSpaceAPI](https://github.com/PixarAnimationStudios/OpenUSD/pull/3716#discussion_r2208711490), which utilizes the interop IDs and these color space definitions.
> 
> Sharing common naming and precise color space definitions would help promote good interop between all of these projects.

I can bring this up today in a meeting we have, especially for `Linear Rec.709 (sRGB)` and `sRGB Encoded Rec.709 (sRGB)` as it does also have an effect on the base extension: https://github.com/KhronosGroup/glTF/pull/2490

For curiosity, there are these image file formats:
https://en.wikipedia.org/wiki/AVIF and https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format
How would the compact name for e.g. Rec.2020 and PQ encoded look like? `pq_rec2020_scene`?

NorbertNopper-Huawei:
@doug-walker We consider to use the values from here also in the base extension:
https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/01_TextureAssetColorSpaces/TextureAssetColorSpaces.md#summary-table--overview-of-the-recommendations
Reviewed in a smaller group on Tuesday and then presented to 3D Formats on Wednesday.

doug-walker:
@NorbertNopper-Huawei, thank you for considering our proposal! 

> How would the compact name for e.g. Rec.2020 and PQ encoded look like? pq_rec2020_scene?

Yes, it could look like that, if it were included in one of the Color Interop Forum recommendations. Otherwise, it would need a namespace prefix to indicate it was an external definition. But I would not recommend adding a scene-referred version of PQ. As described in ISO 22028-5, it is, in theory, possible to use a scene-referred version of PQ (or HLG). However, in practice, the display-referred version is far more commonly used and a renderer would not want to apply additional tone-mapping to such a signal.

When defining the Color Interop Forum color spaces we included the ISO 22028-1 image state (i.e., scene-referred vs. display-referred) as part of the color space encoding definitions. This is an important detail for the renderer to know: If it's display-referred it has already been tone-mapped, whereas if it is scene-referred the tone-mapping is yet to be done. This is obviously something that will very significantly change the look of the final result, so it's important to communicate the intent of the file author.

It seems like most of the color spaces in this PR and the base schema are display-referred (tone-mapping has already been done), so the mapping to the Color Interop Forum ID strings would be as follows:

BT.709-sRGB: srgb_rec709_display
BT.2100-PQ: pq_rec2020_display
BT.2100-HLG: hlg_rec2020_display
Display-P3: srgb_p3d65_display (or srgbe_p3d65_display for the HDR version)

There is currently no CIF ID for "BT.2020-ITU" because Rec.2020 video never became a popular format.

For the two linear options, if the intent is that they are scene-referred, the CIF IDs would be:
BT.709-linear: lin_rec709_scene
BT.2020-linear: lin_rec2020_scene

If it would be helpful for me to attend one of your meetings to answer questions, please let me know.

NorbertNopper-Huawei:
Okay, thanks I got it.

NorbertNopper-Huawei:
I do not know, if finally the naming will be changed in the glTF extensions. At minimum, there should be the mapping possible. As soon as we proceed on this extension, it would make sense that you join the discussions.

NorbertNopper-Huawei:
One problem I do see is, that the list is not yet complete and all the `_display` ones are not listed yet.

NorbertNopper-Huawei:
> If it would be helpful for me to attend one of your meetings to answer questions, please let me know.

@weegeekps Could you please invite him for Tuesday, mainly to clarify the values for the colorSpace https://github.com/KhronosGroup/glTF/pull/2490#issuecomment-3618007931 

From my side, I was not considering the image states.

doug-walker:
As discussed in the meeting today, I followed up on my action-item and added a display-referred linear Rec.709 color space encoding to the [draft ASWF "Color Space Encodings for Displays" recommendation](https://docs.google.com/document/d/1MmBG4a3Dr6S6EO781WjK-xZW7QdHpuo-zd7wtMvG1Rs/edit?usp=sharing). The interop ID for this is "lin_rec709_display".  The initial feedback on the ASWF Slack has been positive, so I think we should be able to include that.

NorbertNopper-Huawei:
From a Vulkan perspective, it does make sense to match most of the given color spaces:
https://docs.vulkan.org/refpages/latest/refpages/source/VkColorSpaceKHR.html

ZehuiLin-Huawei:
> It seems like most of the color spaces in this PR and the base schema are display-referred (tone-mapping has already been done).

From my knowledge, HLG is primarily scene-referred.

doug-walker:
@ZehuiLin-Huawei, indeed some people do consider HLG to be scene-referred. However, that point of view depends on two requirements:
1. For video cameras to strictly implement the HLG OETF.
2. For the HLG OOTF to provide visually pleasing tone-mapping for scene-referred source material.

Unfortunately, neither one of those is true, in practice. Therefore, many people have found it more practical to consider an HLG signal to be display-referred, similar to Rec.2100-PQ. Using the display-referred approach, the HLG EOTF is used to interpret the signal as the colorimetry that should appear on a display. This is typically closer to the intent desired by the content creator than interpreting it as a scene-referred signal via the HLG OETF.

PR #2541, opened 2025-12-07
Changing the masking-shadowing function referenced in the 2.0 specifications

49iohcy:
Hello,

It has recently dawned on me that the GGX visibility term implemented in the glTF-Sample-Renderer as of now uses the *height-correlated* form of the Smith joint masking-shadowing function, instead of the alleged *separable* form in the spec.

I believe this would be a good revision if the intention is to maintain reasonable consistency between the sample implementation in the specification and the glTF-Sample-Renderer project.

Provided are links to the most up-to-date version of `glTF-Sample-Renderer/source/Renderer/shaders/brdf.glsl`, as well as the full expansions of both the *height-correlated* and *separable* forms from Heitz (2014).

Please let me know if you have any questions. Thank you.

* https://github.com/KhronosGroup/glTF-Sample-Renderer/blob/e6b052db89fb2adbaf31da4565a08265c96c2b9f/source/Renderer/shaders/brdf.glsl#L74
* https://observablehq.com/@n-a33/smith-joint-masking-shadowing-function

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -3076,7 +3076,7 @@ with
 
 [latexmath]
 ++++
-G = \frac{\chi^{+}(H \cdot L) \, \chi^{+}(H \cdot V)}{2 \, (\left| N \cdot V \right| \, \sqrt{\alpha^2 + (1 - \alpha^2) (N \cdot L)^2} + \left| N \cdot L \right| \, \sqrt{\alpha^2 + (1 - \alpha^2) (N \cdot V)^2})}
+\nu = \frac{\chi^{+}(H \cdot L) \, \chi^{+}(H \cdot V)}{2 \, (\left| N \cdot V \right| \, \sqrt{\alpha^2 + (1 - \alpha^2) (N \cdot L)^2} + \left| N \cdot L \right| \, \sqrt{\alpha^2 + (1 - \alpha^2) (N \cdot V)^2})}
 ++++
 
 Thus, we have the function


emackey:
Hi @49iohcy, thanks for submitting this!  We talked about this PR a little on today's PBR TSG call, but some of the members wanted more time to review it.  The group agrees that if the glTF Sample Viewer is doing a better job than section B.3 here, then a change like this is likely warranted, particularly if it brings the two into better alignment.

I notice that the height-correlated masking and shadowing function is also mentioned in section [B.3.6.1. Masking-Shadowing Term and Multiple Scattering](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#masking-shadowing-term-and-multiple-scattering).  Does this section need a tweak as well?

proog128:
In [KHR_materials_transmission](https://github.com/KhronosGroup/glTF/blob/7bbd90978cad06389eee3a36882c5ef2f2039faf/extensions/2.0/Khronos/KHR_materials_transmission/README.md#implementation-notes) there is another mention of the G term which should be changed as well.

I did a bit of digging to find out when and why this diverged, but didn't find anything useful. The glTF Sample Viewer implemented the separable function until April 2019, then changed it to height-correlated ([3359d5](https://github.com/KhronosGroup/glTF-Sample-Viewer/commit/3359d5eb83a9e5ac870c33a31e7d6f6a30f43c2b)). The rewrite of Appendix B started off with height-correlated in December 2019 ([d4319f](https://github.com/KhronosGroup/glTF/blob/d4319fae0c8d3d29d7a2b46bac81c9c722aa2e08/specification/2.0/README.md#metal-brdf)), but switched to separable in August 2020 ([f513c9](https://github.com/KhronosGroup/glTF/blob/f513c9edeb423fdf1bce8f2ce7be97e490981f6f/specification/2.0/README.md#specular-brdf)), without mentioning this specifically, neither in the Git commit nor in the pull request (#1717). Most likely an accident.

Anyways, looks good and thanks for bringing it up!

oss-haertl:
It makes sense to adjust this to reflect the current sample viewer implementation.
Filament also defines a [simplified version of the height-correlated form](https://google.github.io/filament/Filament.md.html#listing_approximatedspecularv). This could be added additionally as another alternative.

emackey:
I got the CI to run by pushing a copy of this branch to the main repo.  The section in question is [B.3.2. Specular BRDF](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#specular-brdf). (Also note, the currently-published version is not up-to-date with `main`, and fixes such as https://github.com/KhronosGroup/glTF/pull/2403 were merged but not yet published. The diffs shown here are comparing the top of `main` to what's in this PR, not what's currently published.)

## Diff 1

The first formula changes from this, the current version:

formula 1 current screenshot

To this new version:

formula 1 new screenshot

## Diff 2

The second formula changes from this, the current version:

formula 2 current screenshot

To this new version:

formula 2 current screenshot

(Also note, the above are screenshots of the rendered HTML. I also looked at the rendered PDF versions, and they look the same as their HTML counterparts, no rendering errors.)

Also, much thanks to @proog128 for digging through the history of this!  Do the newly-rendered formulas look OK?

proog128:
@emackey These formulas look good to me. It's mostly a matter of taste, but in the second formula you could consider moving the constant into its own fraction, so that there is no longer a long expression inside brackets and it might be harder to miss that the 2 scales not only the first part, but also the second.

image

PR #2548, opened 2026-01-04
Add MSFT_texture_xnb extension

dellis1972:
This extension adds the ability to specify textures using the XNA Binary format (XNB). An implementation of this extension can use the textures provided in the XNB files as an alternative to the PNG or JPG textures available in glTF 2.0.

This extension is used by users of XNA like frameworks such as MonoGame,FNA and KNI to load glTF files at runtime by including XNB textures.

The specification for the XNB format can be found at [https://docs.monogame.net/articles/getting_to_know/whatis/content_pipeline/XNB%20Format.docx](https://docs.monogame.net/articles/getting_to_know/whatis/content_pipeline/XNB%20Format.docx)

### DIFF
--- a/extensions/2.0/Vendor/MSFT_texture_xnb/README.md
+++ b/extensions/2.0/Vendor/MSFT_texture_xnb/README.md
@@ -14,7 +14,7 @@ Written against the glTF 2.0 spec.
 
 ## Overview
 
-This extension adds the ability to specify textures using the XNA Binary format (XNB). An implementation of this extension can use the textures provided in the XNB files as an alternative to the PNG or JPG textures available in glTF 2.0.
+This extension adds the ability to specify textures using the XNA Binary format (XNB). An implementation of this extension can use the textures provided in the XNB files as an alternative to the PNG or JPG textures available in glTF 2.0. The extension only supports Textures in XNB format, other types such as Audio, Video and 3D Models are not supported.
 
 The extension is added to the `textures` node and specifies a `source` property that points to the index of the `images` node which in turn points to the XNB texture file. A client that does not understand this extension can ignore the XNB file and continue to rely on the PNG or JPG textures specified.
 
--- a/extensions/2.0/Vendor/MSFT_texture_xnb/README.md
+++ b/extensions/2.0/Vendor/MSFT_texture_xnb/README.md
@@ -70,7 +70,11 @@ When used in the glTF Binary (.glb) format the `images` node that points to the 
 
 ## Known Implementations
 
-This extension is used by users of XNA like frameworks such as MonoGame,FNA and KNI to load glTF files at runtime by including XNB textures.
+This extension is used by users of XNA like frameworks such as MonoGame, FNA and KNI to load glTF files at runtime by including XNB textures.
+
+## Resources
+
+* **XNB Format Specification**: [XNB%20Format.docx](https://docs.monogame.net/articles/getting_to_know/whatis/content_pipeline/XNB%20Format.docx)
 
 ## License
 
--- a/extensions/2.0/Vendor/MSFT_texture_xnb/README.md
+++ b/extensions/2.0/Vendor/MSFT_texture_xnb/README.md
@@ -1,4 +1,4 @@
-# XNA_texture_xnb 
+# MSFT_texture_xnb
 
 ## Contributors
 
@@ -14,7 +14,7 @@ Written against the glTF 2.0 spec.
 
 ## Overview
 
-This extension adds the ability to specify textures using the XNa Binary format (XNB). An implementation of this extension can use the textures provided in the XNB files as an alternative to the PNG or JPG textures available in glTF 2.0.
+This extension adds the ability to specify textures using the XNA Binary format (XNB). An implementation of this extension can use the textures provided in the XNB files as an alternative to the PNG or JPG textures available in glTF 2.0.
 
 The extension is added to the `textures` node and specifies a `source` property that points to the index of the `images` node which in turn points to the XNB texture file. A client that does not understand this extension can ignore the XNB file and continue to rely on the PNG or JPG textures specified.
 
@@ -23,7 +23,7 @@ The extension is added to the `textures` node and specifies a `source` property 
     {
         "source": 0,
         "extensions": {
-            "XNA_texture_xnb": {
+            "MSFT_texture_xnb": {
                 "source": 1
             }
         }
@@ -46,7 +46,7 @@ When used in the glTF Binary (.glb) format the `images` node that points to the 
     {
         "source": 0,
         "extensions": {
-            "XNA_texture_xnb": {
+            "MSFT_texture_xnb": {
                 "source": 1
             }
         }
@@ -66,7 +66,7 @@ When used in the glTF Binary (.glb) format the `images` node that points to the 
 
 ## glTF Schema Updates
 
-* **JSON schema**: [glTF.XNA_texture_xnb.schema.json](schema/glTF.XNA_texture_xnb.schema.json)
+* **JSON schema**: [glTF.MSFT_texture_xnb.schema.json](schema/glTF.MSFT_texture_xnb.schema.json)
 
 ## Known Implementations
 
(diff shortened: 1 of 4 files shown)


lexaknyazev:
`XNA` is not a registered vendor prefix. IIRC, it's not an active MS trademark anymore.

If MonoGame can reasonably "own" it, please create an issue with a prefix registration request (see https://github.com/KhronosGroup/glTF/blob/main/extensions/Prefixes.md). In a case of any legal reasons preventing MonoGame from owning the XNA prefix, please register another prefix and rename the extension.

dellis1972:
@lexaknyazev 

I was going to make this a MSFT_texture_xnb but I wasn't sure if I could reasonably do that (even though I do work at MS). 
Does the Prefix "owner" have to own the file format?

lexaknyazev:
> Does the Prefix "owner" have to own the file format?

That's not our concern. The extension spec and the prefix should be owned by the same entity (unless it's `EXT`, which implies some level of multi-vendor interoperability).

@bghgary WDYT?

bghgary:
I'm not aware of any rules from Khronos or Microsoft against putting it as MSFT prefix, but I'm not sure it makes sense since XNA is not being advanced by Microsoft. @dellis1972 Is XNB owned by some entity? Is there a spec for it? I did a quick search and didn't see anything obvious. Feel free to ping me internally if you want to chat.

dellis1972:
> I'm not aware of any rules from Khronos or Microsoft against putting it as MSFT prefix, but I'm not sure it makes sense since XNA is not being advanced by Microsoft. @dellis1972 Is XNB owned by some entity? Is there a spec for it? I did a quick search and didn't see anything obvious. Feel free to ping me internally if you want to chat.

The Spec is licences to Microsoft.  This is a link to the doc on the format. Its a backup from the original MS doc.
https://github.com/SimonDarksideJ/XNAGameStudio/blob/archive/Samples/XNA_XNB_Format/XNB%20Format.docx

bghgary:
If it's a Microsoft spec, then MSFT prefix sounds fine to me. This glTF extension should be specific and point to this XNB spec. If this XNB spec ever advances, we will need a new glTF extension.

lexaknyazev:
I'd strongly suggest hosting that document in a more permanent place than someone's personal GitHub repo.

dellis1972:
> I'd strongly suggest hosting that document in a more permanent place than someone's personal GitHub repo.

The spec used to be hosted on MS servers at one point. I'll chat with the MonoGame maintainers , see if we can host it on our website.

extensions/2.0/Vendor/MSFT_texture_xnb/README.md
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+],
+"images": [
+    {
+        "mimeType": "image/png",
+        "bufferView": 1
+    },
+    {
+        "mimeType": "image/vnd-ms.xnb",
```
bghgary: It doesn't look like this mime type is registered with [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml)? Should it be?

bghgary: It also looks like the convention for this should be:
```suggestion
        "mimeType": "image/vnd.ms-xnb",
```

lexaknyazev: From RFC 6838:
> Vendor-tree registrations will be distinguished by the leading facet "vnd.".  That may be followed, at the discretion of the registrant, by either a media subtype name from a well-known producer (e.g., "vnd.mudpie") or by an IANA-approved designation of the producer's name that is followed by a media type or product designation (e.g., vnd.bigcompany.funnypictures).

If multiple XNA-related media types are expected, they should likely follow the `vnd.ms-xna.*` pattern; otherwise it's okay to use `vnd.ms-xnb` just for this one.

dellis1972: I based this of the .dds spec which uses `image/vnd-ms.dds`, which is also not registered.
I don't even know how to go about registering those mime types lol

lexaknyazev: Although DDS does not have a normative file format spec (it's more like a C header), [this page](https://learn.microsoft.com/en-us/windows/win32/wic/dds-format-overview) says that its media type is `image/vnd.ms-dds`.

It's quite typical to not officially register vendor-specific types.

bghgary: Yup, it looks like the dds mime type is wrongly formatted in [the spec](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/MSFT_texture_dds#overview), but it's probably not worth fixing anymore since it's already being used. Maybe we can accept both and fix the spec? Not sure it's worth it.

lexaknyazev: DDS legacy is not a concern for this PR; the XNB media type used here seems to be fixed.

dellis1972:
So I think this is good to go. We are just waiting on moving the docx for the XNA format to the monogame docs site, so I can link it in this PR.

extensions/2.0/Vendor/MSFT_texture_xnb/README.md
bghgary: Yes, of course not. We should consider creating another issue to discuss.

dellis1972:
Updated PR description with a link to the XNB format spec.

extensions/2.0/Vendor/MSFT_texture_xnb/README.md
```diff
@@ -0,0 +1,77 @@
...
+]
+```
+
+## glTF Schema Updates
+
+* **JSON schema**: [glTF.MSFT_texture_xnb.schema.json](schema/glTF.MSFT_texture_xnb.schema.json)
+
+## Known Implementations
+
+This extension is used by users of XNA like frameworks such as MonoGame,FNA and KNI to load glTF files at runtime by including XNB textures.
```
lexaknyazev: nit: missing space before "FNA"

extensions/2.0/Vendor/MSFT_texture_xnb/README.md
```diff
@@ -0,0 +1,77 @@
...
+        "mimeType": "image/vnd.ms-xnb",
+        "bufferView": 2
+    }
+]
+```
+
+## glTF Schema Updates
+
+* **JSON schema**: [glTF.MSFT_texture_xnb.schema.json](schema/glTF.MSFT_texture_xnb.schema.json)
+
```
lexaknyazev, line 70: Please add a `## Resources` section here with the link to the XNB format document.

lexaknyazev (commented):
Since XNB is a generic data serialization format, the _texture_ glTF extension should be be more precise about the expected structure of referenced XNB files.

extensions/2.0/Vendor/MSFT_texture_xnb/README.md
dellis1972: all done 😄

extensions/2.0/Vendor/MSFT_texture_xnb/README.md
```diff
@@ -0,0 +1,81 @@
...
+## Known Implementations
+
+This extension is used by users of XNA like frameworks such as MonoGame, FNA and KNI to load glTF files at runtime by including XNB textures.
+
+## Resources
+
+* **XNB Format Specification**: [XNB%20Format.docx](https://docs.monogame.net/articles/getting_to_know/whatis/content_pipeline/XNB%20Format.docx)
```
lexaknyazev: Please change `[XNB%20Format.docx]` to `[XNB Format.docx]` since it's rendered literally.

lexaknyazev:
Please address the comment above:
> Since XNB is a generic data serialization format, the _texture_ glTF extension should be be more precise about the expected structure of referenced XNB files.

dellis1972:
> Please address the comment above:
> 
> > Since XNB is a generic data serialization format, the _texture_ glTF extension should be be more precise about the expected structure of referenced XNB files.

@lexaknyazev  not sure if this is what you had in mind? https://github.com/KhronosGroup/glTF/pull/2548/changes#diff-059720a29aac280c4f662c58bf05d6ad686886bb75f57ca52eaed5d5c4bd6d4cR17

vpenades:
I went ahead and SharpGLTF 1.0.7 supports this extension. Assuming it will be merged eventually...

PR #2550, opened 2026-01-08
Adjust Object Model for morph targets

### LINKED ISSUE #2549: Clarify Object Model for morph target weights
From the Object Model spec:

> Upon loading an asset, an implementation registers specific glTF object properties for the Object Model by resolving JSON pointers identified by templates provided by this document to JSON properties of the asset being loaded. Undefined glTF properties that have **schema-default** values are considered defined with their default values.

The list of pointers includes the following entries:
Pointer |
-|
`/meshes/{}/weights`
`/meshes/{}/weights/{}`
`/nodes/{}/weights`
`/nodes/{}/weights/{}`

glTF 2.0 JSON schemas **do not define** default values for the `weights` properties but the spec has the following language:
> A mesh with morph targets **MAY** also define an optional `mesh.weights` property that stores the default targets' weights. These weights **MUST** be used when `node.weights` is undefined. When `mesh.weights` is undefined, the default targets' weights are zeros.

> When an instantiated mesh has morph targets, it **MUST** use morph weights specified with the `node.weights` property. When the latter is undefined, `mesh.weights` property **MUST** be used instead. When both of these fields are undefined, the mesh is instantiated in a non-morphed state (i.e., with all morph weights set to zeros).

Also from the `KHR_animation_pointer` spec:
> The JSON Pointer **MUST** point to a property defined in the asset. A property is considered defined if it is present in the asset explicitly or if it has a **default** value and its enclosing object is present.

---

There is a subtle difference between "default value" and "schema-default value" as the former _may_ imply that `weights` properties not defined in JSON have "runtime-default" values following the base spec language when a mesh has morph targets. On top of this, the base spec **does not** require the `weights` properties to be defined for animating morph weights using the `"weights"` animation channel target path.

The solution here could be to explicitly say that `weights` properties are defined in the Object Model if the associated mesh has morph targets even when not present in JSON.  

/cc @javagl @emackey @bghgary

lexaknyazev:
Fixes #2549.

/cc @aaronfranke @hybridherbst @robertdorn83

### DIFF
--- a/specification/2.0/ObjectModel.adoc
+++ b/specification/2.0/ObjectModel.adoc
@@ -119,7 +119,7 @@ This document defines a set of <<json-pointer,JSON pointers>> that glTF implemen
 
 1. The Object Model is defined only for valid glTF assets. Querying or setting properties of invalid glTF assets are undefined.
 
-2. Upon loading an asset, an implementation registers specific glTF object properties for the Object Model by resolving JSON pointers identified by templates provided by this document to JSON properties of the asset being loaded. Undefined glTF properties that have schema-default values are considered defined with their default values.
+2. Upon loading an asset, an implementation registers specific glTF object properties for the Object Model by resolving JSON pointers identified by templates provided by this document to JSON properties of the asset being loaded. Undefined glTF properties that have default values are considered defined with their default values.
 
 3. Each instance of empty curly braces (`{}`) in the pointer templates is replaced with the corresponding array element index for each glTF asset property matching the template.
 
@@ -185,8 +185,6 @@ The following pointer templates represent mutable properties defined in the core
 | `/materials/{}/pbrMetallicRoughness/baseColorFactor` | `float4`
 | `/materials/{}/pbrMetallicRoughness/metallicFactor`  | `float`
 | `/materials/{}/pbrMetallicRoughness/roughnessFactor` | `float`
-| `/meshes/{}/weights`                                 | `float[]`
-| `/meshes/{}/weights/{}`                              | `float`
 | `/nodes/{}/translation`                              | `float3`
 | `/nodes/{}/rotation`                                 | `float4`
 | `/nodes/{}/scale`                                    | `float3`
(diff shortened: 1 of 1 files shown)


lexaknyazev:
>  but are these default values expected to be _copied_, or are these intended to keep applying if the mesh weights changed somehow?

With this PR, the `mesh.weights` values are no longer accessible let alone mutable so they should not ever change. Their intended use cases are not very strong though:
- More explicit signalling of a mesh's default state, e.g., when all-zero weights are not the intended default look.
- Potential JSON deduplication if a mesh is used multiple times with the same weights and they are not animated.

bghgary:
I can't say I've look at the issues enough to understand the nuances, but we did have issues implementing this in Babylon.js that are still not resolved, so making this simpler is good. Do we have samples/examples that we can use to test this?

lexaknyazev:
> Do we have samples/examples that we can use to test this?

I don't think we ever had sample assets with `/meshes/{}/weights` pointers. That said we need to have the following test assets (they shouldn't block the PR though).

#### For `KHR_animation_pointer`
All these assets must behave the same when the animation is active:
- An asset that uses a `/nodes/{}/weights` animation pointer on a node with a mesh that has morph targets and does not have any static weights.
- An asset that uses a `/nodes/{}/weights` animation pointer on a node with a mesh that has morph targets and only `mesh.weights` static weights.
- An asset that uses a `/nodes/{}/weights` animation pointer on a node with a mesh that has morph targets and only `node.weights` static weights.
- An asset that uses a `/nodes/{}/weights` animation pointer on a node with a mesh that has morph targets and both  `mesh.weights` and  `node.weights` static weights.

#### For `KHR_interactivity`
_This list is not exhaustive but it's a good start._
1. A graph that reads `/nodes/{}/weights.length` from a node that has no mesh to confirm that the pointer is accepted and returns 0.
2. A graph that reads `/nodes/{}/weights.length` from a node that has a mesh with no morph targets to confirm that the pointer is accepted and returns 0.
3. A graph that reads `/nodes/{}/weights.length` from a node that has a mesh with morph targets and no static weights to confirm that the pointer is accepted and returns the number of morph targets.
4. A graph that reads `/nodes/{}/weights/{}` from a node that has no mesh to confirm that the pointer is rejected.
5. A graph that reads `/nodes/{}/weights/{}` from a node that has a mesh with no morph targets to confirm that the pointer is rejected.
6. A graph that reads `/nodes/{}/weights/{}` from a node that has a mesh with morph targets and no static weights to confirm that the pointer is accepted and returns all zeros.
7. A graph that reads `/nodes/{}/weights/{}` from a node that has a mesh with morph targets and static weights defined only in `mesh.weights` to confirm that the pointer is accepted and returns the weights specified.
8. A graph that reads `/nodes/{}/weights/{}` from a node that has a mesh with morph targets and static weights defined only in `node.weights` to confirm that the pointer is accepted and returns the weights specified.
9. A graph that reads `/nodes/{}/weights/{}` from a node that has a mesh with morph targets and static weights defined both in `mesh.weights` and `node.weights` to confirm that the pointer is accepted and returns the weights specified in `node.weights`.

bghgary:
> I don't think we ever had sample assets with /meshes/{}/weights pointers. That said we need to have the following test assets (they shouldn't block the PR though).

I think it's harder to review the changes without actually trying to implement the changes. We should have some samples to be confident with the implementation.

lexaknyazev:
This PR removes an under-defined and previously untested part of the Object Model scope.

bghgary:
That's fair.

bghgary (approved):

robertdorn83:
@lexaknyazev I have some interactivity tests created based on your list. 
Can be found in the interactivity test repo: [Weight Tests](https://github.com/KhronosGroup/glTF-Test-Assets-Interactivity/tree/main/Tests/Interactivity/pointer/get_set_morphtargets)

lexaknyazev:
@robertdorn83 I think the first test is incorrect: `/nodes/{}/weights.length` should be valid even without a mesh because the spec says:
> All pointers named `*.length` return zero if the corresponding array is not defined.

So `/nodes/{}/weights.length` may be invalid only if the node index does not exist.

robertdorn83:
> @robertdorn83 I think the first test is incorrect: `/nodes/{}/weights.length` should be valid even without a mesh because the spec says:
> 
> > All pointers named `*.length` return zero if the corresponding array is not defined.
> 
> So `/nodes/{}/weights.length` may be invalid only if the node index does not exist.

have changed it!

Issue #2547, opened 2025-12-31
Clarify restrictions for nodes with a defined matrix


lexaknyazev:
The base glTF 2.0 spec says:
> When a node is targeted for animation (referenced by an animation.channel.target), `matrix` **MUST NOT** be present.

This statement is misleading because morph weights of such nodes can be animated (although we probably don't have such sample assets).

/cc @javagl @emackey

lexaknyazev:
The Object Model (and therefore `KHR_animation_pointer`) does not explicitly address a case when TRS properties of a node with defined `matrix` property are targeted via pointer strings. This could be seen from two mutually exclusive viewpoints:
1. The animation pointer extension says that an animation channel that targets a TRS property directly is identical to an animation channel that targets the same TRS property via the pointer. Therefore, using pointers to target TRS properties of nodes that have `matrix` defined must be disallowed.
2. The Object Model spec says that TRS properties are mutable and always exist because they have schema-default values. Therefore, it's not relevant whether the node has `matrix` in JSON—engines would have to decompose it to TRS on load and thus pointers may always target them.

The first resolution would require engines to actively reject TRS pointers based on peculiar JSON structure details, which may be lost by the time the pointer is used. This is further complicated in case of interactivity pointers that may be composed at runtime.

The second resolution would always have an implementation-specific behavior because matrix decomposition is ambiguous by definition unless glTF spec decides to enforce a specific interpretation.

lexaknyazev:
The interactivity extension allows reading current property values, which creates another issue here. If a node has its `matrix` defined in JSON, what are the effective TRS values? Again, two mutually exclusive options:
1. They are undefined, i.e., such pointers are invalid (the interactivity extension can handle that gracefully).
2. They are implementation-defined, i.e., such pointers are valid but scale and rotation values may be severely different between engines and thus interactivity graphs cannot rely on them (and they do not even have a way to know). Again, there's an alternative option here to define exact decomposition in the spec.

lexaknyazev:
FWIW, this issue is not uncommon even in the sample assets repo. Consider `Models/Box`: its root node has the following matrix, that reorients the scene to follow glTF conventions:
```math
\begin{pmatrix}
  1 & 0 & 0 & 0 \\
  0 & 0 & 1 & 0\\
  0 & {-1} & 0 & 0\\
  0 & 0 & 0 & 1
\end{pmatrix}
```

This matrix (like any other TRS matrix) has four decompositions:

[...]

PR #2559, opened 2026-02-02
Clarify TRS properties in the Object Model

lexaknyazev:
/cc @aaronfranke @hybridherbst

### DIFF
--- a/specification/2.0/ObjectModel.adoc
+++ b/specification/2.0/ObjectModel.adoc
@@ -192,6 +192,16 @@ The following pointer templates represent mutable properties defined in the core
 | `/nodes/{}/weights/{}`                               | `float`
 |====
 
+The `/nodes/{}/translation`, `/nodes/{}/rotation`, and `/nodes/{}/scale` pointers represent the current TRS properties of the node. If the static `matrix` property is defined on the node object in JSON, the corresponding rotation and scale pointers are undefined.
+
+[NOTE]
+.Rationale
+====
+Since the glTF 2.0 specification allows negative scale factors, TRS matrix decomposition is ambiguous with regard to rotation and scale. Two implementations that choose different scale factor signs for the same matrix may produce two different rotation quaternions for it. Therefore, updating only the rotation or the scale may result in different node transformation matrices in different implementations.
+
+This restriction does not apply to the translation because the translation vector directly corresponds to the first three values of the last column of the transformation matrix regardless of the decomposition process.
+====
+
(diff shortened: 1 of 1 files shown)


javagl (commented):
Fixes https://github.com/KhronosGroup/glTF/issues/2547 ?

"Since **the** glTF 2.0 specification allows..."

It sounds like a reasonable change for me. 

I cannot be sure that I fully understand the implications for _consumers_ of such assets, and even less for _producers_. It may imply that, as a general rule, individual TRS properties should be preferred over a `matrix`. But to my understanding, that would perfectly make sense for anything that is related to interactivity, and have no drawbacks. 

(The difficulties for consumers to track and handle a mix of TRS and `matrix` already exist, independently of the object model)

Tag @emackey because you also had some thoughts in the issue.

lexaknyazev:
Implications for _consumers_ (i.e., engines or loaders):
- Animation pointer & Interactivity: No need to support the ambiguous case when rotation or scale are updated on a node that defines a static matrix.
- Interactivity only: Requirement to _reject_ graph access (both read & write) to rotation and scale properties of such nodes. We should make test assets to assert that.

Implications for _producers_ (i.e., exporters):
- Do not write static matrices for nodes used for rotations or scale animations via animation pointers (it's already the case for base glTF 2.0 animations).
- In case of interactivity, prefer to not export static matrices at all just to be safe.

aaronfranke (commented):
My opinion is that translation should always be allowed, and it's good to specify this.

lexaknyazev:
> My opinion is that translation should always be allowed

That's what the PR says.

aaronfranke (approved):

bghgary (approved):
lgtm

emackey:
It's unfortunate that the spec needs to reference what exists "in JSON" as opposed to some parameter setting or value.  But it's not clear how to get around this without creating more problems, so, I won't object to this.

hybridherbst:
One question for something that's not entirely clear to me, what happens when matrix and translation are both defined? Does translation "win" and override the translation of the matrix? Can I in that case have the matrix _and_ translation and then animate or make interactive just the translation property?

lexaknyazev:
> what happens when matrix and translation are both defined

That's disallowed by the base spec.

hybridherbst:
Thanks, then I'm not sure if the section is clear enough: https://github.com/KhronosGroup/glTF/pull/2559/changes#diff-8f228d937d6ff200a2972f8164aa5f5738bf0377eab6023a9eba4f2b106e6fe5. 

For me it sounds like "translation" can still be utilized with both interactivity and animation data, only rotation and scale are ruled out. But I might be misreading that.

lexaknyazev:
@hybridherbst

- Static JSON use of any TRS property with `matrix` is disallowed in the base spec. That remains unchanged.
- The Object Model now says that rotation and scale pointers are undefined if the static `matrix` property is defined.
- The statement about rotation and scale being undefined does not include translation, and the note below explicitly highlights why `matrix` presence does not affect translation validity.

Do you think more clarifications are needed?

PR #2560, opened 2026-02-03
Further clarify TRS properties in the Object Model

lexaknyazev:
Follow up to #2559 with more explicit language.

/cc @hybridherbst


aaronfranke (approved):

javagl (approved):
When I first read this, I added a comment: 

> ~Why does it not also mention `translation`? 

Then I read the 'Rationale' below, and deleted that comment.

So yes, mentioning `translation` here immediately _can_ avoid a short moment of confusion.

Issue #2562, opened 2026-02-14
Test data sets for `KHR_gaussian_splatting`


javagl:
This issue is intended for collecting ideas about possible test data sets that could be helpful for the upcoming ratification process of the [`KHR_gaussian_splatting`](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_gaussian_splatting) extension.

**Note:** The goal of this issue is **NOT** to collect links to data sets where someone used some Python script to convert some photos into PLY data and somehow convert this to glTF. The primary goal here is to have test data sets that are useful for checking details of the behavior of a renderer, regarding the compliance with the specification, and _specifically_ the behavior of a renderer for splats in glTF. 
(Yes, the test data sets could eventually include a few "real-world" data sets. But these should be carefully selected and curated. Maybe they could contain some water or metal surfaces to show the effects of spherical harmonics. Or to put it that way: The real-world data sets that are used for testing should look "nice" - in line with the idea of the "Showcase" assets in the glTF-Sample-Assets repository)

### Rotation tests

This is important. The coordinate systems of PLY files - oh dear. The following screenshots show some possible test cases. They all contain "unit cube corners" that are important for illustrating the overall _orientation_ of the data. Beyond that, these tests are

- for rotations around x, y, and z
- each contains 10 rotated splats
- the position of the splats is interpolated between -50 and +50 along the respective axis
- the color is interpolated from "white" to the color for the respective axis
- the scaling factor will be z=3 for the x-axis, x=3 for the y-axis, and y=3 for the z-axis
- the rotation angle will be interpolated from 0 to 90 degrees around the respective axis

<img width="304" height="286" alt="Image" src="https://github.com/user-attachments/assets/54139df0-acaa-4904-a9a6-19bc2953be50" />

<img width="278" height="248" alt="Image" src="https://github.com/user-attachments/assets/fce9d3a6-2b76-403a-9d6e-243c680546d2" />

<img width="302" height="260" alt="Image" src="https://github.com/user-attachments/assets/a33d6d64-ee3d-45d3-a58e-d582e9dfe3d7" />

### Mixing meshes and splats

At the time of writing this, I think that there is no renderer that can read and render glTF files that contain meshes and splats. However, the test cases for the basic functionality could be trivial: An example primitive of a "unit cube splat", and a unit cube mesh. In one case, the mesh is scaled to be half the size of the splats. In the other case, it is scaled to be twice the size of the splats.

[...]

PR #2567, opened 2026-04-08
KHR_gaussian_splatting: Editorial review

lexaknyazev:


extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,234 @@ Release Candidate
...
 ```
 
 Where:
+
 - $C$ is the final color of the pixel.
 - $\mathscr{N}$ is the set of splats that contribute to the pixel.
-- $c_i$ is the color of the $i$-th Gaussian. See [Lighting](#lighting) for details on how to compute this color from the spherical harmonics.
-- $\alpha_i = \alpha \cdot G(x)$ where $G(x)$ is the value of the projected 2D Gaussian's probability density function evaluated at the pixel center $x$.
+- $c_i$ is the color of the $i$-th Gaussian.
+- $\alpha_i = \alpha \cdot G(x)$ where $G(x)$ is the value of the projected 2D Gaussian's probability density function evaluated at the pixel center $x$ and \alpha is the splat's opacity attribute value.
```
javagl: Formatting of `\alpha`

javagl: (I'll try to allocate some time for a real review soon - this is just something that I saw during the call)

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
+> [!NOTE]
+> This is the shape used by the original [3D Gaussian Splatting for Real-Time Radiance Field Rendering](https://repo-sam.inria.fr/fungraph/3d-gaussian-splatting/) paper.
+
+### Splat Reconstruction
+
+Each Gaussian splat represents a 3D Gaussian defined by the following equation:
 
 ```math
-G(x) = \exp\left(-\frac{1}{2}(x-\mu)^T \Sigma^{-1} (x-\mu)\right)
+G(x) = \exp\left(-\frac{1}{2}(p-\mu)^T \Sigma^{-1} (x-\mu)\right)
```
weegeekps:
```suggestion
G(x) = \exp\left(-\frac{1}{2}(x-\mu)^T \Sigma^{-1} (x-\mu)\right)
```

Was this $x$ changed to a $p$ in error?

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
-\mathbf{S} = \begin{pmatrix}
-s_x & 0 & 0 \\
-0 & s_y & 0 \\
-0 & 0 & s_z
+\mathbf{C} = \begin{pmatrix}
+s_x (1 - 2(q_y^2 + q_z^2)) & s_y (2(q_x q_y - q_w q_z)) & s_z (2(q_x q_z + q_w q_y)) \\
+s_x (2(q_x q_y + q_w q_z)) & s_y (1 - 2(q_x^2 + q_z^2)) & s_z (2(q_y q_z - q_w q_x)) \\
+s_x (2(q_x q_z - q_w q_y)) & s_y (2(q_y q_z + q_w q_x)) & s_z (1 - 2(q_x^2 + q_y^2))
 \end{pmatrix}
 ```
```
weegeekps, line 203: This is a much more natural way of thinking about this in the context of glTF, but do we need to explain how we got here since this is different than what the Inria paper provides?

lexaknyazev: I think the only explicit difference here from the previous text is the inclusion of the node's transform matrix and showing how it is used to transform the splat's covariance matrix.

That's a generic operation not specific to that paper. One could imagine that the initial covariance matrix is identity, which is transformed by the splat's scale and rotation, and later with the node's matrix.

Conversions of the scale vector and rotation quaternion to scale and rotation matrices and taking their product have to already exist in any glTF viewer implementation.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
-  "extensionsUsed" : [
-    "KHR_gaussian_splatting"
-  ]
-```
+The extension object contains the following properties:
 
-Other extensions that depend on this extension such as 3D Gaussian splatting compression extensions may require that this extension be included in `extensionsRequired`.
+|                   | Type     | Description                                        | Required                       |
+| ----------------- | -------- | -------------------------------------------------- | ------------------------------ |
+| **kernel**.       | `string` | The kernel used to generate the Gaussians.         | :white_check_mark: Yes         |
```
javagl: Dot behind 'kernel'

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
+
+This extension defines only one kernel type called `"ellipse"`, that is a 2D ellipse kernel used to project an ellipsoid shape in 3D space. Additional kernel types can be added over time by supplying an extension that defines an alternative definition. See the section [Ellipse Kernel](#ellipse-kernel) below for the exact definitions.
+
+### Color Space
+
+The `colorSpace` property is a required property that specifies the color space of the reconstructed 3D Gaussian Splat color values. The color space is typically determined by the training process for the splats. This color space value only applies to the 3D Gaussian splatting data and does not affect any other color data in the glTF.
+
+Unless specified otherwise by additional extensions, this color space information refers to the reconstructed splat color values, therefore all intermediate rendering steps (including but not limited to alpha blending) must be performed on before any color gamut or transfer function conversions.
+
+Additional values can be added over time by defining extensions to add new color spaces. See the section, [Extending the Base Extension](#extending-the-base-extension), for more information.
```
javagl: I _think_ there shouldn't be ',' commas before and after the link

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
+
+#### Available Color Spaces
+
+| Color Space           | Description                                   |
+| --------------------- | --------------------------------------------- |
+| `srgb_rec709_display` | BT.709 sRGB (display-referred) color space.   |
+| `lin_rec709_display`  | BT.709 linear (display-referred) color space. |
+
+> [!NOTE]
+> These string values follow the color space identification pattern recommended by the ASWF Color Interoperability Forum (CIF). See: [ASWF Color Space Encodings for Displays](https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/02_DisplayColorSpaces/DisplayColorSpaces.md).
```
javagl, line 78: Maybe this should be a permalink. They don't seem to have tags or versions. In doubt, it may (have to?) be the commit hash...?

(Also in the 'Resources' at the bottom)

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
+
+### Projection
+
+The `projection` property is an optional property that specifies how the Gaussians should be projected onto the kernel shape. This is typically provided by the training process for the splats. This property is meant to be extended in the future as new projections become standardized within the community.
+
+This base extension defines a single projection method, `"perspective"`, which is the default value.
+
+Additional values can be added over time by defining extensions to add new projection methods. See the section, [Extending the Base Extension](#extending-the-base-extension), for more information.
```
javagl: Maybe remove the commas before and after the link?

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
+> [!NOTE]
+> If this extension is not supported, implementations are expected to follow the base glTF specification and thus render the splat primitive as a point cloud. In such a case, the referenced glTF material will be used.
 
-To render a field of 3D Gaussian splats, the renderer must reconstruct each Gaussian splat using the same forward pass algorithm used during training. This involves using the position, scale, rotation, opacity, and spherical harmonics attributes to reconstruct the Gaussian splat in 3D space.
+The global transformation matrix of the node containing a mesh with splat primitives SHOULD conform to all of the following conditions:
 
-There are three key stages to reconstructing and rendering a 3D Gaussian splat:
+- The last row is $(0, 0, 0, 1)$ (implicitly enforced by the base glTF specification).
+- The lengths of the first three columns are finite positive numbers.
+- The determinant of upper-left 3x3 matrix formed be dividing each matrix element by the length of its column is close to positive one.
```
javagl: Maybe "...determinant of _the_ upper-left 3x3 matrix.."
I wondered if there is an easier way to describe the constraints, but it should be OK, particularly with the clarification of the _intention_ behind these constraints that is given below

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
-### 3D Gaussian Representation
+### Attributes
 
-Each Gaussian splat using the default `ellipse` kernel represents a 3D Gaussian defined by the following equation:
+This kernel defines the following attributes for the mesh primitive.
+
+| Splat Data                   | Attribute Semantic                | Accessor Type | Component Type | Required |
+| ---------------------------- | --------------------------------- | - | - | - |
+| Position                     | `POSITION`                        | VEC3   | Inherited from the glTF specification | :white_check_mark: Yes |
+| Rotation                     | `KHR_gaussian_splatting:ROTATION` | VEC4   | _float_ <br/>_signed byte_ normalized <br/>_signed short_ normalized | :white_check_mark: Yes |
```
javagl, line 126: We talked about which attributes should be 'required'. For the SH0/0 one, this was now changed to omit the

> Yes _(unless using a different method for lighting)_

part. And I thought that we wanted to basically make _all_ properties non-required, or at least add a hint that future extensions may omit (some of) these attributes _if_ they encode the required information in a different way. 

For example, https://github.com/KhronosGroup/glTF/pull/2531 is currently omitting the 'OPACITY' (encoding it in the 'COLOR_0' alpha). How can this be made consistent with the base specification?

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
+
+The opacity of a Gaussian splat is defined by the `KHR_gaussian_splatting:OPACITY` attribute. It stores a normalized linear value between _0.0_ (transparent) and _1.0_ (opaque). Out-of-range values are invalid.
+
+> [!NOTE]
+> A sigmoid activation function applied during training ensures the value stays within the valid range. This guarantees that renderers can use the stored opacity directly for alpha blending without any extra processing.
+
+#### Spherical Harmonics Attributes
+
+When spherical harmonics are being used for lighting, the coefficients for the diffuse component MUST be provided using the `KHR_gaussian_splatting:SH_DEGREE_0_COEF_0` attribute semantic. The zeroth-order spherical harmonic coefficients are always required. The `KHR_gaussian_splatting:SH_DEGREE_l_COEF_n` attributes where `l` is greater than zero hold the higher degrees of spherical harmonics data. To use higher degrees of spherical harmonics the lower degrees MUST be defined.
```
javagl: Why "**When**... (they are used)"? The property is required. 

Maybe...

> The `KHR_gaussian_splatting:SH_DEGREE_0_COEF_0` attribute semantic provides the diffuse component coefficients for the spherical harmonics

or so...

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
+Each Gaussian splat represents a 3D Gaussian defined by the following equation:
 
 ```math
 G(x) = \exp\left(-\frac{1}{2}(x-\mu)^T \Sigma^{-1} (x-\mu)\right)
 ```
 
 Where:
-- $G(x)$ is the value of the Gaussian at position $x$.
+
+- $G(x)$ is the value of the Gaussian at position $p$.
```
javagl: Should that be 'position x' ?

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
 
-```math
-\mathbf{J} = \begin{pmatrix}
-\frac{f_x}{z} & 0 & -\frac{f_x x}{z^2} \\
-0 & \frac{f_y}{z} & -\frac{f_y y}{z^2}
-\end{pmatrix}
-```
+> [!NOTE]
+> Since the construction details of the view and perspective projection matrices are implementation-specific, the formal definitions of $\mathbf{W}$ and $\mathbf{J}$ are not provided in this specification.
+
```
javagl, line 220: It's difficult. Maybe some pointer could be added like

> A common way of defining them is [small math block or link to paper] 

?

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
 
 ## Lighting
 
-At the time of writing, the most common method for lighting 3D Gaussian splats is via the real spherical harmonics. This extension defines attributes to store spherical harmonic coefficients for each splat. The zeroth-order spherical harmonic coefficients are always required. Higher order coefficients are optional. Each color channel has a separate coefficient, so for each degree $ℓ$, there are $(2ℓ + 1)$ coefficients, each containing RGB values.
+This extension defines 3D Gaussian splat lighting via the real spherical harmonics. The corresponding attributes store spherical harmonic coefficients for each splat. The zeroth-order spherical harmonic coefficients are always required. Higher order coefficients are optional. Each color channel has a separate coefficient, so for each degree $l$, there are $(2l + 1)$ coefficients, each containing RGB values.
```
javagl: Calling them "RGB" values _might_ be confusing. Maybe something like ~"values contributing to the red, green, and blue color components" or so...?

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -32,157 +32,236 @@ Release Candidate
...
 
-The default `ellipse` kernel based on the original 3D Gaussian splatting paper typically uses the _BT.709-sRGB_ color space with a _display-referred_ image state for training and rendering. This is different than the typical glTF PBR material model, where scene-referred linear color spaces are used. This extension defines two display-referred color spaces but scene-referred color spaces may be added by extensions. See: [Available Color Spaces](#available-color-spaces)
+The default ellipse kernel based on the original 3D Gaussian splatting paper typically uses the _BT.709 sRGB_ color space with a _display-referred_ image state for training and rendering. This is different than the typical glTF PBR material model, where scene-referred linear color spaces are used. This extension defines two display-referred color spaces but scene-referred color spaces may be added by additional extensions. See: [Available Color Spaces](#available-color-spaces).
 
 Implementations are allowed to relight the splats than the one they were trained in, but this may lead to visual differences compared to the original training results.
```
javagl: There seems to be something missing in that sentence - something like

"...relight the splats _to use a different color space_ than the one..."

or so...?

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -252,190 +331,75 @@ Color_{final} =\,&Color_{SH_{0}} + Color_{SH_{1}} + Color_{SH_{2}} + Color_{SH_{
...
+Compression extensions that operate on 3D Gaussian splatting data should extend this base extension to ensure compatibility. Compression extensions must define how the data can be decoded back into the base 3D Gaussian splatting format defined by this extension, but may also allow optimizations specific to their compression method, e.g., passing textures or other data directly to the GPU for decoding.
 
-To use an extensions that extends `KHR_gaussian_splatting`, the extension must be included within the `extensions` property of the `KHR_gaussian_splatting` extension object. The extension must also be listed in `extensionsUsed` at the top level of the glTF.
+To use extensions that extends `KHR_gaussian_splatting`, the extension must be included within the `extensions` property of the `KHR_gaussian_splatting` extension object. The extension must also be listed in `extensionsUsed` at the top level of the glTF.
```
javagl: This was changed from

> To use an extensions that extends `KHR_gaussian_splatting`, the extension must be included ...

but should be

> To use an extension that extends `KHR_gaussian_splatting`, the extension must be included 

(keep the 'an', but remove the 's)

To be perfectly nitpicky, the second part could be extended to

"...the **new** extension must **also** be included..."

(to make clear that this is not about the base extension)

javagl (commented):
I did not verify the math. 
The inlined comments are mainly about typos or wording details. 
Some of the comments may be subjective, and may not have to be addressed.

The most "important" one is probably about which attribute semantics are 'Required'. **Iff** we agreed on making them all required, then that's OK. But I think that it could make sense to point out that they are **only** required for the specific case (ellipse kernel) that is covered here.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
javagl: Formed "**by**" dividing, btw...

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
```diff
@@ -202,17 +281,17 @@ The diffuse color of the splat can be computed by multiplying the RGB coefficien
...
+To keep the spherical harmonics within the [0, 1] range, the forward pass of the training process applies a _0.5_ bias to the zeroth-order component of the spherical harmonics. The rendering process must also apply this bias when reconstructing the color values from the spherical harmonics. This allows the training to occur around 0, ensuring numeric stability for the spherical harmonics, but also allows the coefficients to remain within a valid range for easy rendering.
 
-Ergo, to calculate the diffuse RGB color from the only the DC component, the formula is:
+Ergo, to calculate the diffuse RGB color from only the zeroth-order component, the formula is:
 
 ```math
 Color_{diffuse} = SH_{0,0} * 0.2820947917738781 + 0.5
 ```
 
 Where $SH_{0,0}$ represents the a vector of the RGB coefficients for the zeroth-order real spherical harmonic.
```
javagl: "...represents the a vector..."

Only "the" or "a".

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
javagl, line 126: If quaternions are encoded with 'byte', should they be re-normalized after encoding, or should the (probably non-normal) decoded value be used directly?

lexaknyazev: There's currently no clear language on "using a different method for lighting" in a case the kernel definition remains the same (if it's different, the whole table of attributes does not apply).

Following glTF's backward compatibility practices, the `ellipse` attributes should be kept (even if without data) to not violate implementations supporting only the base extension.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
lexaknyazev: Jacobian matrix here is just another way of saying "taking partial derivates". The intention is that implementors actually do that themselves based on their matrices instead of copy-pasting something that won't work correctly, e.g., because of a mismatching forward direction.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
lexaknyazev: This is a pre-existing text likely about relighting splats to fit a new environment.

@weegeekps do you remember the intention here?

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
lexaknyazev: The base glTF spec does not require renormalization for rotation animations but it does implicitly say that the _stored_ quaternions are unit. In practice, this heavily reduces the number of quaternions representable with such encodings.

NorbertNopper-Huawei:
Regarding the clamping discussion:
`srgb_rec709_display` suggesting to `clamp(0, 1)`

`lin_rec709_display` currently doing some research ...

lexaknyazev:
> Regarding the clamping discussion: `srgb_rec709_display` suggesting to `clamp(0, 1)`
> 
> `lin_rec709_display` currently doing some research ...

Clamping should not depend on the encoding differences between these two options because both of them are defined only for the $[0, 1]$ range.

Please see the current language at the end of the "Splat Lighting" section.

NorbertNopper-Huawei:
> > Regarding the clamping discussion: `srgb_rec709_display` suggesting to `clamp(0, 1)`
> > `lin_rec709_display` currently doing some research ...
> 
> Clamping should not depend on the encoding differences between these two options because both of them are defined only for the [ 0 , 1 ] range.
> 
> Please see the current language at the end of the "Splat Lighting" section.

The point is, that display referred is an image working space and needs a conversion to the final display format anyway. In common case, usually srgb to srgb, we can clamp if really required. If srgb to another display format, there also a conversion is required.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
javagl: OK, then the line

> **This kernel** defines the following attributes for the mesh primitive.

above the table should be enough here.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
javagl: No strong opinion. The _verbal_ description in the bullet points above may be enough here.

NorbertNopper-Huawei:
Furthermore, as we do have display referred data, we are working in the display image space:
https://www.pixelsham.com/2022/12/06/scene-referred-vs-display-referred-color-workflows/?utm_source=copilot.com

For scene referred and lighting, this will come in another extension. So, you can remove the lighting in this section, as this needs to be solved differently.

NorbertNopper-Huawei:
If one is really nitpicking, the display referred data is not allowed to be used in scene referred space.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
weegeekps: The main intent behind the "color space" is to specify the color space the 3D Gaussian splats were trained in. We can't really prevent implementers from relighting them. Relighting, either into the same or a different color space, may lead to visual differences compared to the original training results and it seems like something we should call out. 

This would probably be better as a non-normative note.

NorbertNopper-Huawei:
```mermaid
flowchart TD
  A[lin_rec709_display] <--> B[encoding and decoding]
  B <--> C[srgb_rec709_display]
  C --> D[SRGB Display]

**Comment by lexaknyazev** [comment]
> So, you can remove the lighting in this section, as this needs to be solved differently.

It's not a real scene lighting; "Splat Lighting" there means computing the final color from SH values. This term was used in the previous spec draft so it's still there; feel free to propose a better option.

The whole clamping issue addresses a completely separate topic. Imagine a splat containing only SH0 values (for brevity) and the SH0 values are like `(10.0, 20.0, 30.0)`. After applying the math, the computed diffuse color will be `(3.32, 6.14, 8.96)`. The spec simply says that if the splat's color space is not defined for such values, they must be clamped, in this case to `(1.0, 1.0, 1.0)`.

**Comment by NorbertNopper-Huawei** [comment]
> > So, you can remove the lighting in this section, as this needs to be solved differently.
> 
> It's not a real scene lighting; "Splat Lighting" there means computing the final color from SH values. This term was used in the previous spec draft so it's still there; feel free to propose a better option.
> 
> The whole clamping issue addresses a completely separate topic. Imagine a splat containing only SH0 values (for brevity) and the SH0 values are like `(10.0, 20.0, 30.0)`. After applying the math, the computed diffuse color will be `(3.32, 6.14, 8.96)`. The spec simply says that if the splat's color space is not defined for such values, they must be clamped, in this case to `(1.0, 1.0, 1.0)`.

This can not happen, as the values were trained in a given range.

**Comment by NorbertNopper-Huawei** [comment]
Can not happen to srgb, and if trained for linear, please look at the image above

**Comment by lexaknyazev** [comment]
> This can not happen, as the values were trained in a given range.

Training process is irrelevant. The spec describes what to do with input data and there are only three possible options:
1. Say that out-of-range values are invalid and MUST be rejected (very expensive to implement).
2. Say that out-of-range values produce undefined rendering (will lead to portability issues).
3. Say that out-of-range values MUST be clamped (the safest option).

**Comment by NorbertNopper-Huawei** [comment]
> > This can not happen, as the values were trained in a given range.
> 
> Training process is irrelevant. The spec describes what to do with input data and there are only three possible options:
> 
> 1. Say that out-of-range data is invalid and MUST be rejected (very expensive to implement).
> 2. Say that out-of-range data produces undefined rendering (will lead to portability issues).
> 3. Say that out-of-range data MUST be clamped (the safest option).

Training process is relevant, as we are reproducing input images.

**Comment by NorbertNopper-Huawei** [comment]
We are doing a glTF extension for 3D Gaussian Splatting, so training is relevant.

**Comment by javagl** [comment]
None of the test data sets in https://github.com/KhronosGroup/glTF/issues/2562 was created with a training process. The spec is supposed to say how input data is handled, _regardless_ where and how it has been created.

**Comment by NorbertNopper-Huawei** [comment]
> None of the test data sets in #2562 was created with a training process. The spec is supposed to say how input data is handled, _regardless_ where and how it has been created.

Then ignore these test assets except the real world data ones, also shared there.

**Comment by NorbertNopper-Huawei** [comment]
We are doing a 3D Gaussian Splatting glTF extension related to the paper and not splat everything.

**Comment by NorbertNopper-Huawei** [comment]
So, if one is creating 3D Gaussian Splatting data manually, one has to take the training algorithm into account.

**Comment by javagl** [comment]
I'll just point to https://github.com/KhronosGroup/glTF/pull/2490#issuecomment-3630446891 , so that you can discuss this with the person who wrote that comment, and assume that you're OK with clamping the resulting values.

**Comment by NorbertNopper-Huawei** [comment]
Maybe I was wrong at that point of time and learned a little bit.

**Comment by javagl** [comment]
Or... the opposite? 😁 The point is: There is data out there that does contains values where the color will be out of the [0,1] range. The specification has to say what to do in this case. And I think that clamping to [0,1] is the most sensible option.

**Comment by javagl** [comment]
I looked at the "Cactus" example. These are the ranges of the SH0/0 values:
```
minX -2.2483134
minY -2.3832924
minZ -2.496656
maxX 7.762948
maxY 5.9173965
maxZ 4.580189
```

lexaknyazev:
1. Assume a glTF asset containing out-of-range SH values. Maybe it's been created manually, maybe the training process had a bug, we'll never know.
2. This asset is opened in a glTF splat viewer.
3. Is it feasible to process all SH values and confirm that they would not produce out-of-range color values? Not likely, so the loading completes.
4. The app then proceeds with rendering and those out-of-range values inevitably reach the shader code.
5. If the spec does not require clamping at this step, the rendering results are undefined, precisely because such cases should not happen.

So the spec may either leave that as-is and call out the undefined results (like with negative scale factors on node transforms) or it may require clamping to make the ecosystem more reliable. There's also a middle ground, that is saying "SHOULD" instead of "MUST". In the latter case, implementations would be advised to clamp but not required to.

NorbertNopper-Huawei:
For srgb, the sum plus the 0.5 (this was also a discussion) is between 0 and 1.
For linear, the sum plus the 0.5 is is normally in the 0 and 1 range, however, can be below 0 or above 1 depending on the data.

NorbertNopper-Huawei:
The point is, that the sum of the data is important. And then we come to conversion cases between display referred data as defined in the above graph, which is well defined.

lexaknyazev:
> For srgb, the sum plus the 0.5 (this was also a discussion) is between 0 and 1.

Even if that's the expectation, the spec MUST say what to do if the computed sum is not in that range.

It might be easier to say that clamping happens after multiplying the computed color with $\alpha_i$ because that is exactly what happens today in viewers that output premultiplied colors into 8-bit framebuffers.

NorbertNopper-Huawei:
> > For srgb, the sum plus the 0.5 (this was also a discussion) is between 0 and 1.
> 
> Even if that's the expectation, the spec MUST say what to do if the computed sum is not in that range.
> 
> It might be easier to say that clamping happens after multiplying the computed color with α i because that is exactly what happens today in viewers that output premultiplied colors into 8-bit framebuffers.

for srgb clamp between 0 and 1 and for linear do nothing (the encoding/decoding will do this part)

NorbertNopper-Huawei:
and to have a common wording, we can say data is untouched, as the rest is done in another step.

NorbertNopper-Huawei:
Within display referred, there are well defined conversions. And what Adam is saying, however, you can not prevent others to use the data in a screen referred (and lighting context) context. This can be defined in the wide gamut or another related extension.
```mermaid
flowchart TD
  A[...] --> B[tonemapping]
  B --> C[display referred]
  C --> D[Any display output format]

NorbertNopper-Huawei:
If someone has a srgb sum is equal as rendering a srgb image (e.g. a photo).
If someone has a linear sum is equal as rendering a linear image (e.g. a photo).

DRx3D:
It seems to me that if you wish to claim that the training process only produces valid values, then we need to standardize the training process too and not just the display of data. I think that standardizing training is out of scope of this particular effort.

NorbertNopper-Huawei:
> It seems to me that if you wish to claim that the training process only produces valid values, then we need to standardize the training process too and not just the display of data. I think that standardizing training is out of scope of this particular effort.

This is not required, as the training process automatically produces values in the given range. When you want to do it manually, you just need to be aware what you are doing.

NorbertNopper-Huawei:
So, again, having srgb, if you want to be precise and nitpicking, clamp between 0 and 1.
For linear, leave as it is.

NorbertNopper-Huawei:
@DRx3D Marco was so kind sharing this screenshot here: https://github.com/KhronosGroup/glTF/issues/2562#issuecomment-4150222995

You want to do this manually?

javagl:
>  if you want to be precise and nitpicking

We have to be.

> clamp between 0 and 1.

Yes.

NorbertNopper-Huawei:
Hmmm, looking at my previous messages, there is something contradictive. Let me re-read ...

NorbertNopper-Huawei:
linear should also be clamped between 0 and 1. Sorry, was wrong.
So, both clamp(0, 1)
Hmm, did I say this during the 3D Formats call ...

NorbertNopper-Huawei:
... it has to be like this, as it happens after tone mapping (see picture above).

NorbertNopper-Huawei:
Let me cross check with AI

NorbertNopper-Huawei:
image

NorbertNopper-Huawei:
So, for display referred and current color spaces `srgb_rec709_display` and `linear_rec709_display` in the given glTF extension, we use `clamp(0, 1)`.
For other `colorSpace` values, as a future glTF extension as usual.

NorbertNopper-Huawei:
I implemented something similar a few years ago at UX3D from scratch to understand the color conversion pipeline. Maybe in this code, there is a bug.
For the shared 3DGS viewer, not yet implemented the linear case.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
javagl: I think the question about the 'intention' referred narrowly to the statement there:

> Implementations are allowed to relight the splats than the one they were trained in,

There's a part of the sentence missing. 
I assume that the missing part may be something like

> Implementations are allowed to relight the splats **to use a different color space** than the one they were trained in,

or so.

lexaknyazev:
After further review, the clamping issue seems far from just nitpicking. Not specifying this step may lead to severe output differences when using alpha-premultiplied blending. Consider the following input:
- `srgb_rec709_display`
- Two identical splats in the same place
  - SH0: `(2, 4, 7)`, therefore the initial color value is `(1.06, 1.63, 2.47)`
  - Opacity: `0.5`
- Consider only the pixel corresponding to the mean point

#### Clamping the SH sum
1. Color value becomes `(1.0, 1.0, 1.0)`
2. Fragment output alpha-premultiplied color is `(0.5, 0.5, 0.5, 0.5)` (already in range)
3. Blending result is `(0.75, 0.75, 0.75)` (still in range)

#### Implicit clamping when rendering to a normalized color buffer
1. Color value remains `(1.06, 1.63, 2.47)`
2. Fragment shader alpha-premultiplied color is `(0.53, 0.81, 1.24, 0.5)`
3. Actual fragment color after implicit clamping is `(0.53, 0.81, 1.0, 0.5)`
5. Blending result is `(0.8, 1.22, 1.5)`, further clamped to `(0.8, 1.0, 1.0)` on write

NorbertNopper-Huawei:
For 3D Gaussian Splatting, this should not happen. Because the original images color values were between 0 and 1. So, the sum has to be between 0 and 1. So, if we have something like 1.0000000001, I would consider as valid data. However, if the sum is e.g. 2.0, the data is for sure incorrect.

javagl:
I took an implicit "TODO (when I'm really bored)" to check this, but maybe someone knows from the tip of their head:

The specification currently allows implementations to _ignore_ the higher-degree spherical harmonics. Can there be a case where a color component becomes _smaller_ by taking the higher-degree coefficients into account? Or as pseudocode: Can there be a case where
`colorRed = computeFrom(sh0);` yields `1.5` but
`colorRed = computeFrom(sh0, sh1, sh2);` would yield `0.99`? 

I think that this _can_ be the case, given that the coefficients can be negative, and how they are mushed together in the shader, but I haven't confirmed.

NorbertNopper-Huawei:
> I took an implicit "TODO (when I'm really bored)" to check this, but maybe someone knows from the tip of their head:
> 
> The specification currently allows implementations to _ignore_ the higher-degree spherical harmonics. Can there be a case where a color component becomes _smaller_ by taking the higher-degree coefficients into account? Or as pseudocode: Can there be a case where `colorRed = computeFrom(sh0);` yields `1.5` but `colorRed = computeFrom(sh0, sh1, sh2);` would yield `0.99`?
> 
> I think that this _can_ be the case, given that the coefficients can be negative, and how they are mushed together in the shader, but I haven't confirmed.

You can try it out e.g. in the https://superspl.at/editor
There you can change the SH bands during runtime.

Sorry, do not want to be impolite. Let me check. Or what do you not understand when asking AI?

lexaknyazev:
@NorbertNopper-Huawei I think we have agreed on clamping in principle already. The final confirmation is where exactly it happens:
1. On the SH sum before applying alpha
2. On the SH sum after applying alpha

If we do (1), then (2) is not needed. If we do only (2), then blending would have to be done with alpha-premultiplied values to avoid very convoluted workarounds. Note that (2) technically allows premultiplied color values exceeding the corresponding alpha values, which is generally incorrect but that's what some viewers do today, maybe by mistake.

NorbertNopper-Huawei:
@lexaknyazev What do you suggest? I mean, and I am wondering, why now these questions do pop up now.

I do not know it at the moment, do you know a solution? I try to provide one as well. Looking at my code.

NorbertNopper-Huawei:
@lexaknyazev Can you please share the links of these viewers, where this happens? Want to look at the code and understand.

NorbertNopper-Huawei:
I do not have the clamp in my code, let's look at the original paper, as this is the ground truth.

NorbertNopper-Huawei:
Let's ask my good old friend R2D2, ...

NorbertNopper-Huawei:
... seems in the original paper and implementation, there is no clamp at all. So, a perfect pipeline would not need it. However, I suggest to just tolerate the current implementations and provide a guide, how to do in a correct way.

NorbertNopper-Huawei:
Of course, maybe someone else should look at this as well.

lexaknyazev:
@NorbertNopper-Huawei Your viewer performs color clamping implicitly, like described in (2) above because:
1. The SH sum itself is not clamped.
2. The fragment output and the blending state use premultiplied alpha.
3. The rendering surface is unsigned normalized.

You could try adding `clamp` to the end of the `computeColor` shader function to check what would happen if we standardize on (1).

NorbertNopper-Huawei:
@lexaknyazev The links to the viewers, please.
Also, please show me, where this happens in the original paper and code.

NorbertNopper-Huawei:
Of course, this is a glTF community decision and we will adapt after internal clarifications.

lexaknyazev:
The original research paper provides both sides of the pipeline so it does not need to explicitly call out potential cross-vendor portability issues. Standards like glTF are generally one-sided in that sense.

By "your viewer" I mean the one currently located in the Khronos internal GitLab.

NorbertNopper-Huawei:
> @NorbertNopper-Huawei Your viewer performs color clamping implicitly, like described in (2) above because:

Yes, it would clamp but it does not, as the data itself is in the 0 to 1 range. This is the reason, that (1) and (2) implementors should consider (0) - the original paper implementation - where no clamping is done and required.

NorbertNopper-Huawei:
> ... Standards like glTF are generally one-sided in that sense.

Maybe both sides need to make changes in their implementations? That is also an option.

NorbertNopper-Huawei:
A compromise. Also, future proof for scene referred data.

lexaknyazev:
> Yes, it would clamp but it does not, as the data itself is in the 0 to 1 range.

- The viewer has no control over the input data.
- The viewer (as it stands today) does clamp as in (2). To claim "no clamp", it must be changed to use a floating-point color buffer for blending (which may make rendering much slower).

NorbertNopper-Huawei:
> > Yes, it would clamp but it does not, as the data itself is in the 0 to 1 range.
> 
> * The viewer has no control over the input data.
> * The viewer (as it stands today) does clamp as in (2). To claim "no clamp", it must be changed to use a floating-point color buffer for blending (which may make rendering much slower).

It is assumed, that the 3DGS data in the glTF is valid, of course. And again, the viewer does not clamp, as the values are not above 1 or below 0.

lexaknyazev:
> It is assumed, that the 3DGS data in the glTF is valid, of course.

Unless we provide an exhaustive SH data validity definition, including free view direction selection and toggling of the higher degrees (see Marco's comments above), all SH data is considered as valid.

NorbertNopper-Huawei:
We could add a note, where we say, that by purpose we are not clamping, as a) we rely on the original paper where no clamping happens and b) there are several approaches to handle this issue, not just clamping, and glTF does not want to mandate this. Also, this is a general problem to 3DGS data itself, so not just related to 3DGS in glTF.

NorbertNopper-Huawei:
If we would have perfect data, the data does not need to be clamped, as in the 0 to 1 range. However, as this can not be guaranteed, there are several approaches to handle this issue, where clamping is one of them.
I was not aware of this. Now I am.

lexaknyazev:
> we rely on the original paper where no clamping happens 

This is not a valid argument. Research papers are not concerned with interoperability.

If you want to treat the reference rasterizer as the ground truth, then three rules should be followed:
1. Negative SH sums must be clamped to zero before alpha multiplication (see the `computeColorFromSH` function in the forward pass).
2. All rendering and blending must be done in a floating-point color buffer (because the reference rasterizer accumulates everything in floating-point variables).
3. The very final color values should be clamped to 1.0 (because the image is saved to PNG).

NorbertNopper-Huawei:
> > we rely on the original paper where no clamping happens
> 
> This is not a valid argument. Research papers are not concerned with interoperability.
> 
> If you want to treat the reference rasterizer as the ground truth, then three rules should be followed:
> 
> 1. Negative SH sums must be clamped to zero before alpha multiplication (see the `computeColorFromSH` function in the forward pass).
> 2. All rendering and blending must be done in a floating-point color buffer (because the reference rasterizer accumulates everything in floating-point variables).
> 3. The very final color values should be clamped to 1.0 (because the image is saved to PNG).

Then please make a proposal how you would do it and I trust you, that it will be accepted by everyone.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
lexaknyazev: Since the covariance matrix expects the effective rotation quaternions to be unit, there are only two options here:
1. Say that stored values are unit. Renderers would be able to use rotation values as-is, however this would restrict the representable rotations, especially for the 8-bit case. Note that normalized 16-bit and floating-point accessors would also be required to contain unit quaternions.
2. Say that implementations must renormalize rotations after reading them from accessors. This would likely be redundant for floating-point and 16-bit formats but may provide slightly better results for the 8-bit case. Given that it's generally inconvenient to condition shader code on accessor component types, such a requirement should not depend on the accessors thus it would be wasteful for pre-normalized values.

I'd suggest the first option.

weegeekps: Option 1 is the best option for splatting. I suspect that it's largely what most users and developers expect regarding a 3DGS implementation based on the original paper.

lexaknyazev: Then no further changes are needed. Added a note.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
lexaknyazev: Updated the section to better reflect its intent.

extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
javagl: For lack of a better solution, the non-permalink will have to do it for now. Opened https://github.com/AcademySoftwareFoundation/ColorInterop/issues/12 , though.

weegeekps (approved):
This all looks good to me. Clear, concise, and well defined. Thank you!

PR #2568, opened 2026-04-15
Update extensions registry

lexaknyazev:


javagl (approved):
I tried to find "the" indicator that `KHR_node_visibility` was actually ratified. I found the _submission_ for ratification, but no indicator that it went through. In doubt, _you_ are the indicator.

Issue #2561, opened 2026-02-07
Proposal: Layered Audio Extension Architecture — KHR_audio_emitter + KHR_audio_graph + KHR_audio_environment


rudybear:
Hi all,

  I'd like to share a proposal for a layered architecture that combines the KHR_audio_emitter and KHR_audio_graph extensions, along with a new KHR_audio_environment extension for listener
  and room acoustics.

  Summary

  Rather than having two competing audio extensions, I propose using KHR_audio_emitter as the base layer and building graph-based processing and environmental audio on top of it:

  - **KHR_audio_emitter** (base) — audio data, sources, emitters. One proposed change: node-level emitter (single) → emitters (array) to support multiple emitters per node.
  - **KHR_audio_graph** (extends audio_emitter) — document-level processing graphs with explicit input/output bindings to audio_emitter sources and emitters. Includes gain, delay, 8 filter
  types, waveshaper, channel routing, and graph-only oscillator sources.
  - **KHR_audio_environment** (extends audio_emitter) — explicit listener node (HRTF/equalpower/custom), parametric and IR-based reverb, custom distance models, per-emitter spatialization
  override.

  Key design points

  - Graphs are declared at document level with explicit bindings — not embedded in sources or emitters
  - Oscillators are graph-only source nodes (no base spec change needed)
  - Reverb and spatialization model selection live in the environment layer, not the processing graph
  - All conventions follow audio_emitter: seconds, radians, gain (0, +∞)
  - The only required change to audio_emitter is emitter → emitters (array) on node bindings

  Materials

  Spec proposals and gap analysis:
  https://github.com/rudybear/KHR_audio_proposals

  - https://github.com/rudybear/KHR_audio_proposals/blob/main/analysis/spec_comparison_and_gap_analysis.md
  - https://github.com/rudybear/KHR_audio_proposals/blob/main/specs/KHR_audio_emitter_changes.md
  - https://github.com/rudybear/KHR_audio_proposals/blob/main/specs/KHR_audio_graph.md
  - https://github.com/rudybear/KHR_audio_proposals/blob/main/specs/KHR_audio_environment.md

  Reference implementation (Web Audio API / TypeScript):
  https://github.com/rudybear/AudioGraphJS/tree/feature/layered-extensions

  Includes runtime, 4 layered-format examples, browser demo, 6 test suites, and a validator. Backward compatible with existing KHR_audio_graph examples.

[...]

PR #2572, opened 2026-04-22
Refine KHR_audio_graph spec and schemas

rudybear:
## Summary
- refactor `KHR_audio_graph` around document-level graphs with explicit inputs, outputs, and node schemas
- remove the older graph-owned emitter/listener/audioData model in favor of the updated schema layout
- keep this PR scoped to `KHR_audio_graph` only

## Validation
- `python3 -m json.tool` on all `KHR_audio_graph/schema/*.json` files

## Context
- mirrored from `facebook/glTF` branch `codex/update-khr-audio-graph-only`
- related Facebook PR: https://github.com/facebook/glTF/pull/4


NorbertNopper-Huawei:
@rudybear I did not look at the 42 changed files (look, two times the same number, the number for the answer to the ultimate question of life ☺️) and I assume, we will also add all folks as contributors for the audio extension. Do not have the full list in my mind, however, maybe we can review the glTF roadmap with the given glTF extensions:
https://docs.google.com/spreadsheets/d/1cKQKc0cckowwe8d1cYMNzucgxBY5o8LXox9MKALqClU/edit?gid=334180457#gid=334180457

rudybear:
Self-review after building this spec end-to-end (Web Audio reference runtime with full test coverage, plus a WebGPU viewer integration) and benchmarking against Web Audio 1.0/1.1 and the X3D 4.0 Sound component. Items I intend to address in the next revision — posting publicly for visibility:

**Spec bugs**
1. **`custom` oscillator type has no data model** — add `periodicWave { real[], imag[] }` (Web Audio/X3D PeriodicWave) or drop `custom` from the enum until defined. Same treatment needed for the `custom` gain-interpolation curve and the waveshaper `amount`→curve mapping (currently implementation-defined).
2. **The bundled KHR_audio_emitter copy is stale** vs #2137 (singular `emitter` in the node schema/prose vs the plural `emitters` array adopted upstream in April). Plan: drop the vendored copy and depend on #2137 directly.

**Design items**
3. **Add a `compressor` node** (Web Audio/X3D DynamicsCompressor params). It is the only roster gap vs *both* benchmark standards, and the workhorse for mastering/ducking.
4. **Feedback cycles**: currently DAG-only, which forbids feedback-delay topologies. Web Audio permits cycles when every cycle contains a DelayNode (≥1 render quantum). Leaning toward adopting that rule; validators can check it.
5. **Extended source properties placement**: loopStart/loopEnd/offset/when/duration/priority are playback semantics, not graph semantics — they likely belong in the base layer (or a small playback extension), and `playbackRate` currently exists in both layers. `state` as a mutable string pointer needs normative transitions or replacement with interactivity verbs (see the playable-asset experiment linked from #2561). Needs joint resolution with #2137.
6. **Define the playback clock**: `when` is seconds from an undefined origin. USD's stage-time anchoring is the precedent worth following (schedule relative to a glTF animation timeline).

**Decisions to document (so review doesn't re-litigate them)**: no audio-rate parameter connections (reserve a `to: {node, param}` connection form for future use), no envelopes (gain smoothing is the mechanism), graphs are pre-spatialization (master-bus inserts belong to the environment layer), and implicit fan-in mixing follows Web Audio rules with `speakers` interpretation.

Also: the reference implementation's schema-validation CI now validates against this PR's schemas directly (fixtures modernized from the pre-refactor model). Full gap analysis: https://github.com/rudybear/KHR_audio_proposals/blob/feature/environment-v2/analysis/2026-08-session/03-gap-analysis-audio-graph.md

PR #2576, opened 2026-05-01
Add MPEG lighting extensions

haudiobe:
MPEG_lights_texture_based and MPEG_lights_punctual: When inserting virtual information in a captured real-world environment, lighting is a fundamental cue for providing a realistic experience to the user. Improper lighting and shadows can break the immersive illusion of the inserted virtual objects. In a VR context, accurate lighting models allow achieving a high level of realism. The MPEG-SD lighting extensions allow for dynamic lighting that is more like a video than a static source: coefficients like reflectance can change over time. The light sources can be of two natures: real or virtual. In both cases, the light sources may be represented by the same mathematical models (punctual light, ambient light, etc.).

### DIFF
--- a/.github/COWORKERS
+++ b/.github/COWORKERS
@@ -1,0 +1,1 @@
+/extensions/2.0/Vendor/MPEG_lighting @edrthomas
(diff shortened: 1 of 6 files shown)


pzgulyas:
Isn't there an overlap with KHR_animation_pointers?

PR #2578, opened 2026-05-06
Update axes description

### LINKED ISSUE #2352: Coordinate System and Units
Spec says:
> glTF uses a right-handed coordinate system. glTF defines +Y as up, +Z as forward, and -X as right; the front of a glTF asset faces +Z.

Should it be +X as right?

### LINKED ISSUE #2484: "Positive rotation is counterclockwise" in section 3.4 is an ambiguous statement
In general, words like clockwise and counterclockwise only work in the context of a completely fixed 2D coordinate system, and even then I wouldn't recommend using it because of all the transformations we do!

### The ambiguity
Say, from the perspective of a viewer:
- If Z is forward, X is left, and Y is up, a positive rotation around Z is clockwise 
- If Z is back, X is right, and Y is up, a positive rotation around Z is counterclockwise

This is ambiguous! Note that both of these are right handed, just like the spec says

### Disambiguating
A handedness independent way to describe rotations in, say, an XY or UV plane, is to say that
- In the XY plane, **positive rotations go in the angular direction from X to Y**
- The same rule applies to all other planes too!
- a positive YZ plane rotation goes from Y to Z
- a positive ZX plane rotation goes from Z to X
- a positive UV plane rotation goes from U to V

A more general and accessible way to describe positive rotation direction in any plane/around any axis, without getting bogged down in technical details like cross products or determinants, would be to say that gltf follows the right hand rule for both rotations and triangle winding.

If you make a thumbs up gesture with your right hand, then if your thumb is a rotation axis, the curl of your fingers will tell you the positive rotation direction. This applies to triangle winding as well - if the curl of your fingers follow the vertex definition order spatially, your thumb represents the face normal, if a right handed winding order is used.

Here's an infographic I made [a while back](https://bsky.app/profile/freya.bsky.social/post/3lat6j5d2y225)! As far as I've been able to tell, this is a consistent rule that should apply across the board, from rotations, windings, cross products, 3D space, and UV space. It seems to be the most mathematically pure way to define rotation ordering, and to avoid all sign errors. If you deviate from this, the implementation is likely flawed and will cause sign errors throughout the codebase, which will only cause more headaches down the line in search of landing at the next stop of, an even number of sign errors c:

Image

I saw there was a similar PR on this before but it seems like it didn't go anywhere and also was getting wrapped up in technical solutions for what I think can be more easily explained geometrically!

lexaknyazev:
Fixes #2352.
Fixes #2484.

/cc @aaronfranke


aaronfranke (approved):

aaronfranke:
I wholeheartedly approve this PR. "Front" and "side" are much clearer terms, and the airplane is a much clearer image.

javagl (approved):
This review should have been submitted earlier, but somehow got buried. All fine.

aaronfranke (approved):

PR #2603, opened 2026-06-09
proposal for KHR_mesh_opacity_micromap

pixeljetstream:
Hello,

with the standardization of [Vulkan `VK_KHR_opacity_micromap`](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_opacity_micromap.html) as well as 
[Microsoft DirectX Raytracing — opacity micromaps](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#opacity-micromaps) we are proposing this extension to allow providing micromap data as part of a glTF.

Because the data is typically mesh-specific (micromaps operate in barycentric space) we added a mesh prefix to the extension name, also so that it doesn't clash with the Vulkan extension on searches.

We are looking forward to your feedback.

### DIFF
--- a/extensions/2.0/Khronos/KHR_mesh_opacity_micromap/README.md
+++ b/extensions/2.0/Khronos/KHR_mesh_opacity_micromap/README.md
@@ -1,0 +1,432 @@
+<!--
+Copyright 2018-2026 The Khronos Group Inc.
+SPDX-License-Identifier: LicenseRef-KhronosSpecCopyright
+-->
+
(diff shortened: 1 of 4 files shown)


pixeljetstream:
renamed to EXT  https://github.com/pixeljetstream/glTF/tree/EXT_mesh_opacity_micromap

pixeljetstream:
superceded by https://github.com/KhronosGroup/glTF/pull/2623

PR #2610, opened 2026-07-07
Add KHR_materials_retroreflection extension

mklefrancois:
## Overview

This PR proposes a new Khronos extension, `KHR_materials_retroreflection`, which adds a physically-plausible retroreflective response to any glTF material model.

It is based on the Minimal Retroreflective Microfacet (MRM) model of [Portsmouth et al., JCGT 15(1), 2026](http://jcgt.org/published/0015/01/04/): a BSDF is made retroreflective by substituting the outgoing view direction `V` with its reflection about the surface normal before evaluation and sampling, redirecting the specular peak into the back-scatter direction. This produces the bright highlight seen on high-visibility clothing, glass-bead road markings, and safety signs. The model preserves reciprocity and energy conservation under reflection-symmetric NDFs (GGX, Beckmann, Phong).

A `retroreflectionFactor` linearly blends between the standard material (0.0) and the retroreflective material (1.0), matching OpenPBR's `specular_retroreflectivity` parameter. An optional `retroreflectionTexture` provides a per-texel multiplier.

## Contents

- Extension specification (`README.md`)
- JSON schema (`material.KHR_materials_retroreflection.schema.json`)

retroreflection

## Status

Draft

## Relationship to other work

Aligns with the retroreflection models in [OpenPBR 1.1](https://github.com/AcademySoftwareFoundation/OpenPBR/blob/475bd90d5211d1a4bfb3227d147c692cbd9bc958/index.html#retroreflectivity) and [MaterialX](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2783).

## Known implementations

- [NVIDIA nvpro-samples / vk_gltf_renderer](https://github.com/nvpro-samples/vk_gltf_renderer)

## Contributors

- Martin-Karl Lefrançois, NVIDIA (@mklefrancois)
- Nia Bickford, NVIDIA (@NBickford-NV)


extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,135 @@
...
+* Nia Bickford, NVIDIA, [@NBickford-NV](https://github.com/NBickford-NV)
+
+## Status
+
+Draft
+
+## Dependencies
+
+Written against the glTF 2.0 spec.
+
```
bsdorra, line 17: Taking care of non-supported material models

```suggestion
## Exclusions

- This extension must not be used on a material that also uses `KHR_materials_pbrSpecularGlossiness`.
- This extension must not be used on a material that also uses `KHR_materials_unlit`.
```

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,135 @@
...
+## Overview
+
+This extension adds a physically-plausible retroreflective response to any glTF material model.
```
bsdorra: OpenPBR only changes the core opaque specular BRDFs. Lobes such as clearcoat and sheen remain unchanged. glTF should do the same.

```suggestion
This extension adds a physically-plausible retroreflective response to the metallic and dielectric BRDFs of the glTF metallic-roughness material model.
```

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,135 @@
...
+It is based on the Minimal Retroreflective Microfacet (MRM) model of [Portsmouth et al. (JCGT 15(1), 2026)](http://jcgt.org/published/0015/01/04/),
+which turns a BSDF into a retroreflective one by substituting the outgoing view direction $V$ with its reflection about the surface normal, $V_\text{retro} = 2 N (V \cdot N) - V$, before evaluation and sampling.
+This redirects the specular peak into the back-scatter direction, producing the bright
+retroreflective highlight seen on high-visibility clothing,
+glass-bead road markings, many safety signs, and other materials.
+In addition, MRM preserves reciprocity and energy conservation under reflection-symmetric NDFs (including GGX, Beckmann, and Phong).
+
+A *retroreflection weight* then linearly blends between the regular material model (at a factor of 0) and the retroreflective material model (at a factor of 1). This matches the OpenPBR `specular_retroreflectivity` parameter.
```
bsdorra:
```suggestion
A *retroreflection weight* then linearly blends between the regular metallic and dielectric BRDFs (at a factor of 0) and their retroreflective variants (at a factor of 1). This matches the OpenPBR `specular_retroreflectivity` parameter.
```

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Schema
+
+* [material.KHR_materials_retroreflection.schema.json](schema/material.KHR_materials_retroreflection.schema.json)
+
+## Sample Asset
+
+A [traffic_cone](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/TrafficCone/traffic_cone/traffic_cone.gltf) sample contains a traffic cone with a retroreflective material using the `retroreflectionFactor` and `retroreflectionTexture` parameters, alongside the same cone with a non-retroreflective material for comparison. It includes `KHR_lights_punctual` for directional lighting (note though that `KHR_materials_retroreflection` does not depend on `KHR_lights_punctual`). [The original model](https://sketchfab.com/3d-models/traffic-cone-573feef839d7450cb3e12da9986e7a98) was created by [hinndia](https://sketchfab.com/hinndia), modified by Martin-Karl Lefrançois to add retroreflection, and is licensed under [CC-BY-4.0](http://creativecommons.org/licenses/by/4.0/); see [license.txt](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/TrafficCone/traffic_cone/license.txt).
```
bsdorra: To my knowledge, extensions usually don't refer to sample assets. Remove?

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+The
+same $V \to V_\text{retro}$ substitution is used when sampling $f_\text{retro}$ and when
+evaluating its PDF; see Listing 1 of Portsmouth et al. 2026.
+
+### Note on BTDFs
+
+Implementations may choose to replace $V$ by $V_\text{retro}$ in the BSDF for the entire material to form a retroreflective BSDF for the entire material. This is often simpler to implement because one can apply the $V \to V_\text{retro}$ substitution at a higher level rather than in individual lobes.
+
+However, as described in Section 5 of Portsmouth et al. 2026, this produces nonintuitive results for transmissive materials, especially when the index of refraction $\eta$ is near 1. Therefore, this specification recommends applying retroreflection only to reflective lobes, which also matches OpenPBR and MaterialX's retroreflection implementations.
```
bsdorra: I suggest splitting this section into a normative part that specifies how to integrate the new feature with the base metal-roughness model described in [Appendix B](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#appendix-b-brdf-implementation), and a non-normative part that provides information about the MRM implementation. This keeps the extension flexible by allowing different ways to implement retroreflection.

Example using the notation style and definitions of Appendix B:

---

## Retroreflection

If this extension is present on a material, then it replaces the metallic BRDF and dielectric BRDF of the glTF metallic-roughness material, as defined in [Appendix B](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#appendix-b-brdf-implementation), with blends between the original BRDFs and retroreflective variants.

Let `metal_brdf_retro` and `dielectric_brdf_retro` be retroreflective variants of `metal_brdf` and `dielectric_brdf`, respectively, increasing the reflective response in the back-scatter direction.

```
metal_brdf =
  mix(metal_brdf, metal_brdf_retro, w)

dielectric_brdf =
  mix(dielectric_brdf, dielectric_brdf_retro, w)
```

where $w$ is the per-shading-point weight from the [Extending Materials](#extending-materials) section.

The final material is then evaluated as defined in Appendix B:

```
material = mix(dielectric_brdf, metal_brdf, metallic)
```

## Implementation

*This section is non-normative.*

A straightforward implementation of the retroreflective variants is the Minimal Retroreflective Microfacet (MRM) model of [Portsmouth et al. (JCGT 15(1), 2026)](http://jcgt.org/published/0015/01/04/). Using the notation from Appendix B, MRM substitutes the reflected view direction $V_\text{retro}$ for $V$ when evaluating and sampling the affected BRDFs:

$$V_\text{retro} = \texttt{reflect}(-V, N) = 2 N (V \cdot N) - V.$$

All intermediate quantities derived from $V$, including the half vector $H$ and Fresnel terms, are evaluated using $V_\text{retro}$ for the retroreflective variants. The same $V \to V_\text{retro}$ substitution is used when sampling the retroreflective BRDF and evaluating its PDF; see Listing 1 of Portsmouth et al. 2026. When sampling the blended BRDF as a mixture, its PDF is the corresponding weighted sum.

As described in Section 5 of Portsmouth et al. 2026, applying the substitution to BTDFs produces nonintuitive results for transmissive materials. This extension therefore applies retroreflection only to the metallic and dielectric BRDFs, matching OpenPBR and MaterialX.

extensions/2.0/Khronos/KHR_materials_retroreflection/schema/material.KHR_materials_retroreflection.schema.json
```diff
@@ -0,0 +1,25 @@
...
+            "description": "Linear blend between forward microfacet and retroreflective microfacet.",
+            "default": 0.0,
+            "minimum": 0.0,
+            "maximum": 1.0,
+            "gltf_detailedDescription": "Linear blend between the regular forward microfacet (0.0) and the retroreflective microfacet (1.0). This will be multiplied by the retroreflection texture when present."
+        },
+        "retroreflectionTexture": {
+            "allOf": [ { "$ref": "textureInfo.schema.json" } ],
+            "description": "A texture that defines the retroreflection factor in the red channel.",
+            "gltf_detailedDescription": "A texture that defines a per-texel multiplier for retroreflectionFactor, stored in the red (R) channel with a linear transfer function. This will be multiplied by retroreflectionFactor. Values outside [0, 1] are clamped."
```
lexaknyazev: glTF 2.0 only supports unsigned normalized textures so please remove the statement about clamping.

If a future texture extension would add formats with values outside of the unsigned normalized range, usage of such textures for texture slots like this one (there are lots of them already) would have to be defined elsewhere.

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,135 @@
...
+  "extensionsUsed": [ "KHR_materials_retroreflection" ]
+}
+```
+
+The extension object contains the following properties:
+
+| | Type | Description | Required |
+| -------------------------- | -------- | ----------- | ------------------ |
+| **retroreflectionFactor** | `number` | Linear blend between forward microfacet (0.0) and retroreflective microfacet (1.0). Range [0, 1]. | No, default: `0.0` |
+| **retroreflectionTexture** | [`textureInfo`](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#reference-textureinfo) | Per-texel multiplier for `retroreflectionFactor`, sampled from the **R** channel. Values outside [0,1] are clamped. | No |
```
lexaknyazev: Ditto.

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Overview
+
+This extension adds a physically-plausible retroreflective response to the metallic and dielectric BRDFs of the glTF metallic-roughness material model.
+It is based on the Minimal Retroreflective Microfacet (MRM) model of [Portsmouth et al. (JCGT 15(1), 2026)](http://jcgt.org/published/0015/01/04/),
+which turns a BRDF into a retroreflective one by substituting the outgoing view direction $V$ with its reflection about the surface normal, $V_\text{retro} = 2 N (V \cdot N) - V$, before evaluation and sampling.
+This redirects the specular peak into the back-scatter direction, producing the bright
+retroreflective highlight seen on high-visibility clothing,
+glass-bead road markings, many safety signs, and other materials.
+In addition, MRM preserves reciprocity and energy conservation under reflection-symmetric NDFs (including GGX, Beckmann, and Phong).
```
lexaknyazev: Please expand "NDF".

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+A *retroreflection weight* then linearly blends between the regular metallic and dielectric BRDFs (at a factor of 0) and their retroreflective variants (at a factor of 1). This matches the OpenPBR `specular_retroreflectivity` parameter.
```
lexaknyazev: Please remove the OpenPBR mention from this paragraph as it can't be normative.

If it's useful for readers, it should be a note formatted as a separate statement like this:
```
> [!NOTE]
> This matches the OpenPBR `specular_retroreflectivity` parameter.
```

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+<figure>
+<img src="./figures/retroreflection.png"/>
+<figcaption><em>Left: A traffic cone with <code>KHR_materials_retroreflection</code> on both sleeves (using a <code>retroreflectionFactor</code> of 1, masked by <code>retroreflectionTexture</code>). The lower sleeve appears brighter than the top sleeve because the light and camera are more closely aligned there. Right: The same traffic cone model and material without the extension. Rendered in <a href="https://github.com/nvpro-samples/vk_gltf_renderer">vk_gltf_renderer</a>.</em></figcaption>
```
lexaknyazev, line 39: Should this description be updated following the recent updates of the sample asset?

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,142 @@
...
+## Schema
+
+* [material.KHR_materials_retroreflection.schema.json](schema/material.KHR_materials_retroreflection.schema.json)
+
+## Known Implementations
+
+* [NVIDIA DesignWorks Samples' vk_gltf_renderer](https://github.com/nvpro-samples/vk_gltf_renderer)
+
+## Resources
+
```
lexaknyazev, line 148: This section should likely be marked as non-normative.

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
NBickford-NV: Thank you! I'm not 100% sure how to refer to the sample asset given that a link could be unstable -- e.g. maybe we could say "the TrafficCone asset from the glTF-Sample Assets repository"?

I'll send fixes for your other comments over to MKL to be merged into this PR -- thanks again!

lexaknyazev: This screenshot is from the original two-cone asset, right? IIUC, it has been updated to use `KHR_materials_specular` on the cone without retroreflection.

So my question is: should the screenshot in the extension spec and the description of this screenshot be updated to be in sync with the sample asset?

NBickford-NV: Oh! I see. Yes, I can pull the latest screenshot from https://github.com/KhronosGroup/glTF-Sample-Assets/pull/285/changes#diff-0a2815579d70912e579a8e1cfb02cbbc1ac3169c416f7a835e2c6a1c165a0c56 . It looks like both use `KHR_materials_specular`, though, so the description's technically still accurate -- do you think it'd be wise to mention both use `KHR_materials_specular` or is that sort of superfluous?

DRx3D: @NBickford-NV : You can use the PR link if you need to refer the sample. The restructure PR should go in the next week or so. There will be time after that lands to update this PR with the specific (and final) location.

lexaknyazev: Please don't use PR links in specs; they are temporary by definition.

NBickford-NV: To clarify, what I'm thinking is replacing the image asset with a copy of the image from the other PR, and leaving the text unchanged since the description is technically still accurate. Does that sound OK?

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
NBickford-NV: Done, thanks!

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
NBickford-NV: This is done now, and looks like this:

image

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
NBickford-NV: I've updated the image and left the text unchanged.

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
NBickford-NV: Done, thanks!

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,147 @@
...
+All intermediate quantities derived from `V`, including the half vector `H` and Fresnel terms, are evaluated using `V_retro` for the retroreflective variants. In other words, if the regular BRDF is a function `f(L, V)` taking the light direction `L` and reflected view direction `V` at a point, then the MRM BRDF is `f(L, reflect(-V, N))`.
+
+The same `V` to `V_retro` substitution is used when sampling the retroreflective BRDF and evaluating its PDF; see Listing 1 of Portsmouth et al. 2026.
+
+## Excluding Other Lobes
+
+*This section is non-normative.*
+
+While it is possible to apply the MRM model transformation to any BSDF, this extension intentionally does not apply retroreflectivity to the additional BSDFs introduced by other extensions. This matches the behavior of retroreflection within OpenPBR and MaterialX's implementations.
```
bghgary: Could you clarify how retroreflection composes with extensions that modify the metallic and dielectric BRDFs, especially anisotropy and iridescence? The spec clearly excludes additional lobes such as clearcoat, sheen, and transmission, but does not say whether BRDF modifiers are applied before the `V` substitution. [Sample Renderer PR 50](https://github.com/KhronosGroup/glTF-Sample-Renderer/pull/50) currently evaluates its retro branch without anisotropy or iridescence; [Babylon](https://github.com/BabylonJS/Babylon.js/pull/18829) keeps both inside the base retro variant. A sentence defining that ordering would prevent divergent implementations.

bghgary (commented):
[Reviewed by Copilot on behalf of @bghgary]

Clarification inline.

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
mklefrancois: Thanks for asking for clarification, @bghgary!

Matthias Raab, one of the spec authors, confirmed this. "BRDF modifiers like iridescence should be applied after the V substitution, i.e., using the mirrored view direction, not the original one." The reasoning is that these effects are direction-dependent; therefore, for the combined result to be well-defined, the modifiers need to see the same view vector that the retro lobe itself uses.

Practically, that means:

- Anisotropy and iridescence should be evaluated inside the retro branch using the mirrored V and not be skipped or computed against the pre-mirror V.
- Regarding https://github.com/KhronosGroup/glTF-Sample-Renderer/pull/50, this was an oversight and will be fixed. Thanks for reporting this.

mklefrancois: A fix was applied in https://github.com/KhronosGroup/glTF-Sample-Renderer/pull/50.

Now re-applies whichever BRDF-direction/Fresnel modifiers are active (anisotropy's bent-normal reflection, iridescence's Fresnel tint) using V_retro, instead of falling back to the isotropic non-iridescent base BRDF. Clearcoat/sheen/transmission remain excluded per the spec's "Excluding Other Lobes" section.

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,147 @@
...
+## Known Implementations
+
+* [NVIDIA DesignWorks Samples' vk_gltf_renderer](https://github.com/nvpro-samples/vk_gltf_renderer)
+
+## Resources
+
+*This section is non-normative.*
+
+* [Jamie Portsmouth, Matthias Raab, Laurent Belcour, and Francis Liu, The Minimal Retroreflective Microfacet Model, *Journal of Computer Graphics Techniques (JCGT)*, vol. 15, no. 1, 60-75, 2026](http://jcgt.org/published/0015/01/04/)
+* [Retroreflection in MaterialX](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2783)
```
lexaknyazev: This has been included in MaterialX 1.39.5 so it would be better to refer to the released version.

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
```diff
@@ -0,0 +1,147 @@
...
+* [Retroreflection in the OpenPBR 1.2 draft](https://github.com/AcademySoftwareFoundation/OpenPBR/blob/475bd90d5211d1a4bfb3227d147c692cbd9bc958/index.html#retroreflectivity)
```
lexaknyazev: Direct commit links may become inaccessible and do not give the full context. Ideally we'd link to the OpenPBR 1.2 release but since it's not ready yet, let's use the corresponding retroreflection PR instead.

javagl: Off-topic for this specific PR, but somewhat important in the broader sense:

> Direct commit links may become inaccessible 

Preferably, there should be tags/releases for things like this, but in general (and for the case that something is not _yet_ released/tagged), I'd see links like this as the second best option, and wonder how they could become "inaccessible", given that they are literally referred to as **perma**links...?!

lexaknyazev: FWIW, no link is _truly_ permanent. In this particular case, the commit is a part of a merged PR so it will likely persist but that PR has nothing to do with retroreflectivity so I'd say that the link is misleading. Besides, it does not render well because it uses GitHub's HTML preview instead of actually serving the HTML file (which is only possible for deployed branches).

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
NBickford-NV: For the specification, I think I've added some clarifications and added the right language in my PR [here](https://github.com/mklefrancois/glTF/pull/4): retroreflection applies after any other extensions have done their replacements to `metal_brdf` and `dielectric_brdf`. (If preferred, we could list every extension that currently does this, and say "retroreflection applies after any of these" in case future extensions would like to be applied after retroreflection.)

This was also a really good question in that it made me realize a point of confusion: although we specify in another paragraph that retroreflection (non-normatively) shouldn't apply to transmission, https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_transmission added transmission as a component of `dielectric_brdf`. (Incidentally, this means that `dielectric_brdf` is not necessarily a BRDF, but rather a BSDF!). The clearest patch I can think of (because we need to define retroreflection near that level of the input so Fresnel factors work) is to explicitly say that retroreflection shouldn't apply to transmission.

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
NBickford-NV: I've got this fixed in my PR [here](https://github.com/mklefrancois/glTF/pull/4) to point to the first relevant line in the extension at that tag, https://github.com/AcademySoftwareFoundation/MaterialX/blob/v1.39.5/documents/Specification/MaterialX.PBRSpec.md?plain=1#L203 .

extensions/2.0/Khronos/KHR_materials_retroreflection/README.md
NBickford-NV: I've got this changed in my PR [here](https://github.com/mklefrancois/glTF/pull/4) to point to https://github.com/AcademySoftwareFoundation/OpenPBR/pull/255 instead.

Plus is it's stable if the repo gets its history rewritten; minus is it doesn't include the commit hash so if the PR is deleted then the information's lost. (Looks like GitHub still shows the changes in the PR even if it's not merged in and the the head repository is deleted, which is nice; that didn't use to be the case.) I'm not sure this is worth bikeshedding, though.

Thanks!

lexaknyazev (approved):

PR #2616, opened 2026-07-16
Specify asset version syntax

### LINKED ISSUE #2452: Tighten asset spec version definition
The spec currently says the following about the `asset.version` property:
> The glTF version in the form of `<major>.<minor>` that this asset targets.

Additionally, the JSON schema implies the following regexp pattern for the string value: `^[0-9]+\.[0-9]+$`.

---

Neither of these definitions is precise enough to require exactly `"2.0"` spelling for glTF 2.0 assets. For example, should it be valid to use `"02.00"` as a version string?

@emackey @javagl @bghgary WDYT?

lexaknyazev:
Fixes #2452.

/cc @aaronfranke

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -600,7 +600,61 @@ image::figures/objects.svg[pdfwidth=6in,align=left]
 [[asset]]
 == Asset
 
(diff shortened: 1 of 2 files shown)


aaronfranke (approved):

javagl (approved):
As discussed during the call 👍

Issue #2590, opened 2026-05-27
[glTF 2.1] Unified File References


akmorgan99:
# Unified File References

glTF 2.1 defines a new `files` array for referencing external files in a unified way. This array works like the existing `buffers` and `images` arrays, but can be used for any type of file. Additionally, the `"mimeType"` property is required for files.

The `files` array will be used for all new types of files added in the future. The existing `buffers` and `images` arrays are kept as-is for those file types.

Cyclical references between files are strictly prohibited.

## Use cases

With glTF 2.0, extensions which reference external files need to define their own arrays for those files. This approach quickly runs into 2 problems:

- Extensions need to redefine the same pattern over and over: an array of files, each with `"uri"`, `"bufferView"`, and `"mimeType"` properties.

- Tools cannot reliably discover external file dependencies without understanding every extension.

For example, the `EXT_lights_ies` extension added an additional `lights` array repeating the aforementioned pattern inside the extension object.

glTF 2.1 solves these problems by providing a unified way to reference external files, simplifying both extensions and tools.

There are basic operations that should be possible to do with trivial simple implementations of glTF, but which actually end up being impossible unless they support every single extension that defines external file references. For example:

- A packing tool that packs glTF file(s) into a self-contained `.glb` file.

- A file picker dialog that allows the user to select a `.gltf` file, and then uploads the entire model including all of its dependencies.

- A validation tool that checks if a glTF file is valid and needs to check for all the files used by the glTF.

With glTF 2.1, tools only need to look in 3 places for external files: `buffers`, `images`, and `files`.

## Example

> [!NOTE]
> Examples are illustrative and not exhaustive.

```json
{
    "buffers": [
        {
            "uri": "buffer0.bin"
        }
    ],
    "images": [
        {
            "mimeType": "image/png",
            "uri": "texture.png"
        }
    ],
    "files": [
        {
            "mimeType": "model/gltf-binary",
            "uri": "model.glb"
        },
        {
            "mimeType": "audio/mpeg",
            "uri": "audio.mp3"
        },
        {
            "mimeType": "video/mp4",
            "uri": "video.mp4"
        }
    ]
}
```

## Why this design

Without unified file references, the glTF file would need to look like this instead, which causes the problems described above:

```json
{
    "buffers": [
        {
            "uri": "buffer0.bin"
        }
    ],
    "images": [
        {
            "mimeType": "image/png",
            "uri": "texture.png"
        }
    ],
    "externalAssets": [
        {
[...]

Issue #2586, opened 2026-05-27
[glTF 2.1] External Assets


akmorgan99:
# External assets

External assets, also known as complex scenes, external references, external models, nested models, are the star of the show in glTF 2.1.

External assets allow a glTF file to reference other glTF files, and use them as instantiable models in the scene hierarchy.

Cyclical references between glTF assets are strictly prohibited.

## Use cases

- Composing a large scene out of multiple glTF files.

- Reusing node hierarchies multiple times, for example multiple instances of the same character.

- Deferred loading of content, only loading referenced files when they are needed.

- Packaging up a glTF file with its dependencies into a single container file without altering the content of the original glTF file.

If the nested glTF file contains URIs, and the file is embedded data (base64 data URI or bufferView), those URIs are matched against other objects in the `files` array by `name`. Otherwise, if the nested glTF file is located at an external URI, the URIs in the nested glTF file are resolved relative to that URI. Any glTF file with `uri` properties needs to have them resolve to something, so the `files` array serves as a substitute for the file system for embedded nested glTF files where the file system is not available.

## Example

> [!NOTE]
> Examples are illustrative and not exhaustive.

```json
{
    "files": [
        {
            "mimeType": "model/gltf-binary",
            "uri": "sub_model.glb"
        }
    ],
    "externalAssets": [
        {
            "name": "sub_model_glb",
            "file": 0
        }
    ],
    "nodes": [
        {
            "name": "RootNode",
            "children": [1]
        },
        {
            "name": "SubModel",
            "externalAsset": 0
        }
    ],
    "scene": 0,
    "scenes": [{ "nodes": [0] }]
}
```

This takes a file `sub_model.glb`, references it in the `files` array, uses it as a model in the `externalAssets` array, and instantiates that model in the scene hierarchy as `SubModel`.

## Why this design

The previous design was a separate format, [glTF External Reference](https://github.com/KhronosGroup/glTF-External-Reference). It has been referred to by several names in the past including "glTFX", "glXF", "glTF Experience Format", and "glX". There was extensive discussion about where to place this functionality. Ultimately, it was decided to bring this functionality into the core glTF specification.

[...]

Issue #2446, opened 2024-09-27
Multiple buffers, duplicate URIs?


donmccurdy:
I came across a file recently containing three buffers, all sharing the same URI:

```jsonc
  "buffers": [
    {
      "uri": "glTF.bin",
      "byteLength": 8160
    },
    {
      "uri": "glTF.bin",
      "byteLength": 8160
    },
    {
      "uri": "glTF.bin",
      "byteLength": 8160
    }
  ],
```

No validation errors were found, and viewers I tested (three.js, babylon.js, playcanvas) seemed fine, but it did break processing in glTF Transform, which tried to overwrite the buffer repeatedly. The file also contained valid buffer views pointing to each buffer.

My impression is that this is suboptimal, but technically valid, does that seem correct?

javagl:
I think that there are not contraints that disallow this, so would agree that this is technically valid.

I just created a `Triangle` with such duplicate buffers:

[TriangleDuplicateBuffer.zip](https://github.com/user-attachments/files/17164447/TriangleDuplicateBuffer.zip)

Just for the heck of it, I did two further tests:

I declared them once with `"byteLength": 44` (correct) and once with `"byteLength": 43` ("wrong"). And interestingly, the validator reports **both** to have `"byteLength": 44`. (So it apparently overwrites the _declared_ byte length with the length of the actual `.bin` file...)

I also thought that it could be difficult to clearly define when two buffer declarations are actually **equal**. So (locally) I also tried using 
`"uri": "./subdirectory/../Triangle_data.bin",`
as the URI for the second one. This resolves to the same URI. And in fact, the validator reports this as `"uri": "Triangle_data.bin",` (i.e. some form of "normalized" version of the URI). But there probably are cases where such a normalization is far more difficult (or maybe even impossible). 

So ... when this is allowed, it might raise some questions for implementors. Consider an example like the `Triangle.gltf`, but with two triangles, each referring to its own accessors and buffer views and buffer. And _only on the lowest level_ - namely, on the level of the `uri` of the buffer - these buffers are **equal**. Now, when someone does
`gltf.meshes[0].primitives[0].attributes["POSITION"].set(0, 1234.567);`
will this affect _both_ triangles, because the underlying, most raw `Buffer` that was read from the `.bin` file are **identical** because they have the same URI? 

(This is probably beyond what the glTF spec can cover. But for things like glTF-Transform, there _will_ be an answer to this question, and regardless what the answer is: It might be different from what someone expects...)

[...]

PR #2618, opened 2026-07-17
Draft changes for External Assets and Unified File References in glTF 2.1

weegeekps:
This PR contains the draft changes for External Assets and Unified File References in glTF 2.1. This PR currently points towards the `draft-2.1` branch and should be safe to merge to that branch.

### Relevant Issues:

- https://github.com/KhronosGroup/glTF/issues/2586
- https://github.com/KhronosGroup/glTF/issues/2590

### Summary from Copilot:

**New external asset referencing system:**

* Added top-level `files` and `externalAssets` arrays to the glTF schema, enabling assets to reference external files and compose complex scenes from multiple glTF files. (`glTF.schema.json`, `file.schema.json`, `externalAsset.schema.json`, [[1]](diffhunk://#diff-49c121052eab7e3b5c7961ebdd28551d49efeb1b1eff44e78a592be21b463cabR75-R92) [[2]](diffhunk://#diff-8095630125a1c2d1ae148319ca4bb3cf65e0db151999054569c4732d86f19552R1-R52) [[3]](diffhunk://#diff-3c4ccc786add97902a57f52e2b5f532a004b2dbcff217dfaa5427b9a032d3bbaR1-R16)
* Nodes can now reference external assets via an `externalAsset` property, which cannot be used together with `mesh`. (`node.schema.json`, [specification/2.1/schema/node.schema.jsonL40-R44](diffhunk://#diff-b8cd660fe1441f70434b05a4b5fabafd72415b2dab8deff8c474bfd06a02efe0L40-R44))
* Introduced the concept of file aliases, allowing URIs within external assets to be remapped for efficient resource sharing or overriding. (`file.schema.json`, `file.alias.schema.json`, [[1]](diffhunk://#diff-8095630125a1c2d1ae148319ca4bb3cf65e0db151999054569c4732d86f19552R1-R52) [[2]](diffhunk://#diff-15b5eeeb206e352419208331a1760cc3d1474ed8a968e10c94404bae890d4b2dR1-R19)

**Specification and documentation updates:**

* Updated the specification to describe the new unified file reference system, external asset referencing, and aliasing mechanism, including normative requirements and implementation notes. (`Specification.adoc`, [[1]](diffhunk://#diff-4f9c15fb8bab9b0b7fab6bbd3ed60dc99409bf687a574a0f77b8156473aef95fR2592-R2729) [[2]](diffhunk://#diff-4f9c15fb8bab9b0b7fab6bbd3ed60dc99409bf687a574a0f77b8156473aef95fR755)
* Added a new appendix with diagrams and example JSON to illustrate the expected scene graph and asset hierarchy when using external assets. (`Specification.adoc`, [specification/2.1/Specification.adocR3516-R3644](diffhunk://#diff-4f9c15fb8bab9b0b7fab6bbd3ed60dc99409bf687a574a0f77b8156473aef95fR3516-R3644))

### DIFF
--- a/specification/2.1/Specification.adoc
+++ b/specification/2.1/Specification.adoc
@@ -752,7 +752,7 @@ The node hierarchy is defined using a node's `children` property, as in the foll
 
 The node named `Car` has four children. Each of those nodes could in turn have its own children, creating a hierarchy of nodes.
 
-Nodes **MAY** reference external assets. Nodes which contain an external reference **SHOULD NOT** include a camera or mesh. After loading the external asset file, implementations **SHOULD** connect the external asset to the parent file's node hierarchy by attaching all root nodes in the default scene as children.
+Nodes **MAY** reference external assets. After loading the external asset file, implementations **SHOULD** connect the external asset to the parent file's node hierarchy by attaching all root nodes in the default scene as children.
 
 [[transformations]]
 === Transformations
--- a/specification/2.1/Specification.adoc
+++ b/specification/2.1/Specification.adoc
@@ -752,6 +752,7 @@ The node hierarchy is defined using a node's `children` property, as in the foll
 
 The node named `Car` has four children. Each of those nodes could in turn have its own children, creating a hierarchy of nodes.
 
+Nodes **MAY** reference external assets. Nodes which contain an external reference **SHOULD NOT** include a camera or mesh. After loading the external asset file, implementations **SHOULD** connect the external asset to the parent file's node hierarchy by attaching all root nodes in the default scene as children.
 
 [[transformations]]
 === Transformations
@@ -1083,7 +1084,7 @@ On top of that, a sparse accessor includes a `sparse` JSON object describing the
 
 - `count`: number of displaced elements. This number **MUST NOT** be greater than the number of the base accessor elements.
 
-- `indices`: object describing the location and the component type of indices of values to be replaced. The indices **MUST** form a strictly increasing sequence. The indices **MUST NOT** be greater than or equal to the number of the base accessor elements. 
+- `indices`: object describing the location and the component type of indices of values to be replaced. The indices **MUST** form a strictly increasing sequence. The indices **MUST NOT** be greater than or equal to the number of the base accessor elements.
 
 - `values`: object describing the location of displaced elements corresponding to the indices referred from the `indices`.
 
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
 
 Skinned animation is achieved by animating the joints in the skin's joint hierarchy.
 
+[[file-references]]
+== Unified file references
+
+Files **MAY** be added to a top-level `files` array. The file's data **MAY** be stored externally and referenced by a `uri` or stored as binary in a buffer and referenced by `bufferView` index.
+
+Each file object **MUST** define a `mimeType` property, which may be any media type that matches the file's contents. Cyclical references between files are strictly prohibited.
+
+Files with a `mimeType` of `model/gltf+json` or `model/gltf-binary` **MAY** define a list of `aliases` that can be used to match and redirect a `uri` inside of the referred file. Any `uri` in the referenced file that matches the value `alias` exactly **SHOULD** use the associated `file` instead. This allows for efficient packing of resources shared across several referred files, or overriding inner data. Aliases are not inherited from the parent asset. Extensions may add additional mime types that can define `aliases`.
+
+[NOTE]
+.Implementation Note
+====
+If the same `uri` is used within multiple file references, implementations should not assume that the files are identical without validation.
+====
+
+[source,json]
+----
+{
+    "files": [
+        {
+            "uri": "path-to-a.gltf",
+            "mimeType": "model/gltf+json",
+            "aliases": [
+                {
+                    "alias": "path-to-a.bin",
+                    "file": 1
+                }
+            ]
+        },
+        {
+            "uri": "path-to-a.bin",
+            "mimeType": "application/gltf-buffer"
+        },
+        {
+            "uri": "https://example.org/path-to-a.glb",
+            "mimeType": "model/gltf-binary"
+        },
+        {
+            "bufferView": 0,
+            "mimeType": "model/gltf-binary"
+        }
+    ]
+}
+----
+
+[[external-assets]]
+== Referencing external assets
+
+glTF 2.1 defines the ability to compose multiple glTF files together, by referencing external glTF files from another glTF file. Any glTF file which references another glTF file as an external asset is called a complex scene. Complex scenes are defined using multiple components: file references to define where the data is stored and external asset definitions that define the properties of the asset node.
+
+External glTF assets **MAY** be added to a top-level `externalAssets` array. The external asset **MUST** contain a `file` property that refers to an object in the `files` array by index. Multiple objects in the `externalAssets` array may refer to the same `file`. All external assets **MUST** refer to files with the `"model/gltf-binary"` or `"model/gltf+json"` media types.
+
+[source,json]
+----
+{
+    "files": [
+        {
+            "uri": "path-to-a.gltf",
+            "mimeType": "model/gltf+json"
+        },
+        {
+            "uri": "path-to-a.bin",
+            "mimeType": "application/gltf-buffer"
+        },
+        {
+            "uri": "path-to-b.glb",
+            "mimeType": "model/gltf-binary"
+        },
+        {
+            "bufferView": 0
+            "mimeType": "model/gltf-binary"
+        }
+    ],
+    "externalAssets": [
+        {
+            "file": 0
+        },
+        {
+            "file": 2
+        },
+        {
+            "file": 3
+        }
+    ],
+}
+----
+
+[NOTE]
+.Implementation Note
+====
+When an embedded external asset is embedded via a <<data-uri,Data URI>> or a `bufferView`, any relative URIs it contains should resolve against the same filesystem root as the parent asset, unless the URI is remapped by an alias. _E.g._, If asset `A` embeds external asset `B`, and `B` references textures at the relative path `./textures/**`, those textures resolve relative to the location of `A`, making them accessible to both `A` and `B`.
+====
+
+Nodes *MAY* contain an `externalAsset` property that refers to an external glTF asset in the `externalAssets` array by index.
+
+[NOTE]
+.Implementation Note
+====
+When the same external asset is referenced by multiple nodes, implementations should load independent copies of the external asset. Each copy of the external asset should copy any glTF JSON and hierarchy, but may share buffer data between the copies. This balances compatibility with existing features and extensions while also minimizing the amount of copied data.
+====
+
+Once loaded, the root nodes defined by the external asset's default scene and those root nodes full node hierarchies *SHOULD* be treated as being attached as a child to the node containing the `externalAsset` reference.
+
+[source,json]
+----
+{
+    "nodes": [
+        {
+            "name": "RootNode",
+            "translation": [0, 0, 0],
+            "children": [1, 2]
+        },
+        {
+            "translation": [2, 0, 20],
+            "externalAsset": 0
+        },
+        {
+            "translation": [0, 0, 0],
+            "children": [3],
+            "externalAsset": 1
+        },
+        {
+            "translation": [0, 1, 1],
+            "externalAsset": 2
+        }
+    ]
+}
+----
+
+This instantiates the external asset as this node in the scene hierarchy, such that the root of the file becomes this node. The root nodes in the default `scene` of the nested glTF become children of this external asset node. Assets that have no `scenes` array or lack a default `scene` are valid but have nothing to attach to the node hierarchy and should be treated as such. Extensions or future spec versions **MAY** alter this behavior.
+
+[NOTE]
+.Implementation Note
+====
+Nodes may contain both an `externalAsset` and other `children`. For consistency between implementations, it is recommended that the root node(s) of the external asset are attached after the other children of the node.
+====
+
+See <<appendix-d-external-asset-diagrams, Appendix D>> for detailed diagrams outlining a sample structure and the expected scene graph.
 
 [[specifying-extensions]]
 == Specifying Extensions
@@ -3374,3 +3513,132 @@ This can be achieved by ensuring that stem:[v_k != -v_{k+1}] for all keyframes.
 ====
 
 The first in-tangent stem:[a_1] and last out-tangent stem:[b_n] **SHOULD** be zeros as they are not used in the spline calculations.
+
+[appendix]
+[[appendix-d-external-asset-diagrams]]
+= External Asset Scene Graph Diagrams
+
+== Overview
+
+This section provides informative example JSON and diagrams of External Assets to demonstrate the expected node hierarchy from implementations. For full normative details, see the sections on <<nodes-and-hierarchy>>, <<file-references>>, and <<external-assets>>.
+
+The following three sections represent the same asset and structure. The first two sections are diagrams to visually demonstrate hierarchy, and the third section is the equivalent glTF JSON for the assets represented within the diagrams.
+
+== Expected Scene Graph
+
+.Expected Scene Graph
+image::figures/external-assets-expected-scene-graph.svg[pdfwidth=100%,align=left]
+
+== glTF Structure & Hierarchy
+
+.glTF Structure & Hierarchy
+image::figures/external-assets-hierarchy.svg[pdfwidth=100%,align=left]
+
+== JSON
+
+=== Asset A
+
+[source,json]
+----
+{
+    "asset": {
+        "version": "2.1"
+    },
+    "scene": 0,
+    "scenes": [ { "nodes": [0] } ],
+    "files": [
+        {
+            "name": "File Ref. to Asset B",
+            "uri": "asset-b.glb",
+            "mimeType": "model/gltf-binary"
+        },
+        {
+            "name": "File Ref. to Asset C",
+            "uri": "asset-c.glb",
+            "mimeType": "model/gltf-binary"
+        },
+    ],
+    "externalAssets": [
+        {
+            "name": "External Asset 1",
+            "file": 0
+        },
+        {
+            "name": "External Asset 2",
+            "file": 1
+        }
+    ],
+    "nodes": [
+        {
+            "name": "Root Node A1",
+            "children": [ 1, 2, 3 ]
+        },
+        {
+            "name": "Node A1",
+            "externalAsset": 0
+        },
+        {
+            "name": "Node A2",
+            "externalAsset": 0
+        },
+        {
+            "name": "Node A3",
+            "externalAsset": 1
+        }
+    ]
+}
+----
+
+=== Asset B
+
+[source,json]
+----
+{
+    "asset": {
+        "version": "2.1"
+    },
+    "scene": 0,
+    "scenes": [ { "nodes": [0] } ],
+    "nodes": [
+        {
+            "name": "Root Node B1",
+            "children": [ 1, 2 ]
+        },
+        {
+            "name": "Node B1",
+        },
+        {
+            "name": "Node B2",
+        }
+    ]
+}
+----
+
+=== Asset C
+
+[source,json]
+----
+{
+    "asset": {
+        "version": "2.1"
+    },
+    "scene": 0,
+    "scenes": [ { "nodes": [0, 1] } ],
+    "nodes": [
+        {
+            "name": "Root Node C1",
+            "children": [ 2 ]
+        },
+        {
+            "name": "Root Node C2",
+            "children": [ 3 ]
+        },
+        {
+            "name": "Node C1",
+        },
+        {
+            "name": "Node C2",
+        }
+    ]
+}
+----
--- a/specification/2.1/Specification.adoc
+++ b/specification/2.1/Specification.adoc
@@ -537,7 +537,7 @@ This is typically achieved with algorithms like Grisu2 used by common JSON libra
 [[uris]]
 == URIs
 
-glTF assets use <<uri,URIs>> or <<iri,IRIs>> to reference buffers and image resources. Assets **MAY** contain at least these two URI types:
+glTF assets use <<uri,URIs>> or <<iri,IRIs>> to reference buffers, files, and image resources. Assets **MAY** contain at least these two URI types:
 
 - **Data URIs** that embed binary resources in the glTF JSON as defined by the <<data-uri,RFC 2397>>. The Data URI's `mediatype` field **MUST** match the encoded content.
 +
--- a/specification/2.1/Specification.adoc
+++ b/specification/2.1/Specification.adoc
@@ -752,7 +752,7 @@ The node hierarchy is defined using a node's `children` property, as in the foll
 
 The node named `Car` has four children. Each of those nodes could in turn have its own children, creating a hierarchy of nodes.
 
-Nodes **MAY** reference external assets. After loading the external asset file, implementations **SHOULD** connect the external asset to the parent file's node hierarchy by attaching all root nodes in the default scene as children. See the section on <<external-assets,referencing External Assets>> for further details.
+Nodes **MAY** reference external assets. After loading the external asset file, the root nodes in the external asset's default scene are treated as children of the referencing node. See the section on <<external-assets,referencing External Assets>> for further details.
 
 [[transformations]]
 === Transformations
@@ -2687,10 +2687,10 @@ Nodes *MAY* contain an `externalAsset` property that refers to an external glTF 
 [NOTE]
 .Implementation Note
 ====
-When the same external asset is referenced by multiple nodes, implementations should load independent copies of the external asset. Each copy of the external asset should copy any glTF JSON and hierarchy, but may share buffer data between the copies. This balances compatibility with existing features and extensions while also minimizing the amount of copied data.
+When the same external asset is referenced by multiple nodes, it is recommended that all related JSON objects except for `buffer` objects are cloned. The `buffer` reference may be retained to allow sharing binary data between external asset instances. This maximizes compatibility with existing extensions while also minimizing the amount of duplicated binary data. Future specification versions or extensions may introduce mechanisms to allow better sharing of object definitions across external asset instances.
 ====
 
-Once loaded, the root nodes defined by the external asset's default scene and those root nodes full node hierarchies *SHOULD* be treated as being attached as a child to the node containing the `externalAsset` reference.
+Once loaded, the root nodes defined by the external asset's default scene and their node hierarchies are treated as being attached as children to the node containing the `externalAsset` reference.
 
 [source,json]
 ----
--- a/specification/2.1/Specification.adoc
+++ b/specification/2.1/Specification.adoc
@@ -752,7 +752,7 @@ The node hierarchy is defined using a node's `children` property, as in the foll
 
 The node named `Car` has four children. Each of those nodes could in turn have its own children, creating a hierarchy of nodes.
 
-Nodes **MAY** reference external assets. After loading the external asset file, implementations **SHOULD** connect the external asset to the parent file's node hierarchy by attaching all root nodes in the default scene as children.
+Nodes **MAY** reference external assets. After loading the external asset file, implementations **SHOULD** connect the external asset to the parent file's node hierarchy by attaching all root nodes in the default scene as children. See the section on <<external-assets,referencing External Assets>> for further details.
 
 [[transformations]]
 === Transformations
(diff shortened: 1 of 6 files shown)


specification/2.1/schema/externalAsset.schema.json
```diff
@@ -0,0 +1,16 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "externalAsset.schema.json",
+    "title": "External Asset",
+    "type": "object",
+    "description": "An external asset that references a file in the `files` array and may be instantiated by glTF scene nodes.",
```
EricGriffith: I think 'scene node' is ambiguous. Maybe 'node objects' is clearer?

specification/2.1/schema/file.schema.json
```diff
@@ -0,0 +1,52 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "file.schema.json",
+    "title": "File Reference to External File",
+    "type": "object",
+    "description": "A file reference to an external file. Files **MAY** be referenced by an URI (or IRI) or a buffer view index.",
```
EricGriffith: file reference -> reference

specification/2.1/schema/file.schema.json
```diff
@@ -0,0 +1,52 @@
...
+                {
+                    "type": "string"
+                }
+            ]
+        },
+        "uri": {
+            "type": "string",
+            "description": "The URI (or IRI) of the external file.",
+            "format": "iri-reference",
+            "gltf_detailedDescription": "The URI (or IRI) of the external file.  Relative paths are relative to the current glTF asset.  Instead of referencing an external file, this field **MAY** contain a `data:`-URI. This field **MUST NOT** be defined when `bufferView` is defined.",
```
EricGriffith, line 41: Do we want to emphasize that data URIs should use base64 encoding?

Also, double spaces in "glTF asset.  Instead of"

specification/2.1/schema/file.schema.json
```diff
@@ -0,0 +1,52 @@
...
+            "gltf_uriType": "external-file"
```
EricGriffith, line 42: Do we actually use this gltf_uriType in any meaningful way?

specification/2.1/schema/glTF.schema.json
```diff
@@ -72,6 +72,24 @@
...
             "gltf_detailedDescription": "An array of cameras.  A camera defines a projection matrix."
         },
+        "externalAssets": {
+            "type": "array",
+            "description": "An array of external assets.",
+            "items": {
+                "$ref": "externalAsset.schema.json"
+            },
+            "minItems": 1,
+            "gltf_detailedDescription": "An array of external assets.  An external asset references a file in the `files` array and may be instantiated by glTF scene nodes."
```
EricGriffith, line 82: same comment about 'scene node'

specification/2.1/schema/node.schema.json
```diff
@@ -37,7 +37,11 @@
         },
         "mesh": {
             "allOf": [ { "$ref": "glTFid.schema.json" } ],
-            "description": "The index of the mesh in this node."
+            "description": "The index of the mesh in this node. Cannot be used together with `externalAsset`."
```
EricGriffith: I thought we agreed to not impose this restriction. If we are going to impose a restriction, I think we need to stick to MUST / MAY / SHOULD style language.

specification/2.1/Specification.adoc
```diff
@@ -752,6 +752,7 @@ The node hierarchy is defined using a node's `children` property, as in the foll
 
 The node named `Car` has four children. Each of those nodes could in turn have its own children, creating a hierarchy of nodes.
 
+Nodes **MAY** reference external assets. After loading the external asset file, implementations **SHOULD** connect the external asset to the parent file's node hierarchy by attaching all root nodes in the default scene as children.
```
EricGriffith: We need more unambiguous language than this. The [KHR_node_visibility](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_node_visibility) extension has this wording:

> In other words, a node is visible if and only if its own visible property is true and all its parents are visible.

With this 'SHOULD' wording it is ambiguous whether marking a node that instantiates an external asset as hidden would actually hide the nodes of the instantiated asset.

specification/2.1/Specification.adoc
```diff
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
 
 Skinned animation is achieved by animating the joints in the skin's joint hierarchy.
 
+[[file-references]]
+== Unified file references
```
EricGriffith: This is the only place we mention 'unified' in relation to file references in the schemas / specs. I suggest we drop the unified and just say 'File References'

specification/2.1/Specification.adoc
```diff
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
...
+
+Files **MAY** be added to a top-level `files` array. The file's data **MAY** be stored externally and referenced by a `uri` or stored as binary in a buffer and referenced by `bufferView` index.
```
EricGriffith: Files **MAY** -> File References **MAY**.

We also need to allow for data URIs in this description

specification/2.1/Specification.adoc
```diff
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
...
+
+Each file object **MUST** define a `mimeType` property, which may be any media type that matches the file's contents. Cyclical references between files are strictly prohibited.
```
EricGriffith: We should be consistent about always using file contents vs file data.

I also don't think we need the remark about cyclical references here. That's really for external assets where external assets must not create a cycle.

specification/2.1/Specification.adoc
```diff
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
...
+
+Files with a `mimeType` of `model/gltf+json` or `model/gltf-binary` **MAY** define a list of `aliases` that can be used to match and redirect a `uri` inside of the referred file. Any `uri` in the referenced file that matches the value `alias` exactly **SHOULD** use the associated `file` instead. This allows for efficient packing of resources shared across several referred files, or overriding inner data. Aliases are not inherited from the parent asset. Extensions may add additional mime types that can define `aliases`.
```
EricGriffith: referred file -> referenced file.

Also, I think the behavior of aliases should be described in (and scoped to) externalAssets.

specification/2.1/Specification.adoc
```diff
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
...
+            "mimeType": "model/gltf-binary"
+        }
+    ]
+}
+----
+
+[[external-assets]]
+== Referencing external assets
+
+glTF 2.1 defines the ability to compose multiple glTF files together, by referencing external glTF files from another glTF file. Any glTF file which references another glTF file as an external asset is called a complex scene. Complex scenes are defined using multiple components: file references to define where the data is stored and external asset definitions that define the properties of the asset node.
```
EricGriffith: I propose this instead to be a bit more future proof (specifically, I don't want to say that two glTF files referencing each other is always and forever a complex scene):

> glTF 2.1 defines the ability to compose complex scenes. A complex scene is when one glTF file includes a scene from another glTF file by referencing that glTF as an external asset. Complex scenes are defined using multiple components: file references to define where the data is stored and external asset definitions that define the scene from the referenced glTF file to include.

specification/2.1/Specification.adoc
```diff
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
...
+====
+When an embedded external asset is embedded via a <<data-uri,Data URI>> or a `bufferView`, any relative URIs it contains should resolve against the same filesystem root as the parent asset, unless the URI is remapped by an alias. _E.g._, If asset `A` embeds external asset `B`, and `B` references textures at the relative path `./textures/**`, those textures resolve relative to the location of `A`, making them accessible to both `A` and `B`.
+====
+
+Nodes *MAY* contain an `externalAsset` property that refers to an external glTF asset in the `externalAssets` array by index.
+
+[NOTE]
+.Implementation Note
+====
+When the same external asset is referenced by multiple nodes, implementations should load independent copies of the external asset. Each copy of the external asset should copy any glTF JSON and hierarchy, but may share buffer data between the copies. This balances compatibility with existing features and extensions while also minimizing the amount of copied data.
```
EricGriffith, line 2690: I thought we had agreed that we needed to support both shared and separate instances. That is, if you reference a glTF with some animation (e.g., "Running"), we wanted a way to support both of these scenarios:
 - Multiple instances of this can be in the scene, each with independent state for its Running animation (i.e., all the instances are animated independently)
 - Multiple instances of this can be in the scene, all of which share the same state for the Running animation (i.e., all the instances are animated in sync)

I think in the calls, we had agreed that this should be done via the files property: for each instance you want to have independent state, you should have a separate entry in the files array (and therefore also a separate entry in the externalAsset array). However, as I think about it more, I think it should be sufficient to just say that each entry in the externalAsset array should have its own state that is shared by any node that references it:
```json
{
    "files": [
        {
            "uri": "path-to-a.gltf",
            "mimeType": "model/gltf+json"
        },
    ],
    "externalAssets": [
        {
            "name": "State Group 1",
            "file": 0
        },
        {
            "name": "State Group 2",
            "file": 0
        },
        {
            "name": "State Group 3",
            "file": 0
        }
    ],
    "nodes": [
        {
            "children": [1, 2, 3, 4, 5, 6]
        },
        {
            "name": "State Group 1, Instance 1",
            "externalAsset": 0
        },
        {
            "name": "State Group 1, Instance 2",
            "externalAsset": 0
        },
        {
            "name": "State Group 2, Instance 1",
            "externalAsset": 1
        },
        {
            "name": "State Group 3, Instance 1",
            "externalAsset": 2
        },
        {
            "name": "State Group 3, Instance 2",
            "externalAsset": 2
        },
        {
            "name": "State Group 3, Instance 3",
            "externalAsset": 2
        },
    ],
}
```

specification/2.1/Specification.adoc
```diff
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
...
+        },
+        {
+            "translation": [0, 1, 1],
+            "externalAsset": 2
+        }
+    ]
+}
+----
+
+This instantiates the external asset as this node in the scene hierarchy, such that the root of the file becomes this node. The root nodes in the default `scene` of the nested glTF become children of this external asset node. Assets that have no `scenes` array or lack a default `scene` are valid but have nothing to attach to the node hierarchy and should be treated as such. Extensions or future spec versions **MAY** alter this behavior.
```
EricGriffith: Instead of:

> This instantiates the external asset as this node in the scene hierarchy, such that the root of the file becomes this node. The root nodes ...

I suggest:

> This instantiates the external asset as a child in the scene hierarchy. The root nodes ...

specification/2.1/Specification.adoc
```diff
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
...
+
+Nodes *MAY* contain an `externalAsset` property that refers to an external glTF asset in the `externalAssets` array by index.
+
+[NOTE]
+.Implementation Note
+====
+When the same external asset is referenced by multiple nodes, implementations should load independent copies of the external asset. Each copy of the external asset should copy any glTF JSON and hierarchy, but may share buffer data between the copies. This balances compatibility with existing features and extensions while also minimizing the amount of copied data.
+====
+
+Once loaded, the root nodes defined by the external asset's default scene and those root nodes full node hierarchies *SHOULD* be treated as being attached as a child to the node containing the `externalAsset` reference.
```
EricGriffith, line 2693: Same comment as above in terms of needing something more unambiguous than SHOULD.

specification/2.1/schema/file.alias.schema.json
```diff
@@ -0,0 +1,19 @@
...
+        "alias": {
+            "type": "string",
+            "description": "String alias matched against an inner path."
+        },
+        "file": {
+            "$ref": "glTFid.schema.json",
+            "description": "The index of the file in the files array."
+        }
+    },
+    "required": [ "alias", "file" ]
```
EricGriffith, line 18: Did you mean to make `alias` required?

specification/2.1/schema/file.schema.json
```diff
@@ -0,0 +1,52 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "file.schema.json",
+    "title": "File Reference to External File",
```
EricGriffith: File Reference to External File -> File Reference

EricGriffith (commented):
We also need to tidy up the description for Buffer in its JSON schema and in the Concepts part of the spec. The buffer schema currently says:

> A buffer points to binary geometry, animation, or skins.

The concepts part says:

> The buffer MAY contain any combination of geometry, animation, skins, and images

specification/2.1/schema/externalAsset.schema.json
weegeekps: Yeah, "scene node" is certainly ambiguous. What about just using "node"? Per the [spec terminology](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#introduction-technical-terminology), a node is defined as:

> An object defining the hierarchy relations and the local transform of its content.

Something like this may work:

```suggestion
    "description": "An external asset that references a file in the `files` array and may be instantiated by any node.",
```

What do you think?

specification/2.1/schema/file.schema.json
weegeekps: I don't think that emphasis is required. We cover this in [§2.8](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#uris) of the specification and we don't emphasize this in any other existing schema.

specification/2.1/schema/file.schema.json
weegeekps: Honestly, I have no idea. My understanding is that this was added to help with validation, but I don't think anyone is actually using it, and a quick search on GitHub supports that assertion.

@lexaknyazev, do you think we need to define a `gltf_uriType` here or can we just drop it?

specification/2.1/Specification.adoc
weegeekps: One of the drafts I presented earlier on in the DTSG used "MUST" instead of "SHOULD", but the feedback given at that time was that it was too limiting.

If an implementation chooses to not connect the node hierarchy of the parent, then the link between the parent file's node and the external asset's node(s) is broken, and changes to the parent node would not apply. That would be the expected behavior.

Given this is the core specification, we need to use a normative keyword. Either "SHOULD" or "MUST". I can add some clarity in either case, but we need to make a decision on this.

cc: @lexaknyazev, @aaronfranke in case either of you have a strong opinion on this. I believe it was one of you two who pushed back on the "MUST" language here.

specification/2.1/schema/file.alias.schema.json
aaronfranke: Yes, because the point of objects in this array is to provide a mapping. Only one of the fields would be useless.

specification/2.1/Specification.adoc
weegeekps: Initially, yes, but then at the DTSG call on the 7/7 we discussed more details and agreed that a copy is the _only_ consistent path without us diving deep into GPU instancing. The group agreed that we'll look at a comprehensive GPU instancing extension after glTF 2.1. This decision was driven partially by initial feedback we were given from early implementers.

specification/2.1/Specification.adoc
aaronfranke: I have a strong opinion on this. I want to see a `KHR_single_root` extension, which would require that the root node of the external file is directly used as the parent file's node, rather than added as a child. Even without `KHR_single_root`, the same interaction with glTF 2.1 will exist with `GODOT_single_root`.

As far as I know, it is not valid for an optional extension (just in `"extensionsUsed"`) to override core spec "MUST" language, and it would be undesired to require all files with `KHR_single_root` to mark it as a required extension.

specification/2.1/Specification.adoc
aaronfranke: This is not necessarily a one-or-the-other situation. The specifics of the behavior here would need to be defined in future spec versions or extensions, and such extensions can choose whichever behavior is desired.

With your example file @EricGriffith, there is a difference between "playing an animation on node 1" which would only affect that instance, and "playing an animation on externalAsset 0" would show up on the instances on nodes 1 and 2. Both of these are useful, which is effectively the whole reason why the `externalAssets` array exists at all (otherwise the nodes could just have `"file": 0` or something).

specification/2.1/Specification.adoc
aaronfranke: > Also, I think the behavior of aliases should be described in (and scoped to) externalAssets.

No, I don't like that. Such a design would prevent glTF from doing alias remaps when referencing a non-scene file, including non-scene glTF files, and non-glTF files.

specification/2.1/Specification.adoc
aaronfranke: Sure. For context, "Unified file references" is how the feature is described in the explainers, but that is not necessarily the right language for glTF itself. Moving forward, glTF itself can just have the concept of file references. The term "Unified file references" is mostly useful in the near future to explain the difference between this system and the previous approach of fragmented file references (such as each extension making its own array of files).

specification/2.1/Specification.adoc
weegeekps: Right, thank you for refreshing my memory.

It is preferable to use "SHOULD" over "MUST" here to allow extensions to override this without making the extension "required".

I'm adding a link to the External Assets section as that clearly defines and reinforces this recommendation further. I'm not really sure that does enough, and I'm not sure how else to clarify this.

@javagl or @lexaknyazev, any ideas?

specification/2.1/Specification.adoc
weegeekps: I don't believe we need to do this. There is a blanket statement regarding binary resources in [§2.4](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#gltf-basics) that covers this.

specification/2.1/Specification.adoc
weegeekps: Moved this to the external assets section.

specification/2.1/Specification.adoc
weegeekps: Yeah, we can't rescope this to external assets. Leaving this as part of the "files" array makes it so extensions that may do different things with glTF that are outside of the scope of external assets, such as a material library, is still allowed to do remapping through aliases in a consistent manner.

specification/2.1/Specification.adoc
weegeekps: Good suggestion. I made a few additional tweaks.

specification/2.1/Specification.adoc
weegeekps: I think this section as a whole makes it sufficiently clear that this is a load bearing piece of the specification for external assets. [RFC 2119's definition of "SHOULD"](https://datatracker.ietf.org/doc/html/rfc2119#section-3) makes it clear that this is meant to be read as a recommendation that you should only ignore if you fully understand the ramifications.

I'm open to additional suggestions for improvement though.

weegeekps (commented):
Pushed some changes.

specification/2.1/Specification.adoc
EricGriffith: @aaronfranke I'm aligned with you but it sounds we're not aligned with @weegeekps.

specification/2.1/Specification.adoc
EricGriffith: I think we need to separate what a given engine's internal representation of the scene graph is from what the logical structure of the scene graph is as exposed to various glTF extensions like interactivity / node visibility / etc.

I suggest we say something like:

> Regardless of how an implementation chooses to internally represent the complex scene, implementations MUST treat the root nodes of the external asset's default scene as children of the node.

wdyt?

specification/2.1/schema/file.schema.json
EricGriffith: I can live with out this.

EricGriffith: @weegeekps btw I think I'm missing some permissions needed to resolve threads. Are you able to grant them?

EricGriffith:
Should this also update the object model?

specification/2.1/Specification.adoc
EricGriffith: We can't leave it at the files level description. We have no idea what sort of files this may be used for in the future and what sort of URIs they may include in them.

specification/2.1/Specification.adoc
weegeekps: We can't support shared state across external assets right now because doing so would break a lot of assumptions that are based on this rule in glTF 2.x ([§3.5.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#nodes-and-hierarchy)):

> The node hierarchy MUST be a set of disjoint strict trees. That is node hierarchy MUST NOT contain cycles and each node MUST have zero or one parent node.

Sharing node state breaks several core concepts of the node hierarchy that existing features and extensions depend on for implementation. Animations and skin poses are actually the most visible victims, but it also breaks punctual lights, material variants, interactivity, and more.

External assets unfold into children of the referencing parent node, so if we don't strongly recommend a copy then we've broken compatibility with many features and forward-compatibility as well.

The proper way for us to do what you're asking is to release glTF 2.1 (with UIDs) and then create an extension that allows sharing state between disjoint node sub-trees. My guess is that the composition extension we've discussed about early on in the DTSG calls will be the best place to solve this problem.

That's why on the 7/7 DTSG call we agreed that a copy, for now, is the only consistent path. We're still allowing for data instancing since buffers are more-or-less immutable, but any sort of potentially mutable state in the JSON has to be copied.

specification/2.1/Specification.adoc
weegeekps: I updated this based on our conversation from earlier today. Let me know if there are any tweaks you'd like to see. I tried to go the route that @javagl suggested and described the expected behavior rather than normatively defining the implementation.

specification/2.1/Specification.adoc
```diff
@@ -2588,6 +2589,144 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
...
+====
+When an embedded external asset is embedded via a <<data-uri,Data URI>> or a `bufferView`, any relative URIs it contains should resolve against the same filesystem root as the parent asset, unless the URI is remapped by an alias. _E.g._, If asset `A` embeds external asset `B`, and `B` references textures at the relative path `./textures/**`, those textures resolve relative to the location of `A`, making them accessible to both `A` and `B`.
+====
+
+Nodes *MAY* contain an `externalAsset` property that refers to an external glTF asset in the `externalAssets` array by index.
+
+[NOTE]
+.Implementation Note
+====
+When the same external asset is referenced by multiple nodes, it is recommended that all related JSON objects except for `buffer` objects are cloned. The `buffer` reference may be retained to allow sharing binary data between external asset instances. This maximizes compatibility with existing extensions while also minimizing the amount of duplicated binary data. Future specification versions or extensions may introduce mechanisms to allow better sharing of object definitions across external asset instances.
```
weegeekps, line 2690: @EricGriffith, @lexaknyazev, @aaronfranke, this is what I've come up with to improve this implementation note. I removed any language that could be taken as normative (since this is informative) and then made it clear that this is a recommendation. I also added a statement at the end that there may be future mechanisms to define how we can share object definitions across the external asset instances.

One thing that is _slightly_ different than what we discussed is that I did state that you don't need to clone the `buffer` objects. I went back through my notes, and this shouldn't be an issue. We already allow multiple `bufferViews` to point to a single `buffer` today, and this is already used by various creation tools and loaders in production.

Let me know what you think.

weegeekps (commented):
Ope... Started a review instead of pressing the comment button last week.

specification/2.1/Specification.adoc
javagl: It sounds like it was related to https://github.com/KhronosGroup/glTF/issues/2446 , and specifically, the point from my first comment there: When someone loads a glTF asset, and then blatantly does some
`gltf.meshes[0].primitives[0].attributes["POSITION"].set(0, 1234.567);`

The question is: Will this affect meshes that (via this long chain of indirections) refer to the exact same bytes of the `buffer`? 

On the level of core glTF 2.0, one could handwavingly leave this to the _intent_ of the author, by saying: If you do _not_ want such changes to affect other primitives, then duplicate your buffer data. But for external file references, it's not that clear.

This _might_ not have to hold up the changes here. It is hardly relevant for pure _renderers_, and it is not even clear whether or not it should be _specified_ at all. But for libraries that allow modifications of glTF (i.e. basically _all_ authoring applications), this is a crucial piece of information.

lexaknyazev: > When someone loads a glTF asset, and then blatantly does some
`gltf.meshes[0].primitives[0].attributes["POSITION"].set(0, 1234.567);`

I'd say that operation cannot exist because an accessor is conceptually a "provider", it cannot write anything back because that write destination may not even exist in cases like sparse and/or compressed data.

javagl: I'd rather consider an accessor (that is associated with a buffer view) as a "(typed) view on that buffer view" (aka `DataView`), and in doubt, one could ask about
`gltf.meshes[0].primitives[0].attributes["POSITION"].bufferView!.set(0, 1234.567);`
`                                                    ^^^^^^^^^^^                  `

To be clear: The suggested behavior is that 1. 'writing to an accessor' is either  completely ruled out, or 2. **iff** someone writes into an accessor (or a buffer view) that is sharing the bottommost `buffer` with another one, then the write operation **will** affect that other object. I think that achieving this behavior in libraries like glTF-Transform is difficult or impossible (given the lack of a 'buffer view' representation to begin with), but no strong opinion from my side for now.

specification/2.1/schema/file.schema.json
EricGriffith: This old commit message explains the intent behind gltf_uriTYpe:
https://github.com/KhronosGroup/glTF/commit/a382e2f73f79a6d16ed5a083a8c2a3edbd4d680d

> I used strings from the data uris to identify the type of the contents.
EX: "data:text/plain;base64, -> text.  We can then use this while
parsing the schema to select an appropriate data type to attempt to
parse the contents of the URI as. (C# EX: byte[] for application, string
for text, and Bitmap for image)

So it looks like the intent is to use the prefix before the `/` in a mime type here. Given this is a generic file, we should either use "application" or we should omit it entirely because it could be purpose specific like "image" if we eventually allow putting image files here, etc.

Not treating this as blocking for this PR tho.

specification/2.1/Specification.adoc
EricGriffith: @javagl This gets to the tension between glTF just being a delivery file format vs. it being something more (via Object Model / Interactivity). Given that there's no currently defined way (in the object model or via interactivity) to write data to an accessor, what happens to the buffer data after a glTF is ingested into a runtime becomes an implementation question for that runtime. If we want to be prescriptive about what what runtime behavior should be, that belongs in a new extension. The same holds for the JSON object properties discussed here.

EricGriffith: Continuing the discussion on this topic in the DTSG meetings.

EricGriffith (approved):

specification/2.1/Specification.adoc
javagl: Yes. The linked issue ( https://github.com/KhronosGroup/glTF/issues/2446 ) is related to the core spec, and already pretty fundamental. For the case of external assets, the same question receives a whole new _dimension_. But it's conceptually not so different from the one that we had in the core spec already.

To be a bit more specific: 

I created a test asset that defines two accessors with two buffer views that point to the same section of a buffer. 

With glTF-Transform, modifying the accessor data via
```
const accessor = document.getRoot().listAccessors()[1];
accessor.setElement(3, [1.0, 0.5, 0.0]);
```
will only affect that one accessor (because upon writing, the accessors are assembled into new bufferView/buffer structures). 

With JglTF, modifying the accessor data via 
```
AccessorFloatData accessorData = (AccessorFloatData)gltfModel.getAccessorModels().get(1).getAccessorData();
accessorData.set(3 * 3 + 1, 0.5f);
```
and writing this out (_without_ rebuilding the buffer structures!) will affect _both_ accessors, because they refer to the same buffer range, and that's what the data contains now. 

To be fair: The default behavior _is_ to rebuild the buffer structures. Creating buffer views that share a section of a buffer is unusual enough to not _directly_ be supported to begin with. But according to the core spec, there could be buffer views that share a section of a buffer, or accessors that share a section of a buffer view, and this implies some degrees of freedom for the behavior of authoring applications. (For example: Authoring applications might very well perform some magic "deduplication", see that two buffer views share some data, and let them therefore share some buffer section as a straightforward optimization...)

However, in that linked issue, the bottom line of my first comment was

> This is probably beyond what the glTF spec can cover. But for things like glTF-Transform, there will be an answer to this question, and regardless what the answer is: It might be different from what someone expects...

(I could have used the word "fragmentation" here, but we cannot specify _everything_...)

Issue #2588, opened 2026-05-27
[glTF 2.1] Shapes


akmorgan99:
# Shapes

glTF 2.1 adds a new `shapes` array to the top-level. This provides the same data as the pending `KHR_implicit_shapes` extension, which is now superseded by glTF 2.1.

The base glTF 2.1 specification includes `"box"`, `"capsule"`, `"cylinder"`, `"plane"`, and `"sphere"` shape types. Data for these shapes, such as size and radius, can be specified in sub-objects with keys matching the type name. Extensions may define additional shape types.

## Use cases

Shapes are useful for a variety of use cases, such as:

- Bounding volumes (core to glTF 2.1)
  - Useful for culling, deferred loading, volume restriction, semantic labeling of volumes, and so on.
- Physics colliders (not core)
  - Solid objects, triggers, bone colliders, and so on.
- Procedural mesh generation (not core)

## Example

> [!NOTE]
> Examples are illustrative and not exhaustive.

```json
{
    "shapes": [
        {
            "type": "box",
            "box": {
                "size": [2.5, 3.0, 2.0]
            }
        }
    ]
}
```

## Why this design

Ratifying and using the existing `KHR_implicit_shapes` extension was considered. However, there was a strong desire for bounding volumes to be a core glTF 2.1 feature, and it would be inappropriate for a core feature to depend on an extension. Additionally, shapes are broadly useful, so it makes sense to define them in the core specification, even if they were not used for bounding volumes.

javagl:
While drafting support for glTF 2.1 (and shapes in particular) in JglTF, one example that fell out of that: It is a glTF that defines all sorts of shapes. 

[example_2_1_ShapesAllWithGeometry-2026-09-08.zip](https://github.com/user-attachments/files/31977117/example_2_1_ShapesAllWithGeometry-2026-09-08.zip)

<sup>(EDIT: Note that it still says version 2.0, but that's a detail for now...)</sup>

It includes their respective visual representation as a mesh, which may be ... helpful, in some way, because it can be rendered:

<img width="484" height="262" alt="Image" src="https://github.com/user-attachments/assets/e07d6df5-a793-4e95-a597-730ce69661c3" />

(Note: The capsule mesh is _slighly_ wrong, because it is not a true convex hull - just two hemispheres connected with some "cone", but I tried to conceal that by using similar top- and bottom radii for now...)

[...]

Issue #2587, opened 2026-05-27
[glTF 2.1] Bounding Volumes


akmorgan99:
# Bounding Volumes and BVH

glTF 2.1 adds a new `"boundingVolume"` property to nodes. A bounding volume is a simple geometric shape, optionally transformed, that should completely enclose the node's content.

Multiple bounding volumes defined in a node hierarchy form a bounding volume hierarchy (BVH). Bounding volumes hierarchies should have spatial coherence, meaning that a node's bounding volume completely encloses the content of its descendants, and a node's content completely fits within the bounding volumes of its ancestors.

This bounding volume feature depends on the glTF 2.1 shapes feature.

## Use cases

- Bounding volume hierarchies can be used to accelerate frustum culling, occlusion culling, ray tracing, and so on.

- Bounding volumes may be used to defer loading of content based on application-specific criteria, such as the angular size on a player's camera, or any other metric.

- Bounding volumes may be used to restrict rendering, interactions, or anything else to a specific volume of space.

- Implementations may choose to use bounding volumes for any purpose.

The tightness or looseness of the bounding volume may be customized by the asset author depending on the use case. For example, a bounding volume used for a skinned mesh may wish to be large enough to contain the mesh in all reasonable deformation states, such as the possible poses of a character. Similarly, a bounding volume placed on a parent with multiple animated children may wish to be large enough to contain all of the children in all reasonable animation states. Alternatively, such a parent could simply not define a bounding volume, and dynamically rely on the bounding volumes of its children, if any.

## Example

> [!NOTE]
> Examples are illustrative and not exhaustive.

```json
{
    "nodes": [
        {
            "boundingVolume": {
                "shape": 0
            }
        }
    ],
    "shapes": [
        {
            "type": "box",
            "box": {
                "size": [2.5, 3.0, 2.0]
            }
        }
    ]
}
```

Bounding volumes are per-node, and can be placed on the same node as meshes, lights, cameras, or any other content, or on an empty node, even if it has no children.

## Why this design

Bounding volumes are considered broadly useful and important for large-scale scenes, so they are a core feature of glTF 2.1.

Bounding volumes use shapes with transforms, rather than being restricted to axis-aligned bounding boxes, to enable finer control over the bounding volumes.

Bounding volumes are done at the node level, rather than the mesh level, to allow bounding volumes to be defined for any content, not just meshes.

bjornblissing:
Could the specification clarify how a node's transform applies to its boundingVolume?

[...]

PR #2619, opened 2026-07-17
Draft changes for Shapes and Bounding Volumes in glTF 2.1

weegeekps:
This PR contains the draft changes for Shapes and Bounding Volumes in glTF 2.1. This PR currently points towards the `draft-2.1` branch and should be safe to merge to that branch.

These changes were authored by @lilleyse.

### Relevant Issues:

- https://github.com/KhronosGroup/glTF/issues/2588
- https://github.com/KhronosGroup/glTF/issues/2587

### Summary from Copilot:

**Major Additions:**

**1. Bounding Volumes for Nodes**
- Added the `boundingVolume` property to the `node` schema, allowing nodes to specify a local-space bounding volume for culling, loading, or other spatial operations. The bounding volume references a shape and can be transformed relative to the node. (F06053e9R1, [specification/2.1/schema/boundingVolume.schema.jsonR1-R47](diffhunk://#diff-c68e27eb4fadec3a1cd7aa046a3c67f0211ffbcd508217e77bcab7c415445606R1-R47))
- Documented how bounding volumes are defined, how they are used in hierarchies (BVH), and best practices for asset authors.

**2. Shape Definitions and Registry**
- Introduced a new top-level `shapes` array in the glTF schema, enabling authors to define reusable shape objects.
- Documented the available built-in shape types (`box`, `sphere`, `capsule`, `cylinder`, `plane`), their parameters, and requirements for validity.

**3. Shape and Bounding Volume Schemas**
- Added a general `shape.schema.json` describing the structure of a shape, including type and type-specific parameters.
- Added individual schemas for each built-in shape type: `box`, `sphere`, `capsule`, `cylinder`, and `plane`, each with their own required parameters and validation rules. [[1]](diffhunk://#diff-9ac16541a196a7c165028f0c8457da124d66062f2d3fc2358e8c1117004667f0R1-R21) [[2]](diffhunk://#diff-a5a84039dceca264843e0958f39aed0eb868c3374e8ba2039e8200aa4b831e87R1-R16) [[3]](diffhunk://#diff-d1f946d530abf4a783021bda0b7629b8dfe813b51f00348aa9c0f1880d7c43a6R1-R28) [[4]](diffhunk://#diff-d31afd6509a74a5ff9d52e462ab4a3dd3a8ef063c1a895ba685bd9f2c8401cfcR1-R28) [[5]](diffhunk://#diff-65fadc39fe3f12e000bc452d5038fb990f86f3cdf9be034d801db3b51668d2feR1-R20)
- Added a `boundingVolume.schema.json` describing how a bounding volume references a shape and specifies its transform.

**Summary of Most Important Changes:**

**Bounding Volumes**
* Added `boundingVolume` property to nodes, referencing a shape and specifying transform (rotation, scale, translation). (F06053e9R1, [specification/2.1/schema/boundingVolume.schema.jsonR1-R47](diffhunk://#diff-c68e27eb4fadec3a1cd7aa046a3c67f0211ffbcd508217e77bcab7c415445606R1-R47))
* Documented the usage, best practices, and implementation notes for bounding volumes and bounding volume hierarchies in the glTF spec.

**Shape System**
* Introduced a top-level `shapes` array in glTF, allowing reusable, parameterized shape definitions.
* Documented built-in shape types and their parameters, including validation requirements and extension mechanisms.

**Schema Additions**
* Added `shape.schema.json` and individual schemas for each built-in shape type (`box`, `sphere`, `capsule`, `cylinder`, `plane`) to validate shape definitions. [[1]](diffhunk://#diff-c173b788eae3a9c25a15e2a0b44711d2317b11b906977f5ff670ae7f6ffb4faaR1-R66) [[2]](diffhunk://#diff-9ac16541a196a7c165028f0c8457da124d66062f2d3fc2358e8c1117004667f0R1-R21) [[3]](diffhunk://#diff-a5a84039dceca264843e0958f39aed0eb868c3374e8ba2039e8200aa4b831e87R1-R16) [[4]](diffhunk://#diff-d1f946d530abf4a783021bda0b7629b8dfe813b51f00348aa9c0f1880d7c43a6R1-R28) [[5]](diffhunk://#diff-d31afd6509a74a5ff9d52e462ab4a3dd3a8ef063c1a895ba685bd9f2c8401cfcR1-R28) [[6]](diffhunk://#diff-65fadc39fe3f12e000bc452d5038fb990f86f3cdf9be034d801db3b51668d2feR1-R20)
* Added `boundingVolume.schema.json` to validate bounding volume definitions and their relationship to shapes.


specification/2.1/schema/shape.box.schema.json
```diff
@@ -0,0 +1,21 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "shape.box.schema.json",
+    "title": "glTF Box Shape",
```
EricGriffith: Do we need glTF in the title of all of these shapes?

specification/2.1/schema/shape.box.schema.json
```diff
@@ -0,0 +1,21 @@
...
+    "type": "object",
+    "description": "Parameters describing a box shape.",
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
+    "properties": {
+        "size": {
+            "type": "array",
+            "description": "The size (diameter) of the box in each axis in local space.",
```
EricGriffith: I don't think diameter is a common term for the size of a box.

specification/2.1/schema/shape.box.schema.json
```diff
@@ -0,0 +1,21 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "shape.box.schema.json",
+    "title": "glTF Box Shape",
+    "type": "object",
+    "description": "Parameters describing a box shape.",
```
EricGriffith: I suggest we use the fuller description from the spec instead:

> An axis-aligned box with a size per-axis, centered at the origin in local space, with normals facing outwards along each axis.

specification/2.1/schema/shape.capsule.schema.json
```diff
@@ -0,0 +1,28 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "shape.capsule.schema.json",
+    "title": "glTF Capsule Shape",
+    "type": "object",
+    "description": "Parameters describing a capsule shape.",
```
EricGriffith: > A capsule, centered at the origin in local space with potentially different radii at each end, equivalent to the convex hull of two spheres located along the Y axis (in local space) at a specified distance, with normals facing outwards.

specification/2.1/schema/shape.cylinder.schema.json
```diff
@@ -0,0 +1,28 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "shape.cylinder.schema.json",
+    "title": "glTF Cylinder Shape",
+    "type": "object",
+    "description": "Parameters describing a cylinder shape.",
```
EricGriffith: > A cylinder, centered at the origin in local space with potentially different radii at each end, equivalent to the convex hull of two circles in the X/Z plane positioned along the Y axis at a specified distance, with the top cap normal facing along the +Y axis, the bottom cap normal facing along the -Y axis, and the side normals facing outwards along the X/Z plane.

specification/2.1/schema/shape.plane.schema.json
```diff
@@ -0,0 +1,20 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "shape.plane.schema.json",
+    "title": "glTF Plane Shape",
+    "type": "object",
+    "description": "Parameters describing a plane shape.",
```
EricGriffith: > A plane centered at the origin in local space, optionally with finite extents, with normal along the +Y axis in local space. The plane is one-sided.

specification/2.1/schema/shape.schema.json
```diff
@@ -0,0 +1,66 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "shape.schema.json",
+    "title": "glTF Shape Resource",
+    "type": "object",
+    "description": "Parameters describing a shape used for physics, BVH, or any other use case.",
```
EricGriffith: Suggest this:

> A generic, re-usable shape definition with no inherent meaning attached to it.

specification/2.1/schema/shape.schema.json
```diff
@@ -0,0 +1,66 @@
...
+    "allOf": [ { "$ref": "glTFChildOfRootProperty.schema.json" } ],
+    "properties": {
+        "type": {
+            "description": "Specifies the shape type.",
+            "anyOf": [
+                {
+                    "const": "box",
+                    "description": "An axis-aligned box with a size per-axis, centered at the origin in local space, with normals facing outwards along each axis."
```
EricGriffith: In other places we don't put these detailed descriptions in the enum values and they don't show up in the generated docs. I propose we just have the description at the type level and rely on the more detailed description for the actual shape type schemas.

specification/2.1/schema/shape.schema.json
```diff
@@ -0,0 +1,66 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "shape.schema.json",
+    "title": "glTF Shape Resource",
+    "type": "object",
+    "description": "Parameters describing a shape used for physics, BVH, or any other use case.",
+    "allOf": [ { "$ref": "glTFChildOfRootProperty.schema.json" } ],
+    "properties": {
+        "type": {
+            "description": "Specifies the shape type.",
```
EricGriffith: Suggest this instead:

> The type of shape this shape object represents. If any properties are explicitly specified for the shape, the shape properties object MUST match the shape type.

specification/2.1/schema/shape.schema.json
```diff
@@ -0,0 +1,66 @@
...
+                    "description": "A sphere, centered at the origin in local space with a radius equal in all directions, with normals facing outwards."
+                },
+                {
+                    "type": "string"
+                }
+            ]
+        },
+        "box": {
+            "type": "object",
+            "description": "A set of parameter values that are used to define a box shape.",
```
EricGriffith: Maybe just this: "Box shape parameters"

specification/2.1/schema/shape.sphere.schema.json
```diff
@@ -0,0 +1,16 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "shape.sphere.schema.json",
+    "title": "glTF Sphere Shape",
+    "type": "object",
+    "description": "Parameters describing a sphere shape.",
```
EricGriffith: > A sphere, centered at the origin in local space with a radius equal in all directions, with normals facing outwards.

specification/2.1/Specification.adoc
```diff
@@ -845,6 +845,39 @@ The next example defines the transform for a node with attached camera using the
 ----
 
 
+[[bounding-volumes]]
+=== Bounding Volumes
+
+Any glTF node **MAY** define a local space bounding volume using a node's `boundingVolume` property. A bounding volume is a simple geometric shape, optionally transformed, that **SHOULD** completely enclose the node's geometry.
+
```
EricGriffith, line 853: Suggest this blurb:

> Bounding volumes for skinned meshes SHOULD be placed on the skin nodes (or their ancestors) rather than on the node with the skinned mesh.

specification/2.1/Specification.adoc
```diff
@@ -2589,6 +2622,49 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
 Skinned animation is achieved by animating the joints in the skin's joint hierarchy.
 
 
+[[shapes]]
+== Shapes
+
+glTF defines an array of shapes. All built-in shape types are "implicit" shapes, meaning that they have a surface defined by a mathematical function, and they have a distinct inside and outside.
```
EricGriffith, line 2629: Since we're planning to add mesh shape back in, not sure they will all be implicit

specification/2.1/Specification.adoc
```diff
@@ -2589,6 +2622,49 @@ See <<appendix-c-interpolation,Appendix C>> for additional information about int
...
+| `"plane"`    | A plane centered at the origin in local space, optionally with finite extents, with normal along the +Y axis in local space. The plane is one-sided.
+| `"sphere"`   | A sphere, centered at the origin in local space with a radius equal in all directions, with normals facing outwards.
+|====
+
+
+A sub-object with a key matching the shape type **MAY** be used to specify the size parameters of the shape. When the `type` of a shape refers to a built-in shape type, the shape must not supply a sub-object for a different shape type: for example, a shape whose `type` is "box" must not have a `sphere` property defined.
+
+By default, planes have infinite extents along their tangent and bitangent vectors, specified as X and Z in local space, respectively. A plane may optionally provide `sizeX` and `sizeZ` properties to specify a finite total size along these axes. The center of the plane's surface is at the origin in local space. The entirety of the space behind the plane, in its local -Y direction, is considered inside of the plane.
```
EricGriffith: +X and +Z ?

EricGriffith (commented):
Do we want to include mesh shape before we publish or do we want to incrementally publish?

specification/2.1/schema/shape.schema.json
weegeekps: That's too much of a fragment, which can be difficult for non-native speakers. I went with:

"Parameter values defining a box shape."

specification/2.1/Specification.adoc
weegeekps: Yeah, great callout. I've added this blurb.

specification/2.1/Specification.adoc
weegeekps: I think we should defer this language change until we have the mesh shape design fully ready. That way we only need to adjust the language once.

specification/2.1/schema/shape.box.schema.json
weegeekps: No, I removed it.

bjornblissing:
_Duplicating my comment from the issue:_

Could the specification clarify how a node's transform applies to its boundingVolume?

My understanding is that the shape is defined in canonical local space, then the optional boundingVolume scale / rotation / translation are applied relative to the node's local coordinate system, and finally the node's own matrix or TRS transform (plus ancestor transforms) places the volume in world space.

In other words, is the intended relationship:
`shape local space → bounding-volume transform → node transform → parent transforms`

If so, it may be helpful to state this explicitly.

specification/2.1/Specification.adoc
aaronfranke: As discussed in the group, replaced with this text:

> Skinned mesh nodes with bounding volumes **SHOULD** be placed as children of the skeleton root node (if the skin defines a skeleton root node), or as a sibling of a root of its joint hierarchy, not as scene root nodes. This allows the bounding volume to be transformed together with the skeletal hierarchy, such that root motion is applied to the bounding volume. This keeps the skinned mesh's bounding volume consistent with its skin-joint-node-determined visual representation.

specification/2.1/schema/shape.schema.json
aaronfranke: Yes, this is good. Note that the box schema has a more complete description:

> An axis-aligned box with a size per-axis, centered at the origin in local space, with normals facing outwards along each axis.

specification/2.1/schema/shape.box.schema.json
aaronfranke: The reasoning here is that some applications define how big a box is using "extents" instead of "size". The term "size" is hopefully unambiguous, so it's fine to remove "diameter", but it was meant to clarify this.

specification/2.1/Specification.adoc
lexaknyazev: Please remove the "not as scene root nodes" part. It wasn't present in the glTF 2.0 spec and does not add any value to the normative statement that comes right before it.

specification/2.1/Specification.adoc
```diff
@@ -845,6 +845,41 @@ The next example defines the transform for a node with attached camera using the
...
+
+[options="header"]
+|====
+| Property      | Description
+| `shape`       | The index of the shape in the top-level `shapes` array. Required.
+| `rotation`    | The bounding volume's unit quaternion rotation relative to the node's local coordinate system.
+| `scale`       | The bounding volume's scale along the rotated local X, Y, and Z axes.
+| `translation` | The bounding volume's translation along the node's local X, Y, and Z axes.
+|====
+
```
bjornblissing, line 863: Suggested addition to clarify the transform order for bounding volumes. Without this, implementations may place volumes differently - especially with nested nodes, non-uniform scale, or rotation - causing inconsistent culling, picking, collision, and other spatial-query results.


```suggestion

The shape is defined in its own local coordinate system. Any transform defined by the bounding volume's TRS properties **MUST** be applied before the node's global transformation matrix is applied, transforming the shape first into the node's local coordinate system and then into global space, as described in the <<transformations,Transformations>> section.

```

EricGriffith: @bjornblissing Taking this back to the DTSG to discuss.

EricGriffith (approved):

EricGriffith:
> _Duplicating my comment from the issue:_
> 
> Could the specification clarify how a node's transform applies to its boundingVolume?
> 
> My understanding is that the shape is defined in canonical local space, then the optional boundingVolume scale / rotation / translation are applied relative to the node's local coordinate system, and finally the node's own matrix or TRS transform (plus ancestor transforms) places the volume in world space.
> 
> In other words, is the intended relationship: `shape local space → bounding-volume transform → node transform → parent transforms`
> 
> If so, it may be helpful to state this explicitly.

I agree with this, but we'll discuss in the DTSG.

PR #2624, opened 2026-08-03
Cleanup matrix data layout figures

lexaknyazev:
Recreated the matrix padding layout figures using idiomatic SVG. New figures are much smaller in file size and can be directly edited if needed.


javagl (approved):
The SVG looks much cleaner and more compact. Now I'm nearly wondering where the original lines like
`<path fill="#F0F0F0" fill-rule="evenodd" d="M3 833.701h259.843v129.921H3Zm259.843 0h259.842v129.921H262.843Zm259.842 0h259.843v129.921H522.685Zm259.843 0h259.842v129.921H782.528Zm259.842 0h259.84v129.921h-259.84Zm259.84 0h259.85v129.921h-259.85Z"/>` for drawing a single box even came from, but likely a suboptimal export from some drawing program...

lexaknyazev:
That quoted path explicitly draws six rectangles (although with rather odd numbers); new images use `defs` to avoid such explicit repeats. FWIW, original images were exported from PowerPoint.

javagl:
PowerPoint could explain it - it's not an SVG-centric editor after all. (It would have surprised me if that had been created from InkScape, but even there, depending on how the drawing was created, odd numbers and repetition from copy+paste could easily happen)

PR #2627, opened 2026-08-17
KHR_interactivity: Address implementation feedback

lexaknyazev:
- Clarified `variable/interpolate`, `pointer/interpolate`, and `event/receive` operations.
- Added an optional `name` property to variables and events.

### DIFF
--- a/extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
+++ b/extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
@@ -3798,13 +3798,27 @@ The `useSlerp` configuration value **MUST** be a boolean literal and it **MUST N
 
 This operation has no internal state.
 
+This operation uses an easing function to map input progress positions to output progress positions. That easing function is derived from a cubic Bézier curve defined on a 2D plane as follows:
+
+1. The points latexmath:[P_0] located at latexmath:[(0, 0)] and latexmath:[P_3] located at latexmath:[(1, 1)] are the curve's start and end respectively.
+2. The control points latexmath:[P_1] and latexmath:[P_2] define the curve's tangent vectors latexmath:[\vec{P_0P_1}] and latexmath:[\vec{P_2P_3}]. The latexmath:[x] coordinates of latexmath:[P_1] and latexmath:[P_2] are restricted to the latexmath:[[0, 1\]] range.
+3. As such, the parametric form of the curve is:
++
+[latexmath]
++++++
+f_{x,y}(u) = (3P_1 - 3P_2 + 1) \cdot u^3 + (-6P_1 + 3P_2) \cdot u^2 + (3P_1) \cdot u
++++++
++
+where latexmath:[u \in [0, 1\]] is the parameter and latexmath:[f_{x,y}(u)] returns a point that belongs to the cubic Bézier curve on a 2D plane.
+4. The corresponding easing function is then defined as latexmath:[q = g(t)] where latexmath:[t] is the input progress position, latexmath:[q] is the output progress position, and latexmath:[g(t) = f_y(f_x^{-1}(t))].
+
 The internal state of an execution graph having one or more interactivity nodes with the `variable/interpolate` operation includes an implementation-defined _variable interpolation state dynamic array_ each element of which contains the following data:
 
 - The reference to the variable being interpolated
 - Implementation-defined high precision timestamp value representing the interpolation start time
 - Interpolation duration value converted to the implementation-defined high precision time type
 - Variable value at the time of the successful interactivity node activation
-- Information needed for cubic Bézier spline evaluation derived from the interactivity node's input values
+- Information needed for evaluation of the easing function derived from the interactivity node's input values
 - Target variable value
 - Implementation-specific pointer to the `done` output flow of the interactivity node that has added this entry
 
@@ -3821,7 +3835,7 @@ When the `in` input flow is activated:
 .. activate the `err` output flow and skip the next steps.
 5. If the _variable interpolation state dynamic array_ contains an entry with the same variable reference,
 .. remove it from the array.
-6. Using the implicitly-defined end points stem:[P_0 (0, 0)] and stem:[P_3 (1, 1)] together with the control points stem:[P_1] and stem:[P_2] provided via the input values construct a cubic Bézier easing function for the stem:[[0, 1]] input range.
+6. Using the cubic Bézier curve control points stem:[P_1] and stem:[P_2] provided via the input values, construct the corresponding easing function as described above.
 7. Add a new entry to the _variable interpolation state dynamic array_ filling it with the required information based on the evaluated input values.
 8. Activate the `out` output flow.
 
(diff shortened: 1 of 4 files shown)


extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -3822,6 +3822,8 @@ When the `in` input flow is activated:
 5. If the _variable interpolation state dynamic array_ contains an entry with the same variable reference,
 .. remove it from the array.
 6. Using the implicitly-defined end points stem:[P_0 (0, 0)] and stem:[P_3 (1, 1)] together with the control points stem:[P_1] and stem:[P_2] provided via the input values construct a cubic Bézier easing function for the stem:[[0, 1]] input range.
++
+The easing function constructed from these four control points can be represented as latexmath:[y=f(x)] where latexmath:[x \in [0; 1]] is the input progress value and latexmath:[y] is the output progress value. Control points latexmath:[P_0] and latexmath:[P_3] correspond to the function at the bounds of the input progress range, i.e., latexmath:[f(0)=0] and latexmath:[f(1)=1]. The latexmath:[\vec{P_0P_1}] and latexmath:[\vec{P_2P_3}] vectors define the curve's tangents.
```
bghgary: Two things here: the easing description leaves the evaluation ambiguous, and there are some minor rendering issues on this line and nearby.

The curve parameter is not the input progress value -- it has to be solved from `x(u) = x` before taking `y(u)`. Everything stated here is equally true of evaluating the curve at `u = x`, so the wording does not exclude that misreading.

```suggestion
The four control points define a parametric curve latexmath:[(x(u), y(u))] where latexmath:[u \in [0; 1\]] is the curve parameter, latexmath:[P_0] and latexmath:[P_3] are its end points, and the latexmath:[\vec{P_0P_1}] and latexmath:[\vec{P_2P_3}] vectors define its tangents at those end points. Since the first components of latexmath:[P_1] and latexmath:[P_2] are restricted to the latexmath:[[0; 1\]] range, latexmath:[x(u)] is non-decreasing, so the curve defines an easing function latexmath:[y=f(x)] where latexmath:[x \in [0; 1\]] is the input progress value and latexmath:[y] is the output progress value, with latexmath:[f(0)=0] and latexmath:[f(1)=1]. Evaluating latexmath:[f] at an input progress value requires first solving latexmath:[x(u)=x] for latexmath:[u] and then computing latexmath:[y(u)]; the curve parameter latexmath:[u] is not the input progress value.
```

We hit this in the Babylon.js implementation: for valid control points such as `P1 = (1, 1)` and `P2 = (0, 1)`, `dx/du` is zero at the root and a naive Newton inversion returns NaN there. Maybe worth an implementation note.

On rendering: an inline macro ends at the first `]`, so an interval's own closing bracket terminates it early. This line renders as `\(x \in [0; 1\)]` in `interactivity-outputs/Specification.html` on this PR's build -- closing delimiter inside the interval, stray `]` after it. The suggestion escapes it as `\]`, which the document already does at 1199 (`\arccos(a) \in [0; \pi\]`). Three pre-existing lines render the same way and are untouched by it: `stem:[[0, 1]]` at 3824 and 4235, and `stem:[[0, T]]` at 4290.

On notation: intervals appear in two forms -- `;` inside `\in` expressions (1184, 1199, 1214, 3840, 4251, and this line) and `,` in "the ... range" phrasing (2841, 2855, 3824, 4235, 4290). This line follows the existing `\in` convention, but 3824 and 3826 put both forms one line apart.

The tick steps below call the input and output progress values `t` and `q`, while this paragraph calls them `x` and `y`; worth using one set of names, especially since `t` usually denotes the curve parameter.

bghgary: The curve parameter is not `x`: it has to be solved from `x(u) = x` before taking `y(u)`. Everything stated here -- `f(0)=0`, `f(1)=1`, the tangents -- is equally true of evaluating the curve at `u = x`, so this wording does not rule out the parametric misreading. Worth stating the inversion explicitly.

A note on the inversion would help too, because it is not numerically trivial: `dx/du` can vanish at an interior point for control points this operation accepts. `P1 = (1, 1)`, `P2 = (0, 1)` gives `x(u) = 3u - 6u^2 + 4u^3`, so `dx/du = 3(1 - 2u)^2`, which is zero at `u = 0.5`. Step 3 does not reject those inputs, and a plain Newton solve returns NaN there.

One more: this names the input and output progress `x` and `y`, while the tick steps below use `t` and `q` for the same two quantities -- and `t` is the conventional name for the curve parameter.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -5556,6 +5569,8 @@ the behavior graph is invalid and thus cannot be used; this has no effect on oth
 .. if the `nodes` array length is greater than the implementation-specific limit on the number of interactivity nodes,
 ... _reject the graph_;
 .. validate each element of the `nodes` array as described in the <<validation-node-object,Node Object Validation>> section.
+7. If "`the graph`" object has the `name` property,
```
bghgary, line 5584: The graph object's `name` gets a validation step here, but not the other two things variables and events get in this PR: the JSON Syntax "Graphs" section still does not mention `name`, so its uniqueness and execution semantics are left unstated, and `graph.KHR_interactivity.schema.json` still refs `glTFChildOfRootProperty.schema.json` rather than the new `nameableProperty.KHR_interactivity.schema.json`.

If the reason for the new schema is that the core `gltf_detailedDescription` describes root-level objects, that applies to the graph as well. Intentional?

bghgary (commented):
[Reviewed by Copilot on behalf of @bghgary]

The `name` additions read correct, and the `variable/interpolate` clarification matches what Babylon.js already implements. One wording suggestion and some minor points inline.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
lexaknyazev: Yes, this is intentional. A `graph` object follows the regular "child of root" rules inherited from the main spec. The validation step is added just to ensure that the procedure covers all properties.

extensions/2.0/Khronos/KHR_interactivity/Specification.adoc
```diff
@@ -3798,13 +3798,27 @@ The `useSlerp` configuration value **MUST** be a boolean literal and it **MUST N
...
 
+This operation uses an easing function to map input progress positions to output progress positions. That easing function is derived from a cubic Bézier curve defined on a 2D plane as follows:
+
+1. The points latexmath:[P_0] located at latexmath:[(0, 0)] and latexmath:[P_3] located at latexmath:[(1, 1)] are the curve's start and end respectively.
+2. The control points latexmath:[P_1] and latexmath:[P_2] define the curve's tangent vectors latexmath:[\vec{P_0P_1}] and latexmath:[\vec{P_2P_3}]. The latexmath:[x] coordinates of latexmath:[P_1] and latexmath:[P_2] are restricted to the latexmath:[[0, 1\]] range.
+3. As such, the parametric form of the curve is:
++
+[latexmath]
++++++
+f_{x,y}(u) = (3P_1 - 3P_2 + 1) \cdot u^3 + (-6P_1 + 3P_2) \cdot u^2 + (3P_1) \cdot u
```
bghgary: Item 4 uses `f_x` and `f_y`, which are not introduced anywhere. Please define them as the components of `f_{x,y}`.

The formula is written to be read per component -- `+ 1` is correct that way because `P_3 = (1, 1)` -- but line 3812 calls the result a point, so nothing states that it is applied componentwise.

bghgary (approved):

Issue #2611, opened 2026-07-07
[glTF 2.1] Multiple Binary Chunks in GLB


akmorgan99:
# More flexible GLB buffers and chunk layout

The binary file format (GLB / `.glb`) version 3, being introduced with glTF 2.1, will have the following additional features:

- Multiple chunks are allowed to be used as glTF buffers.

- glTF buffers explicitly refer to chunks in the GLB file by index in a new `"chunk"` property.

- Definition change for the location of the glTF JSON data: The first chunk *that is a JSON chunk* is the glTF data (rather than requiring the glTF JSON chunk to be the first chunk in the file).

- The order and layout of chunks in the GLB file is more flexible, allowed to be any order, so long as there isn't a non-glTF JSON chunk before the JSON chunk containing the main glTF data.

The existing behavior of glTF 2.0's first buffer implicitly referring to chunk 1 in the `.glb` file (when `"uri"` is not set) will still be supported for backward compatibility. This implicit behavior ONLY happens with files where the JSON chunk is at chunk 0, the buffer chunk is at chunk 1, there is only one buffer without a `"uri"`, and that buffer is buffer 0.

## Use cases

Using a single buffer chunk in a GLB file can lead to issues when the buffer is very large, such as:

- Compatibility with languages that have a maximum size of managed memory byte arrays.

- Compatibility with JSON parsers interpreting the `"byteLength"` property as a double-precision float.

Furthermore, the more flexible chunk layout allows placing custom chunks before the glTF JSON chunk, which can be useful for cases where data is needed at the start of the file, such as streaming use cases. glTF 2.1 does not define any such cases, but this flexibility allows for extensions, applications, or future glTF versions to define this without breaking compatibility with binary file format version 3. For example, having a thumbnail binary chunk at the start of the file is being considered.

## Why this design

Extending the implicit behavior of glTF 2.0 to allow multiple buffers to implicitly refer to multiple chunks was considered, but the explicit behavior makes things clear and enables use cases such as using a pre-glTF-JSON chunk as a buffer chunk, which would be impossible with the implicit behavior.

javagl:
I'm nearly a bit surprised to see this  as its own issue, because a considerable part of this was already anticipated and discussed as part of https://github.com/KhronosGroup/glTF/issues/2594#issuecomment-4583019141 , but it can certainly make sense to make this _explicit_ (but also to have that link to the other issue, because they are at least related)

Issue #2594, opened 2026-05-27
[glTF 2.1] 64-bit upgrade to binary format


akmorgan99:
# 64-bit upgrade to the binary format

glTF 2.0's binary format (`.glb`) uses 32-bit integers for lengths. glTF 2.1 provides a upgraded binary format that uses 64-bit integers for lengths. This is known as binary format version 3, and the existing binary format (version 2) is still defined.

Additionally, a new chunk encoding field is added to the chunk header for version 3 only. For glTF 2.1, this field is always zero. Future versions of glTF or extensions may use this field.

Importers must support both binary format version 2 and version 3. Exporters may choose to keep using version 2 binary format for compatibility, so long as the files are smaller than 4 GiB and do not use the new chunk encoding field.

## Use cases

The 64-bit binary format allows for files larger than 4 GiB. The upgraded limit is 8 EiB, with the most significant bit reserved for future use.

The chunk encoding field may be used in the future for chunk-level compression, encryption, or other features. glTF 2.1 does not define any of those, however glTF extensions already allow for compression at other levels.

## Example

> [!NOTE]
> Examples are illustrative and not exhaustive.

<img width="2232" height="290" alt="Example Image" src="https://github.com/user-attachments/assets/60482f33-db7e-452c-a44f-e475455a346f" />

## Why this design

The 8 EiB limit ensures that the most significant bit is always zero, allowing glTF binary parsers to safely use signed 64-bit integers, as well as unsigned 64-bit integers. This allows for easier implementation in signed-only languages, allows `-1` to be used as a sentinel or error value, and reserves the most significant bit for future use if needed.

DRx3D:
Fixed title from "64-big" to "64-bit".

DRx3D:
Fixed NOTE highlight to be in a blockquote. ([reference](https://github.com/orgs/community/discussions/16925))

vpenades:
As I understand it, this new GLB binary format is:
- Incompatible with older versions of GLB that use 32bit length
- Is still limited to a single binary chunk

Am I right in those assumptions?

aaronfranke:
@vpenades Yes and yes (pending discussion). Implementations can handle both, but this has to be explicitly added to the implementation.

vpenades:
I would suggest that, now the binary format is going to be updated, to support multiple binary chunks. Not to increase the binary size limit, but to allow simple roundtrips between glTF with multiple binary chunks and GLB

javagl:
[...]

PR #2628, opened 2026-08-18
Draft changes for 64-bit Upgrade to Binary Format in glTF 2.1

aaronfranke:
This PR contains the draft changes for the 64-bit upgrade to the binary file format. The 32-bit format is still defined. This PR currently points towards the `draft-2.1` branch and should be safe to merge to that branch.

**Note:** This text may be replaced in the future, as @lexaknyazev is planning on rewriting the binary file format section, likely splitting into separate files. Regardless of the progress on that rewrite, this PR exists as a modification to the existing text, which the glTF 2.1 TSG in Khronos has agreed to, and can be merged in advance.

### Relevant Issues:

- https://github.com/KhronosGroup/glTF/issues/2594
- https://github.com/KhronosGroup/glTF/issues/2611

### Summary of changes:

GLB binary format version 2 uses 32-bit integers for lengths. GLB binary format version 3 provides an upgraded binary format that uses 64-bit integers for lengths. The upgraded file size limit is 8 EiB.

glTF 2.1 data saved in GLB binary format version 3 allows for buffers to explicitly refer to chunks by index in a new `"chunk"` property, allowing for multiple binary buffer chunks, and a more flexible GLB chunk layout.

GLB binary format version 3 introduces a new chunk encoding field to the chunk header. For glTF 2.1, this field is always zero. Future versions of glTF or extensions may use this field.

Importers must support both binary format version 2 and version 3. Exporters may choose to keep using version 2 binary format for compatibility, so long as the files are smaller than 4 GiB and do not use the new chunk encoding field.


specification/2.1/schema/buffer.schema.json
```diff
@@ -6,18 +6,23 @@
     "description": "A buffer points to binary geometry, animation, or skins.",
     "allOf": [ { "$ref": "glTFChildOfRootProperty.schema.json" } ],
     "properties": {
+        "byteLength": {
```
EricGriffith: Do we need to change the order of byteLength relative to the other properties? Not everything is alphabetized.

specification/2.1/Specification.adoc
```diff
@@ -930,11 +930,13 @@ Buffers and buffer views do not contain type information. They simply define the
 
 The glTF asset **MAY** use the GLB file container to pack glTF JSON and one glTF buffer into one file. Data for such a buffer is provided via the GLB-stored `BIN` chunk.
 
-A buffer with data provided by the GLB-stored `BIN` chunk, **MUST** be the first element of `buffers` array and it **MUST** have its `buffer.uri` property undefined. When such a buffer exists, a `BIN` chunk **MUST** be present.
+A buffer **MAY** have its data provided by a GLB-stored chunk. This may be done explicitly by specifying the `buffer.chunk` property in the buffer object. When a buffer references a chunk by index, that chunk **MUST** be present in the GLB.
```
EricGriffith: This may be done -> This is done

specification/2.1/Specification.adoc
```diff
@@ -930,11 +930,13 @@ Buffers and buffer views do not contain type information. They simply define the
...
+
+Alternatively, for compatibility with glTF 2.0, the first buffer in the `buffers` array **MAY** have its data implicitly provided by a GLB-stored `BIN` chunk when neither the `buffer.chunk` nor `buffer.uri` properties are defined. When such a buffer exists at buffer index 0, a `BIN` chunk **MUST** be present at chunk index 1 in the GLB file, immediately following the JSON chunk at chunk index 0. Other buffers do not have this implicit behavior, and **MUST** only have explicit references to its data via one of the `buffer.chunk` or `buffer.uri` properties.
```
EricGriffith: for compatibility -> for backwards compatibility

Also suggest this for closer alignment with the previous version:
the first buffer in the `buffers` array **MAY** have its data implicitly provided by the GLB-stored `BIN` chunk at chunk index 0. When doing this, the buffer **MUST** have both its `buffer.chunk` and `buffer.uri` properties undefined, and the GLB file **MUST** have at least two chunks: a JSON chunk at index 0 and a `BIN` chunk at index 1. Other buffers do not have this implicit behavior, and **MUST** have one of its `buffer.chunk` or `buffer.uri` properties defined.

(Do we want to make chunk and uri mutually exclusive?)

specification/2.1/Specification.adoc
```diff
@@ -930,11 +930,13 @@ Buffers and buffer views do not contain type information. They simply define the
...
 
 Any glTF buffer with undefined `buffer.uri` property that is not the first element of `buffers` array does not refer to the GLB-stored BIN chunk, and the behavior of such buffers is left undefined to accommodate future extensions and specification versions.
```
EricGriffith: Any glTF buffer with undefined `buffer.uri` property
|
V
Any glTF buffer with undefined `buffer.chunk` property and undefined `buffer.uri`

specification/2.1/Specification.adoc
```diff
@@ -2683,18 +2686,25 @@ The registered media type is `model/gltf-binary`.
 [[binary-gltf-layout-overview]]
 === Overview
 
-Binary glTF is little endian. The figure below shows an example of a Binary glTF asset.
+Binary glTF is little endian. The figure below shows an example of the binary format version 2 layout:
 
-.Binary glTF Layout
+.Binary glTF version 2 Layout
```
EricGriffith: version -> Version

specification/2.1/Specification.adoc
```diff
@@ -2683,18 +2686,25 @@ The registered media type is `model/gltf-binary`.
...
 image::figures/glb2.svg[pdfwidth=6in,align=left]
 
+The figure below shows an example of the binary format version 3 layout:
+
+.Binary glTF version 3 Layout
```
EricGriffith: version -> Version

specification/2.1/Specification.adoc
```diff
@@ -2703,11 +2713,22 @@ uint32 version
...
+[source,c]
+----
+uint32 magic
+uint32 version
+uint64 length
+----
+
 * `magic` **MUST** be equal to equal `0x46546C67`. It is ASCII string `glTF` and can be used to identify data as Binary glTF.
-* `version` indicates the version of the Binary glTF container format. This specification defines version 2.
+* `version` indicates the version of the Binary glTF container format. The glTF 2.1 specification defines version 2 and version 3.
```
EricGriffith, line 2943: General question here: Do we want to restate the version 2 spec in the 2.1 spec or do we want to just point to the glTF 2 spec for the version 2 spec?

specification/2.1/Specification.adoc
```diff
@@ -2725,19 +2748,30 @@ uint32 chunkType
...
+ubyte[] chunkData
+----
+
 * `chunkType` indicates the type of chunk. See <<table-chunktypes>> for details.
+* `chunkEncoding` indicates the encoding method used for the chunk data. See <<chunk-encoding>> for details.
+* `chunkLength` is the length of `chunkData`, in bytes. In binary format version 3, the most significant bit **MUST** be 0, which limits chunk sizes to 8 EiB (exbibytes).
 * `chunkData` is the binary payload of the chunk.
 
-The start and the end of each chunk **MUST** be aligned to a 4-byte boundary. See chunks definitions for padding schemes. Chunks **MUST** appear in exactly the order given in <<table-chunktypes>>.
+In binary format version 2, the start and the end of each chunk **MUST** be aligned to a 4-byte boundary. In binary format version 3, the start of each chunk **MUST** be aligned to an 8-byte boundary. The start of each chunk is the next 4 or 8 byte aligned boundary after the end of the previous chunk. See chunks definitions for padding schemes. Chunks **MUST** appear in exactly the order given in <<table-chunktypes>>.
```
EricGriffith: The start of each chunk is the next 4 or 8 byte aligned boundary after the end of the previous chunk.
|
V
The start of each chunk is the next 4-byte (v2) or 8-byte (v3) aligned boundary after the end of the previous chunk.

specification/2.1/Specification.adoc
```diff
@@ -2754,22 +2788,37 @@ This chunk holds the glTF JSON, as it would be provided within a .gltf file.
 In a JavaScript implementation, the `TextDecoder` API can be used to extract the glTF content from the ArrayBuffer, and then the JSON can be parsed with `JSON.parse` as usual.
 ====
 
-This chunk **MUST** be the very first chunk of a Binary glTF asset. By reading this chunk first, an implementation is able to progressively retrieve resources from subsequent chunks. This way, it is also possible to read only a selected subset of resources from a Binary glTF asset.
+In binary format version 2, this chunk **MUST** be the very first chunk of a Binary glTF asset. In binary format version 3, the first chunk that is a JSON chunk is the glTF JSON chunk, allowing for non-JSON chunks to precede the glTF JSON chunk. By reading this chunk first, an implementation is able to progressively retrieve resources from subsequent chunks. This way, it is also possible to read only a selected subset of resources from a Binary glTF asset.
```
EricGriffith, line 3008: The language on line 2766 says the chunks must appear in the order shown in the table on line 2768. So we need to reconcile this difference. This also relates to the language in buffer about implicit chunk indices. If we end up allowing thumbnail data to be at index 0, then that disallows implicit chunk indexing when using a thumbnail (which is probably OK).

specification/2.1/Specification.adoc
```diff
@@ -2754,22 +2788,37 @@ This chunk holds the glTF JSON, as it would be provided within a .gltf file.
...
 [[binary-buffer]]
 ==== Binary buffer
 
 This chunk contains the binary payload for geometry, animation key frames, skins, and images. See <<glb-stored-buffer,GLB-stored Buffer>> for details on referencing this chunk from JSON.
 
-This chunk **MUST** be the second chunk of the Binary glTF asset.
+In binary format version 2, this chunk **MUST** be the second chunk of the Binary glTF asset. In binary format version 3, binary buffer chunks are explicitly referenced by index from the JSON chunk's buffers array, allowing for multiple binary buffer chunks to be present in a Binary glTF asset.
 
-This chunk **MUST** be padded with trailing zeros (`0x00`) to satisfy alignment requirements.
+If the chunk is padded, either inside the binary chunk content, or outside the chunk length, the padding **MUST** be trailing zeros (`0x00`).
```
EricGriffith: We should not prescribe how to pad inside chunk content for binary data; we should treat it as a black box:

If the chunk has padding after the end of the content, the padding **MUST** be trailing zeros (`0x00`).

specification/2.1/Specification.adoc
```diff
@@ -2754,22 +2788,37 @@ This chunk holds the glTF JSON, as it would be provided within a .gltf file.
...
 
 When the binary buffer is empty or when it is stored by other means, this chunk **SHOULD** be omitted.
 
+
+[[chunk-encoding]]
+==== Chunk Encoding
+
+In binary format version 3 only, between each chunk's type and size, the `chunkEncoding` field indicates the encoding method used for the chunk data, such as compression or encryption.
+
+The value `0x00000000` indicates that the chunk data is plainly encoded, meaning that it is not compressed or encrypted. Other values indicate that the chunk data is encoded with a specific method, such as compressed and/or encrypted. Alternate encoding schemes **SHOULD** be defined in glTF extensions, however such extensions are not required to be marked as used in the JSON chunk.
```
EricGriffith: What do you think about this?

The value `0x00000000` indicates that the chunk data is plainly encoded, meaning the chunk data does not need to be decoded (e.g., decompressed or decrypted) before use. Other values indicate that the chunk data does require decoding before use. Alternate encoding schemes **SHOULD** be defined in glTF extensions, however such extensions are not required to be marked as used in the JSON chunk.

Chunk-level encoding is distinct from any encoding that may be applied to data within chunks (e.g., storing a `.zip` file within a buffer in a `BIN` chunk).

specification/2.1/Specification.adoc
```diff
@@ -2754,22 +2788,37 @@ This chunk holds the glTF JSON, as it would be provided within a .gltf file.
...
+
+Each chunk may use the same or a different encoding method compared to other chunks, as provided by the flexibility of this field being per-chunk. Additionally, since the file and chunk headers themselves are always encoded the same (never compressed or encrypted), this allows applications to read enough of the file to determine its overall layout even when the chunk contents cannot be read, ensuring that file identification utilities can still recognize the file as a Binary glTF asset.
```
EricGriffith: Each chunk may use the same or a different encoding method compared to other chunks, as provided by the flexibility of this field being per-chunk. Additionally, since the file and chunk headers themselves are always encoded the same (never compressed or encrypted),
|
V
Each chunk specifies its own encoding, which allows chunks within the same Binary glTF asset to use different encoding methods. The chunk encoding does not apply to chunk headers. This allows ...

specification/2.1/schema/buffer.schema.json
aaronfranke: I think we should do a pass on all files to alphabetize the properties. For this PR, I am only cleaning up the files I am touching, but we should do follow-ups later.

EricGriffith: I think it will just create unnecessary churn for folks looking to diff 2.0 vs. 2.1

aaronfranke: Alphabetizing the properties is a non-breaking change, so we can also change this in the glTF 2.0 specification.

specification/2.1/Specification.adoc
aaronfranke: Multiple chunks are allowed, so it should be "a" chunk, not "the" chunk. And the chunk index of this implicit chunk is 1, not 0.

For the wording, I prefer my version - I think it's clearer to separate the JSON part as the first sentence and the GLB part as the second sentence.

specification/2.1/Specification.adoc
aaronfranke: I'd imagine this will be resolved with @lexaknyazev's reorganization later.

specification/2.1/Specification.adoc
aaronfranke: "The chunk encoding does not apply to chunk headers" is not needed. It's impossible to encode the chunk header with the chunk header, because you would need to read the chunk header to know how the chunk header is encoded. My old text was redundant in this area, too.

specification/2.1/schema/buffer.schema.json
EricGriffith: It's fine with me if we change glTF 2.0 first, but I'd prefer not to make this change in 2.1

specification/2.1/figures/glb3.svg
EricGriffith, line 1: The parentheses aren't properly enclosing the labels on the second row of text in the blocks for many of the blocks.

Image

specification/2.1/Specification.adoc
```diff
@@ -2942,19 +2965,30 @@ uint32 chunkType
...
+ubyte[] chunkData
+----
+
 * `chunkType` indicates the type of chunk. See <<table-chunktypes>> for details.
+* `chunkEncoding` indicates the encoding method used for the chunk data. See <<chunk-encoding>> for details.
+* `chunkLength` is the length of `chunkData`, in bytes. In binary format version 3, the most significant bit **MUST** be 0, which limits chunk sizes to 8 EiB (exbibytes).
 * `chunkData` is the binary payload of the chunk.
 
-The start and the end of each chunk **MUST** be aligned to a 4-byte boundary. See chunks definitions for padding schemes. Chunks **MUST** appear in exactly the order given in <<table-chunktypes>>.
+In binary format version 2, the start and the end of each chunk **MUST** be aligned to a 4-byte boundary. In binary format version 3, the start of each chunk **MUST** be aligned to an 8-byte boundary. The start of each chunk is the next 4-byte (v2) or 8-byte (v3) aligned boundary after the end of the previous chunk. See chunks definitions for padding schemes. Chunks **MUST** (v2) or **SHOULD** (v3) appear in exactly the order given in <<table-chunktypes>>.
```
EricGriffith: chunks definitions -> chunk definitions

specification/2.1/Specification.adoc
EricGriffith: Added it to our meeting agenda to discuss.

specification/2.1/schema/buffer.schema.json
EricGriffith: I reverted the reorder of byteLength for now; we can discuss alphabetizing properties in the DTSG.

specification/2.1/figures/glb3.svg
EricGriffith: I fixed this

specification/2.1/Specification.adoc
EricGriffith: I fixed this too

PR #2635, opened 2026-08-24
Make JSON valid in `shape.schema.json`

javagl:
https://github.com/KhronosGroup/glTF/pull/2619 introduced invalid JSON

(Eventually, we should run a formatter, but "being valid" is some sort of baseline...)


weegeekps (approved):
Yes, a formatter or linter for the JSON schemas is certainly needed. Thanks for fixing this.

PR #2641, opened 2026-09-02
Update Binary Data Storage

### LINKED ISSUE #2117: Clarification of alignment requirement for matrix columns
The [specifcation, 3.6.2.4. Data Alignment](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#data-alignment) says:

> When two or more vertex attribute accessors use the same bufferView, its byteStride MUST be defined.
> ...
> For performance and compatibility reasons, each element of a vertex attribute MUST be aligned to 4-byte boundaries [...]
> 
> Accessors of matrix type have data stored in column-major order; start of each column MUST be aligned to 4-byte boundaries. [...]

The first statements clearly refer to _vertex attributes_. And knowing about the care and diligence that went into all this, it's reasonable to assume that these requirements do _not_ apply to data that is _not_ used as vertex attributes (otherwise, the statements would not be restricted to these).  

The last statement does **not** explicitly refer to vertex attributes. It seems to talk about accessors _in general_. 

So the question is:

Does the alignment requirement for matrix columns apply to **ALL** accessors, or **ONLY** to vertex attributes?

If it referred only to vertex attributes, the statement should be changed to

> Accessors of matrix type have data stored in column-major order. **When they are used as vertex attributes, then the** start of each column MUST be aligned to 4-byte boundaries. [...]

If it referred to all accessors, it might make sense to say that explicitly, to avoid confusion. Maybe by just adding something like "**(regardless of whether they are used as vertex attributes or not)**" as a clarification, unless someone comes up with a more elegant way of phrasing that...

### LINKED ISSUE #2152: Clarify bufferView.target for sparse accessors
_https://github.com/KhronosGroup/glTF/issues/1440#issuecomment-1110785623 by @hybridherbst_

> I find the [description in the spec](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#:~:text=When%20a%20buffer%20view%20is%20used%20by%20vertex%20indices%20or%20attribute%20accessors%20it%20SHOULD%20specify%20bufferView.target%20with%20a%20value%20of%20element%20array%20buffer%20or%20array%20buffer%20respectively.) confusing in that regard:
> 
> > When a buffer view is used by vertex indices or attribute accessors it SHOULD specify bufferView.target with a value of element array buffer or array buffer respectively
> 
> as one could easily assume that indices and values in sparse accessors are bufferViews used by vertex indices or attribute accessors. I think the correct reading would be that they're not _directly_ used by the accessor but _indirectly_ used by the accessor.

### LINKED ISSUE #2169: Clarify handling of meshes with undefined data
> There's no such requirement in the spec, so data remains undefined.

Okey, so, if I take one of these examples (provided the node and the mesh are instanced in a scene) and I try to load the model in a game engine or something.... which is the expected behavior?  crash on load?, silently skip the mesh?  I think this should be clarified

_Originally posted by @vpenades in https://github.com/KhronosGroup/glTF-Validator/issues/189#issuecomment-1148566794_

### LINKED ISSUE #2198: What does "normalized" mean for morph target weights?
In the spec https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#animations, for `"weights"` there's written:

>float
signed byte normalized
unsigned byte normalized
signed short normalized
unsigned short normalized

What does "normalized" mean in this context? The [SimpleMorphTarget](https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_017_SimpleMorphTarget.md) example inlcudes weights that neither sum to one nor have unit L2 norm.

Does "normalized" mean "values restricted to those between zero and one"?

### LINKED ISSUE #2477: Question regarding accessor's byte offset when sparse data is available
I have been trying to figure that one out from the specifications but I can't seem to quite confirm the expected behaviour here for a glTF importer.

Currently, I apply the byte offset property from the accessor to both the "regular" data and the sparse data, like so:

```
buffer.m_data = gltfBuffer.data.data() + gltfBufferView.byteOffset + gltfAccessor.byteOffset;
buffer.m_sparseData = gltfSparseBuffer.data.data() + gltfSparseBufferView.byteOffset + gltfAccessor.byteOffset;
```

Is this correct? or is the byte offset property only meant to be applied to the "regular" data pointer?
I couldn't find any glTF sample file so far that'd have non-zero byte offsets while using sparse accessors.

### LINKED ISSUE #2577: Is it possible for `bufferView.byteStride` to be smaller than accessor element size?
Let's say the accessor has component type FLOAT (size 4 bytes) and type VEC3 (3 components => element size 12 bytes).

Is it valid for `bufferView.byteStride` that the accessor references to be equal to 4?

The specification says:

> When byteStride of the referenced bufferView is not defined, it means that accessor elements are tightly packed, i.e., effective stride equals the size of the element. When byteStride is defined, it MUST be a multiple of the size of the accessor’s component type.

> When accessor values are not tightly-packed (i.e., bufferView.byteStride is greater than element’s byte length), iteration over the created data view would need to take interleaved values into account (i.e., skip them).

From the latter I would infer that the intent was to ensure that `byteStride` is greater than or equal to element's byte length. However, that's not explicitly specified in the first sentence, which only restricts the `byteStride` to be divisible by the size of the *component* type, and says nothing about the element type, so from these two parts alone, normatively speaking, it does not seem like anything prevents the aforementioned setup to be valid. Am I missing the relevant language somewhere else?

lexaknyazev:
This PR expands the "Binary Data Storage" section to cover various minor issues accumulated since the initial glTF 2.0 specification release.

Although the update is mostly editorial, it contains the following important changes:
- The description of buffer resources now consistently mentions that referenced resources can be larger than the value of the buffer's `byteLength` property. Previous language about the GLB's `BIN` chunk being potentially "up to 3 bytes bigger" has been removed because that language was not normatively accurate anyway.
- The exact layout of 2x2 and 3x3 matrix accessors when not used for vertex attributes is now fully defined. Previous language was ambiguous and self-contradicting to some extent.
- Many new statements about data types, byte offsets, lengths, and strides were added for better clarity however they could be generally derived from the old language.

Fixes #2117
Fixes #2152
Fixes #2169
Fixes #2198
Fixes #2477
Fixes #2577

### DIFF
--- a/specification/2.0/Specification.adoc
+++ b/specification/2.0/Specification.adoc
@@ -895,24 +895,23 @@ The next example defines the transform for a node with attached camera using the
 == Binary Data Storage
 
 
-[[buffers-and-buffer-views]]
-=== Buffers and Buffer Views
+[[buffers]]
+=== Buffers
 
-
-[[buffers-and-buffer-views-overview]]
+[[buffers-overview]]
 ==== Overview
 
-A *buffer* is arbitrary data stored as a binary blob. The buffer **MAY** contain any combination of geometry, animation, skins, and images.
+A _buffer_ is arbitrary data stored as a binary blob. Buffers **MAY** contain any combination of any data used by the glTF asset.
 
 Binary blobs allow efficient creation of GPU buffers and textures since they require no additional parsing, except perhaps decompression.
 
-glTF assets **MAY** have any number of buffer resources. Buffers are defined in the asset's `buffers` array.
+A buffer is defined by the `byteLength` and `uri` properties that specify the byte size of the buffer and the URI to the buffer data, respectively. glTF assets **MAY** have any number of buffers. Buffers are defined in the asset's `buffers` array.
 
-While there's no hard upper limit on buffer's size, glTF assets **SHOULD NOT** use buffers bigger than 2^53^ bytes because some JSON parsers may be unable to parse their `byteLength` correctly. Buffers stored as <<glb-file-format-specification,GLB>> binary chunk have an implicit limit of 2^32^-1 bytes.
+While there's no hard upper limit on a buffer's byte size, glTF assets **SHOULD NOT** use buffers bigger than 2^53^-1 bytes because some JSON parsers may be unable to parse the value of their `byteLength` property correctly. Buffers stored as <<glb-file-format-specification,GLBv2>> binary chunk have an implicit size limit of 2^32^-1 bytes.
 
-All buffer data defined in this specification (i.e., geometry attributes, geometry indices, sparse accessor data, animation inputs and outputs, inverse bind matrices) **MUST** use little endian byte order.
-
-The following example defines a buffer. The `byteLength` property specifies the size of the buffer file. The `uri` property is the URI to the buffer data.
+[NOTE]
+.Example
+====
 [source,json]
 ----
 {
@@ -924,26 +923,69 @@ The following example defines a buffer. The `byteLength` property specifies the 
    ]
 }
 ----
+====
 
-The byte length of the referenced resource **MUST** be greater than or equal to the `buffer.byteLength` property.
+The byte length of the referenced resource **MUST** be greater than or equal to the buffer's `byteLength` property. Only the resource's byte range from zero (inclusive) to `byteLength` (exclusive) is considered referenced by the buffer object.
 
-Buffer data **MAY** alternatively be embedded in the glTF file via `data:` URI with base64 encoding. When `data:` URI is used for buffer storage, its `mediatype` field **MUST** be set to `application/octet-stream` or `application/gltf-buffer`.
+Buffer data **MAY** alternatively be embedded in the glTF file via `data:` URIs. When a `data:` URI is used for buffer storage, its `mediatype` field **MUST** be set to `application/gltf-buffer` or `application/octet-stream`.
 
-A _buffer view_ represents a contiguous segment of data in a buffer, defined by a byte offset into the buffer specified in the `byteOffset` property and a total byte length specified by the `byteLength` property of the buffer view.
+If the `uri` property is undefined, the buffer's data is provided by other means, such as a GLBv2 chunk (see below) or additional extensions. In absence of data sources, the buffer's content is undefined and handling of such buffers is implementation-dependent.
 
-Buffer views used for images, vertex indices, vertex attributes, or inverse bind matrices **MUST** contain only one kind of data, i.e., the same buffer view **MUST NOT** be used both for vertex indices and vertex attributes.
+[[glb-stored-buffer]]
+==== GLB-stored Buffer
 
-When a buffer view is used by vertex indices or attribute accessors it **SHOULD** specify `bufferView.target` with a value of _element array buffer_ or _array buffer_ respectively.
+The glTF asset **MAY** use the GLBv2 file container format to pack glTF JSON and one glTF buffer into the same file. Data for that buffer is provided by the GLB-stored `BIN` chunk.
+
+When the glTF asset is stored inside a GLBv2 file, a glTF buffer that has index zero, i.e., the first element of the `buffers` array, and has its `uri` property undefined represents the GLB-stored buffer. The GLB file **MUST** have a `BIN` chunk in this case.
+
+[NOTE]
+.Example
+====
+In the following snippet, the first buffer object refers to GLB-stored data, while the second buffer uses an external resource.
+
+[source,json]
+----
+{
+    "buffers": [
+        {
+            "byteLength": 35884
+        },
+        {
+            "byteLength": 504,
+            "uri": "external.bin"
+        }
+    ]
+}
+----
+====
+
+Any glTF buffer with undefined `uri` property that is not the first element of the `buffers` array does not refer to the GLB-stored BIN chunk, and the behavior of such buffers is left undefined to accommodate future extensions and specification versions.
+
+The byte length of the GLB's `BIN` chunk **MUST** be greater than or equal to the byte length of the corresponding glTF buffer.
 
 [NOTE]
 .Implementation Note
 ====
-This allows client implementations to early designate each buffer view to a proper processing step, e.g, buffer views with vertex indices and attributes would be copied to the appropriate GPU buffers, while buffer views with image data would be passed to format-specific image decoders.
+Not requiring strict equality of chunk's and buffer's lengths is consistent with how buffers use external resources and slightly simplifies glTF to GLBv2 conversion: buffer's `byteLength` does not need to be updated after applying GLBv2 padding.
 ====
 
-The `bufferView.target` value uses integer enums defined in the <<_bufferview_target,Properties Reference>>.
+See <<glb-file-format-specification,GLB File Format Specification>> for details on GLBv2 File Format.
 
-The following example defines two buffer views: the first holds the indices for an indexed triangle set, and the second holds the vertex data for the triangle set.
+
+[[buffer-views]]
+=== Buffer Views
+
+[[buffer-views-overview]]
+==== Overview
+
+A _buffer view_ represents a span of bytes defined by the `byteLength`, `buffer`, `byteOffset`, `byteStride`, and `target` properties. The `byteLength` property defines the buffer view's byte length, the `buffer` and `byteOffset` properties define the buffer to use as the data source and the byte offset within it, respectively, the `byteStride` property defines the stride when the buffer view is used for vertex attributes, and the optional `target` property hints at the buffer usage when transferring mesh data. Buffer views are defined in the asset's `bufferViews` array.
+
+If the `byteOffset` property is not defined, it is assumed to be zero.
+
+[NOTE]
+.Example
+====
+The following snippet defines two buffer views: the first holds the indices for an indexed triangle set, and the second holds the vertex data for the triangle set.
 
 [source,json]
 ----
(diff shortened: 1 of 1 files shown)


specification/2.0/Specification.adoc
```diff
@@ -924,26 +923,69 @@ The following example defines a buffer. The `byteLength` property specifies the
...
-The byte length of the referenced resource **MUST** be greater than or equal to the `buffer.byteLength` property.
+If the `uri` property is undefined, the buffer's data is provided by other means, such as a GLBv2 chunk (see below) or additional extensions. In absence of data sources, the buffer's content is undefined and handling of such buffers is implementation-dependent.
 
-Buffer data **MAY** alternatively be embedded in the glTF file via `data:` URI with base64 encoding. When `data:` URI is used for buffer storage, its `mediatype` field **MUST** be set to `application/octet-stream` or `application/gltf-buffer`.
+[[glb-stored-buffer]]
+==== GLB-stored Buffer
+
+The glTF asset **MAY** use the GLBv2 file container format to pack glTF JSON and one glTF buffer into the same file. Data for that buffer is provided by the GLB-stored `BIN` chunk.
+
+When the glTF asset is stored inside a GLBv2 file, a glTF buffer that has index zero, i.e., the first element of the `buffers` array, and has its `uri` property undefined represents the GLB-stored buffer. The GLB file **MUST** have a `BIN` chunk in this case.
```
javagl, line 939: I'd have to cross-check that with the old state. But when reading this paragraph "from scratch", I came up with the nitpick of what happens when a glTF in a GLBv2 has a `buffer[0]` with a URI that is _not_ `undefined`. But with the current wording, this is allowed, and would represent a GLB that _only_ contains the glTF JSON (and refers to an _external_ buffer, i.e. simply does not (and may not) have a `BIN` chunk), and there's no reason to disallow that.

specification/2.0/Specification.adoc
```diff
@@ -965,47 +1007,29 @@ The following example defines two buffer views: the first holds the indices for
...
 
-Any glTF buffer with undefined `buffer.uri` property that is not the first element of `buffers` array does not refer to the GLB-stored BIN chunk, and the behavior of such buffers is left undefined to accommodate future extensions and specification versions.
+Buffer views used by images or any other objects introduced by extensions or future Specification versions that refer to a buffer view as a whole, i.e., without specifying byte offset and/or byte length, **MUST NOT** be used by any objects that can refer to buffer view regions, such as accessors.
 
-The byte length of the `BIN` chunk **MAY** be up to 3 bytes bigger than JSON-defined `buffer.byteLength` value to satisfy GLB padding requirements.
+When a buffer view is directly used by vertex indices accessors or vertex attribute accessors, it **MAY** define the `target` property with a value of _element array buffer_ or _array buffer_, respectively. The `target` value uses integer enums defined in the <<_bufferview_target,Properties Reference>>. Buffer views used for other kinds of data **MUST NOT** define the `target` property.
```
javagl, line 1018: Trying to match that with the issue that is addressed here: Assuming that this paragraph fixes https://github.com/KhronosGroup/glTF/issues/2152 , I wonder if it could make sense to make the meaning of "direct" explicit, 

(Also, I have to assume that the change from **SHOULD** to **MAY** is intentional, but It surprised me a bit)

specification/2.0/Specification.adoc
```diff
@@ -1014,15 +1038,24 @@ See <<glb-file-format-specification,GLB File Format Specification>> for details
...
+The dimensionality of accessor elements is defined by the `type` property and the data type of those elements is defined by the `componentType` and `normalized` properties.
 
-All accessors are stored in the asset's `accessors` array.
+If the `bufferView` accessor property is defined, the accessor elements are sourced from the referenced buffer view and the `byteOffset` property defines the byte offset of the first accessed element within it. If the `byteOffset` property is not defined, it is assumed to be zero. If the buffer view's `byteStride` property is present, it defines the stride in bytes between the first bytes of consecutive accessor elements.
 
-The following example shows two accessors, the first is a scalar accessor for retrieving a primitive's indices, and the second is a 3-float-component vector accessor for retrieving the primitive's position data.
+If the `bufferView` accessor property is not defined, the accessor elements are sourced from an infinite sequence of zero bytes and the `byteOffset` property **MUST NOT** be defined.
```
javagl: Reading over it, this seems to address https://github.com/KhronosGroup/glTF/issues/2169 . 
(Just as a note - I assume that the case that the _actual_ data may come from an extension is handled elsewhere)

specification/2.0/Specification.adoc
```diff
@@ -1014,15 +1038,24 @@ See <<glb-file-format-specification,GLB File Format Specification>> for details
...
+
+If the `sparse` accessor property is defined, the accessor elements are selectively replaced based on the properties of the `sparse` object.
+
+Optional `min` and `max` properties provide minimum and maximum values that the accessor can provide, respectively.
```
javagl, line 1055: Note: I assume that the fact that they are required for POSITION is handled elsewhere.

specification/2.0/Specification.adoc
```diff
@@ -1033,107 +1066,271 @@ The following example shows two accessors, the first is a scalar accessor for re
...
+
+The following sections provide more details on each aspect of accessors.
 
 [[accessor-data-types]]
 ==== Accessor Data Types
 
-[options="header"]
+The effective component type of accessor elements is defined by the combination of the enumerated `componentType` and boolean `normalized` properties.
+
+The following table lists effective component types with their corresponding accessor properties and short names used in the subsequent sections of the Specification.
```
javagl, line 1094: Just a ping @aaronfranke : I think that you wanted to use these "short names" somewhere recently (some explainer or extension...? Or that 'int32' attribute thingy?...). Maybe now you could use them.

specification/2.0/Specification.adoc
```diff
@@ -1033,107 +1066,271 @@ The following example shows two accessors, the first is a scalar accessor for re
...
+All multi-byte component types use little endian byte order.
 
-Floating-point data **MUST** use <<ieee-754,IEEE-754>> single precision format.
+Signed integers use two's complement representation.
 
-Values of `NaN`, `+Infinity`, and `-Infinity` **MUST NOT** be present.
+The `normalized` property **MUST NOT** be true when the `componentType` is `5125` or `5126`.
 
-[options="header"]
+Only the values of the `componentType` property present in the table above are in scope of this Specification; support for others **MAY** be added by extensions of future Specification versions.
```
javagl: "...by extensions **or** future Specification versions."
???

specification/2.0/Specification.adoc
```diff
@@ -1033,107 +1066,271 @@ The following example shows two accessors, the first is a scalar accessor for re
...
 |====
-| `type`     | Number of components
-| `"SCALAR"` | 1
-| `"VEC2"`   | 2
-| `"VEC3"`   | 3
-| `"VEC4"`   | 4
-| `"MAT2"`   | 4
-| `"MAT3"`   | 9
-| `"MAT4"`   | 16
+| `type`     | Component size in bytes | Element size in bytes
```
javagl, line 1191: This addresses #2117 and makes the decisions there explicit.

specification/2.0/Specification.adoc
```diff
@@ -1033,107 +1066,271 @@ The following example shows two accessors, the first is a scalar accessor for re
...
+            "normalized": true,
             "type": "VEC3"
         }
     ]
 }
 ----
-
-In this accessor, the `componentType` is `5126` (_float_), so each component is four bytes.  The `type` is `"VEC3"`, so there are three components.  The size of each element is 12 bytes (`4 * 3`). Thus, the accessor takes 7020 bytes (`[7032 ... 14051]` inclusive range of the buffer view).
-
+====
```
javagl, line 1316: Checked 👍 

Image

specification/2.0/Specification.adoc
```diff
@@ -1033,107 +1066,271 @@ The following example shows two accessors, the first is a scalar accessor for re
...
+| `"MAT3"`   | 2                       | 24
+| `"MAT3"`   | 4                       | 36
+| `"MAT4"`   | 1                       | 16
+| `"MAT4"`   | 2                       | 32
+| `"MAT4"`   | 4                       | 64
 |====
 
-Element size, in bytes, is
-`(size in bytes of the 'componentType') * (number of components defined by 'type')`.
+Data layouts for accessors of `"MAT2"` type and any one-byte component type as well as accessors of `"MAT3"` type and any one- or two-byte component types include extra padding bytes (marked as `X`) as follows.
```
javagl, line 1203: Random aside: There are **no** statements about the values of the padding bytes, neither here nor in the "Sparse Accessors" section. So I assume that the values are arbitrary, and things like a sparse matrix accessor replacing a matrix
`m00 m10  XX  XX m01 m11 XX  XX` with a matrix
`m00 m10  YY  YY m01 m11 YY  YY`
(i.e. changing the padding bytes) are valid.

specification/2.0/Specification.adoc
javagl: Update: The possibility that the data is sourced from an extension is now only mentioned **very** casually and hidden, in the last paragraph of the "Accessor Bounds" section, which says

> When neither sparse nor bufferView properties are defined, [...] This is intended for use cases when data is supplied by external means (e.g., via extensions) ...

I did stumble over this (and it's still vaguely open via https://github.com/javagl/JglTF/issues/128 ), and it caused quite a few headaches (particularly, when trying to implement meshopt). 

I think that it should be mentioned more prominently **here**. 

**Very** roughly...

> If the `bufferView` accessor property is not defined, the accessor elements are not sourced from a buffer view. They can be sourced from an extension. If they are not sourced from an extension, then they must be treated as being sourced from an infinite sequence of zero bytes and the `byteOffset` property **MUST NOT** be defined.

(Wording to be polished, of course)

javagl (commented):
The change is big and largely a rewrite of the section, with paragraphs being moved around, so it's not easy to match the _old_ and _new_ state. 

However, I did a first pass. This mainly consisted of _reading_ the new state, and trying to match it with the issues that are supposed to be resolved with this. There are a few inlined comments. Some of them are really just comments/pointers to the issues that are addressed (maybe helpful for other reviewers). 

Two things that I did **not** see being addressed explicitly (but I might just have overlooked them): 

- The constraint that the byte stride may not be smaller than the element size, as of https://github.com/KhronosGroup/glTF/issues/2577 . (It did not stand out when reading the new text. Text-searches for `byteStride` did not reveal a prominen place. Maybe it is said "implicitly" somewhere, maybe just using the wording of `"byte stride"` or just `"stride"` instead of `"byteStride"`....)
- (Referring to one inlined comment: ) The requirement that the `POSITION` must have its `min` and `max` defined is mentioned in the "Geometry" section. One could consider to add a hint to this at the point where `min` and `max` are explained, because that section otherwise sounds like they are _always_ optional...

specification/2.0/Specification.adoc
javagl: (This is mentioned only in the "Geometry" section. It may be worth mentioning here that these values are not _always_ optional...)

specification/2.0/Specification.adoc
lexaknyazev: Correct, and that was not disallowed before.

specification/2.0/Specification.adoc
lexaknyazev: > change from **SHOULD** to **MAY**

I think there's a general agreement among WG members that missing `target` is not an issue.

lexaknyazev: Sections about sparse data explicitly mention that `target` is not for them.

specification/2.0/Specification.adoc
lexaknyazev: This is mentioned explicitly in the "Accessor Bounds" section. I'd prefer to not overload the "Overview" section.

specification/2.0/Specification.adoc
lexaknyazev: Yes, a few glTF 2.1 features would benefit from standardized names here.

specification/2.0/Specification.adoc
lexaknyazev: There are no requirements on specific replacement values, so it's indeed valid.

specification/2.0/Specification.adoc
javagl: (Aside: Feel free to mark pure comments as "Resolved" whenever you want. Otherwise, I'd leave them open (maybe just as pointers or heads-ups for other reviewers), and mark them as "resolved" once it approaches the state where it's supposed to be merged)

specification/2.0/Specification.adoc
lexaknyazev: Done.

specification/2.0/Specification.adoc
lexaknyazev: Fixed.

lexaknyazev:
@javagl The two specific issues (a typo and a better definition of accessors without buffer view sources) from the inlined comments have been addressed; would you do another pass?

---
> Two things that I did not see being addressed explicitly
> - The constraint that the byte stride may not be smaller than the element size
> - The requirement that the `POSITION` must have its `min` and `max` defined

L1258 and L1476.

specification/2.0/Specification.adoc
```diff
@@ -1014,15 +1038,26 @@ See <<glb-file-format-specification,GLB File Format Specification>> for details
...
+If the `bufferView` accessor property is defined, the accessor elements are sourced from the referenced buffer view and the `byteOffset` property defines the byte offset of the first accessed element within it. If the `byteOffset` property is not defined, it is assumed to be zero. If the buffer view's `byteStride` property is present, it defines the stride in bytes between the first bytes of consecutive accessor elements.
 
-All accessors are stored in the asset's `accessors` array.
+If the `bufferView` accessor property is not defined, the accessor elements are not sourced from a buffer view and the `byteOffset` property **MUST NOT** be defined.
 
-The following example shows two accessors, the first is a scalar accessor for retrieving a primitive's indices, and the second is a 3-float-component vector accessor for retrieving the primitive's position data.
+An accessor element source **MAY** be defined by an extension. In absence of element sources, accessor elements are sourced from an infinite sequence of zero bytes.
```
javagl, line 1051: I think that this sufficiently emphasizes the possibility of `bufferView` being undefined now, and that the actual source may then be provided by an extension.

javagl (approved):
I'll 'Approve' this from my perspective now. **But** I think that it should remain open for a little longer. It's a really big change, and others should have a chance to take a closer look.

specification/2.0/Specification.adoc
```diff
@@ -924,26 +923,69 @@ The following example defines a buffer. The `byteLength` property specifies the
...
-Buffer data **MAY** alternatively be embedded in the glTF file via `data:` URI with base64 encoding. When `data:` URI is used for buffer storage, its `mediatype` field **MUST** be set to `application/octet-stream` or `application/gltf-buffer`.
+[[glb-stored-buffer]]
+==== GLB-stored Buffer
+
+The glTF asset **MAY** use the GLBv2 file container format to pack glTF JSON and one glTF buffer into the same file. Data for that buffer is provided by the GLB-stored `BIN` chunk.
+
+When the glTF asset is stored inside a GLBv2 file, a glTF buffer that has index zero, i.e., the first element of the `buffers` array, and has its `uri` property undefined represents the GLB-stored buffer. The GLB file **MUST** have a `BIN` chunk in this case.
+
+[NOTE]
+.Example
```
emackey, line 942: Is there a pattern to which examples end up inside `[note]` and which do not?

emackey (approved):
Looks good to me.  One question about formatting.

emackey:
In case anyone else needs this, here are the rendered spec outputs from main vs this branch as of today:

[spec-main-2026-09-04.zip](https://github.com/user-attachments/files/31853070/spec-main-2026-09-04.zip)

[spec-branch-2026-09-04.zip](https://github.com/user-attachments/files/31853073/spec-branch-2026-09-04.zip)

specification/2.0/Specification.adoc
lexaknyazev: All examples _should_ eventually be inside `[note]` because they aren't normative, however it's not the case yet. As this PR addresses only one section, it only updates the relevant examples.

emackey: Got it, thanks.

lexaknyazev:
Thanks for the reviews! Merging this now to unblock further updates. Any additional feedback could be addressed in follow-up PRs.

Issue #2542, opened 2025-12-08
Clarification of "two implementations" requirement for ratification


javagl:
### Summary (and Disclaimer)

There is a requirement for ratification that is sometimes informally summarized as ~_"there must be two implementations (of that 'thing' that should be ratified)"_. In the context of glTF, this is mainly relevant for _extensions_. But it is not clear what this should mean.

This is, in many ways, not really important for me. I'm certainly not a fan of ~"over-formalizing processes", to the point where "following the process" implies more work than _just doing it_. But these questions are not so much about the process (_how_ things are done), but rather about the tasks (_what_ has to be done). Given the implications for a formal ratification, there should be clear answers to that.

### Context

I brought up the question of what "an implementation" should mean, recently [in comment in the `KHR_gaussian_splatting PR`](https://github.com/KhronosGroup/glTF/pull/2490#issuecomment-3589363862), and a [comment in the `KHR_meshopt_compression` PR](https://github.com/KhronosGroup/glTF/pull/2517#pullrequestreview-3548434351). It is somehow related to (not so recent) attempts to [clarify some aspects of the extension development process](https://github.com/KhronosGroup/glTF/pull/2225), even though that PR is closed in the meantime, and I omitted formal details about the ratification process there.

I think it could make sense to funnel these considerations and questions into a dedicated issue.

### The Core

The ratification requirement that "there must be two (independent) implementations" probably has its roots in the [Khronos Group Operational Guidelines (PDF)](https://www.khronos.org/files/working-group-guidelines.pdf), which lists...

> 3. Specification Development
> 3.1. Ratification Criteria and Process
> [...]
> Ideally, when a specification is submitted for ratification, the Working Group will be aware of at least two independent conformant implementations of the specification 
> [...]
> Pragmatically, final conformant implementations and conformance tests are interdependent and insisting on conformant implementations and proven tests before ratification could delay the specification to the extent of damaging market acceptance. In which case, the Board typically look for the following milestones to be met before ratification:
>     - The Working Group is confident that all areas of the specification have been independently implemented twice so that ‘specification bugs and flaws that become obvious through implementation’ have been discovered. The implementations do not have to be complete in themselves - multiple partial implementations from different members can be used to generate specification coverage, additionally, the implementations do not have to be released, or announced or conformant – or even available to the Working Group – working on trust to take implementers at their word.

[...]

PR #2642, opened 2026-09-02
Update extensions registry

lexaknyazev:
@weegeekps could we add anything to the line 536?

### DIFF
--- a/extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
+++ b/extensions/2.0/Khronos/KHR_gaussian_splatting/README.md
@@ -26,7 +26,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 ## Status
 
-Release Candidate
+Complete, Ratified by the Khronos Group
 
 ## Dependencies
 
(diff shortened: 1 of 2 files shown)


weegeekps:
@lexaknyazev Yes, maybe these?

- CesiumJS: https://github.com/CesiumGS/Cesium
- Cesium Native: https://github.com/CesiumGS/cesium-native
- @javagl's JSplat: https://github.com/javagl/JSplat
- @NorbertNopper-Huawei's 3dgs_ply2gltf tool: https://github.com/NorbertNopper-Huawei/3dgs_ply2gltf

Esri may have something to share for ArcGIS too, but I am unsure if any of their implementations are open source. Not sure what link they would want to use.

lexaknyazev:
We don't need to list all known implementations, just something to replace the placeholder.

javagl:
We had some discussions about this, and right now, I don't want to engage in arguments about what "what something being supported" means (and if so, that could/should happen at https://github.com/KhronosGroup/glTF/issues/2542). But people might want to try and load https://github.com/KhronosGroup/glTF/issues/2562 and expect the implementation to load them, and if this is not possible then it should be made clear. (I don't even know what kind of error message to expect there right now, but I'd place my bets on a `DeveloperError` being thrown)

lexaknyazev:
@javagl The question for this PR is much simpler: would you prefer JSplat to be mentioned in the extension or not?

javagl:
Prefer - compared to what?

Many extensions don't list "Known Implementations". Some just list "References", linking to papers, or GitHub permalinks to readers/writers (but from a quick sample, the latter is not used for the ratified ones). I'm still convinced that things like Draco don't even _have_ two independent implementations. 

In any case, this list could only be a "snapshot", as it will not be updated, and it could only be a formality, because it doesn't point to the relevant place in the code or make claims about whether the implementation is a reader, writer, or renderer.

So if a tool that can _write_ the extension (but not read it) and a renderer that can render the SPZ extension (but not the base extension) count as "two implementations of the (base?) extension", in the spirit of the ratification conditions of the "Khronos Operational Guidelines", then that may be enough. JSplat can read, write, and render the base extension (but no other aspects of glTF, because it's focussed on splats). I frankly don't care whether or not it is included.

Bottom line: I have no strong preference here.

lexaknyazev:
@weegeekps PTAL.

weegeekps (approved):
LGTM. 🚢

PR #2644, opened 2026-09-08
Generalize skinned mesh placement guidance in relation to skeleton node

Kjakubzak:
This PR adds long-missing guidance for where skinned meshes live in the node hierarchy, and defines what skin.skeleton means:                                                             
                                                                                                                                                                                            
  • Where to put the mesh node. A skinned mesh's bounding volume follows the mesh node's own position in the scene tree (not the joints), so the mesh node belongs under whichever animated node moves the character. (Displacement on the root joint, sometimes called "root motion," only reaches volumes on that node or below it; volumes aren't skeleton-aware, and per-frame re-posed bounds need runtime handling the spec doesn't define.).                                                                                                                         
  • One principle for all skinned meshes, not just ones with bounding volumes: keep the mesh node under a shared animated ancestor of the joints. Skins sharing one skeleton belong to the same skeleton.                                                                          
  • What skeleton is. It's now the common root of the joints hierarchy or an ancestor thereof. Declaring one is recommended but optional; ideally it also covers the meshes using the skin, and it may itself be a joint. This replaces the bounding-volume placement rule from #2619, whose sibling-of-joint-root option breaks under the common Blender-style layout (mesh next to the hips, locomotion on the hips) and whose "root motion" rationale was never defined.                                                                                                                                
                                                                                                                                                                                            
Deliberately conservative: every new rule is a SHOULD, never a MUST, so all existing valid 2.0 assets stay valid.


javagl:
Just a drive-by comment to express my confusion that the CLA bot is claiming that the GitHub user who opened the PR that caused the CLA bot to run is not a GitHub user. Websearches (with the nearly inevietable AI summary) indicate that this is a misconfiguration of some e-mail settings...

aaronfranke:
The CLA assistant error is because the commit is authored with `kenj@meta.com` but GitHub is configured to use `kjakubzak@gmail.com` as the email. GitHub allows adding multiple emails, so it should be an easy fix on Ken's side.

specification/2.1/Specification.adoc
```diff
@@ -873,7 +873,15 @@ Bounding volumes may be used to defer loading of content based on application-sp
...
+[NOTE]
+.Implementation Note
+====
+A bounding volume defined on a skinned mesh node is transformed by the mesh node's world transform (the composition of the node's local transform with all ancestor transforms), not by the skin's joint transforms. Because a skinned mesh node's own transform is ignored for the skinning computation, the mesh node's position in the scene hierarchy affects only the bounding volume, not the deformed geometry.
+
+If an ancestor node's animated transform displaces the character (for example, a translation animation targeting a joint or a navigation node), the skinned mesh node needs to be a descendant of that animated node for its bounding volume to track the displacement. A mesh node placed as a sibling of the displaced node will not inherit that displacement, and the bounding volume will remain stationary while the deformed mesh moves away from it. Displacement animated on the root joint — sometimes called "root motion" — therefore reaches the bounding volume only when the volume is defined on the displaced node itself or on one of its descendants. See <<joint-hierarchy,Joint Hierarchy>>.
```
aaronfranke: 1. Do we want to mention "the character" in this way? I already added text "such as the possible poses of a character" elsewhere, but this feels more concrete.
2. I am not sure if we should explicitly define root motion as the motion of the root joint - root motion might be done with animation of a glTF node that is the ancestor of all joints.

specification/2.1/Specification.adoc
```diff
@@ -873,7 +873,15 @@ Bounding volumes may be used to defer loading of content based on application-sp
...
+
+Node placement determines the bounding volume's world transform (translation, rotation, and scale inherited from ancestors); it does not account for skinned deformation. Bounding volumes are not skeleton-aware: a volume follows joint motion only through the transforms inherited from its node's ancestors. Per-frame tracking of skinned deformation (for example, recomputing bounds from posed joint transforms) requires runtime handling not defined by this specification. Skinned deformation may cause vertices to exceed any statically-defined volume regardless of the mesh node's placement.
```
aaronfranke:
```suggestion
Node placement determines the bounding volume's world transform (translation, rotation, and scale inherited from ancestors); it does not account for skinned deformation. Bounding volumes are not skeleton-aware: a volume follows joint motion only through the transforms inherited from its node's ancestors. Per-frame tracking of skinned deformation (for example, recomputing bounds from posed joint transforms) would require runtime handling not defined by this specification. Skinned deformation may cause vertices to exceed any statically-defined volume regardless of the mesh node's placement.
```

aaronfranke (approved):

PR #2645, opened 2026-09-11
Update Nodes, Cameras, and Animations

### LINKED ISSUE #2101: Composing Node's local transformation matrix
From [3.5.3 Transformations](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#transformations)

> To compose the local transformation matrix, TRS properties MUST be converted to matrices and postmultiplied in the T * R * S order; first the scale is applied to the vertices, then the rotation, and then the translation.

Does this need to specify `column-major` matrices? If the user converted the TRS properties to `row-major` matrices, I assume they would need to be premultiplied in the S * R * T order? e.g. [XMMATRIX](https://docs.microsoft.com/en-us/windows/win32/api/directxmath/ns-directxmath-xmmatrix)

### LINKED ISSUE #2140: Single or double precision node transforms
Does the specification indicate whether the node matrix and translation properties support double precision numbers? 

https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/node.schema.json

I think accessors are specifically limited to single precision and it seems that some viewers, such as Babylon.js and Unity, have issues rendering transform hierarchies with double precision. As such I’m curious whether there is an expectation that these values be kept in single precision range?

### LINKED ISSUE #2217: What should clients do "When the provided camera's aspect ratio does not match the aspect ratio of the viewport"?
The spec says this about perspective and orthographic cameras:

> When the provided camera's aspect ratio does not match the aspect ratio of the viewport, client implementations **SHOULD NOT** crop or perform non-uniform scaling ("`stretching`") to fill the viewport.

https://github.com/KhronosGroup/glTF/blob/8e798b02d254cea97659a333cfcb20875b62bdd4/specification/2.0/Specification.adoc#3103-projection-matrices

So… What should be done instead?

I personally interpret this to implicitly specific either letterboxing or "zooming out" to fit the entire specified camera frame. But it should probably be made explicit I guess.

Currently Babyon and Three both seem to stretch, while the Khronos reference viewer doesn't.

EDIT: It appears this may have been left deliberately open for the application to choose: https://github.com/KhronosGroup/glTF/pull/2053#issuecomment-933017285

### LINKED ISSUE #2430: Clarify animation channel target path validity
The current spec says:
> When `node` isn’t defined, channel **SHOULD** be ignored. Valid path names are `"translation"`, `"rotation"`, `"scale"`, and `"weights"`.

Additionally, the JSON schema for the `path` property allows any string value for extensibility.

---

The `KHR_animation_pointer` spec says:
> To use this extension, the animation channel target path **MUST** be set to `"pointer"` and the actual JSON pointer value is provided in the extension object's `pointer` property. The animation channel `node` property **MUST NOT** be set.

This extension spec does not (and cannot) cover the inverse case: if the `path` is set to `"pointer"` but the extension object is not present.

---

**Proposed base spec update**
> When the `node` property is defined, valid values for the `path` property are `"translation"`, `"rotation"`, `"scale"`, and `"weights"`.
>
> When the `node` property is not defined, the channel **SHOULD** be ignored and the `path` property **MAY** have any value.

@emackey @javagl WDYT?

### LINKED ISSUE #2468: TRS decomposition with negative scale makes camera matrix non-unique
As discussed in a [similar but closed issue](https://github.com/KhronosGroup/glTF/issues/2467), a TRS decomposition of a matrix is not unique due to the possibility of negative scales. Typically this doesn't present any problems, since only the product of these 3 transformations is of interest (except for animations, where one needs to animate each property separately, but animated nodes are explicitly forbidden to use the `matrix` property and should store the decomposed TRS directly).

However, when computing the camera view matrix, we are effectively required to do the TRS decomposition and then discard the scaling part, as per [3.10.2. View Matrix](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#view-matrix):

> The view matrix is derived from the global transform of the node containing the camera with the scaling ignored.

However, negative scaling makes the TRS decomposition non-unique, meaning that the camera view matrix is also non-unique.

Of course, I do realize that most sane glTF exporters won't add negative scaling effects to the camera matrix, and most glTF loaders will simply ignore the "with the scaling ignored" and use the node matrix directly as the camera view matrix. However, it does seem like a (tiny, admittedly) oversight in the specification.

Maybe 3.10.2 should say something like

> The view matrix is derived from the global transform of the node containing the camera with the scaling ignored, ***assuming that the scaling is strictly positive***.

?

### LINKED ISSUE #2547: Clarify restrictions for nodes with a defined matrix

lexaknyazev:
Fixes #2101
Fixes #2140
Fixes #2217
Fixes #2430
Fixes #2468
Fixes #2547


aaronfranke (approved):

weegeekps (approved):

DRx3D (approved):
Reviewed by committee - LGTU

PR #2646, opened 2026-09-18
Update Geometry

### LINKED ISSUE #2111: Disambiguation of attribute semantic names

lexaknyazev:
Fixes #2111.


specification/2.0/Specification.adoc
```diff
@@ -1613,101 +1615,137 @@ Accessors of the _uint32_ effective component type **MUST NOT** be used for any
...
+
+This kind of data layout is often called "`non-indexed`" because there is no indirection between the primitive's vertices and the attribute values. It is typically used for topology types that do not reuse the same vertex multiple times, such as point lists, line strips, or triangle strips (see below).
+====
+
+If the primitive's `indices` property is defined, it refers to the accessor that provides vertex indices, and the mesh primitive is assembled by taking the elements of the vertex indices accessor and fetching the corresponding elements of the attribute accessors in the order as they are provided by the vertex indices accessor. The total number of the primitive's vertices is equal to the value of the `count` property the vertex indices accessor.
```
javagl: One could _try_ to squeeze in the role of the `mode` here. But describing this in a way that is precise and still dealing with the fact that a `TRIANGLE_FAN` creates triangles where the 0th vertex is "implicit" for all of them (except the first one) could be more confusing than helpful. The `mode` is explicitly addressed in the next section, so that should be fine.

javagl (approved):
Most of this was discussed in the call. I looked at some of the points that had been open (e.g. "number of vertices" -> "amount of geometry"), and it looks fine for me.

specification/2.0/Specification.adoc
lexaknyazev: That's largely orthogonal. The role of `indices` is to pull and "instantiate" vertex data regardless of the topology type, and the role of `mode` is to connect those "instantiated" vertices.

